Issue #53, opened 2024-12-03
get rid of arg_utils and merge run.py and merge.py into the codebase


LionSR:

LionSR:
closed by #54

PR #54, opened 2024-12-04
Clean up CLI interface to depreciate argparse interface

LionSR:


LionSR:
this closes #53
although for future possibility of calling from python interface arg_utils is not removed

PR #86, opened 2025-03-03
Fix/claude pdf image (native pdf?)

LionSR:
in the reflection round


LionSR:
again got shot in the foot by async...

PR #131, opened 2025-05-18
Implement AbortController-based stream cancelation

LionSR:
## Summary
- introduce AbortController handling in BaseReflectionAgent and interrupt()
- add optional `signal` parameter to ModelHandler.createResponse and implementations
- pass AbortController signal to model handlers to enable real-time request abort
- document stop button behaviour with providers supporting AbortController
- add DOM lib typings to `tsconfig.json` for AbortController support

## Testing
- `npm run format`
- `npm test` *(fails: `vscode-test` not found)*

### DIFF
--- a/docs/guide/progress-board.md
+++ b/docs/guide/progress-board.md
@@ -40,7 +40,7 @@ The header provides a summary and actions for the selected stream:
   - **Red (Error)**: The agent encountered an error during execution.
   - **Yellow (Ready/Initial)**: The view is ready, but no stream is active yet.
 - **Stream Header Actions**:
-  - <i class="codicon codicon-debug-stop"></i> **Stop**: Attempts to gracefully stop the currently running task for this stream. Note that the current API call (if any) will complete first.
+  - <i class="codicon codicon-debug-stop"></i> **Stop**: Attempts to gracefully stop the currently running task for this stream. If the underlying model provider supports `AbortController` (such as OpenRouter), the ongoing request is cancelled immediately; otherwise the current API call will finish first.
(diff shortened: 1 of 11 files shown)


LionSR:
example abort anthropic


#!/usr/bin/env -S npm run tsn -T

import Anthropic from '@anthropic-ai/sdk';
const client = new Anthropic();

/**
 * This script demonstrates two ways of cancelling a stream,
 * by racing to see whether some Rust code prints "unwrap"
 * before 1.5 seconds or not.
 *
 * The most common is simply to `break` from the loop,
 * but you can also call `stream.controller.abort()` from outside the loop
 * if you need to.
 */
async function main() {
  const question = 'Hey Claude! How can I recursively list all files in a directory in Rust?';

  const stream = await client.messages.create({
    model: 'claude-3-5-sonnet-latest',
    stream: true,
    max_tokens: 500,
    messages: [{ role: 'user', content: question }],
  });

  // If you need to, you can cancel a stream from outside the iterator
  // by calling "stream.controller.abort()"
  const timeout = setTimeout(() => {
    console.log('\nCancelling after 1.5 seconds.');
    stream.controller.abort();
  }, 1500);

  for await (const event of stream) {
    if (event.type === 'content_block_delta' && event.delta.type === 'text_delta') {
      process.stdout.write(event.delta.text);

      // Most typically, you can cancel the stream by using "break"
      if (event.delta.text.includes('unwrap')) {
        console.log('\nCancelling after seeing "unwrap".');
        clearTimeout(timeout);
        break;
      }
    }
  }
}

main();

LionSR:
Help me finish it @claude

PR #189, opened 2025-05-24
Simplify progress view groups

LionSR:
## Summary
- simplify log group header markup to use `<summary>`
- replace custom expand/collapse code with `<details>` elements
- update logic for inserting groups and restoring their toggle state

## Testing
- `npm run format`
- `npm run lint`
- `npm test` *(fails: Module not found)*


LionSR:
@claude can you review it again for me? make sure everything would still work, and there is no other files that need to be changed too.

PR #205, opened 2025-05-24
Avoid message when packing only input file

LionSR:
## Summary
- suppress "Files packed" info when only the source .tex is found

## Testing
- `npm run format`
- `npm run lint`
- `npm test` *(fails: VS Code unavailable)*


LionSR:
can you fix it for me @claude

PR #209, opened 2025-05-24
feat: display token usage summary

LionSR:
## Summary
- show token usage and cost in progress header
- send token usage stats from OutputHandler
- persist usage stats in ProgressViewProvider
- update webview and DOM handlers to display new summary

## Testing
- `npm run format`
- `npm run lint`
- `npm test` *(fails: vscode-test requires network)*


LionSR:
@claude review this. Also think we can use some interface to make it more graceful

LionSR:
@claude fix, do as you say

LionSR:
Actually I think we should provide per round stats and show them in a per round fashion...

At the very top don't seem to be the best idea (as there might be many logGroup)

@claude

PR #225, opened 2025-05-28
Consolidate progress view groups

LionSR:
## Summary
- avoid nested progress view subgroups for output handling
- keep XML, file and diff processing under the same OutputProcessing group
- log statistics directly in the parent group

## Testing
- `npm run format`
- `npm run lint`
- `npm test`


LionSR:
after testing I think you misunderstood. I still want a outputhandler subgroup, just not outputhandler a/outputhandler b/outputhandler c; now all of these are under parent. change @claude

PR #231, opened 2025-06-02
Auto-refresh diff view

LionSR:
## Summary
- refresh diff editors after window size or word wrap changes
- wire refresh listener into compare command
- dispose refresh listeners once diff closes

## Testing
- `npm run format`
- `npm run lint`
- `npm test` *(fails: no network access)*

------
https://chatgpt.com/codex/tasks/task_e_683db54020bc8324affa5dcb1c72658b

### DIFF
--- a/src/commands/compareCommands.ts
+++ b/src/commands/compareCommands.ts
@@ -10,10 +10,10 @@ import * as logger from '../logger/logUtils';
 
 // Local imports - utilities
 import {
-  getWorkspacePath,
   getFullPathFromWorkspace,
   fileExists,
 } from '../utils/workspaceFileUtils';
+import { registerDiffRefresh } from '../utils/diffViewUtils';
 
 const CHANNEL = 'CompareCommands';
 logger.initialize(CHANNEL);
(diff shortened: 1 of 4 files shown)


LionSR:
can you add this constant at the constant.ts that already exists? @claude

LionSR:
I think the problem occurs when the switch between the Split View and the inline diff view happens, and then all the lines become automatically not wrapped, and the lines become infinitely long. can you find a smart workaround within our extension? @claude

LionSR:
the problem is:

Thanks, the [ConfigurationScope](https://code.visualstudio.com/api/references/vscode-api#ConfigurationScope) does let me get different values for different documents! This doesn't seem to pick up on toggling by the editor.action.toggleWordWrap command, though; when I toggle it and check the state again using the code you've given here, the value hasn't changed, and the listener code you gave never fires. Does that command somehow do something different instead of actually changing the configuration itself?

so we should just not try to get this config.

can you fix. also time constant should be in constant.ts. 85 is not very different from 95. 

@claude

LionSR:
you need to dramatic simplify this, we do not need onVisibleEditorsChange anymore. just detect the column change. also the part I commented out was done for a reason - we do not need that path!!! @claude

LionSR:
you need to dramatic simplify this, we do not need onVisibleEditorsChange anymore. just detect the column change. also the part I commented out was done for a reason - we do not need that path!!! @claude

PR #347, opened 2025-06-15
Fix progress view cleanup with override

LionSR:
## Summary
- ensure output name override is included when building stream IDs
- adjust pack and clean commands to pass the overridden path

## Testing
- `npm run format`
- `npm run lint`
- `npm test` *(fails: no output due to environment limitations)*

------
https://chatgpt.com/codex/tasks/task_e_684f1efa1e9083249cf0ec53f49646ff


LionSR:
be careful '' should not be passed if outputNameOverride is '' though.

maybe better to extract a common utility function



@claude

PR #611, opened 2025-07-03
Fix Progress View stream tab display issues

LionSR:
## Summary
- Stream tabs now appear immediately when tasks are launched
- Fixed duplicate task groups caused by race conditions
- Content is properly cleared when deleting the last stream tab

### 1. Stream tabs not appearing immediately
The issue was that `setActiveStream` was called from `executeAgent.ts` before any log messages existed, so the stream wasn't in `streamTabs` yet. The fix adds `ensureStream()` to create empty streams when `setActiveStream` is called.

### 2. Duplicate task groups
Race condition between `UPDATE_LOGS` (which sends all groups) and individual `ADD_TASK_GROUP` events. Fixed by checking if a group already exists in the DOM before creating it.

### 3. Content not cleared when deleting last stream
`WebviewUpdater.updateAll()` only updated content when there was an active stream. Now it explicitly clears content when no active stream exists.

## Test Plan
- [x] Launch a new task and verify the stream tab appears immediately
- [x] Delete the last stream tab and verify content is cleared
- [x] Run multiple tasks concurrently and verify no duplicate task groups appear
- [x] Switch between stream tabs and verify content updates correctly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/progressView/ProgressViewProvider.ts
+++ b/src/progressView/ProgressViewProvider.ts
@@ -182,9 +182,8 @@ export class ProgressViewProvider
       webviewView.webview.onDidReceiveMessage(async (message) => {
         if (message.command === COMMANDS.WEBVIEW_READY) {
           this._webviewReady = true;
-          if (this._pendingUpdate) {
-            this.updateWebview();
-          }
+          // Always update webview when it becomes ready to ensure all streams are shown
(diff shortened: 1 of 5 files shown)


LionSR:
bugbot run

PR #804, opened 2025-07-31
Add response saving for debugging

LionSR:
## Summary
- extend debugMessageSaver utility to save arbitrary objects
- save raw model responses with new maybeSaveResponse
- log API responses in ToolUse and standard agent cycles
- document `texra.debug.saveResponseObjects` configuration

## Testing
- `npm run lint`
- `npm test` *(fails: Critical dependency warning with webpack)*

------
https://chatgpt.com/codex/tasks/task_e_688b0b11d30483249174d491eb073b84

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -194,13 +194,11 @@ Control logging behavior:
 
 ```json
 "texra.logger.debugMode": false,
-"texra.debug.saveMessageObjects": false,
-"texra.debug.saveResponseObjects": false
+"texra.debug.saveDebugObjects": false
 ```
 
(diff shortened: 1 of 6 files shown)


LionSR:
@claude maybeSaveDebugObject

Can we maybe use one function and one config? I mean like in the package.json we just have to set one and if it's set then both will be saved. To manipulate this because I think there are two separate interfaces which are very very similar this will make it a little bit meaningless and hard to maintain. So also there are huge number of functional arguments to this too. Can we somehow group them or I don't know make it graceful and also make them more ordered? This would make it much better I think. Yes to be maintained basically this is the idea.

PR #846, opened 2025-08-18
fix: resize large images before encoding

LionSR:
## Summary
- resize images above 2000px using ImageMagick/GraphicsMagick before base64 encoding
- detect image dimensions and create temporary resized copies when needed
- remove temporary files after encoding to keep workspace clean

## Testing
- `npm run lint`
- `npm test` *(fails: libatk-1.0.so.0: cannot open shared object file)*

------
https://chatgpt.com/codex/tasks/task_e_68a31cc986148324a54a0379121aa8de

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Features
+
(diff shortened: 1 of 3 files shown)


LionSR:
@claude fix according to comments and add to changeling

PR #1037, opened 2025-09-18
fix: show tool name in tool use header

LionSR:
## Summary
- add a dedicated label element in the tool use log template so the header can be updated dynamically
- adjust the tool use formatter to populate the header with the parsed tool name when available

## Testing
- npm run format
- npm run lint
- npm test *(fails: VS Code binary requires libatk-1.0.so.0 in this environment)*

------
https://chatgpt.com/codex/tasks/task_e_68cc2331110883248ccf10f585b94330

### DIFF
--- a/src/progressView/index.html
+++ b/src/progressView/index.html
@@ -198,7 +198,7 @@
             <summary class="details-summary">
               <i class="toggle-icon"></i>
               <i class="codicon codicon-wrench"></i>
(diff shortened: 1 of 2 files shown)


LionSR:
@claude can we make sure tool name do not get displayed twice? One in the tool use title is enough

LionSR:
@codex @cursor review again

PR #1051, opened 2025-09-19
fix: load multiple output agent definitions

LionSR:
## Summary
- allow the agent loader to prefer `_multiple` definitions when multiple outputs are requested
- update runtime execution and resume flow to pass the new preference flag
- add unit coverage proving `_multiple` variants are selected with a fallback to the base agent

## Testing
- npm run lint
- npm test *(fails: vscode binary requires libatk-1.0.so.0 in container)*

------
https://chatgpt.com/codex/tasks/task_e_68cc3ce61c5483249eddf737ca833482


LionSR:
@Claude I do not think multiple should be propagated for inheritance, thus resolve the issue. Document this

LionSR:
@codex I do not think multiple should be propagated for inheritance, thus resolve the issue. Document this

PR #1088, opened 2025-09-22
fix: log agent errors to stream logger

LionSR:
## Summary
- cache the computed stream tab id and dedicated AgentLogger across executeAgentWithLogging so they are available to the outer catch block
- reuse the active agent log group when handling errors or create a temporary error group before emitting the failure message
- keep the existing shared channel logging so the progress view still receives the error notification

## Testing
- npm run format
- npm run compile
- npm run lint
- npm test 2>&1 | tail -n 200 (fails: missing libatk-1.0.so.0 when launching VS Code test harness)

------
https://chatgpt.com/codex/tasks/task_e_68d17095a2bc8324994a233528290013


LionSR:
@claude fix potential issues for me.

PR #1089, opened 2025-09-22
feat: enrich tool result summaries

LionSR:
## Summary
- add summary-aware payload helpers to `ToolResult` and update the base tool plumbing
- enrich the file operation tool and tool-use logging with structured log payloads and summaries
- surface tool summaries in the progress view formatter and update the tool-use tests

## Testing
- npm run format
- npm run lint
- npm run compile
- npm test *(fails: vscode-test missing libatk-1.0.so.0)*

------
https://chatgpt.com/codex/tasks/task_e_68d17304a26c8324a9ae8f904ed29019

### DIFF
--- a/src/agent/core/ToolUseCycle.ts
+++ b/src/agent/core/ToolUseCycle.ts
@@ -9,6 +9,21 @@ import type { ExtendedTokenUsageStats } from '@agent/types/UsageTypes';
 import { maybeSaveDebugObject } from '@agent/utils/debugMessageSaver';
 // Utility to run iterative tool-use cycles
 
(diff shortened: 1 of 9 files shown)


LionSR:
@claude I feel the current implementation is too complex. can you simplify in a graceful way while achieving what I want?

all these toPayLoad etc are very complex.
I think this is only a display issue....

- the tool output should not be too long even if there are multiple lines (there should be a maximal vertical length)
- when there is summary, only summary is displayed.
- when there are errors, display that in the tool use title. never just show isError: False/True.

PR #1106, opened 2025-09-23
fix: escape history renderer user strings

LionSR:
## Summary
- import the `he` encode helper and add utilities to sanitize history renderer output
- encode all user-provided fields, including comma-separated file lists, before inserting HTML
- escape tool configuration keys and values so snippets render safely

## Testing
- npm run format
- npm run compile
- npm run lint
- npm test *(fails: VS Code binary is missing libatk-1.0.so.0 in the container)*

------
https://chatgpt.com/codex/tasks/task_e_68d2adb61d188324b3304ca7f1270226

### DIFF
--- a/src/common/modules/htmlEncoding.js
+++ b/src/common/modules/htmlEncoding.js
@@ -1,0 +1,28 @@
+// Third-party imports
+import { encode as encodeEntities, decode as decodeEntities } from 'he';
+
+/**
+ * Encode a value for safe HTML insertion.
+ * @param {unknown} value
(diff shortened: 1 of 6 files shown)


LionSR:
@claude
The encodeListForHtml function could benefit from a configurable separator:

export const encodeListForHtml = (values, separator = ', ') => {
  if (!Array.isArray(values) || values.length === 0) {
    return '';
  }
  return values.map((entry) => encodeHtml(entry)).join(separator);
};

PR #1669, opened 2025-11-30
fix: handle Anthropic tool-use responses without text blocks

LionSR:
The Anthropic SDK's finalText() method throws an error when the
response contains only thinking + tool_use blocks with no text.
This happens during tool-use flows where the model decides to
call tools without any accompanying text.

Instead of using finalText(), manually extract text from the
response content array, which gracefully handles the case of
no text blocks by returning an empty string.

Fixes streaming error: "stream ended without producing a content
block with type=text"

---

> [!NOTE]
> Avoids SDK finalText() errors by manually extracting text from content blocks and reusing this logic in response extraction.
> 
> - **Anthropic Model Handler** (`src/agent/modelHandlers/modelHandlerAnthropic.ts`):
>   - **Text extraction utility**: Add `extractTextFromContent(content, trim?)` to concatenate `text` blocks from `BetaContentBlock[]`.
>   - **Streaming path**: Replace `stream.finalText()` with manual extraction from `response.content` to prevent errors when responses contain only `thinking`/`tool_use` blocks.
>   - **Response parsing**: Simplify `extractResponse` to use `extractTextFromContent(..., true)` and keep end-tag handling/replacements.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eea21d4e382704659a4cebc4d689680d8e7edd83. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/agent/modelHandlers/modelHandlerAnthropic.ts
+++ b/src/agent/modelHandlers/modelHandlerAnthropic.ts
@@ -166,6 +166,23 @@ const isAnyThinkingBlockParam = (
 const isBetaToolUseBlock = (block: BetaContentBlock): block is ToolUseBlock =>
   block.type === 'tool_use';
 
+/** Extract concatenated text from content blocks, optionally trimming each block */
+const extractTextFromContent = (
+  content: BetaContentBlock[] | undefined,
+  trim = false,
+): string => {
+  if (!Array.isArray(content)) {
+    return '';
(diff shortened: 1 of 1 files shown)


LionSR:
@codex review

PR #1723, opened 2025-12-08
Fix agent recovery from history streams

LionSR:
When SET_AGENT_OPTIONS arrives after STATE_RESTORE, if the saved agent value doesn't match any option in the dropdown (e.g., remote agent from another session, deleted custom agent), _restoreSelectValue would fall back to the first enabled option (default agent).

This fix:
- Makes _restoreSelectValue return boolean indicating if a match was found
- In _restoreAgentSelection, if no match was found but we have a saved value, recreate the placeholder option to preserve the selection

This ensures agent selection is properly preserved when restoring from history or streams, even when the agent isn't currently in the options.

---

> [!NOTE]
> Preserves agent selection when options are missing and refactors state restoration to set DOM directly, infer session type, restore checkboxes, and properly rebuild file lists.
> 
> - **State restoration**:
>   - New `_determineSessionType` and `_extractAgentValue` to reliably infer session type and agent values.
>   - `_handleStateRestoration` now sets DOM fields directly, saves state, and calls `applySessionType({... skipSave: true })` instead of `restore()`.
>   - `_restoreFormFields` returns `sessionType`, sets agents via `_setAgentValue`, restores `model`, instruction, single-file inputs, and checkbox values using `CHECK_BOXES` and `safeSetElementChecked`.
>   - `_restoreFileArrays` always clears lists, updates container visibility and toggle icon, then repopulates files.
> - **Agent selection recovery**:
>   - `_restoreSelectValue` now returns boolean; `_restoreAgentSelection` creates a placeholder option via `_setAgentValue` when no option matches saved/previous value.
> - **Imports**:
>   - Add `CHECK_BOXES` and `safeSetElementChecked`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1994630b6c6790c0c368e72f66d9869caf64b10f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/webview/modules/messageHandlers.js
+++ b/src/webview/modules/messageHandlers.js
@@ -785,105 +785,143 @@ export class MainViewMessageHandler extends BaseWebviewMessageHandler {
     }
   }
 
+  /**
+   * Restore the main view state from a TaskState object.
+   *
+   * Flow:
+   * 1. Extract session type from canonical session descriptor
+   * 2. Set all form field values directly in DOM
+   * 3. Set file arrays in DOM
+   * 4. Store state for persistence
+   * 5. Apply session type UI changes (visibility, disabled states)
+   *
+   * Note: We intentionally do NOT call mainViewState.restore() here because
+   * we already set all DOM values directly. Calling restore() would redundantly
+   * re-read state and re-apply to DOM, plus clear and rebuild file lists twice.
+   */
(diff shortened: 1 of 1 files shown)


LionSR:
@claude fix according to PR comments

PR #1729, opened 2025-12-09
fix(streamtabs): use full file path for stream tab deduplication

LionSR:
Previously, stream tab IDs used only the file basename for identification,
causing collisions when files with the same name existed in different
directories (e.g., project1/paper.tex and project2/paper.tex). Now the
full file path is used, ensuring each file gets its own unique stream tab.

---

> [!NOTE]
> Switches stream tab IDs to use workspace-relative file paths (normalized to forward slashes) to avoid basename collisions, with updated tests.
> 
> - **Logger**:
>   - **`getStreamTabId`**: Use full workspace-relative `inputFile` path instead of basename for IDs to avoid collisions.
>     - Normalize Windows backslashes to forward slashes in returned IDs.
>     - Preserve multiple-output suffix handling via `getMultipleName`.
>     - Tool-use stream IDs unchanged (`agent@model#shortId`).
> - **Tests**:
>   - Update expectations to use full paths and verify multiple-output suffix behavior.
>   - Add cases to distinguish same filenames in different directories.
>   - Add case to verify Windows path normalization.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 08df8b713a49803d4d58b4f7f75e1c0f4b10a5cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/logger/streamUtils.ts
+++ b/src/logger/streamUtils.ts
@@ -44,6 +44,7 @@ export function getStreamTabId(
   const agentName = options.useMultipleOutputs
     ? getMultipleName(cleanAgent)
     : cleanAgent;
-  const baseName = inputFile ? path.basename(inputFile) : '';
-  return `${agentName}@${model}: ${baseName}`;
+  // Use full path for deduplication to distinguish files with same name in different directories
+  const filePath = inputFile ?? '';
+  return `${agentName}@${model}: ${filePath}`;
 }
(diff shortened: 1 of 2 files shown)


LionSR:
@claude fix

LionSR:
@Claude think we are not using the full absolute path for this? In most case should've been relative to workspace

PR #1959, opened 2026-01-02
Fix streaming progress view initialization

LionSR:
When initializing a stream for a task group, refreshStreamSurface was called with forceRebuild: true BEFORE the group was added to state. This caused UPDATE_LOGS to be sent with empty groups, which made the frontend clear all groups added via ADD_TASK_GROUP - resulting in the "init" group and messages under other groups disappearing.

Changed to forceRebuild: false to use incremental update mode, which doesn't clear existing groups. The task group will arrive via the ADD_TASK_GROUP message immediately after.

---

> [!NOTE]
> Fixes race during new stream initialization so groups aren't cleared by a full refresh.
> 
> - In `TaskGroupEvents`, add the `TaskGroup` to state and set `activeRunId` before calling `initializeStreamForTaskGroup`; then send `ADD_TASK_GROUP` and await persistence
> - In `ProgressEventHandler.initializeStreamForTaskGroup`, keep `forceRebuild: true` and document that the group must already exist in state so `UPDATE_LOGS` includes it
> - Results: prevents empty `UPDATE_LOGS` from wiping groups and ensures correct initial render of the new task group
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5b52c39e043034fe0a9388e119d8d1f5ad9ecf69. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/progressView/events/ProgressEventHandler.ts
+++ b/src/progressView/events/ProgressEventHandler.ts
@@ -394,7 +394,9 @@ export class ProgressEventHandler {
     this.setStreamStatus(stream, status);
 
     if (this.webviewUpdater.isAvailable()) {
-      // Force rebuild since this is a new stream initialization
+      // Force rebuild to clear any previous stream's content. The new task
+      // group must be added to state BEFORE this call (in TaskGroupEvents)
+      // so UPDATE_LOGS includes it and the frontend renders it correctly.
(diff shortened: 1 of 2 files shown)


LionSR:
@codex review

PR #2078, opened 2026-01-08
Fix follow-up message disappearing after reload

LionSR:
Remove log content clearing from _clearSessionKindState() to fix data loss when sending follow-up messages after webview reload.

Root cause: When follow-up triggers session resume, setActiveStream event causes UPDATE_STREAMS (which may incorrectly detect session kind change due to missing task state) followed by UPDATE_LOGS with forceRebuild: false. The session kind change would clear log content, but the incremental UPDATE_LOGS wouldn't restore it.

Fix: Log content clearing should only happen in handleUpdateLogs with forceRebuild: true, which properly coordinates the clear operation with message re-rendering. Task group clearing remains in place as they're still incompatible between session kinds.

---

> [!NOTE]
> Addresses data loss when switching/resuming sessions and improves cleanup.
> 
> - Stop clearing `log` content in `_clearSessionKindState`; only clear during `handleUpdateLogs` with `forceRebuild: true`. Also clear logs when no active stream in `handleUpdateStreams`
> - Always send `todos` (default to `[]`) on stream refresh; send context state only if defined
> - Add `clearPendingTaskGroups(stream)` and `clearAllPendingTaskGroups()` in `ProgressEventHandler`, and invoke from `DELETE_STREAM`/`DELETE_ALL` handlers to prevent memory leaks
> - Keep full rebuilds on deletion via `updateWebview({ forceRebuild: true })`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 434966fbe01be36adaf24ee757f1e642ce52d303. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/progressView/ProgressViewMessageHandler.ts
+++ b/src/progressView/ProgressViewMessageHandler.ts
@@ -179,6 +179,8 @@ export class ProgressViewMessageHandler extends BaseViewMessageHandler {
   private async handleDeleteStream(message: any): Promise<void> {
     // Delete persisted session data if any
     await this.deleteSessionSnapshot(message.stream);
+    // Clear pending task groups to prevent memory leaks
+    this.provider.eventHandler.clearPendingTaskGroups(message.stream);
     await this.provider.state.clearStream(message.stream);
     // Force rebuild since we deleted a stream
     this.provider.updateWebview({ forceRebuild: true });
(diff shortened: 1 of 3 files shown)


LionSR:
@claude @codex review

PR #2924, opened 2026-03-22
Simplify output filename generation for workflow agents

LionSR:
## Summary
Refactored output filename generation to remove unnecessary agent and model information from filenames. Since workflow agent outputs are stored in task run storage that already provides execution context, filenames now only include the round number via a round-specific subdirectory.

## Key Changes
- **Simplified `getOutputFileName()`**: Removed `agent`, `model`, and `editedFile` parameters. Output files now follow the layout `r{round}/output.{ext}` (round is encoded in the directory, not the filename).
- **New `getExtractedDocOutputFileName()` function**: Extracted logic for generating paths for documents extracted from multi-document XML output. Files are written as siblings of the main output in the same round directory — layout `r{round}/{sourceName}.{ext}`. The round is provided by the caller's round directory, so no round suffix is added to the filename itself.
- **Updated `XmlOutputManager`**: Replaced filename parsing logic with direct use of `getExtractedDocOutputFileName()`, eliminating the need to extract agent/model from the output filename. Derives the round directory via `getFileDirectory(outputLocation)` so extracted docs inherit the parent's location kind (`runStorage` / `workspace` / `external`) when routed through `TaskRunFileService.createLocation()`.
- **Updated `runReflectionFlow()`**: Simplified output file location generation by removing model name parameter and `editedFile` logic from `getOutputFileName()` call.

## Implementation Details
- Removed dependency on `getAgentFirstNameChunk()` utility and `extractLastRoundMatch()` function as they're no longer needed.
- Output directory resolution still uses the input file's directory as a fallback, maintaining backward compatibility with the `outputDir` option.
- The refactoring reduces coupling between filename generation and agent/model metadata, making the code more maintainable and aligned with the actual storage structure.

https://claude.ai/code/session_01VFpyhiXsLxLCaSptCvHU6N

---

> [!NOTE]
> **Medium Risk**
> Changes output path/filename conventions and forces run-storage routing, which can break callers or tooling that relied on the previous naming/location scheme for workflow artifacts.
>
> **Overview**
> Workflow agent outputs are now always written to task-run storage and use a simplified per-round directory layout (`r{round}/output.{ext}`) instead of filenames that embed agent/model/edited-file metadata.
>
> `XmlOutputManager` no longer parses agent/model out of output filenames when splitting multi-document XML; extracted documents are written alongside the round output using a new `getExtractedDocOutputFileName()` helper, and `executeAgent` stops passing the removed `isSubagent` flag into `runReflectionFlow`.
>
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit be9589981ec9174251f70b4eacff851f687d19c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/implementations/flows/reflection/runReflectionFlow.ts
+++ b/src/agent/implementations/flows/reflection/runReflectionFlow.ts
@@ -61,8 +61,6 @@ export interface RunReflectionFlowInput<
   getOutputFileLocation?: (round: number) => AgentFileLocation;
   usageMonitor?: UsageMonitor;
   onRoundCompleted?: (roundIndex: number, totalRounds: number) => void;
-  /** When true, outputs are routed to task-run storage by default. */
-  isSubagent?: boolean;
 }
 
 export interface RunReflectionFlowResult {
@@ -174,14 +172,11 @@ export async function runReflectionFlow<C = unknown>(
   const { useScratchpad, shouldEnsureXmlStructure, totalRounds, outputExt } =
     deriveConfig(setting, prompt);
 
+  const inputDir = path.dirname(config.inputFile);
   const getOutputFileLocation =
     input.getOutputFileLocation ??
     ((round: number): AgentFileLocation => {
-      const fileName = getOutputFileName(
-        outputExt,
-        round,
-        config.inputFile,
-      );
+      const fileName = getOutputFileName(inputDir, outputExt, round);
       return fileService.createLocation(fileName) as AgentFileLocation;
     });
 
--- a/src/agent/implementations/flows/reflection/runReflectionFlow.ts
+++ b/src/agent/implementations/flows/reflection/runReflectionFlow.ts
@@ -138,8 +138,9 @@ export async function runReflectionFlow<C = unknown>(
   let shared: ReflectionFlowShared | undefined;
   let services: ReflectionServices<C> | undefined;
 
+  // Workflow outputs always go to run storage, never to the user's workspace.
   const fileService = new TaskRunFileService(executionId, {
-    forceRunStorage: input.isSubagent,
+    forceRunStorage: true,
   });
 
   const baseFiles: WorkspaceFileLocation[] = (
@@ -181,11 +182,7 @@ export async function runReflectionFlow<C = unknown>(
         round,
         config.inputFile,
       );
-      // Always route to run storage — raw/intermediate output should never
-      // land in the user's workspace.
-      return fileService.createRawOutputLocation(
-        fileName,
-      ) as AgentFileLocation;
+      return fileService.createLocation(fileName) as AgentFileLocation;
     });
 
   const interruptible: IInterruptible = {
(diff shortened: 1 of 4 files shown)


LionSR:
While simplifying, noticed `getExtractedDocOutputFileName(source, round, outputDir)` was re-appending `r${round}/` even though the call site in `XmlOutputManager.processMultipleLatexDocuments` passes `path.dirname(outputLocation.absolutePath)` — which is already the `r{round}/` folder created by `getOutputFileName`. Result on disk would have been doubly nested: `r{round}/r{round}/{name}.{ext}`, contrary to this PR's stated layout.

Fix: dropped the redundant `round` parameter and the duplicate `r${round}` join, so the helper just places the extracted doc inside the supplied round directory. Renamed the local at the call site to `roundDir` so the contract is obvious. Net -1 line, build + typecheck clean.

LionSR:
Pushed `1d41bf1af` addressing the Codex P1 and the two Copilot doc-mismatch comments.

**P1 regression (Codex + Copilot, `XmlOutputManager.ts` line 230)**

Honest explanation: my prior simplifier pass replaced the round-directory derivation with `path.dirname(outputLocation.absolutePath)` and fed that into `getExtractedDocOutputFileName` → `TaskRunFileService.createLocation()`. When run-storage mode is on, `outputLocation.absolutePath` is an absolute path under `<storage>/executions/<id>/r{round}/…`, which lives outside the workspace. `WorkspaceFS.locatePath()` rightly classified it as `external`, so `createLocation()` returned an `ExternalFileLocation` for each extracted document — dropping `runStorage` tracking. Downstream UIs then showed full absolute run-storage paths and treated the files as untracked.

**Fix**

Switched `processMultipleLatexDocuments` to derive the round directory via `getFileDirectory(outputLocation)` (already imported, already used in `splitScratchpadOutputXml` for the same reason). `getFileDirectory`:
- returns `path.dirname(location.relativePath)` for `workspace` / `runStorage` parents → yields a workspace-relative `r{round}/`, which `createLocation()` re-routes to runStorage when enabled
- falls back to `path.dirname(location.absolutePath)` only for `external` parents → preserved end-to-end

So the extracted document's `kind` now matches the parent `outputLocation`:
- runStorage parent → runStorage extracted doc (tracking preserved)
- workspace parent → workspace extracted doc
- external parent → external extracted doc

Pattern is now consistent with `splitScratchpadOutputXml`.

**Copilot doc-mismatch comments (lines 22 and 41)**

Updated the PR body rather than the code, since the simplification's shipping design is round-via-directory and callers depend on it:
- `getOutputFileName` → `r{round}/output.{ext}`
- `getExtractedDocOutputFileName` → `r{round}/{sourceName}.{ext}` (round supplied by the caller's round directory, not a filename suffix)

**Verification**
- `npm run compile:fast` — clean
- `npm run typecheck` — only the pre-existing `tsconfig.json` deprecation notices; no errors introduced by this change

PR #3033, opened 2026-04-16
fix: accept internal field in AgentSettingSchema

LionSR:
## Summary

- **Fix**: `latexFixer` (and any other agent with `settings.internal: true`) crashes silently on launch because `z.strictObject` in `AgentSettingBaseSchema` rejects the unknown key.
- **Approach**: Declare `internal: z.boolean().optional()` on `AgentSettingBaseSchema` so the field is valid at schema time. Simpler than a load-boundary strip, and it also covers `validateAgentYamlContent` — the parallel path raised by both Copilot and cursor[bot] reviews.

## Why the schema approach over stripping

The original approach stripped `internal` in `loadAgentSettingAndPrompts`. Reviewers (rightly) pointed out that `validateAgentYamlContent` parses `data.settings` with the same schema and would still reject YAMLs containing `internal`. Adding the field to the schema fixes both call sites in one place with no runtime logic.

Registry scanning in `agentRegistry.ts` reads the raw YAML before schema parsing (`rawSettings.internal === true`), so that flow is unchanged. The runtime keeps a typed `internal?: boolean` on the merged `AgentSetting` which can be consumed if needed; nothing reads it today.

## Files

- `src/agent/core/AgentDataclass.ts`: adds `internal: z.boolean().optional()` to `AgentSettingBaseSchema`.
- `src/agent/runtime/agentLoad.ts`: reverts the `stripYamlMetadata` helper and the central `YAML_METADATA_FIELDS` set — no longer needed.
- `src/test/agent/runtime/agentLoad.test.ts`: test case that loads a YAML with `settings.internal: true` and asserts the field round-trips through parsed settings without throwing (addresses Copilot's test-coverage ask).

## Test plan

- [x] `npm run typecheck` passes
- [x] New unit test covers `settings.internal: true` round-trip
- [ ] Launch `latexFixer` via "Fix Compilation" — should run instead of producing an empty stream
- [ ] Launch `latexFixer` from the Launcher with an instruction — should work
- [ ] Launch other tool-use agents (chat, research, etc.) — no regression
- [ ] Launch workflow agents (correct, devise, etc.) — no regression
- [ ] Agent with `inherits` — parent settings should still merge correctly

---

> [!NOTE]
> **Low Risk**
> Low risk: this only broadens agent YAML settings validation to accept an optional boolean field and adds a regression test, without changing load/merge behavior.
> 
> **Overview**
> Fixes agent YAML parsing/validation to **accept `settings.internal`** by adding an optional `internal?: boolean` field to `AgentSettingBaseSchema` (previously rejected by `z.strictObject`).
> 
> Adds a unit test in `agentLoad.test.ts` ensuring an agent definition containing `settings.internal: true` loads successfully and the value round-trips through `loadAgentSettingAndPrompts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91c9fb62555f895dbcbee780bc95b788b57527dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/core/AgentDataclass.ts
+++ b/src/agent/core/AgentDataclass.ts
@@ -32,6 +32,8 @@ export const AgentSettingBaseSchema = z.strictObject({
     )
     .prefault([]),
   tools: z.array(ToolDefinitionSchema).prefault([]),
+  /** Registry metadata: hides the agent from default launcher listings. */
+  internal: z.boolean().optional(),
 });
 
 const XmlStructureMode = z.enum(['never', 'scratchpadOnly', 'always']);
--- a/src/agent/runtime/agentLoad.ts
+++ b/src/agent/runtime/agentLoad.ts
@@ -100,23 +100,6 @@ export function ensureAgentCategoryForSource<
   return settings;
 }
 
-/**
- * YAML settings fields that are registry metadata (read by scanYaml)
- * and not part of the runtime AgentSettingSchema.
- * Add new YAML-only metadata fields here to prevent z.strictObject rejection.
- */
-const YAML_METADATA_FIELDS: ReadonlySet<string> = new Set(['internal']);
-
-function stripYamlMetadata(
-  raw: Record<string, unknown>,
-): Record<string, unknown> {
-  const result: Record<string, unknown> = {};
-  for (const [k, v] of Object.entries(raw)) {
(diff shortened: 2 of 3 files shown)


LionSR:
@claude fix

PR #3046, opened 2026-04-18
feat: show welcome view instead of disabling extension when no workspace

LionSR:
When no folder is open, the extension used to return early with only a
dismissible toast, leaving the TeXRA sidebar empty. Register a lightweight
welcome webview in the sidebar with "Open Folder" / "Clone Repository"
buttons so new users get a clear call-to-action, and reload the window
when a folder is added so full activation runs cleanly.

https://claude.ai/code/session_01FxXGNAmuovbWvJg2sqGHw3

---

> [!NOTE]
> **Medium Risk**
> Mostly UI/UX wiring, but it changes activation flow and adds new listeners/providers tied to the event bus and VS Code context keys, which could affect extension startup behavior and performance.
> 
> **Overview**
> When TeXRA is opened with **no folder** or a **multi-root workspace**, the extension now registers a lightweight `texra.welcomeView` with `viewsWelcome` copy (Open Folder/Clone Repo/tour/docs) instead of showing a toast and leaving the sidebar blank; once a single folder is added it triggers a window reload and sets the `texra.activated` context to swap to the main webview.
> 
> Adds an Explorer decoration provider that listens for `addOutputFiles` events and marks any workspace files involved in agent outputs/edits with a session-scoped **"T" badge**.
> 
> Declares `capabilities.untrustedWorkspaces.supported = false` so VS Code shows the standard workspace trust prompt with a clear rationale.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 041cd9fa9795f0685f3d66ebb0b437bd47b86fc6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3071, opened 2026-04-20
fix(palette): gate command palette entries behind texra.activated

### LINKED ISSUE #3052: Gate command palette entries behind `texra.activated`
## Context
Follow-up from #3046 (feat: show welcome view instead of disabling extension when no workspace).

#3046 added a `texra.activated` context key and gated `view/title` menu contributions behind it so toolbar buttons don't appear before `registerCommands(context)` runs. However, commands surfaced in the Command Palette (e.g. `texra.showMainView`, `texra.mainView.reset`, `texra.showDashboard`, `texra.cleanOutput`) are still listed in no-workspace mode and will fail with "command not found" if invoked, because `registerCommands(context)` returns early in the no-workspace activation path.

This isn't a regression introduced by #3046 (the palette leak existed before), but the PR surfaces it and it's the natural next step after gating the toolbar.

## Details
Add `"when": "texra.activated"` to `commandPalette` menu contributions in `package.json` for each command that requires the workspace-activated path. This ensures palette entries only appear after full activation has run, matching the `view/title` gate that #3046 introduced.

## References
- PR: #3046
- Raised in the Claude bot review on PR #3046

LionSR:
## Summary
- Pre-activation (no folder open) the welcome view runs and `registerCommands(context)` returns early, leaving every `texra.*` command unregistered. Until now they still appeared in the Command Palette and would fail with "command not found" when invoked.
- Add explicit `commandPalette` menu entries with `"when": "texra.activated"` for every command that `registerCommands()` registers, matching the `view/title` gate added in #3046.
- Debug-only commands (`testConnection`, `testAgentLoading`, `loadSpecificAgent`, `testTextEditor`, `parseXml`, `parseYaml`, `showLinterMessages`, `countLinterMessages`) keep their `config.texra.logger.debugMode` requirement, now combined with the new gate (`texra.activated && config.texra.logger.debugMode`).

Closes #3052

## Test plan
- [ ] Open VS Code with **no folder** → confirm `Cmd+Shift+P` and type `texra` shows zero TeXRA commands (only the welcome view in the sidebar).
- [ ] Open a single-folder workspace → confirm the palette lists all TeXRA commands as before (and toolbar buttons appear in `view/title`).
- [ ] Toggle `texra.logger.debugMode` on with a workspace open → confirm the 8 debug commands appear; toggle off → they disappear.
- [ ] Build: `npm run compile:fast` succeeds.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only VS Code contribution visibility/activation timing to prevent invoking unregistered commands; main risk is unintentionally hiding palette entries or keybindings if `texra.activated` is not set in some activation paths.
> 
> **Overview**
> Prevents "command not found" errors by **gating all TeXRA Command Palette entries, keybindings, editor/explorer menus, and the Getting Started walkthrough** behind the `texra.activated` context so they only appear after a single-folder workspace is open and commands are registered.
> 
> Moves the `setContext('texra.activated', true)` call to the end of `activate()` (after all `registerCommand` calls) and tweaks the empty-workspace welcome copy to clarify that TeXRA commands/tour become available only once a folder is opened.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41ee6743a7ec9ebf779dc137c29edadd25b8566c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "Data Science"
   ],
   "activationEvents": [
+    "onStartupFinished",
     "onColorTheme"
   ],
   "main": "./dist/extension.js",
--- a/package.json
+++ b/package.json
@@ -1091,6 +1091,7 @@
         "id": "texra.gettingStarted",
         "title": "Get started with TeXRA",
         "description": "Get TeXRA up and running in a few minutes. Connect your account, pick your field, and run your first paper through the orchestrator.",
+        "when": "texra.activated",
         "steps": [
           {
             "id": "texra.gettingStarted.sampleWorkspace",
@@ -1242,7 +1243,7 @@
     "viewsWelcome": [
       {
         "view": "texra.welcomeView",
-        "contents": "TeXRA is a multi-agent orchestration system for LaTeX research. An orchestrator coordinates specialized agents that derive results, verify proofs, run symbolic computation, generate figures, and assemble manuscripts — every step auditable, every citation grounded.\nOpen a folder to begin.\n[Open Folder](command:workbench.action.files.openFolder)\n[Clone Repository](command:git.clone)\nNew to TeXRA? [Take the tour](command:workbench.action.openWalkthrough?%22texra.gettingStarted%22) or [read the docs](https://github.com/LionSR/TeXRA#readme).",
+        "contents": "TeXRA is a multi-agent orchestration system for LaTeX research. An orchestrator coordinates specialized agents that derive results, verify proofs, run symbolic computation, generate figures, and assemble manuscripts — every step auditable, every citation grounded.\nOpen a folder to begin. The TeXRA tour and commands become available once a workspace is open.\n[Open Folder](command:workbench.action.files.openFolder)\n[Clone Repository](command:git.clone)\nWant to learn more first? [Read the docs](https://github.com/LionSR/TeXRA#readme).",
         "when": "workbenchState == empty"
       },
       {
@@ -1254,32 +1255,32 @@
     "menus": {
       "editor/title": [
         {
-          "when": "editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
+          "when": "texra.activated && editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
           "command": "texra.indentCurrentTeX",
           "group": "navigation"
         },
         {
-          "when": "editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
+          "when": "texra.activated && editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
           "command": "texra.applyReplacements",
           "group": "navigation"
         },
         {
-          "when": "editorLangId == latex",
+          "when": "texra.activated && editorLangId == latex",
           "command": "texra.fixCompilation",
           "group": "navigation"
         },
         {
-          "when": "editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
+          "when": "texra.activated && editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
           "command": "texra.getTeXCount",
           "group": "navigation"
         },
         {
-          "when": "editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
+          "when": "texra.activated && editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
           "command": "texra.cleanOutput",
           "group": "2_texra_housekeeping@1"
         },
         {
-          "when": "editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
+          "when": "texra.activated && editorLangId =~ /^latex$|^latex-expl3$|^doctex$/",
           "command": "texra.cleanBuild",
           "group": "2_texra_housekeeping@2"
         }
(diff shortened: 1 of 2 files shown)


LionSR:
Thanks for the careful review. Pushed `8ffc842` addressing both gaps:

**Keybindings & non-palette menus** — `keybindings`, `editor/title`, `editor/context`, and `explorer/context` entries all now AND `texra.activated` into their `when` clauses (existing filters like `editorLangId =~ /^latex$|^latex-expl3$|^doctex$/` and `focusedView == texra.mainView` are preserved). Pre-activation the keybindings stay inert and the editor/title/context menu items no longer appear, so they cannot fire commands that `registerCommands()` skipped.

**Walkthrough** — gated the entire "Get started with TeXRA" walkthrough with `"when": "texra.activated"` since every step links to `command:texra.*`. The pre-activation welcome view used to include a "Take the tour" link; that link would have dead-ended once the walkthrough was gated, so I replaced it with a sentence telling the user the tour and commands become available after they open a folder. The `[Read the docs]` external link is preserved so users still have somewhere to go pre-activation.

`view/title` entries already had `view == texra.mainView` filters which only render post-activation, so no change needed there.

Verified: `node -e "JSON.parse(require('fs').readFileSync('package.json','utf8'))"` parses, `npm run compile:fast` succeeds.

LionSR:
Thanks for the careful review — you're right that this is worth being defensive about, so I've pushed a follow-up: **`881d3cd8a fix(activation): add onStartupFinished so palette gating cannot strand users`**.

### Investigation

Pre-fix, `activationEvents` was just `["onColorTheme"]`, and `texra.activated` is set inside `activate()` at `src/extension.ts:113` — *after* the early return for the no-folder/multi-root case at lines 108–111. So the gating only takes effect once `activate()` has run and reached line 113.

In practice the extension was still activating in single-folder windows even before this commit, because:

1. `onColorTheme` fires very early in the workbench startup sequence (theme application happens before the user can interact with the palette), and
2. The TeXRA Activity Bar view container + `texra.welcomeView` / `texra.mainView` registrations cause activation as soon as that sidebar is rendered.

So the worst-case window where palette entries would be hidden was extremely narrow (the brief moment between workbench paint and theme apply, in a workspace where the sidebar was never opened). Not a hard regression in the field, but the contract was fragile and depended on an undocumented side-effect of `onColorTheme`.

### Fix

Added `onStartupFinished` to `activationEvents` so the extension is guaranteed to activate (and `texra.activated` to be set) after the workbench finishes loading, regardless of theme behavior or sidebar visibility. This is the canonical solution and costs nothing — `onStartupFinished` is explicitly designed for "activate after startup, but don't block startup."

The no-folder / multi-root fallback path (welcome view with "Open Folder" / "Clone Repository" link) is unchanged and remains driven by `viewsWelcome` with `when: workbenchState == empty`, so users in those workspaces still have a clear entry point even though `texra.activated` will not be set.

JSON validated, `npm run compile:fast` clean.

LionSR:
Thanks for catching this. Pushed 97e08a058 which addresses the race directly (Option A): the `setContext('texra.activated', true)` call has been moved out of the top of `activate()` and now runs immediately after `registerCommands(context)` and `registerFileDecorations(context)`. Since `registerCommands()` also registers the `texra.mainView` webview view provider, by the time the context key flips to `true` every command handler and view provider that the gated UI contributions (commandPalette, keybindings, editor/explorer menus, walkthrough, views) reference is wired up.

Net effect: `texra.activated` now means "TeXRA is fully wired up" rather than "TeXRA's `activate()` started running", so the commandPalette / keybinding entries no longer briefly become invokable while their handlers are still missing. No second context key was needed since nothing in `activate()` between the old and new locations depends on `texra.activated` already being set (the only other reference in the codebase is a doc comment in `welcomeView.ts`).

LionSR:
Good catch — the prior fix (`97e08a058`) was incomplete.

That commit moved `setContext('texra.activated', true)` to right after `registerCommands(context)`, but I missed that `texra.showMainView` and `texra.toggleView` are NOT registered inside `registerCommands()`. They are registered later in `activate()` via two direct `vscode.commands.registerCommand` calls (previously around lines 343-344). Since both have `commandPalette` entries gated on `texra.activated`, the original race window persisted for those two commands.

New fix in `3f694501d`: moved the `setContext('texra.activated', true)` call to the very end of `activate()`, after every `registerCommand` site in this file (verified with grep over `src/extension.ts`, `src/welcomeView.ts`, and `src/fileDecorations.ts`). The accompanying comment was rewritten to be explicit about why the call must be last and to name the late-registered commands so future edits don't reintroduce the same race.

Build verified with `npm run compile:fast`.

LionSR:
Addressed outstanding review comments:

- **cursor[bot] / chatgpt-codex-connector (walkthrough opened before `texra.activated` is set)** — fixed in `41ee6743a`. Moved `setContext('texra.activated', true)` above the first-run walkthrough launch. Every `registerCommand` call in `activate()` still runs before the context flips, so the original ordering guarantee is preserved.

Previously addressed on this branch:

- **Copilot (keybindings / editor menus not gated)** → commit `8ffc842f4`.
- **Copilot (walkthrough contained `command:texra.*` links but was reachable pre-activation)** → commit `8ffc842f4` gated the walkthrough itself.
- **Copilot & chatgpt (texra.activated flipped too early)** → commits `97e08a058` and `3f694501d` moved setContext past all registrations.
- **chatgpt P1 (bootstrap — users can't activate from palette)** → commit `881d3cd8a` added `onStartupFinished` activation event.
- **cursor[bot] (`texra.showMainView` / `texra.toggleView` registered after gate was set)** → commit `3f694501d`.

Build verified locally (`npm run compile:fast`).

PR #3077, opened 2026-04-20
fix(streamTabs): hide expanded child streams in compact sidebar

### LINKED ISSUE #2986: Fix child streams remaining visible in compact sidebar mode
## Context
Follow-up from #2984 (perf+ux: optimize stream tabs for 100+ child streams).

The refactor in #2984 replaced the old `hasChildren && !this.compact` rendering guard with `children && childCount > 0`, dropping the `!this.compact` check. Similarly, the `expanded` flag doesn't factor in `this.compact`. If a parent tab is expanded before the sidebar narrows into compact mode, child streams remain visible with no way to collapse them (the chevron is correctly hidden in compact mode via the `stream-tab` template).

## Details
- In `StreamTabs` render, the child-streams `<div>` should be hidden when `this.compact` is true, either by:
  - Adding `|| this.compact` to the `?hidden` condition: `?hidden=${!expanded || this.compact}`
  - Or factoring `!this.compact` into the `expanded` computation
- Medium severity — causes layout issues when the sidebar is resized to compact width while children are expanded

## References
- PR: #2984
- Review comment: https://github.com/LionSR/TeXRA/pull/2984#discussion_r3035787024
- File: `src/progressView/frontend/components/StreamTabs.ts` ~L599-609

LionSR:
## Summary

Follow-up to #2984. The refactor dropped the `!this.compact` check from the child-streams render guard, so a parent tab expanded before the sidebar narrows to compact mode kept its children visible with no way to collapse them (the chevron is correctly hidden in compact mode).

Adds `|| this.compact` to the `?hidden` condition on the `.child-streams` div so children are hidden whenever the layout is compact, independent of `expanded` state. Locally scoping the layout-mode check to the render guard keeps the `expanded` flag's semantics intact.

Closes #2986

## Test plan

- [ ] Open Progress View with an orchestrator that has 2+ child streams.
- [ ] Expand a parent tab via the chevron with the sidebar at normal width; verify children appear.
- [ ] Drag the sidebar narrower until it enters compact mode; verify child streams are hidden and the chevron disappears.
- [ ] Widen the sidebar back to normal mode; verify the parent is still expanded and children re-appear.

---

> [!NOTE]
> **Low Risk**
> Small, UI-only change that alters conditional rendering for child streams; risk is limited to compact-mode display/regression in expand/collapse behavior.
> 
> **Overview**
> Updates `StreamTabs` rendering so when the sidebar is in *compact* mode it forces `childCount` to `0`, preventing the `.child-streams` subtree from rendering at all.
> 
> This ensures child streams can’t remain visible in compact layout and also reduces DOM/memory usage for sessions with many child streams, while preserving the existing expand/collapse behavior in non-compact mode.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4309b717ff094a459773122a741ad38062342fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/progressView/frontend/components/StreamTabs.ts
+++ b/src/progressView/frontend/components/StreamTabs.ts
@@ -611,7 +611,7 @@ export class StreamTabs extends LitElement {
               (stream) => stream.name,
               (stream) => {
                 const children = this.childStreamsByParent.get(stream.name);
-                const childCount = !this.compact ? (children?.length ?? 0) : 0;
+                const childCount = children?.length ?? 0;
                 const expanded =
                   childCount > 0 && this.expandedParents.has(stream.name);
 
--- a/src/progressView/frontend/components/StreamTabs.ts
+++ b/src/progressView/frontend/components/StreamTabs.ts
@@ -616,7 +616,7 @@ export class StreamTabs extends LitElement {
                   childCount > 0 && this.expandedParents.has(stream.name);
 
                 // prettier-ignore
-                return html`<stream-tab .info=${stream} .compact=${this.compact} .status=${this.getStatus(stream.name)} .lastTimestamp=${this.getTimestamp(stream.name)} ?active=${stream.name === this.activeStreamId} .hasPendingApproval=${this.pendingApprovalStreamIds.has(stream.name)} .childCount=${childCount} ?expanded=${expanded}></stream-tab>${children && childCount > 0 ? html`<div class="child-streams" ?hidden=${!expanded}>${repeat(children, (child) => child.name, (child) => {
+                return html`<stream-tab .info=${stream} .compact=${this.compact} .status=${this.getStatus(stream.name)} .lastTimestamp=${this.getTimestamp(stream.name)} ?active=${stream.name === this.activeStreamId} .hasPendingApproval=${this.pendingApprovalStreamIds.has(stream.name)} .childCount=${childCount} ?expanded=${expanded}></stream-tab>${children && childCount > 0 ? html`<div class="child-streams" ?hidden=${!expanded || this.compact}>${repeat(children, (child) => child.name, (child) => {
                   const childStatus = this.getStatus(child.name);
                   const isFinished = !ACTIVE_CHILD_STATUSES.has(childStatus);
                   // prettier-ignore
--- a/src/progressView/frontend/components/StreamTabs.ts
+++ b/src/progressView/frontend/components/StreamTabs.ts
@@ -611,12 +611,20 @@ export class StreamTabs extends LitElement {
               (stream) => stream.name,
               (stream) => {
                 const children = this.childStreamsByParent.get(stream.name);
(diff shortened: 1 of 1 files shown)


LionSR:
Thanks @cursor and @copilot — both reviews were correct that `|| this.compact` was unreachable on the previous commit. The outer ternary (`children && childCount > 0 ? html : nothing`) was evaluated against `childCount = !this.compact ? ... : 0`, so when the sidebar was compact the whole `<div class="child-streams">` was already being torn down before `?hidden` ever ran.

The earlier short-circuit also silently regressed the original design intent from #2984 ("Use `?hidden` instead of conditional template for child-streams div, preserving DOM nodes across expand/collapse toggles") whenever the layout went narrow.

New commit `8975f3d5c` removes the dead branch in `childCount` and lets the `?hidden=${!expanded || this.compact}` attribute do the work it was meant to do. Concretely:

- `childCount = children?.length ?? 0` — purely from the data, so the `<div class="child-streams">` always mounts when children exist.
- The chevron toggle inside `StreamTab` is still independently gated by `hasChildren = this.childCount > 0 && !this.compact`, so it stays hidden in compact mode.
- `?hidden=${!expanded || this.compact}` now actually fires and hides the child list when the sidebar narrows, without destroying 100+ child `StreamTab` components.

Build (`npm run compile:fast`) and `npm run typecheck` are clean (only the pre-existing `src/latex/arxivProcessor.ts:133` axios error remains).

LionSR:
Addressed Codex's P2 review in `e4309b71`.

**Fix:** Restored the `childCount = !this.compact ? (children?.length ?? 0) : 0` short-circuit so the entire `<div class="child-streams">` subtree is unmounted in compact mode (rather than mounted-then-hidden). Also dropped the now-redundant `|| this.compact` from the `?hidden` attribute.

**Why this is the right balance:**
- Issue #2986 is still fixed: in compact mode `childCount` is 0, so the children div is never rendered (and the chevron is gated by `hasChildren = childCount > 0 && !this.compact` inside `StreamTab`).
- PR #2984's DOM-preservation goal is preserved for the common case it targeted: user-driven expand/collapse toggles in non-compact mode. In non-compact mode the div always mounts with `?hidden=${!expanded}`, so the chevron click no longer tears down 100+ child `stream-tab` components.
- The compact boundary is a sidebar-resize event (infrequent compared to expand/collapse clicks), so unmount/remount across that boundary is an acceptable tradeoff and frees the memory.
- `StreamTabs.compact` is already declared as `@property({ type: Boolean, reflect: true })` and propagated by `ProgressApp` via `.compact=${compactTabs}`, so reactivity is sound and the conditional cleanly tears the subtree down when the layout flips to compact.

Verified with `npm run compile:fast` and `npm run typecheck` (only pre-existing axios error in `src/latex/arxivProcessor.ts:133`, unrelated).

PR #3079, opened 2026-04-20
fix(prPolling): paginate check-runs so terminal CI fires on >100 runs

### LINKED ISSUE #3072: Paginate check runs so terminal CI events fire on PRs with >100 checks
## Context
Follow-up from #3064 (feat(pr-activity): emit terminal CI events + surface review thread replies).

#3064 added one-shot `CI complete` / `CI passed` events per head SHA in `PRPollingSource`. To avoid premature emission from a truncated check-runs page, the gate was tightened to require a complete page:

```ts
runs.length > 0 &&
runs.length >= checksRes.data.total_count &&
runs.every((r) => r.status === 'completed')
```

The check-runs request is fixed at `per_page=100` (no pagination loop). When `total_count > 100`, `runs.length >= total_count` is always false, so neither `CI completed` nor `CI passed` can ever emit for that SHA. Agents waiting on terminal CI to gate a merge will stall indefinitely on PRs with large CI matrices, even after every check finishes green.

This was flagged by the Codex reviewer on the merged commit (`42f33206`) — see [discussion_r3107776483](https://github.com/LionSR/TeXRA/pull/3064#discussion_r3107776483). The current TeXRA CI matrix is well under 100 runs, so the gap is latent for now.

## Details
- File: `src/tools/github/PRPollingSource.ts`, around the check-runs fetch (line ~395) and both terminal-event gates (initial seeding ~440, emit path ~535).
- Walk all pages of `/repos/{owner}/{repo}/commits/{sha}/check-runs?per_page=100` (Link `rel="next"` or `?page=N` until `runs.length >= total_count`) and apply the terminal-state gate against the assembled full list.
- Re-evaluate the seeding branch the same way so it doesn't suppress real transitions on the first tick.
- ETag handling needs care: subsequent pages should use their own `If-None-Match` if we want to keep 304s on unchanged subsequent pages, but the simplest correct change is to drop ETag short-circuiting on multi-page fetches (page 1 304 may not imply page 2 unchanged in practice).

## References
- PR: #3064
- Review comment: https://github.com/LionSR/TeXRA/pull/3064#discussion_r3107776483
- Code: `src/tools/github/PRPollingSource.ts` (check-runs fetch + terminal-event gates)

LionSR:
## Summary

- The check-runs endpoint caps at `per_page=100`. The single fetch in `PRPollingSource` left `runs.length >= total_count` permanently false on PRs with more than 100 registered check runs (large monorepos, matrix builds), so the terminal `CI completed` / `CI passed` events never emitted and agents subscribed via PR-activity tools were stranded waiting on CI.
- Added a `fetchAllCheckRuns` helper that walks pages until the full set is collected, then both terminal-state gates (initial seeding and emit) are evaluated against the assembled list.
- ETag handling preserves the cheap 304 cache hit for the common single-page case (`total_count <= 100`); on multi-page responses the stored ETag is dropped because a 304 on page 1 doesn't imply pages 2+ are unchanged and GitHub doesn't expose a combined ETag.

Closes #3072

## Test plan

- [x] `npm run typecheck` passes (the pre-existing `src/latex/arxivProcessor.ts:133` axios error is unrelated).
- [x] `npm run compile:fast` succeeds.
- [ ] Manual verification on a PR with >100 check runs is hard to reproduce locally; confirmed by code review:
  - Single-page (`total_count <= 100`): unchanged fast path, ETag is preserved, next tick can 304.
  - Multi-page: page 1 is conditional, pages 2..N are unconditional GETs, results concatenated, terminal gates fire once `runs.length >= total_count`.
  - 304 on page 1 still short-circuits the whole call (covers the common steady-state).

---

> [!NOTE]
> **Medium Risk**
> Touches core polling/state logic and adds multi-request pagination with caching, so bugs could cause missed CI transitions or increased GitHub API load; changes are localized to `PRPollingSource`.
> 
> **Overview**
> Fixes a long-standing bug where PR subscriptions could *never* emit terminal CI events when a commit had more than 100 check-runs by adding a paginated `fetchAllCheckRuns` path and evaluating CI completion/pass gates against the fully-assembled run set.
> 
> Introduces per-page ETag/page-content caching for check-runs (with a single-page 304 fast path), dedupes runs by id to handle mid-pagination churn, caps maximum pages to limit runaway API fanout, and defers cache commits until after successful diff/emit processing (also clearing the cache on new head SHA).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce222da6d5782c60a9cdc76d899f5daad8f44a6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/tools/github/PRPollingSource.ts
+++ b/src/tools/github/PRPollingSource.ts
@@ -484,9 +484,11 @@ export class PRPollingSource implements AsyncEventSource {
         // runs.length > 0: an empty array is ambiguous (no CI configured vs.
         // runs not yet registered), so "done" isn't meaningful and seeding
         // would suppress the real terminal event once runs appear. Also
-        // require we have the full set (length >= total_count); fetchAllCheckRuns
-        // returns the latest observed total_count, so this also covers runs
-        // added mid-walk that bumped the page count.
+        // require we have the full set (length >= total_count). The runs
+        // array from fetchAllCheckRuns is deduped by id and total_count is
+        // the latest server-reported value, so this comparison is safe
+        // against mid-walk page shifts that would otherwise let a duplicate-
+        // padded array trip the gate before a newly-registered run was seen.
         if (
           state.headSha &&
           runs.length > 0 &&
@@ -585,10 +587,12 @@ export class PRPollingSource implements AsyncEventSource {
       // Gate on `runs.length > 0`: an empty array is ambiguous (no CI
       // configured vs. runs not yet registered), so "done" isn't meaningful.
       // Also require we have the full set (length >= total_count). The API
-      // caps at per_page=100; `fetchAllCheckRuns` walks pagination AND
-      // returns the most recent `total_count` it observed (not the stale
-      // page-1 value), so this comparison correctly guards against runs
-      // added mid-walk that bumped `total_count` higher than page 1 reported.
+      // caps at per_page=100; `fetchAllCheckRuns` walks pagination, dedupes
+      // runs by id (so page-shift duplicates can't pad the count), AND
+      // returns the latest `total_count` from this tick's 200 responses
+      // (not a stale page-1 snapshot or a stuck monotonic max from a prior
+      // mid-walk inflation). A duplicate-filled buffer can therefore never
+      // trick this gate into firing "CI complete" before every run is seen.
       if (
         state.headSha &&
         runs.length > 0 &&
@@ -634,15 +638,29 @@ export class PRPollingSource implements AsyncEventSource {
    * gates (`runs.length >= total_count`) stuck false forever, stranding agents
    * waiting on "CI complete" / "CI passed".
    *
-   * ## Mid-walk additions
+   * ## Mid-walk churn (additions, removals, dedupe)
    *
-   * `total_count` can grow between pages if GitHub registers new runs while
-   * we're paginating. We re-read `total_count` from every 200 response,
-   * recompute `totalPages` against the maximum we've ever seen, and continue
-   * until we've collected at least that many runs (or hit the safety cap).
-   * The returned `total_count` is the latest observed value, never page 1's
-   * stale snapshot — downstream `runs.length >= total_count` gates can rely
-   * on it.
+   * `total_count` and the per-page set of runs can both shift while we're
+   * paginating: GitHub may register new runs (pushing earlier runs onto a
+   * later page → duplicates across our pages), or runs may be removed/replaced
+   * (a page shrinks). We defend against both:
+   *
+   * - We accumulate runs into a `Map<id, GhCheckRun>` so duplicates collapse.
+   *   The terminal gate downstream (`runs.length >= total_count`) is evaluated
+   *   against the deduped count, not the raw appended count, so a page-shift
+   *   that fills our buffer with duplicates can't trick us into emitting
+   *   "CI complete" before a newly-registered run has actually been seen.
+   *
+   * - We re-read `total_count` from every 200 response and use the LAST seen
+   *   value (not the historical max). The server's `total_count` is its
+   *   current authoritative count at the moment of that page-N fetch; if a
+   *   transient mid-walk inflation later settles back down, taking the max
+   *   would strand the gate forever (`runs.length` can never reach an
+   *   inflated, no-longer-true total).
+   *
+   * The returned `total_count` is the latest observed server value (or, on
+   * a full-304 tick, the previously-committed value carried via cache),
+   * never page 1's stale snapshot.
    *
    * ## ETag caching (multi-page)
    *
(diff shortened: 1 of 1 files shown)


LionSR:
Addressed both review concerns in 10f91f7.

## Concern A (P1): stale `total_count` mid-walk

`fetchAllCheckRuns` now re-reads `total_count` from every 200 response, recomputes `totalPages` against the **maximum** we've ever seen, and keeps walking until we've collected at least that many runs. The returned `total_count` is the latest observed value — never page 1's stale snapshot — so the downstream `runs.length >= total_count` gate in `pollOne` is reliable even when GitHub registers new check runs while we're mid-walk.

Added a runaway guard: `MAX_CHECK_RUNS_PAGES = 50` (5,000 runs). If `total_count` would exceed that, we log a warning and stop. Better to under-report on a degenerate PR than fan out into hundreds of GETs every 30s.

## Concern B: rate-limit risk from dropped ETag

Picked **strategy (a) — per-page ETags with cached page data**. The single-string `state.etags.checkRuns` was replaced with a richer per-subscription cache:

```ts
checkRunsCache?: {
  etagsByPage: Map<number, string>;
  pagesByPage: Map<number, GhCheckRun[]>;
  lastTotalCount: number;
};
```

Each tick now issues conditional `If-None-Match` requests **per page**. A 304 reuses the cached page contents; a 200 refreshes that page only. In steady state (no new push, all runs settled) every page returns 304 and we surface a true end-to-end 304 to the caller — N HEAD-cheap requests instead of N full-page GETs.

Why (a) over (b): even with a smart "page 1 304 means everything's unchanged" composite, the moment a single run on any page transitions state we'd refetch everything. Per-page conditional means we only refetch pages that actually changed, which on a settling matrix build of 250 runs is usually 0–1 pages per tick. Worth the modest extra plumbing.

Other refinements:
- `state.checkRunsCache` is dropped when `headSha` changes (new push), so the post-push tick doesn't waste a round of mismatched `If-None-Match` headers before the cache naturally refreshes.
- Single-page fast path (`<=100` runs) is behaviorally unchanged: page-1 ETag short-circuits the whole call on 304.
- Per-tick caches are rebuilt from scratch (`nextEtags`, `nextPages` committed at end), so any pages that disappear (e.g. page count shrank from 5 to 3) are evicted, not leaked.

## Verification

- `npm run typecheck`: clean (only pre-existing `@openrouter/sdk` and `arxivProcessor.ts:133` errors remain, both unrelated).
- `npm run lint`: 0 warnings on `PRPollingSource.ts`.
- Prettier formatted.

LionSR:
Addressed in 78b89992e by deferring the cache write until the tick has successfully consumed the result (pattern (b) — staged value, no callback).

**What changed**

- `fetchAllCheckRuns` no longer mutates `state.checkRunsCache`. It returns `{ response, stagedCache }`, where `stagedCache` is the new `{ etagsByPage, pagesByPage, lastTotalCount }` value built from the per-page fetches.
- `pollOne` destructures `checksOutcome.response` into `checksRes` (used by the existing diff/seed branches unchanged) and `checksOutcome.stagedCache` into a local. It assigns `state.checkRunsCache = stagedCheckRunsCache` only at the end of the success path — both at the tail of the first-tick seeding branch and at the tail of the diff branch.
- A sibling rejection in the `Promise.all` now throws before either commit point is reached, so the previous cache stays in place and the next tick will re-fetch (no phantom 304, no swallowed transitions).

**Edge cases verified**

- Single-page 304 fast path returns `{ response: { status: 304 } }` with no `stagedCache` — the existing cache is still valid and was never replaced, so there's nothing to commit.
- The `state.headSha !== newHead` reset (line 386) still clears `state.checkRunsCache = undefined` unconditionally, before the parallel fetch. If a sibling then rejects on the fresh SHA, the cache stays empty — page 1 fetches unconditionally next tick (no ETag to send, no phantom 304). Correct.
- All-pages-304 still surfaces a true 304 to the caller; the staged cache (which carries forward the same per-page ETags) is committed on success so subsequent ticks keep getting 304s in steady state.

Type-checks clean (only the pre-existing `@openrouter/sdk` and `arxivProcessor.ts:133` errors remain).

LionSR:
Addressed both paginated check-runs races in `ce222da6d`. Both fixes live in `src/tools/github/PRPollingSource.ts::fetchAllCheckRuns`.

**Concern 1 (Codex P1, line 775) — duplicate-padded `allRuns`**

Replaced the `GhCheckRun[]` accumulator with `Map<id, GhCheckRun>`. Each page's runs are now `runsById.set(r.id, r)`, so when GitHub shifts a run across page boundaries mid-walk we collapse the duplicate instead of double-counting. The downstream terminal gate `runs.length >= total_count` evaluates against the unique run count (`runsById.size`, surfaced as the array length of the values), so a duplicate-filled buffer can no longer trip "CI complete" / "CI passed" before a newly-registered run has actually been observed.

**Concern 2 (Cursor medium, line 791) — `observedTotal` stuck on a stale max**

Dropped the `Math.max`-style monotonic tracker. Now:
- `latestTotal` is set only from this tick's 200 responses (in page order; the LAST 200 wins, since GitHub's `total_count` is its current authoritative value at fetch time).
- 304 pages explicitly return `total: undefined`, never `cache.lastTotalCount` — that fallback was the channel by which a transient prior-tick inflation persisted forever once page 1 started returning 304.
- The walk-length recomputation also adapts to a *shrunken* total, not just a grown one.
- Final reported total resolves in priority order: latest 200 → cached `lastTotalCount` (only when every page was 304, i.e. truly unchanged) → deduped run count (degenerate case).

The seed estimate at the start of the walk still uses `cache.lastTotalCount` as a starting page-count guess when page 1 is 304, but the moment any 200 arrives, `latestTotal` takes over for both walk-extension and the staged cache write — so a settled-down lower value can now take effect.

**Audit of the three locations Cursor flagged**
- `L778-L791` (loop): now uses `latestTotal = result.total` (last-wins) and recomputes `totalPages` accordingly.
- `L736-L737` (init from cache): we no longer seed `observedTotal` from `cache.lastTotalCount`; only the page-walk estimate (`seedTotal`) consults the cache, and only as a starting guess.
- `L801-L802` (stage cache write): `lastTotalCount: reportedTotal` where `reportedTotal` is derived strictly from this tick's responses (or the cached value only when all-304).

Comments at the two consumer sites (`!state.initialized` seed branch, diff-and-emit branch) were also updated to reflect dedupe + last-wins-total semantics.

**Cache-commit deferral preserved** — `stagedCache` is still only assigned to `state.checkRunsCache` at the tail of `pollOne`'s success branches, exactly as `78b89992e` established. No regression there.

**Build/typecheck** — typecheck reports only the two pre-existing errors (`@openrouter/sdk` resolution and `arxivProcessor.ts:133` axios). Lint is clean for the changed file.

PR #3081, opened 2026-04-20
fix: aux-bar persistence, selected-tab contrast, and tool-filter noise

LionSR:
## Summary

Three small user-visible fixes in the TeXRA main view and tool-use flow:

1. **TeXRA panel auto-restores on the auxiliary sidebar.** After #3046 split the `texra` container into two sibling views gated by `when: texra.activated` / `!texra.activated`, VS Code lost the user's saved aux-bar position for `texra.mainView` — at startup the context key is false until the end of `activate()`, so when VS Code restored layout the `mainView` id did not yet exist. Collapse back to a single always-visible view id (`texra.mainView`) and serve the welcome screen from a lightweight `WebviewViewProvider` in the no-workspace path. View-location state now persists across sessions.

2. **Selected stream tab matches VS Code's list selection pattern.** Active-tab styling relied on `color: inherit` cascading into `.tab-meta`'s span/codicon descendants. The meta row rendered dim against the blue selection background ("overshadowed"). Force the selection foreground on every descendant via a `*` selector so title, meta, `.last-active`, `.model`, and codicon glyphs flip together. Destructive-action cue on `.tab-delete:hover` preserved.

3. **Silence per-cycle tool-filter noise.** `resolveTools` emitted yellow warnings every tool-use cycle for intentional filtering outcomes ("excluded: disabled in Settings > Tools", "excluded: external dependency not installed", "not found in registry"). Drop them. Agent YAML typos are already warned once at load time in `agentLoad.ts` / `RemoteAgentLoader.ts`; missing external dependencies are surfaced via `notifyUnavailableTools`. Keeps the "Memory tool not found in registry" warning — genuine bug signal.

## Changes

- `package.json`: drop `texra.welcomeView` + `viewsWelcome` block; remove `when: texra.activated` gate from `texra.mainView`.
- `src/frontend/ui/welcomeView.ts`: rewrite as a `WebviewViewProvider` for `texra.mainView` serving welcome HTML with native `command:` links; still reloads the window when workspace becomes single-folder.
- `src/progressView/frontend/components/StreamTabs.ts`: force selection foreground onto every descendant of `.tab-container.is-active`; preserve delete-hover error cue.
- `src/agent/implementations/flows/tooluse/runToolUseFlow.ts`: drop the three routine filter-warning log calls in `resolveTools`.

## Test plan

- [ ] Drag TeXRA to the auxiliary sidebar, close + reopen VS Code → panel auto-restores on the right.
- [ ] Launch VS Code with no folder open → welcome screen renders with working Open Folder / Clone / Docs links.
- [ ] Open a folder from the welcome screen → window reloads into the full TeXRA UI.
- [ ] Select a stream tab → title, time, model, and category icon all render in the selection foreground (no dim meta row).
- [ ] Hover the close icon on a selected tab → flips to error foreground.
- [ ] Disable a tool in Settings > Tools or start a tool-use agent whose YAML references a tool missing a dependency → no yellow warning bubble per tool-use cycle; genuine typos still warn once at agent load.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1226,33 +1226,13 @@
     "views": {
       "texra": [
         {
-          "type": "tree",
-          "id": "texra.welcomeView",
-          "name": "TeXRA",
-          "contextualTitle": "TeXRA",
-          "when": "!texra.activated"
-        },
-        {
           "type": "webview",
           "id": "texra.mainView",
           "name": "TeXRA",
-          "contextualTitle": "TeXRA",
-          "when": "texra.activated"
+          "contextualTitle": "TeXRA"
         }
       ]
     },
-    "viewsWelcome": [
-      {
-        "view": "texra.welcomeView",
-        "contents": "TeXRA is a multi-agent orchestration system for LaTeX research. An orchestrator coordinates specialized agents that derive results, verify proofs, run symbolic computation, generate figures, and assemble manuscripts — every step auditable, every citation grounded.\nOpen a folder to begin. The TeXRA tour and commands become available once a workspace is open.\n[Open Folder](command:workbench.action.files.openFolder)\n[Clone Repository](command:git.clone)\nWant to learn more first? [Read the docs](https://texra.ai).",
-        "when": "workbenchState == empty"
-      },
-      {
-        "view": "texra.welcomeView",
-        "contents": "TeXRA supports one folder at a time. Open a single folder to continue.\n[Open Folder](command:workbench.action.files.openFolder)",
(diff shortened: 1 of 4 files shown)


LionSR:
Follow-up: tightened the three WHY comment blocks touched by this PR (welcome-view docstring, tool-filter warning rationale, and active-tab styling) while preserving all the non-obvious rationale. No behavior changes; `compile:fast` and `typecheck` are clean (modulo the pre-existing `arxivProcessor.ts:133` axios error).

PR #3094, opened 2026-04-21
fix(latex): mirror figure deps into run-storage workspace

LionSR:
## Summary

- Workflow-mode agents with `storageMode = taskRunStorage` (and all subagents, which force run storage) mimic the workspace under `executions/<id>/` via symlinks. Figures referenced via `\includegraphics` weren't being mirrored, so `latexmk` inside the run dir failed when any figure was referenced.
- `extractFigurePathsFromLatex` took `path.dirname(file.absolutePath)` as the search root — for a symlinked `.tex` that's the run-storage dir, where the figures don't exist, so extraction returned empty. The same bug was in `mirrorFigureDependencies`: even a correct figure list produced run-storage absolute paths, which `pathToLocation` classified as external and `mirrorWorkspaceFile` then skipped.
- `processOutputFiles` (round 1+) disabled both figure and `\input` mirroring, so figures newly referenced by agent edits weren't available at compile time either.

## Changes

- New `resolveLatexDir()` helper in `src/latex/latexParsingUtils.ts` that `fs.realpath`-resolves a tex file's directory (with a graceful fallback for missing files). Mirrors the pattern already used by `extractFileDependencies`.
- `extractFigurePathsFromLatex`, `mirrorFigureDependencies`, and the `extractFiguresFromFiles` inner loop all use it — figure paths now resolve against the real workspace dir when the tex is symlinked into run storage.
- New private `mirrorFiguresForFiles()` mirrors figure deps without touching the model's vision context (so output files don't re-send figures as images each round).
- Replaced the `includeFigureExtraction` / `mirrorFigureDeps` boolean pair with a single `figureMode: 'extract' | 'mirror' | 'none'` to make the mutual exclusivity explicit.
- `processOutputFiles` now runs `figureMode: 'mirror'` + `mirrorFileDependencies: true` so newly-referenced figures and `\input` targets are symlinked before PDF compilation.

## Test plan

- [ ] Set `texra.agentOutputs.storageMode = taskRunStorage` and run a workflow agent on a doc that references `\includegraphics{figures/plot.pdf}` and `\input{subfile.tex}`. Confirm `executions/<id>/figures/plot.pdf` and `…/subfile.tex` are symlinks after round 0, and `executions/<id>/build/<paper>.pdf` compiles.
- [ ] Round 1+: prompt the agent to introduce a new `\includegraphics` reference. Confirm the figure is symlinked and the output PDF compiles.
- [ ] Subagent: trigger a subagent workflow (always forces run storage). Same checks as above.
- [ ] Regression: with default `storageMode = workspace`, confirm behavior unchanged (realpath is a no-op; `mirrorWorkspaceFile` remains idempotent).
- [ ] `npm run typecheck` and `npm run compile:fast` pass.

---

> [!NOTE]
> **Medium Risk**
> Changes how LaTeX figure/dependency paths are resolved and mirrored when files are symlinked into run storage, which can affect PDF compilation behavior across workflow rounds. Also adjusts Progress view child-stream auto expand/collapse logic, potentially impacting sidebar UX state handling.
> 
> **Overview**
> Fixes LaTeX compilation in `taskRunStorage` mode by resolving `.tex` directories via realpath (new `resolveLatexDir`) so `\\includegraphics` figure paths and mirroring target the underlying workspace files instead of run-storage symlink paths.
> 
> Updates output-round processing to *mirror* newly referenced figures and `\\input`/include/bib dependencies into run storage without re-adding figures to the vision context, and replaces the previous figure booleans with an explicit `figureMode: 'extract' | 'mirror' | 'none'`.
> 
> Refactors Progress view utilities by centralizing `setsEqual` and revises `StreamTabs` parent expansion to be derived from stream state (active/unknown/finished) with a per-parent user override, reducing unnecessary re-renders and making expand/collapse behavior more predictable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ddfbe9a8b8ebef41fb524395835cf0fe44d05cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/latex/LatexMediaManager.ts
+++ b/src/latex/LatexMediaManager.ts
@@ -204,8 +204,13 @@ export class LatexMediaManager {
       return;
     }
 
+    const texFiles = files.filter((file) =>
+      hasExtension(file.absolutePath, '.tex'),
+    );
+    if (texFiles.length === 0) return;
+
     await pMap(
-      files,
+      texFiles,
       async (file) => {
         try {
           const figures = await extractFigurePathsFromLatex(file);
--- a/src/latex/LatexMediaManager.ts
+++ b/src/latex/LatexMediaManager.ts
@@ -18,6 +18,7 @@ import { hasExtension } from '@utils/core/pathCore';
 // Local file imports
 import { extractLatexFileDependencies } from './extractFileDependencies';
 import { extractFigurePathsFromLatex } from './extractFigure';
+import { resolveLatexDir } from './latexParsingUtils';
 import { tikzPictureManager } from './TikzPictureManager';
 import { compileLatex2Pdf } from './texTools';
 
@@ -47,7 +48,10 @@ export class LatexMediaManager {
       return;
     }
 
-    const baseDir = path.dirname(latexFile.absolutePath);
+    // Resolve against the workspace directory so figure paths from a
+    // run-storage symlink map back to real workspace files (otherwise
+    // mirrorWorkspaceFile would classify them as external and skip).
+    const baseDir = await resolveLatexDir(latexFile.absolutePath);
     const targetLocations = new Set<FileLocation>();
     for (const relative of figures) {
       const trimmed = relative?.trim();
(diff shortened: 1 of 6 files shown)


LionSR:
Addressed Copilot review:
- **`.tex` filter in `mirrorFiguresForFiles`** — fixed in 9f3737b (avoid regex-scanning non-tex outputs like PDFs in later rounds).
- **Figure path dedupe** — fixed in c7ba5f5. `Set<FileLocation>` didn't dedupe since `pathToLocation()` returns a fresh object per call; now deduped by absolute-path string before wrapping.

PR #3194, opened 2026-04-27
fix: clarify delegation depth gating

LionSR:
## Summary

- evaluate delegation depth through a shared policy helper, snapshotting only the max-depth config at tool-use flow setup so prompt pruning and runtime enforcement derive from the same policy without carrying a second depth value
- tailor unknown-lineage remediation and diagnostics so unknown depth is reported as `unknown` instead of leaking the internal sentinel
- share external tool availability probe results across callers so duplicate GitHub prerequisite checks do not run during one refresh

## Validation

- `git diff --check`
- npm-based checks were not run because this worktree environment has no `npm` binary available

---

> [!NOTE]
> **Medium Risk**
> Touches multi-agent delegation gating and resume behavior; incorrect policy snapshotting or sentinel handling could wrongly allow/block delegation. External tool availability probing changes could affect tool enable/disable reporting during refreshes.
> 
> **Overview**
> Clarifies and unifies nested-delegation depth enforcement by introducing `evaluateDelegationGate`, adding an explicit `UNKNOWN_DELEGATION_DEPTH` sentinel, and returning structured block reasons (`max_depth_reached` vs `unknown_depth`). Tool-use flow setup now snapshots `delegationConfig` and threads it through `ToolFileInteractionContext` so tool filtering and `DelegationTools` runtime checks use the *same* policy decision, with improved error text and diagnostics when delegation is blocked.
> 
> External tool availability checks now support an optional shared `probe` result that is reused across `check`, `statusLabel`, and `detailCheck` callbacks (notably for GitHub PR prerequisites), reducing duplicated async work during a single refresh.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce2d5f58848842a0f4504af79a93bb7253ef6a7b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/core/flows/ToolUseCycleFlow.ts
+++ b/src/agent/core/flows/ToolUseCycleFlow.ts
@@ -795,6 +795,7 @@ class ToolUseDispatchNode<C> extends Node<
           agentName: options.agentName,
           workingDirectory: options.workingDirectory,
           delegationDepth: options.delegationDepth,
+          delegationConfig: options.delegationConfig,
           onExecutionReady,
           onToolOutput,
         },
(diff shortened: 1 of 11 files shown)


LionSR:
why are there two depth numbers!

We need single source of truth

LionSR:
Worried about the single source of truth here 

May need a followup

Issue #18, opened 2024-11-15
support google models via openai apis


LionSR:
https://developers.googleblog.com/en/gemini-is-now-accessible-from-the-openai-library/

PR #17, opened 2024-11-13
add a editor page for editing prompts loaded from xml

LionSR:

PR #3255, opened 2026-05-02
docs: add PRD for TeXRA Electron app

LionSR:
Initial draft of the product requirements document for shipping TeXRA as
a standalone Electron desktop app alongside the existing VS Code extension.
Covers tech stack picks (electron-vite + electron-builder + Lit reuse),
code-reuse boundary, platform-impl scope, migration phases, and risks.
Picks are grounded in a parallel codebase scout; revisions to follow as
remaining tooling-research scouts complete.

https://claude.ai/code/session_0126E5jgiizC3gz6MwS2H2hS

---

> [!NOTE]
> **Low Risk**
> Documentation-only change adding a draft PRD; no runtime code or build artifacts are modified.
> 
> **Overview**
> Adds a new draft product requirements document `prds/prd-electron-app.md` describing a proposed Electron desktop app alongside the existing VS Code extension.
> 
> The PRD outlines planned architecture and repo split (`core`/`extension`/`desktop`), code-reuse boundaries, IPC/security model, platform adapter scope, migration phases, and key risks/mitigations (packaging, updates, auth, diff/Monaco, secrets).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 765f03ff2729af098e23d8a735be94581d40c2db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+| **`asar` packing misses native binaries** (Codex SDK, Monaco workers, codicon font) | Medium     | High     | `electron-builder.yml` `asarUnpack` glob covers `@openai/codex-*/**`, Monaco worker chunks, the codicon TTF. Verified per-platform in CI.                                                                                                                                |
+| **Codex SDK platform-binary mismatch** under mac-universal builds                   | Low        | High     | Confirm `@openai/codex-sdk`'s install-time platform detection works under `electron-builder` lipo'd builds. Test ARM64 + x64 explicitly.                                                                                                                                 |
+| **`safeStorage` Linux fallback to "basic" mode** = secrets with a hardcoded key     | Medium     | High     | Detect via `getSelectedStorageBackend()`. Surface a one-time warning. Document keyring install. Don't silently degrade.                                                                                                                                                  |
```
LionSR: The fallback name regressed here: the risk row still says safeStorage falls back to `"basic"` mode, while the corrected API value elsewhere in the PRD is `basic_text`. Since this table is likely what implementers scan during packaging/auth hardening, leaving the stale sentinel here can reintroduce the same bug the earlier sections fixed. Please make this row use `basic_text` too.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+| **Corp proxy / SSL inspection** breaks model API or auto-update                     | Medium     | Medium   | Respect `HTTP_PROXY` / `HTTPS_PROXY`. Electron's `net` module does by default; verify SDKs do too. Document for IT admins.                                                                                                                                               |
+
+## 12. Success criteria
+
+- v1 ships signed installers: macOS (Universal DMG + ZIP), Windows (NSIS x64 + arm64), Linux (AppImage + deb).
+- A user can: sign in (or set API key), open a project folder, pick an agent and model, execute, view progress, approve tool edits via the new diff component, see final output. End-to-end without VS Code installed.
+- No regression in the VS Code extension. Same `pnpm --filter extension build` produces a working VSIX from the same `packages/core/`.
+- Total Electron-side new code (in `packages/desktop/`) under ~3,000 LOC. Diff in `packages/core/` and `packages/extension/` (excluding the monorepo move itself) under ~500 LOC. Anything more means an abstraction leak.
```
LionSR: This success criterion no longer matches the plan above. It caps non-move `core`/`extension` changes at ~500 LOC, but §14.1 now budgets ~1,355 net-new plus ~1,941 modified LOC there, and §9 promotes several of those pre-refactorings as recommended/required. As written, the project would fail its own success criteria even if implemented exactly as the PRD prescribes. Either update the criterion to match the current budget or separate a hard v1 gate from optional pre-refactoring totals.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+
+#### LOC by component (`packages/desktop/src/main/` + `renderer/`)
+
+| Component                                                                         | New LOC          | Notes                                                                                                 |
+| --------------------------------------------------------------------------------- | ---------------- | ----------------------------------------------------------------------------------------------------- |
+| `main/index.ts` (lifecycle, single-instance lock, fix-path top-of-file)           | ~120             | Cold-start bookkeeping, single-instance, `fix-path()` cached.                                         |
+| `main/platform/` (8 files: config, state×2, log, fs, workspace, storage, secrets) | 250–400          | Mirrors VS Code impls in `src/frontend/vscode/` (~300 LOC today).                                     |
+| `main/ipc/` (typed channels + dispatcher adapter)                                 | ~150             | Wraps Zod schemas from `core/shared/`; one new transport adapter file.                                |
```
LionSR: This LOC table still reflects the pre-collapse layout. §7.1 says the `ipc/` directory was collapsed into a single `main/ipc.ts` and there is no extra transport-adapter file in `core`, but this row uses `main/ipc/` and describes a separate transport adapter. The same table also still has `renderer/components/TexraDiffView.ts` even though §7.1 says `renderer/components/` was dropped. Please align the component budget paths with the canonical layout so implementers do not create the directories the architecture just removed.

LionSR (commented):
Fresh pass on the current head. A few remaining PRD consistency issues are still actionable.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,927 @@
...
+| `vscode.window.showInformationMessage` (et al.)                             | `dialog.showMessageBox` from main; in-app toast for non-blocking                                                                                                                 |
+| Status bar                                                                  | Footer in main window (already mocked in webview frontend)                                                                                                                       |
+| Walkthrough (`getting-started.md`)                                          | First-run modal rendering the same markdown                                                                                                                                      |
+| Command palette                                                             | Lit palette (Cmd/Ctrl-Shift-P) over the existing `src/commands.ts` registry; `globalShortcut` only when window focused                                                           |
```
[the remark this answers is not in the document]
LionSR: This is still present in two other current-head sites too: the stack table still says “custom Lit palette over existing `src/commands.ts` registry” at line 160, and Phase 3 repeats “Lit command palette over the existing `src/commands.ts` registry” at line 725. Since `src/commands.ts` imports `vscode` and registers extension/webview providers, all three references should be replaced with a host-agnostic command manifest/catalog.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,930 @@
...
+
+`pnpm` over npm: 2–3× installs, `workspace:*` protocol pins internal versions cleanly, `--filter desktop` and `--filter extension` for targeted builds. Turborepo's task pipeline is overkill for three packages — we run two parallel build commands and that's enough. Project references via `composite: true` in TS keep type-checking incremental.
+
+This split is the **largest mechanical change** in the project. Do it before the desktop host has any code, not after.
+
+### 6.5 Why roll-own deep-link
+
+Existing `src/auth/UriHandler.ts` already handles the `vscode://vscode.texra/auth-callback?code=...&state=...` parse-and-dispatch logic. The Electron equivalent registers `texra://` and reuses the same parse/dispatch on the URL. Three platform-specific gotchas worth budgeting:
+
+- **macOS:** `app.on('open-url')` fires for already-running app; for cold-start the URL arrives in `process.argv`. Capture synchronously at module top.
```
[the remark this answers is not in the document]
LionSR: The same platform split should be reflected in the risk table as well: line 801 still frames `process.argv` cold-start handling as a three-OS concern including macOS. Electron’s `open-url` guidance for macOS is to register the listener very early, before `ready`, because URLs that launch the app are delivered through that event; `process.argv` is the Windows/Linux cold-start path. Please update that row so it does not steer macOS implementation toward argv parsing.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+| Total TS/TSX files in `src/`                                         | 853                                             |
+| Files importing `vscode`                                             | 106 (12.4%)                                     |
+| `Platform` interface LOC                                             | ~470                                            |
+| Existing VS Code platform impl LOC                                   | ~303 (6 files)                                  |
+| Estimated Electron platform impl LOC                                 | ~250–400 (8 files)                              |
+| Lines in `nativeToolEditApproval.ts` (the diff blocker)              | 439                                             |
+| Estimated `<texra-diff-view>` Lit wrapper LOC (Monaco hosted inside) | ~200–400                                        |
+| Lines in `SupabaseAuthProvider.ts`                                   | 943                                             |
+| Estimated reuse from `SupabaseAuthProvider.ts`                       | ~80%                                            |
```
LionSR: This appendix estimate is stale after the §9 #14 pivot. The plan now says to extract `SupabaseSession` so there is “no 80% reuse estimation” and just two thin host wrappers, but the appendix still reports “Estimated reuse from `SupabaseAuthProvider.ts`: ~80%.” There are also earlier remnants at lines 53 and 453. Please replace the percentage framing with the concrete extraction plan, otherwise readers get two different auth-port strategies from the same PRD.

LionSR (commented):
One more residual stale estimate from the auth refactor pivot.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+| Token rename (`--vscode-*` → `--texra-*` mapping layer)                              | 450    | 1.5%  |
+| API swap (`postMessage` wrapper transport)                                           | 490    | 1.6%  |
+| Reimplementation (terminal font init, markdown error color, HTML token substitution) | 141    | 0.5%  |
+
+**The minimum Electron-side changeset to mount all three frontends:**
+
+- 1 file modified: `src/shared/vscode.ts` (~45 LOC) — feature-detect Electron, route through `window.electron` instead of `acquireVsCodeApi`. Component code unchanged.
+- 3 new files (~230 LOC total) in `desktop/src/`: `main/ipc.ts`, `preload/index.ts`, `renderer/main.ts`. Window creation is inline in `main/index.ts` (per §7.1).
+- 1 new file: `desktop/src/renderer/themeTokens.css` defining the 53 `--vscode-*` token values for light/dark/high-contrast themes (the existing fallbacks document the defaults).
+- 0 component template changes, 0 signal/context-architecture changes, 0 changes to message dispatchers.
```
LionSR, line 109: This is the biggest underestimation in the PRD: the Lit component templates may be reusable, but the host-side message handlers are not a no-change layer. `MainViewMessageHandler` (~440 LOC), `ProgressViewMessageHandler` (~1,366 LOC), and `SettingsViewMessageHandler` (~1,745 LOC) all import `vscode` and contain the real handlers for file selection, command routing, settings writes, auth/API-key dialogs, open-external, terminal creation, and push updates. If desktop does not import `common/webview` / `frontend` (§7.2), the plan needs an explicit host-neutral controller/service extraction for these handlers; otherwise Phase 2/3 is missing several thousand LOC of porting work behind the “bridge-and-bootstrap” wording.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+
+We do **not** use `extraResources` for v1 — it places files in `process.resourcesPath` (a third pathing convention) with no benefit for our asset shapes. Reserved for the future only if we ever ship user-patchable assets meant to be edited out-of-band.
+
+## 9. Pre-refactorings — land these in the extension first
+
+These changes are safe to ship in the VS Code extension today. Each one shrinks the Electron port's blast radius, none of them require Electron to land. Tier 1 are high-leverage; pick those first.
+
+### Tier 1 — high leverage, low risk
+
+**1. Move `src/utils/config/configUtils.ts` behind `ConfigProvider`.** _(~126 LOC refactored, ~40 LOC added to `ConfigProvider`.)_
```
LionSR: Moving `configUtils.ts` behind `ConfigProvider.get()` is not enough for the settings flow the PRD promises. The current config utility surface also has `updateConfig`, `inspectConfig`, `isConfigExplicitlySet`, and `watchConfig`, with consumers in the settings view, setup flow, file lister, and main view. The current `ConfigProvider` interface only exposes `get<T>()`, so Phase 3’s “settings round-trip through conf” cannot be implemented through the proposed abstraction. Please make the pre-refactor include write/inspect/watch semantics, or call out a separate settings-store adapter.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+
+**14. Extract host-agnostic OAuth state machine from `SupabaseAuthProvider`.** _(~750 LOC moved into a new `core/auth/SupabaseSession.ts`; ~190 LOC of VS Code-specific glue stays in `extension/`.)_
+Today `src/auth/SupabaseAuthProvider.ts` (943 LOC) interleaves OAuth state-machine logic with VS Code-specific surfaces (`vscode.AuthenticationProvider`, `vscode.UriHandler`, `context.secrets`). Pull the state machine — sign-in initiation, PKCE handling, token storage, refresh, GitHub token exchange — into a host-agnostic `SupabaseSession` class in `core/`. Inject the host-specific surfaces as constructor dependencies: `(secrets: PlatformSecrets, openExternal: (url) => void, onAuthCallback: Listener<AuthCallbackResult>)`. The VS Code-side wrapper becomes ~190 LOC of glue: register `AuthenticationProvider`, register `UriHandler`, wire its callback into `SupabaseSession`. **Why now:** the Electron auth surface becomes ~50 LOC of glue against the same `SupabaseSession` class — no parallel implementation, no "80% reuse" estimation; **it's literally one class with two thin host wrappers**. Subsumes #5 (the URL parser is part of the state machine).
```
LionSR: The auth extraction needs to include `SupabaseClient`, not just `SupabaseAuthProvider`. Today `SupabaseClient.ts` imports `vscode`, stores a `vscode.ExtensionContext`, reads session tokens from `context.secrets`, and `isReady()` requires `vscodeProviderRegistered` to be true. Remote-agent loading calls `SupabaseClient.getSessionTokens()`, so an Electron `SupabaseSession` wrapper alone will still leave tier checks / remote agents coupled to VS Code readiness and secret storage. Please add `SupabaseClient` to the host-agnostic refactor scope or define a token-provider boundary it consumes.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+
+**Streaming large payloads** (Phase 5+): for diffs >1MB or transcripts that risk hitting Electron's IPC size limit (~100MB), bypass the push channel and use a `MessageChannelMain` / `MessagePort` between main and renderer. Streamed via the same Zod schemas, just on a different transport.
+
+Net change to handler code in `core/`: zero. New code is the two preload functions (`invoke`, `on`) and the two main-process listeners.
+
+### 7.5 Process model
+
+- **Main process** — app lifecycle, window mgmt, native menu, auto-update, protocol handler, platform impls.
+- **Renderer (one per window)** — Lit UI; `nodeIntegration: false`, `contextIsolation: true`, `sandbox: true`. Talks to main via preload bridge.
+- **Utility process** (Phase 4+) — long agent runs spawned from main on `texra:execute`, streaming progress events back via `MessagePort`. Crashes recoverable; renderer survives. Phase 1–3 runs agents in main for simplicity.
```
LionSR: Running agents in a utility process is not just a spawn detail. The current agent/runtime/tool code emits progress through an in-memory singleton `ProgressEventBus`, while the progress view subscribes in the extension host process. If Electron moves long agent runs into a utility process, those `bus.emit(...)` calls will not reach the main process or renderer unless every progress/approval/log event is bridged over `MessagePort` and the utility process initializes its own `platform()` services. The PRD should budget and specify that cross-process event/state bridge, or keep v1 execution in the main process.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,972 @@
...
+- **Exit criteria:** Signed installers from CI on all three platforms. Auto-update from `v0.0.1 → v0.0.2` works.
+
+### Phase 7 — Beta, polish, docs (2 weeks)
+
+- First-run walkthrough modal.
+- Sentry Electron SDK opt-in, native crashes only, `tracesSampleRate: 0`. `beforeSend` strips file paths outside workspace root.
+- Telemetry parity with extension (or explicit opt-out path).
+- In-app log viewer pane.
+- Documentation site updates; new download page.
+- Migration doc: how to import API keys / settings from the VS Code extension's `globalState`.
```
LionSR: This migration path is misleading for credentials. API keys, GitHub tokens, and the Supabase session are stored in VS Code `SecretStorage` (`SecretManager` / `context.secrets`), not in `globalState`; an external Electron app cannot read those secrets from the VS Code globalState JSON. The plan should either require an extension-side export handoff, prompt users to re-enter/re-authenticate secrets, or explicitly scope credentials out of the importer. Otherwise users will expect API-key migration that the desktop app cannot perform.

LionSR (commented):
Deeper implementation-risk pass. These are the areas I’d expect to cause schedule or architecture trouble if the PRD is used as the implementation source of truth.

LionSR (commented):
Deeper pass on packaged-runtime risks. These are the remaining places where the PRD can read cleanly but still fail once implemented against the current TeXRA code paths.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,986 @@
...
+
+### 7.3 Platform impls (Electron)
+
+Eight files, ~250–400 LOC total. Each mirrors an existing VS Code impl in `src/frontend/vscode/`.
+
+| Interface                | VS Code (today)                                             | Electron                                                                          |
+| ------------------------ | ----------------------------------------------------------- | --------------------------------------------------------------------------------- |
+| `ConfigProvider`         | `vscode.workspace.getConfiguration` w/ 3-namespace fallback | `conf` instance + Zod schema mirroring `package.json` `contributes.configuration` |
+| `StateStore` (global)    | `ExtensionContext.globalState`                              | `conf` (file: `state.global.json`) under `app.getPath('userData')`                |
+| `StateStore` (workspace) | `ExtensionContext.workspaceState`                           | `conf` per-project: `<project>/.texra/state.json`                                 |
```
LionSR: This default changes the semantics of workspace state in a way that can hurt users. VS Code `workspaceState` is app-private Memento storage, but `<project>/.texra/state.json` writes TeXRA runtime state into the user's repository. The current workspace keys include agent history, stream logs/tabs, task states, output-file metadata, run instructions, enabled agents, Git author settings, and migration markers; writing those under the project will dirty repos, fail on read-only/shared folders, and risk committing local run history unless every project ignores `.texra/`. Use `app.getPath('userData')` keyed by a stable workspace hash for the default, and reserve project-local state for an explicit settings-as-code/export feature.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,986 @@
...
+- `chokidar` — pure JS in v4.
+
+**Runtime closure has no `.node` files.** Verified by walking `package-lock.json` — every package containing native bindings (`@rolldown/binding-*`, `lightningcss-*-*`) is marked `dev: true, optional: true`, so they're Vite/Rolldown build-time only and never ship in the production bundle. To prevent regression, Phase 6 packaging adds a CI check that fails if `electron-builder`'s emitted asar contains any `.node` file we didn't explicitly allow-list (Codex SDK is the only exception, and ships unpacked from asar via `asarUnpack` — see §8.4).
+
+A build-time guard plugin (custom esbuild plugin: any `import 'vscode'` in `packages/desktop/` or `packages/core/` fails the build) prevents leakage as the codebase grows.
+
+### 8.4 Resources
+
+`resources/agents/`, `resources/walkthroughs/`, `resources/logo-128x128.png`, replacement rules — bundled **inside `app.asar`** via `electron-builder`'s `files` glob (so `app.getAppPath() + '/resources/...'` resolves correctly in dev and after packaging). Existing code reads via `platform().fs.readFile`, so no path code changes.
```
LionSR, line 627: The resource-loading plan is missing the actual default-agent bootstrap. Current TeXRA does not just read `resources/agents` through `platform().fs.readFile`: `copyDefaultAgents(context)` copies `context.extensionPath/resources/{agents,tool_use_agents}` into `GlobalStorageFS` on version changes/missing files, and `AgentDirectoryManager` then serves built-in/custom directories from global storage while also importing `vscode`. If desktop only bundles these YAMLs inside `app.asar`, built-in agent discovery/update can be missing or stale. The PRD should include a host-agnostic `AgentDirectories`/resource-sync adapter that copies bundled agents from `app.getAppPath()` into `userData` on version bump, mirroring the extension path.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,986 @@
...
+
+**Two distinct asset locations to keep straight in `electron-builder.yml`:**
+
+- **`files`** (default — bundled inside `app.asar`): YAML agent definitions, walkthrough markdown, logos, replacement rules, every TS/JS we ship. Path: `app.getAppPath()` + relative.
+- **`asarUnpack`** (extracted from asar at install time, lives under `process.resourcesPath/app.asar.unpacked/...`): native binaries that subprocess-spawn won't tolerate inside an asar. Currently just `**/node_modules/@openai/codex-*/**`. Resolve via `process.resourcesPath + '/app.asar.unpacked/...'`.
```
LionSR, line 632: `asarUnpack` by itself is not enough for the current Codex path. TeXRA constructs the SDK with `codexPathOverride: findCodexBinaryPath()`, and that helper resolves platform packages via `createRequire(path.join(__dirname, '..'))`, then falls back to global npm and PATH. In a packaged Electron app, unpacking `@openai/codex-*` does not automatically make this helper return a spawnable `process.resourcesPath/app.asar.unpacked/...` binary; it can resolve an `app.asar` path, miss the unpacked tree, or silently fall back to an older global CLI. The PRD should choose one packaged strategy explicitly: either use the `@openai/codex/bin/codex.js` shim and unpack its vendor/platform deps, or teach `BinaryResolver`/`findCodexBinaryPath()` to resolve from `app.asar.unpacked` before global/PATH fallback and verify that exact path in the Phase 6 harness.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1085 @@
...
+
+### 4.1 Why this is tractable — measured
+
+A six-front parallel scout of the codebase confirmed:
+
+- **853 TypeScript files** in `src/`. Only **106** import `vscode`. The remainder access host services through `platform()` from `@platform`.
+- **The `Platform` interface is tiny.** Total surface across `config`, `state`, `log`, `filesystem`, `workspace`, `storage`, `secrets`: ~470 LOC of interfaces. The existing VS Code implementation is ~300 LOC across 6 files (`src/frontend/vscode/`). The Electron-side mirror is ~200–300 LOC of glue.
+- **Webviews are pure Lit.** No React/Vue/Svelte. All three (`webview`, `progressView`, `settingsView`) extend `LitElement`, use `@lit-labs/signals`, communicate via Zod-validated message schemas in `src/shared/`. The transport wrapper at `src/shared/vscode.ts` already includes a fallback API for non-webview contexts — the Electron transport is essentially a one-file swap.
+- **Agent runtime is `vscode`-free.** All ~141 files in `src/agent/` confirmed. Streaming uses callbacks; cancellation uses standard `AbortController`; persistence is filesystem-based JSON validated by Zod. Drops into an Electron main or utility process unchanged.
```
LionSR: This sentence is now stale after the §7.5 update. The PRD now explicitly says v1 does **not** run agents in a utility process because the current `ProgressEventBus` is an in-memory singleton and would need a MessagePort/platform bridge. Keeping "Drops into an Electron main or utility process unchanged" here reintroduces the exact implementation trap later sections fixed. Please narrow this to "main process at v1" and point utility-process execution to §13.1.

LionSR (commented):
Fresh pass on current head `8e50af57`. Most earlier implementation-risk issues were incorporated, but a few stale cross-references and budget numbers remain.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1085 @@
...
+| Workspace (`workspace.fs`, `getConfiguration`, `workspaceFolders`)    | 4–5   | 20   | small            | Already wrapped — Electron impls of `FileSystemProvider`/`WorkspaceProvider`                                          |
+| Editor (`TextDocument`, `Range`, `showTextDocument`, `vscode.diff`)   | 10+   | 56+  | **medium-large** | Lit `<texra-diff-view>` wrapping Monaco's diff editor (lazy-loaded); file preview defers to OS via `shell.openPath()` |
+| Commands (`registerCommand`, `executeCommand`)                        | 56    | 152+ | medium           | Custom command registry + IPC dispatch                                                                                |
+| Webviews (`WebviewView`, `WebviewPanel`, `asWebviewUri`)              | 14+   | 37   | small–medium     | `BrowserWindow` + `contextBridge`                                                                                     |
+| Auth/Secrets (`authentication`, `SecretStorage`, `UriHandler`)        | 20    | 15+  | **large**        | `safeStorage` + custom protocol handler; reuse 80% of `SupabaseAuthProvider`                                          |
```
LionSR: This coupling-inventory row still uses the old "reuse 80% of `SupabaseAuthProvider`" framing, but §9 #14 now correctly scopes the auth work as extracting both `SupabaseSession` and `SupabaseClient` with no percentage-based reuse estimate. There is another stale copy at line 453. Please replace both with the concrete extraction plan; otherwise readers still get two different auth-port strategies from the same PRD.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1085 @@
...
+**One transport file.** The renderer-side transport is the preload's `contextBridge`-exposed API; no separate file in `core/`. The earlier "transport adapter" was a phantom file — the bridge IS the transport.
+
+```
+RPC:   renderer ─ipcRenderer.invoke('texra:rpc', msg)→ preload bridge → ipcMain.handle('texra:rpc') → core dispatcher → @agent/*
+push:  @agent/* → eventBus → main process listener → webContents.send('texra:push', msg) → preload bridge → ipcRenderer.on('texra:push') → renderer dispatcher
+```
+
+**Streaming large payloads** (Phase 5+): for diffs >1MB or transcripts that risk hitting Electron's IPC size limit (~100MB), bypass the push channel and use a `MessageChannelMain` / `MessagePort` between main and renderer. Streamed via the same Zod schemas, just on a different transport.
+
+Net change to handler code in `core/`: zero. New code is the two preload functions (`invoke`, `on`) and the two main-process listeners.
```
LionSR: This no-handler-change claim no longer matches the updated plan. §9 #18 now says the three host-side message handlers total ~3,551 LOC and must be split into host-neutral controllers before Phase 2/3 can work. That means the dispatcher/schema shape can stay, but handler/controller code does change materially. Please update this paragraph so it does not promise "Net change to handler code in `core/`: zero" while the later pre-refactoring section makes #18 a gate.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1085 @@
...
+
+## 12. Success criteria
+
+- v1 ships signed installers: macOS (Universal DMG + ZIP), Windows (NSIS x64 + arm64), Linux (AppImage + deb).
+- A user can: sign in (or set API key), open a project folder, pick an agent and model, execute, view progress, approve tool edits via the new diff component, see final output. End-to-end without VS Code installed.
+- No regression in the VS Code extension. Same `pnpm --filter extension build` produces a working VSIX from the same `packages/core/`.
+- Total **net-new** code in `packages/desktop/` under ~3,000 LOC (the v1 hard gate; per §14.1 we sit at 2,180–2,900). The pre-refactorings in §9 land in `packages/core/` and `packages/extension/` and are budgeted separately at ~1,355 net-new + ~1,941 modified LOC; they are _prep work_, not v1 surface. Going over the ~3,000-LOC desktop cap means an abstraction leak — that's the gate.
```
LionSR: These pre-refactoring budget numbers are stale after the latest additions. §14.1 now budgets `packages/core` + `extension` at ~1,995 net-new and ~5,901 modified LOC, but the success criterion still says ~1,355 net-new + ~1,941 modified LOC. As written, the PRD's own success criteria understate the required prep work by several thousand modified lines. Please align this line with the current §14.1 totals.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1085 @@
...
+| `renderer/themeTokens.css`                                                          | ~150             | 53 tokens × 3 themes (light/dark/HC).                                                                                                   |
+| `renderer/index.html`                                                               | ~30              | Single-window shell.                                                                                                                    |
+| Beta polish (walkthrough modal, log-viewer pane, migration importer)                | ~370             | Phase 7 deliverables.                                                                                                                   |
+| **Subtotal**                                                                        | **~2,500–3,000** |                                                                                                                                         |
```
LionSR: The component subtotal still does not add up to the listed rows. Summing the component table gives 1,990–2,340 LOC (120 + 250–400 + 150 + 100 + 80 + 80 + 30 + 80 + 50 + 120 + 80 + 100 + 200–400 + 150 + 30 + 370), not 2,500–3,000. Either correct the subtotal or add the missing components that account for the extra ~510–660 LOC.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1087 @@
...
+    scope?: 'global' | 'workspace',
+  ): Promise<void>;
+  inspect<T>(key: string): {
+    default?: T;
+    global?: T;
+    workspace?: T;
+    effective?: T;
+  };
+  isExplicitlySet(key: string): boolean;
+  watch(key: string | RegExp, listener: () => void): Disposable;
```
LionSR, line 667: This interface needs to name a host-neutral disposable contract before it can live in core. Right now the only `Disposable` types near this layer are `vscode.Disposable` returns such as `configUtils.ts:114` and webview/provider disposables, so copying this signature into `packages/core` either leaks a VS Code type or relies on an implicit structural type that each adapter guesses at. Define something like `interface Disposable { dispose(): void }` in the platform/core layer and use it for both `ConfigProvider.watch()` and `WorkspaceProvider.watch()`; then the VS Code adapter can return `vscode.Disposable` structurally and the Electron adapters can wrap `conf`/`chokidar` unsubscribe functions.

LionSR (commented):
Deeper architecture pass on the latest head. Claude addressed several earlier concrete contradictions, but there are still a few boundary issues that will matter if this is meant to become a clean implementation plan rather than just a feasibility doc.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1087 @@
...
+
+extension/handlers/MainViewMessageHandler.ts  (~120 LOC, parses msg → calls controller → wraps reply)
+extension/handlers/ProgressViewMessageHandler.ts  (~250 LOC)
+extension/handlers/SettingsViewMessageHandler.ts  (~380 LOC)
+
+desktop/main/ipc.ts                            (~150 LOC total for all three — generic dispatcher
+                                                routes Zod-typed messages to controller methods)
+```
+
+The host glue is reduced to: validate message via Zod, look up controller method by message type, call it, send response (extension uses `webview.postMessage`; desktop uses `webContents.send`). Real business logic lives in the controllers, takes `platform()` for any host capabilities (file pickers, openExternal, secret input).
```
LionSR: This is still too broad for a clean boundary. The current `Platform` interface is deliberately tiny (`config`, state, log, fs, workspace, storage, secrets), but the handlers use UI/host effects such as file pickers, `openExternal`, input boxes, warning dialogs, terminal creation, editor opening, command execution, clipboard, and webview push. If the new controllers simply take `platform()` for all of that, `Platform` becomes a UI mega-facade and the core layer inherits exactly the host coupling this refactor is trying to remove. The cleaner split is to keep pure controller/domain logic behind `platform()` plus explicit narrow ports for UI effects (`PromptHost`, `ExternalOpener`, `EditorHost`/`DiffViewHost`, `TerminalHost`, `CommandHost`, etc.), or keep those UI decisions in the thin per-host handler and move only host-neutral operations into controllers. The PRD should make that boundary explicit before claiming the desktop side is just generic IPC routing.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1087 @@
...
+
+### Suggested ordering
+
+If we land them all, ~7–9 engineering weeks total, parallelizable. Suggested order: **3 → 1 → 4 → 16 → 2 → 14 → 18 → 17 → 6 → 11 → 15 → 9 → 19 → 7 → 8 → 13 → 10 → 12**. Mechanical fixes first (#3 EventEmitter), then the foundational items (#1 ConfigProvider expansion, #4 watch, #16 FakePlatform — pre-test infrastructure pays off immediately), then the structural extractions in dependency order (#2 DiffViewHost, #14 SupabaseSession+Client, #18 host-neutral controllers, #17 command catalog), then the lint guards (#11 and #15) to lock in agnostic-zone + composition-root purity, then the schema unification (#9), resource-sync adapter (#19), CSS shim (#7), binary resolver (#8), dispatcher cleanup (#13), audit/rehearsal items.
+
+Each Tier 1 item is independently shippable to the extension and produces a smaller, cleaner extension regardless of the Electron decision. After Tier 1 lands, the Electron port is mostly "implement the four interfaces (`ConfigProvider`, `DiffViewHost`, `WorkspaceProvider`-watch, `SupabaseSession` host wrapper) against Electron primitives" — and we can verify each implementation against `FakePlatform`'s invariants.
```
LionSR: This summary is now stale relative to the deeper refactor list. After Tier 1 lands, the desktop port is not mostly the four interfaces listed here: Phase 1 also needs the `AgentDirectories` resource-sync adapter to load bundled/default agents correctly, and Phase 2/3 depend on #18 host-neutral controllers plus #17 command catalog rather than just `ConfigProvider`/`DiffViewHost`/workspace watch/auth glue. For a clean plan, make those items explicit gates in the phase dependencies and exit criteria; otherwise the phase table still hides the largest boundary work even though §9 now documents it.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1087 @@
...
+
+### Phase 0 — Monorepo split (1.5–2 weeks)
+
+The biggest mechanical change. Do this first; everything else is downstream.
+
+- Migrate root → pnpm workspaces. Three packages: `core`, `extension`, `desktop` (`desktop` initially empty).
+- Move source files per the layout in §7.1. Update path aliases in `tsconfig.base.json`. Update import sites if needed (most stay the same via aliases).
+- Refactor `src/utils/config/configUtils.ts` behind `ConfigProvider` (or accept that it remains in `extension/`).
+- Swap `vscode.EventEmitter` → Node `EventEmitter` in `src/auth/tier/` and `src/auth/serverKeys/`.
+- Update CI: `pnpm install`, `pnpm --filter extension build`. Verify VSIX is byte-equivalent (modulo timestamps) to pre-split build.
```
LionSR: `byte-equivalent` is the wrong Phase 0 acceptance test once the repo moves into pnpm workspaces and the extension package is relocated. VSIX byte output can legitimately change due to zip ordering, package metadata, path roots, sourcemap paths, lockfile/package-manager metadata, or generated manifests even when runtime behavior is identical. Conversely, byte-ish equality would not prove that the new package layout resolves resources correctly. Use deterministic behavioral gates instead: `pnpm --filter extension typecheck/compile`, package and smoke-load the VSIX, compare the extension manifest/contributions and bundled runtime asset set against the pre-split build, and run a minimal activation/webview smoke.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1146 @@
...
+| `StateStore` (global)    | `ExtensionContext.globalState`                              | `conf` (file: `state.global.json`) under `app.getPath('userData')`                                                                                                                                                                                                                                                                                                                                                                                                              |
+| `StateStore` (workspace) | `ExtensionContext.workspaceState` (app-private Memento)     | `conf` keyed by **stable hash of the project path**, stored under `app.getPath('userData')/workspace-state/<sha256(projectPath)>.json`. **Not** in `<project>/.texra/` — workspace state contains agent history, stream logs, task states, and run instructions; writing those into the user's repo would dirty git, fail on read-only/shared folders, and risk committing local run history. Project-local state is reserved for an explicit settings-as-code feature (§13.1). |
```
LionSR: This is mixing two different storage classes into one `conf`-backed bucket. Workspace settings / small memento state are fine as JSON. Agent history, stream logs, task states, and run instructions are high-churn, potentially large, and crash-sensitive; putting those in a single `conf` JSON file means long sessions rewrite the same blob over and over and a crash can take out the whole workspace state. Keep `conf` for config and small mementos, but put run artifacts under an append-oriented per-run directory in `userData/projects/<hash>/runs/...` (or reuse the existing run-dir shape) with explicit retention/compaction. Otherwise the desktop app ships with a state store that works in demos and becomes garbage under real agent sessions.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1146 @@
...
+| **Renderer → host (RPC)**  | `vscode.postMessage()` → `BaseViewMessageHandler` | `ipcRenderer.invoke()` ⇄ `ipcMain.handle()` (request/response) | User actions: "execute agent", "save settings", "approve diff", "open project."                      |
+| **Host → renderer (push)** | `webview.postMessage()` from `WebviewUpdater`     | `webContents.send()` → `ipcRenderer.on()` (fire-and-forget)    | Streamed logs, progress events, status updates, approval requests, theme changes, state restoration. |
+
+**One schema layer.** Both directions use the existing Zod-validated message schemas in `core/shared/`. Inbound schemas (renderer→host) become `ipcMain.handle('texra:rpc', validate(payload))`. Outbound schemas (host→renderer) become `webContents.send('texra:push', validate(message))`. The dispatcher pattern in `mainViewDispatcher.ts` / `messageDispatcher.ts` is unchanged — it still receives a single message argument and routes by discriminator.
```
LionSR: Zod validation is not an authority model. This says every renderer action goes through one generic `texra:rpc` channel, and later the renderer gets a command palette over `CommandHost`; that means any renderer compromise can ask main to execute validly-shaped privileged actions. These views render markdown, model/tool output, logs, and file-derived content, so treating XSS as impossible is bad architecture. The preload should expose narrow capability methods or per-view allowlisted RPCs, and `ipcMain` should check the sending `webContents`/frame and allowed command set before dispatch. `contextIsolation` and `sandbox` do not save you if the bridge itself is a generic command pipe.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1146 @@
...
+
+### 7.5 Process model
+
+- **Main process** — app lifecycle, window mgmt, native menu, auto-update, protocol handler, platform impls.
+- **Renderer (one per window)** — Lit UI; `nodeIntegration: false`, `contextIsolation: true`, `sandbox: true`. Talks to main via preload bridge.
+- **No utility process at v1.** Agents run in the main process. The earlier "Phase 4+ utility-process split" was misleading — the existing `ProgressEventBus` is an in-memory singleton, and moving agents to a utility process means bridging every `bus.emit(...)` over `MessagePort` (otherwise progress, logs, and approvals don't reach main or renderer). The utility process would also need its own `initPlatform()` with proxied platform interfaces. Real work, no v1 benefit. Deferred to §13.1 future divergence; v2 candidate.
```
LionSR: Running agents in the main process for v1 may be a pragmatic implementation choice, but the architecture should not bake that choice into the core event path. Right now the document says the `ProgressEventBus` singleton is why utility-process execution is deferred; that is exactly the abstraction leak to remove before desktop code depends on it. Define an `AgentRuntimeHost`/`ProgressSink` boundary now: v1 can implement it in-process in main, and v2 can implement the same contract over `MessagePort`. If you wait until after the desktop shell exists, the future utility-process work becomes a cross-cutting rewrite of progress, approval, cancellation, logging, and platform access.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1146 @@
...
+- Runs `pnpm --filter desktop build` (electron-vite produces the same JS bundle on every OS).
+- Runs `electron-builder` with that OS's targets only (`-m` / `-w` / `-l`).
+- Signs locally using OS-resident credentials (mac uses notarytool; win uses Azure Trusted Signing; linux is unsigned but checksummed).
+- Uploads installers + blockmaps + per-OS manifest as workflow artifacts.
+
+A fourth aggregator job (Linux runner) downloads all artifacts, merges manifests into `latest-mac.yml` / `latest.yml` / `latest-linux.yml`, and pushes to the public release repo. Each OS job runs independently — a flake on the Windows runner doesn't block the mac release.
+
+**Local dev:** developers running `pnpm --filter desktop build` on their own machine get only their OS's installer; that's expected and fine for testing. Cross-OS builds happen only in CI.
+
+- Both share the same `node_modules` (workspace hoisting via pnpm) and the same `tsconfig.base.json` aliases. No shared `dist/`, no shared cache, no shared bundler config.
```
LionSR: This is too hand-wavy for a release build. A pnpm workspace can hoist dependencies during development, but the Electron package must have an explicit production dependency closure. If `desktop` imports a package only declared by `extension` or only reachable through a hoisted root dependency, dev builds pass and packaged installers fail or silently miss files after pruning. The plan needs a hard rule that each package declares every runtime dependency it imports, plus a CI check against the packaged asar/installer rather than relying on shared `node_modules` as an architectural property.

prds/prd-electron-app.md
```diff
@@ -0,0 +1,1146 @@
...
+Cleaner split: keep `Platform` as **OS primitives**, and introduce a separate set of **narrow UI effect ports** in `core/hosts/` that controllers take alongside `platform()`. Each port is small, has one job, and adapts cleanly to both VS Code and Electron primitives:
+
+```ts
+// core/hosts/index.ts
+export interface UIHosts {
+  prompt: PromptHost; // confirm / info / warning / error / input
+  externalOpener: ExternalOpener; // openExternal(url), openPath(file)
+  diff: DiffViewHost; // (already defined in §9 #2)
+  terminal: TerminalHost; // create / send / dispose terminal
+  commands: CommandHost; // runtime invoker (catalog from #17 is metadata)
```
LionSR: `CommandHost` is not a narrow UI port; it is a command bus escape hatch. If core controllers can call `commands.execute(id)`, the core has learned about host command names and we have recreated VS Code's dynamic command system inside the supposedly host-neutral layer. The clean direction is one-way: the outer shell maps catalog IDs from menus/palettes to typed controller methods or explicit ports. Controllers should depend on `PromptHost`, `ExternalOpener`, `DiffViewHost`, etc., not on a stringly command executor. Otherwise the catalog becomes both metadata and runtime authority, and static boundaries disappear.

LionSR (commented):
Architecture pass on the current head. The doc has improved, but the remaining problems are not wording problems; they are boundary problems that will become expensive if this turns into implementation as written.

PR #3872, opened 2026-05-12
refactor: introduce AgentError base class for runtime throws

LionSR:
## Summary

- Adds a tiny `AgentError` base class in `src/common/errors/agentErrors.ts` (12 lines, single class — no subclasses, no fields beyond what `Error` already provides).
- Replaces seven plain `throw new Error(...)` sites in `src/agent/runtime/executeAgent.ts` with `throw new AgentError(...)`.
- The catch-and-rethrow at the end of `runFlowWithLifecycle` now preserves typed agent errors instead of re-wrapping them in a fresh `Error` (which would erase the discriminant for any future `instanceof` check upstream).

## Why

Today every runtime/control-flow failure in `executeAgent` bubbles up as a bare `Error`, so callers either regex on `error.message` or fall back to `'unexpected'` in `classifyAgentError`. Introducing a discriminant unblocks future work (e.g. suppressing redundant toasts when a credit-exhausted error already surfaced via the in-flow `RetryRequestCoordinator`) without changing anything observable today.

## Non-goals (intentionally not in this PR)

- **No subclasses.** Earlier draft had `ModelProviderError` / `AgentAbortError` / `MaxTurnsExceededError` — explicitly removed. Each would have duplicated existing machinery (`ProviderError` schema, `isUserAbort`, etc.) or sat as dead infrastructure with no thrower/catcher. Subclasses will be added when a concrete thrower **and** catcher land together.
- **No behavior change.** No retry policy change, no toast policy change, no `WAITING`-vs-`ERROR` routing change, no `AbortSignal` plumbing change.
- **No max-turn guard.** Out of scope by design.

## Test plan

- [x] `npm run typecheck` — passes across all workspaces
- [x] `npm run lint` — passes
- [ ] Sanity-check a normal agent run from the extension host (any workflow agent, any tool-use agent) — confirm no visible change in error toasts or stream-status transitions
- [ ] Trigger one of the typed-throw paths (e.g. supply an unknown agent name to a delegation call) and confirm the toast / log still reads correctly with `AgentError` as the constructor name

---

> [!NOTE]
> **Low Risk**
> Low risk refactor that changes the error type thrown by `executeAgent` paths; behavior should be the same but any callers relying on `Error` identity/serialization could be affected.
> 
> **Overview**
> Introduces a host-neutral `AgentError` base class in `@common/errors` and exports it from the errors barrel.
> 
> Updates `executeAgent` runtime to throw `AgentError` (instead of plain `Error`) for common launch/validation/control-flow failures (missing agent/model, category mismatch, stream lock contention, invalid resume), and rethrows failures from `runFlowWithLifecycle` as `AgentError` with the original error attached as `cause`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2bbdc6b44a1f3505bd96856027a78a02e61338a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/runtime/executeAgent.ts
+++ b/src/agent/runtime/executeAgent.ts
@@ -41,6 +41,7 @@ import { buildUserVars } from '@agent/utils/userVars';
 import { UsageMonitor } from '@agent/utils/UsageMonitor';
 import { agentConfigToTaskState } from '@agent/utils/agentConfigToTaskState';
 import {
+  AgentError,
   classifyAgentError,
   getSdkErrorMessage,
   toErrorMessage,
@@ -155,7 +156,7 @@ export async function getAgentPath(
     'showAgentConfigBanner',
     { agentName: agentIdentifier },
   );
-  throw new Error(`Could not find agent: ${agentIdentifier}`);
+  throw new AgentError(`Could not find agent: ${agentIdentifier}`);
 }
 
 async function validateModelExists(
@@ -176,7 +177,7 @@ async function validateModelExists(
     ],
     showSuppress: false,
   });
-  throw new Error(`Model ${modelName} not found in MODEL_CONFIGS`);
+  throw new AgentError(`Model ${modelName} not found in MODEL_CONFIGS`);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


LionSR:
Addressed in 2e99300d8.

Both review bots correctly flagged that `if (err instanceof AgentError) throw err;` silently discarded the contextualized `errorMsg` ("Error executing agent <name>: …") built in the catch block — especially harmful for nested subagent runs where the outer wrap is exactly what surfaces the outer agent's name to `formatSubagentError` and other `err.message` consumers.

The type-preservation it bought was speculative (no upstream uses `instanceof AgentError` today, and `cause` already carries the original error). The fix simply drops the guard and always wraps as `AgentError(errorMsg, { cause: err })`. Behavior is now identical to pre-PR except the wrapper type is `AgentError` rather than `Error`.

Issue #3372, opened 2026-05-03
Electron cleanup: finish ProgressSink and runtime-host boundary


LionSR:
## Scope

Complete the runtime progress boundary so agent/runtime code reports progress through injected host capabilities instead of reaching around the composition root.

## Background

The Electron PRD relies on the runtime taking an AgentRuntimeHost or ProgressSink boundary so the desktop app can swap the VS Code event bus for an Electron adapter. Several follow-up slices have reduced this coupling, but the boundary still has shallow pass-through areas and default fallbacks.

## Work

- Audit remaining agent/runtime imports of progress-event and event-bus surfaces.
- Route output-file, flow-node, inquiry, approval, and coordinator progress through the injected runtime host or sink.
- Remove default global progress fallbacks where callers can provide an explicit runtime host.
- Keep the extension adapter as the VS Code composition root for progress events.

## Acceptance Criteria

- Runtime code has a single explicit host/sink path for progress publication.
- Electron can provide an alternate host without importing VS Code or the progress event bus singleton.
- Focused tests cover the moved coordinator and output-event paths.

## Validation

- npm run compile:safe
- npm run compile-tests
- npm run lint

LionSR:
Scouting note (2026-05-13): current main is not cleanly closable for this issue. There is progress toward an AgentRuntimeHost boundary, but runtime and tool paths still contain getAgentRuntimeHost fallbacks and some direct ProgressEventBus reach-throughs. The issue label risk:high is accurate; this should stay open until a focused boundary cleanup removes or documents the remaining fallbacks and tests the moved progress paths.

LionSR:
important

LionSR:
Progress note (2026-05-13): landed commit 1444fa5cb as a narrow boundary cleanup. The accepted-file workspace notification helper now emits workspaceFilesWritten through an AgentRuntimeHost instead of importing ProgressEventBus directly, and the desktop bridge passes its own runtime host into that helper. Added a focused test for the explicit-host path. This is partial progress only; #3372 should remain open because GitHub subscription events, approval paths, flow fallbacks, and other runtime-host/default-host reach-throughs still remain.

LionSR:
[...]

PR #3956, opened 2026-05-14
perf: bound execution listing storage work

LionSR:
## Summary\n\n- Cap per-execution storage work in execution-history listing so a large history does not enqueue one meta/config read pair per entry at once.\n- Apply the same bounded storage work to delete-all and legacy history backfill paths.\n- Keep the existing listing shape, cache behavior, sorting, and per-entry corrupt-history handling unchanged.\n\n## Design note\n\nThis is deliberately a small performance boundary. The execution listing module still owns directory scanning and KV reads; the change only controls fan-out pressure on the extension host. It avoids adding a new cache, resolver, or pagination contract.\n\nRefs #3372.\n\n## Validation\n\n- npm run typecheck\n- npm run compile:fast\n- ../../node_modules/.bin/eslint src/agent/storage/executionListing.ts\n- ../../node_modules/.bin/prettier --check src/agent/storage/executionListing.ts\n- git diff --check

---

> [!NOTE]
> **Medium Risk**
> Introduces bounded-concurrency storage operations and changes bulk paths to continue on individual failures, which could mask per-execution errors or alter timing under load.
> 
> **Overview**
> Adds bounded concurrency (via `p-map`, default `32`) for per-execution storage reads when building the execution listing, preventing large histories from spawning unbounded parallel `meta`/`config` reads.
> 
> Applies the same bounded-concurrency approach to `deleteAllExecutions` and legacy `backfillEntries`, with `stopOnError: false` and a `finally`-guarded cache invalidation so bulk operations attempt all entries while still clearing the listing cache.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b6d62cb83e288688af84c0eb119a6cab61d47ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/storage/executionListing.ts
+++ b/src/agent/storage/executionListing.ts
@@ -9,6 +9,8 @@
  * into per-execution KV stores, then deletes the legacy sources.
  */
 
+import pMap from 'p-map';
+
 import { type AgentConfig, AgentConfigSchema } from '@agent/core/AgentConfig';
 import * as logger from '@agent/core/logger';
 import { getWorkspaceState } from '@agent/core/stateStore';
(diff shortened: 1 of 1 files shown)


LionSR:
Addressed the Cursor cache-invalidation thread in 2b6d62cb: delete-all now invalidates the execution listing cache in a finally block, so partial delete success cannot leave stale cached entries even when pMap reports an AggregateError after finishing the batch.

Issue #3925, opened 2026-05-13
Audit non-boundary fallbacks in agent runtime and progress state


LionSR:
The cleanup work in #3397 and #3372 should also cover excessive fallback chains. These often encode uncertainty about ownership: a value may come from the current run context, a default singleton, a persisted hint, a task state, or a legacy source, and callers have to know the ordering.

## Initial scouting notes

Concrete places worth auditing:

- `src/agent/runtime/AgentRuntimeHost.ts`: `getAgentRuntimeHost()` falls back from `RunContext` to the default host. This is useful at host boundaries, but runtime code should increasingly receive an explicit host from the launch context.
- `src/agent/runtime/BasePromiseCoordinator.ts`: coordinators accept an optional host and otherwise use the ambient host. This is a likely follow-up after coordinator construction is owned by the launch context.
- `src/agent/runtime/executionRegistry.ts`: `runtimeHostForParent()` scans handles and then falls back to the ambient host. This is closely related to #3921.
- `packages/extension/src/progressView/events/ProgressEventHandler.ts`: stream category is derived from task state, hints, and finally `Workflow`. The true owner of stream category should become clearer.
- `src/controllers/progressView/ProgressFollowUpController.ts`: compile-fixer inputs merge output source candidates with original input fallbacks. This may be legitimate recovery logic, but it should be named and justified at the boundary.
- `src/agent/remote/RemoteAgentLoader.ts`: legacy remote-agent fallback and schema-column fallback are compatibility fallbacks. They should remain only while the legacy data shape is supported, and should be isolated from normal load semantics.

## Desired direction

- Distinguish compatibility fallbacks from ordinary execution paths.
- Move fallbacks to load/migration/boundary modules when they are truly needed.
- Prefer one canonical owner for runtime host, stream category, execution id, model/provider identity, and progress state.
- Remove fallback chains that merely compensate for pass-through layers or missing ownership.

This issue is intentionally an audit/cleanup issue, not a request for one large rewrite. It can be split into smaller PRs when a local owner is clear.

LionSR:
Coordination note for the current refactor surface:

- #3928 owns the CLI runtime cleanup in `packages/cli/src/runtime/*`, `packages/cli/src/commands/root.ts`, and CLI chat rendering. Avoid additional CLI approval/runtime-host cleanup until that PR settles.
[...]

PR #3994, opened 2026-05-14
refactor: require explicit runtimeHost in executeAgent

LionSR:
Eliminate the ambient host fallback at executeAgent.ts:325 — the chain
`options?.runtimeHost ?? tryUseRunContext()?.runtimeHost ?? defaultHost`
encoded three sources for one value, exactly the pattern #3925 calls out.

All five production call sites (extension setup, CLI commands, CLI chat,
runValidatedExecutionRequest, DelegationTools) already pass runtimeHost
explicitly, so removing the fallback is a no-op for runtime behavior but
removes a global mutable state read from the SDK entry point.

Changes:
- ExecuteAgentOptions.runtimeHost is now required.
- Delete getAgentRuntimeHost() and its tryUseRunContext.runtimeHost path.
- executeSubagent throws if invoked outside a parent tool run context
  (was silently falling back to ambient host).

Retained: setDefaultAgentRuntimeHost / getDefaultAgentRuntimeHost and
the defaultAgentRuntimeHost global remain in place — they are the lazy
provider for singleton coordinators (planApprovalCoordinator,
proposalCoordinator, retryCoordinator). Those singletons + the
CoordinatorRuntimeHost union are a separate refactor that needs to move
coordinator ownership fully into the launch context first.

Aligns with #3372 (single explicit host/sink path) and partially closes
the host-resolution leg of #3925.

https://claude.ai/code/session_011tk5cuEbZ59AjmWqLCv5VV

---

> [!NOTE]
> **Medium Risk**
> This is an API-breaking refactor that removes ambient/global fallback host resolution, so any missed call sites will now fail at compile time or throw at runtime for subagent execution. Behavior should be unchanged where callers already pass a host, but it tightens invariants around runtime context usage.
> 
> **Overview**
> `executeAgent` now **requires** an explicit `runtimeHost` (and an `options` object) instead of lazily falling back to run-context/default globals, removing the `getAgentRuntimeHost()` helper and related `tryUseRunContext()` path.
> 
> Delegation now fails fast when invoked without a parent tool-run context: `executeSubagent` throws if `runtimeHost` is missing rather than silently continuing with an ambient host. `AgentRuntimeHost` docs were updated to clarify the remaining `set/getDefaultAgentRuntimeHost` global is legacy-only for singleton coordinators.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a5456d0058b936c3bbd2e8e75f73264b9ca1c24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/runtime/AgentRuntimeHost.ts
+++ b/src/agent/runtime/AgentRuntimeHost.ts
@@ -1,7 +1,5 @@
 import type { ProgressEventPayloads } from '@eventBus/ProgressEventBus';
 
-import { tryUseRunContext } from './RunContext';
-
 export interface AgentRuntimeHost {
   emit<K extends keyof ProgressEventPayloads>(
     event: K,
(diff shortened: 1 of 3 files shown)


LionSR:
Closing this as superseded by the merged runtime-host cleanup in #3991. Current main already requires an explicit runtimeHost for executeAgent and executeSubagent now fails fast when no parent tool runtime host is present. The remaining global coordinator ownership question should stay as a separate #3372/#3925 follow-up rather than being reintroduced through this stale branch.

PR #4109, opened 2026-05-16
chore(cli): simplify runtime layer

LionSR:
## Summary

Code-simplifier pass over the CLI runtime layer (added in cli-tui phases 0–6). No behavior changes; +74/-148 across 7 files.

- **logSinks**: collapse 5 stream writers onto a single `writeRaw(key, text)` over a `closed: { stdout, stderr }` map (drops ~40 lines of duplicated try/catch + flag glue).
- **initPlatform**: factor `logAt(level, channel, message)` shared by debug/info/warn arms of `cliPlatformLog`; merge duplicated Supabase + server-side-key init blocks; drop dead intermediate `const globalState`.
- **approvalAdapter**: replace 4-clause `isApprovalEvent` boolean chain with `APPROVAL_EVENTS` tuple + `.includes` (`ApprovalEvent` derived from the tuple); extract `humanInputDenialFeedback` to deduplicate yolo-vs-policy denial in `handleExternalInquiry` and `handleUserQuestion`.
- **approvalPolicy**: drop unused `parseCliApprovalPolicy` (zero callers anywhere in repo).
- **cliContext**: inline single-use `resolveCwdFlag`.
- **runtimeHost**: re-order `requestShowError` handling before the quiet-logs cutoff; drop `quietLogs` local shadow.
- **commands/root.ts**: `flatMap` over agent categories in `listAgents` instead of duplicated `.map(...)` per category; move misplaced JSDoc onto the function it describes.

No exported symbols changed.

## Test plan
- [x] \`npx tsc --noEmit -p packages/cli/tsconfig.json\` — clean
- [ ] CI green
- [ ] Smoke: \`texra chat\` boots and a slash command (e.g. \`/help\`, \`/login\`) round-trips

---

> [!NOTE]
> **Low Risk**
> Low risk refactor focused on deduplicating CLI runtime/logging and approval-handling code; primary risk is subtle behavior changes in logging/quiet-mode and denial feedback strings/paths.
> 
> **Overview**
> **Simplifies the CLI runtime implementation without changing public APIs.**
> 
> Refactors CLI stream/log writing to a single best-effort `writeRaw` path, deduplicates platform log formatting and server-side key initialization in `initCliPlatform`, and cleans up approval-event handling by deriving `ApprovalEvent` from a constant list and centralizing human-input denial feedback.
> 
> Also removes the unused `parseCliApprovalPolicy`, inlines `cwd` flag resolution in `buildCliContext`, adjusts `runtimeHost` quiet-log gating so `requestShowError` is emitted before quiet suppression, and simplifies agent listing to a `flatMap` over categories while moving the `runCli` JSDoc to the correct location.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8d53a72bcb6e167823764873776ee735628165a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/commands/root.ts
+++ b/packages/cli/src/commands/root.ts
@@ -116,16 +116,10 @@ const agentsListCommand = defineCommand({
 async function listAgents(context: CliContext): Promise<number> {
   await initCliPlatform({ ...context, quietLogs: true });
   await loadAgents();
-  const agents = [
-    ...getVisibleAgents(AgentCategory.Workflow).map((agent) => ({
-      ...agent,
-      category: AgentCategory.Workflow,
-    })),
-    ...getVisibleAgents(AgentCategory.ToolUse).map((agent) => ({
-      ...agent,
-      category: AgentCategory.ToolUse,
-    })),
-  ];
+  const agents = [AgentCategory.Workflow, AgentCategory.ToolUse].flatMap(
+    (category) =>
+      getVisibleAgents(category).map((agent) => ({ ...agent, category })),
+  );
 
   if (context.outputFormat === 'json') {
     writeTextStdout(JSON.stringify(agents, null, 2));
@@ -695,12 +689,6 @@ function isCliError(error: unknown): error is Error & { code?: string } {
 }
 
 /**
- * Re-implement the surface citty's `runMain` provides — `--help` / `--version`
- * detection plus error handling around `runCommand` — so usage errors (missing
- * required flag, unknown subcommand, invalid enum value) preserve our
- * canonical `CliExitCode.Usage` (2) instead of citty's hard-coded `exit(1)`.
- */
-/**
  * Walk `rawArgs` positional-by-positional through the subcommand tree to find
  * the deepest matched command. Used to scope `--help` to the subcommand the
  * user typed rather than always showing root-level usage.
@@ -738,6 +726,12 @@ async function resolveDeepestSubCommand(
   return [cmd, parent];
 }
 
+/**
+ * Re-implement the surface citty's `runMain` provides — `--help` / `--version`
+ * detection plus error handling around `runCommand` — so usage errors (missing
+ * required flag, unknown subcommand, invalid enum value) preserve our
+ * canonical `CliExitCode.Usage` (2) instead of citty's hard-coded `exit(1)`.
+ */
 export async function runCli(
   argv?: readonly string[],
 ): Promise<{ exitCode: number }> {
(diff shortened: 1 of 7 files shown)


LionSR (commented):
Reviewed against the TeXRA code-review checklist (`.claude/skills/code-review/SKILL.md`). Diff is mechanical and clean — every change is structural with no behavior delta, no platform leakage. No blockers.

### Notes

- `packages/cli/src/runtime/initPlatform.ts:98-105` — the original conflated platform-init and server-side-key init into one branch ("either init both, or init only keys"). Splitting them into two independent `if` guards is functionally equivalent today, but it now relies on the invariant that `initPlatform()` is synchronous and side-effect-only between the two guards. True in current code; worth keeping in mind if anyone later inserts an `await` between them.
- `packages/cli/src/runtime/approvalAdapter.ts:30-43` — `(APPROVAL_EVENTS as readonly string[]).includes(event)` is the standard widening dance for \`readonly\` tuple \`.includes()\` on a narrower input. Fine as written. A `satisfies` alternative isn't available here because the predicate takes `keyof ProgressEventPayloads`, not a string.
- `packages/cli/src/runtime/logSinks.ts:7-23` — `closed[key]` works because `key` is a narrowed `'stdout' | 'stderr'` literal. TS infers `process[key]` as the union of the two streams; fine today.
- `packages/cli/src/runtime/runtimeHost.ts:50-55` — reading `context.quietLogs` directly drops the previous `?? false` floor. Identical for `boolean | undefined`. If `quietLogs` ever becomes a string/enum, the floor disappears.
- `packages/cli/src/commands/root.ts:691-731` — the citty/\`runMain\` JSDoc moving from above `resolveDeepestSubCommand` to above `runCli` is a correctness fix (old placement described the wrong function). Worth noting in the PR description as a doc fix, not just a mechanical move.

### Verified

- Confirmed the two-guard split in `initPlatform.ts` preserves first-call behavior (both branches fire) and idempotent re-entry behavior; `tryPlatform()?.globalState` resolves to the freshly created store after `initPlatform()`.
- Confirmed all five public writers in `logSinks.ts` route through `writeRaw`; the \`closed\` map is module-local; the "best effort" why-comment is preserved.
- Greped `parseCliApprovalPolicy` repo-wide — zero remaining callers, deletion is safe.
- Confirmed no `vscode` imports in any touched file (`packages/cli/src/runtime/*` and `commands/root.ts` are host-neutral by zone).

PR #4110, opened 2026-05-16
chore(cli): simplify chat orchestration

LionSR:
## Summary

Code-simplifier pass over the non-TUI chat loop and ANSI renderer (built across cli-tui phases 0–4). No observable output change; ~39 lines net deletion.

- **terminalRenderer**: collapse two-pass approval dispatch (the \`isApprovalEvent\` predicate manually listing 6 event names, followed by \`renderApprovalEvent\` switching over the same 6) into a single switch inside \`renderProgressEvent\`. Drops the two helper methods and an unreachable \`default\` branch.
- **runChat**: convert the 9-clause \`else if (command === '...')\` slash-command chain into a single \`switch (command)\`. The \`/exit\` | \`/quit\` early-return stays above the switch since it \`break\`s the outer for-await loop.

No exported symbols changed.

## Test plan
- [ ] CI green
- [ ] Smoke: run \`texra chat\` in a non-TTY (or with TUI disabled) and exercise \`/help\`, \`/clear\`, \`/login\`, \`/tools\`, \`/exit\`
- [ ] Smoke: trigger a bash/edit/plan approval and verify the ANSI card renders identically

---

> [!NOTE]
> **Low Risk**
> Low risk refactor that restructures slash-command handling and progress-event routing without changing the underlying behaviors; primary risk is small control-flow regressions (e.g., missing `break`/early-return) in the CLI chat loop and renderer.
> 
> **Overview**
> Simplifies the plain-mode chat loop by replacing the long `/` slash-command `else if` chain with a single `switch (command)` in `runChat.ts`, keeping `/exit`/`/quit` as an early exit path and preserving per-command behavior/usage messaging.
> 
> Simplifies ANSI progress rendering in `terminalRenderer.ts` by inlining approval-event dispatch directly into `renderProgressEvent` and deleting the redundant `isApprovalEvent`/`renderApprovalEvent` helpers, reducing duplicated event-name lists and unreachable branches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb95153957ca14dcb3cd1ee4c700cb252d2a8f3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/runChat.ts
+++ b/packages/cli/src/chat/runChat.ts
@@ -538,58 +538,70 @@ export async function runChat(
           requestChatExit();
           break;
         }
-        if (command === 'help') {
-          renderer.printHelp();
-        } else if (command === 'clear') {
-          renderer.printClearScreen();
-        } else if (command === 'agent') {
-          if (session.runPromise) {
-            renderer.warn('The active session already owns its agent.');
-          } else if (rest) {
-            agent = rest;
-            renderer.success(`Agent set to ${agent}.`);
-          } else {
-            renderer.warn('Usage: /agent <name>');
-          }
-        } else if (command === 'model') {
-          if (session.runPromise) {
-            renderer.warn('The active session already owns its model.');
-          } else if (rest) {
-            if (await modelAvailable(rest, renderer)) {
-              model = rest;
-              await setCliHelperModel(model);
-              renderer.success(`Model set to ${model}.`);
+        switch (command) {
+          case 'help':
+            renderer.printHelp();
+            break;
+          case 'clear':
+            renderer.printClearScreen();
+            break;
+          case 'agent':
+            if (session.runPromise) {
+              renderer.warn('The active session already owns its agent.');
+            } else if (rest) {
+              agent = rest;
+              renderer.success(`Agent set to ${agent}.`);
+            } else {
+              renderer.warn('Usage: /agent <name>');
             }
-          } else {
-            renderer.warn('Usage: /model <name>');
+            break;
+          case 'model':
+            if (session.runPromise) {
+              renderer.warn('The active session already owns its model.');
+            } else if (rest) {
+              if (await modelAvailable(rest, renderer)) {
+                model = rest;
+                await setCliHelperModel(model);
+                renderer.success(`Model set to ${model}.`);
+              }
+            } else {
+              renderer.warn('Usage: /model <name>');
+            }
+            break;
+          case 'tools': {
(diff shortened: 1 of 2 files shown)


LionSR (commented):
Reviewed against the TeXRA code-review checklist. Switch conversion + collapse of a two-pass approval dispatch into one switch. No behavior changes, no platform leakage. No blockers.

### Notes

- `packages/cli/src/chat/terminalRenderer.ts:137-160` — the deleted `renderApprovalEvent` had a `default:` branch (`renderApprovalCard('approval requested', [event])`) that was already unreachable because `isApprovalEvent` filtered first. Dropping it is fine, but losing the safety net means a future approval event added to `isApprovalEvent`'s successors without being added to this switch would silently fall through to the main event switch below. A `satisfies never` exhaustiveness check would catch that at compile time — **however**, implementing it would require re-introducing a type-guard helper (the predicate over the 6 names), which is exactly what this PR just deleted. The simplification wins; this is left as a documented trade-off.
- `packages/cli/src/chat/runChat.ts:541-654` — the `default:` branch correctly replaces the old trailing `else`, so unknown slash commands are still surfaced via `renderer.warn`. Equality-tested branches translate one-to-one to `case` labels.
- `packages/cli/src/chat/runChat.ts:582,602` — block scopes (`case 'tools': { ... }`, `case 'login': { ... }`) correctly contain locals. Other cases without locals correctly omit the block. Consistent.
- `packages/cli/src/chat/runChat.ts:600-630` — the `login` case replaces `if/else-if/else` with `if (...) { ...; break; }` early-returns. Behavior-equivalent and easier to extend.
- Neither file imports `vscode`; both live in the host-neutral CLI zone.

### Verified

- Read the diff in full; confirmed every original `else if` branch has a one-to-one `case` with `break`, that `tools`/`login` use block scopes for their locals, and the early-`break` rewrite of `login` is behavior-equivalent.
- Confirmed the new switch in `renderProgressEvent` covers the same 6 event names as the deleted `isApprovalEvent` predicate, with no overlap against the existing main switch below `toolDisplay === 'hidden'` (main switch handles tool/log events, not approvals).
- Greped repo for `isApprovalEvent` / `renderApprovalEvent` — only `dist/bin/texra.js` build artifacts and the un-modified baseline files reference them; no external callers (both methods were `private`).
- Spot-checked: `renderToolEditApproval` etc. accept `payload: unknown`, so passing `ProgressEventPayloads[K]` is sound.

PR #4111, opened 2026-05-16
chore(cli-tui): simplify render & modal components

LionSR:
## Summary

Code-simplifier pass over the Ink TUI render & component layer (built across cli-tui phases 1–6). No observable UI change; +121/-225 across 10 files (1 new shared helper).

- **modals**: extract shared \`ConfirmCard\` (bordered Box + colored title + \`ConfirmInput\` + y/n \`KeyHints\` footer) used by \`BashApproval\`, \`AgentProposal\`, \`RetryRequest\`, and \`EditApproval\`. \`PlanApproval\` and \`ApprovalModal\` intentionally untouched (their layout doesn't fit the y/n confirm pattern). \`ExternalInquiry\` left alone (uses a text input, not confirm).
- **App.tsx**: drop dead \`useApp()\` import + \`void exit;\` discard (no exit handler wired here); collapse \`showSideColumn\` from a ternary-always-returning-boolean to a single boolean expression.
- **ModelForm**: extract local \`ModelFrame\` for the duplicated loading/error scaffolding; fix stale comment pointing at a removed \`applyModelSelection\` symbol.
- **KeyHints**: inline the single-use \`Hint\` sub-component into the \`.map\`; lift \`DEFAULT_TAIL\` (the \`Enter confirm · Esc cancel\` pair) to a module-level constant.
- **SubagentList.lastLines**: replace the reverse-loop with \`split → map(trimEnd) → filter → slice(-N)\`.
- **ansiMarkdown**: inline the once-used named \`RenderRule\` constants (\`headingOpen\`, \`codeInline\`, \`fence\`, \`codeBlock\`, \`listItemOpen\`, \`paragraphOpen\`, \`paragraphClose\`, \`textRule\`, \`image\`) directly into the \`r.*\` assignments. The non-obvious "why" comments (\`paragraph_open\` hidden-token rule, \`list_item_open\` continuation gutter) are preserved.

No exported prop contracts changed (\`AppProps\`, \`BashApprovalProps\`, \`AgentProposalProps\`, \`RetryRequestProps\`, \`EditApprovalProps\`, \`ModelFormProps\`, \`KeyHintsProps\` unchanged).

## Test plan
- [x] \`npx vitest run src/test-kernel/cli/\` — 34/34 pass
- [ ] CI green
- [ ] Smoke: open TUI (\`texra chat\` in TTY), trigger each modal type (bash, edit, plan, retry, agent-proposal) and confirm visual layout matches main
[...]

PR #4118, opened 2026-05-16
chore(cli-tui): simplify state layer

LionSR:
## Summary

Code-simplifier pass over the CLI Ink TUI state/subscription layer (built across cli-tui phases 1–5). No subscription contracts, signal shapes, or exported APIs changed; +65/-87 across 3 files.

This is the fourth and last of the four-area CLI simplification sweep (after #4109 runtime, #4110 chat orchestration, #4111 render & modals). The original parallel run lost this work without committing; this re-run lands the same wins.

- **cliState**: inline single-use \`withSliceUpdate\` into \`patchStream\` (the only caller). Replace the two-pass \`removeStream\` parent-map cleanup (\`parents.has(id) || [...parents.values()].includes(id)\` followed by nested loops) with a single forward iteration that builds the new map lazily only when there's actually an edge to drop.
- **subscribeApprovals**:
  - Drop unused \`PolicyDecision\` import alias (never referenced after switching to the inferred return type of \`immediateDecision\`).
  - Single source of truth for \`APPROVAL_EVENTS\` (was duplicated between a string-literal \`ApprovalEvent\` union AND a \`new Set<ApprovalEvent>([...])\` instantiation).
  - Flatten \`routeWithPolicy\` from \`(context, dispatch, payload, toQueuePayload)\` to \`(context, kind, payload, dispatch)\` — each call site loses its \`(p) => ({ kind: 'bash', payload: p })\` closure builder. \`Extract<...>\` narrows the queue payload to the picked kind.
  - Collapse \`setToolEditApprovalHandler\`'s split policy/enqueue paths into a single decision-then-return shape.
  - Extract \`userQuestionFeedback\` from the nested ternary in \`handleUserQuestion\`.
- **subscribeRuntimeHost**: clean up \`updateActiveProcesses\`'s prune loop — drop the redundant \`s.processOutput.size > 0\` guard (the loop is a no-op on an empty map anyway), rename \`mutated\` to \`pruned\`, use \`??=\` for the copy-on-first-write step.

## Test plan
- [ ] CI green (typecheck + format + vitest cli kernel)
- [ ] Smoke: open TUI, trigger each approval kind (bash/plan/agent/retry/external inquiry/user question) and confirm queue + immediate-policy paths still route correctly
- [ ] Smoke: kill a subprocess while a parent stream is active and confirm the process-output map prunes without UI glitches
- [ ] Smoke: close a subagent stream and confirm parent-map edges are cleaned up (no stale focus cycle targets)

---

> [!NOTE]
> **Low Risk**
> Low risk refactor of CLI TUI state/approval plumbing; behavior should be unchanged but small mistakes could surface as missed approvals or stale stream focus cleanup.
> 
> **Overview**
> Simplifies the CLI Ink TUI state/subscription layer without changing event contracts or exported APIs.
> 
> In `cliState`, inlines the single-use stream update helper into `patchStream` and streamlines `removeStream` parent-edge cleanup to a single lazy copy-on-write pass.
> 
> In approvals, centralizes approval-event constants, simplifies `routeWithPolicy`/tool-edit decision flow, and extracts user-question feedback generation into a helper. Runtime-host subscription pruning for `updateActiveProcesses` is also tightened to lazily clone `processOutput` only when entries are actually removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51946337bef4f15aa6b6b608b4194fbea780876f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/tui/state/cliState.ts
+++ b/packages/cli/src/chat/tui/state/cliState.ts
@@ -118,26 +118,17 @@ function emptySlice(streamId: StreamTabId): StreamSlice {
   };
 }
 
-function withSliceUpdate(
-  current: ReadonlyMap<StreamTabId, StreamSlice>,
-  streamId: StreamTabId,
-  update: (slice: StreamSlice) => StreamSlice,
-): ReadonlyMap<StreamTabId, StreamSlice> {
-  const slice = current.get(streamId) ?? emptySlice(streamId);
-  const next = update(slice);
-  if (next === slice) return current;
-  const out = new Map(current);
-  out.set(streamId, next);
-  return out;
-}
-
 export function patchStream(
   streamId: StreamTabId,
   update: (slice: StreamSlice) => StreamSlice,
 ): void {
-  cliState.streams.set(
-    withSliceUpdate(cliState.streams.get(), streamId, update),
-  );
+  const current = cliState.streams.get();
+  const slice = current.get(streamId) ?? emptySlice(streamId);
+  const next = update(slice);
+  if (next === slice) return;
+  const out = new Map(current);
+  out.set(streamId, next);
+  cliState.streams.set(out);
 }
 
 export function setParentStream(
@@ -163,14 +154,13 @@ export function removeStream(streamId: StreamTabId): void {
   // Drop any parent-map edges that touched this stream so the focus cycle
   // never lands on a stale id.
   const parents = cliState.parentStream.get();
-  if (parents.has(streamId) || [...parents.values()].includes(streamId)) {
-    const nextParents = new Map(parents);
-    nextParents.delete(streamId);
-    for (const [child, parent] of nextParents) {
-      if (parent === streamId) nextParents.delete(child);
-    }
-    cliState.parentStream.set(nextParents);
+  let nextParents: Map<StreamTabId, StreamTabId> | undefined;
+  for (const [child, parent] of parents) {
+    if (child !== streamId && parent !== streamId) continue;
+    if (!nextParents) nextParents = new Map(parents);
+    nextParents.delete(child);
   }
+  if (nextParents) cliState.parentStream.set(nextParents);
 }
 
 export function resetCliState(): void {
--- a/packages/cli/src/chat/tui/state/subscribeApprovals.ts
+++ b/packages/cli/src/chat/tui/state/subscribeApprovals.ts
@@ -34,7 +34,6 @@ import {
   denyMessage,
   immediateDecision,
   markApprovalDenied,
-  type ApprovalDecision as PolicyDecision,
 } from '../../../runtime/approvalAdapter';
 import { assertNever } from '../assertNever';
 import { enqueueApproval, type ApprovalDecision } from './approvalQueue';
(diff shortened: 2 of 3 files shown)


LionSR (commented):
Reviewed against the TeXRA code-review checklist (`.claude/skills/code-review/SKILL.md`). Three structural simplifications that preserve subscription contracts and signal shapes. No blockers.

### Notes

- `packages/cli/src/chat/tui/state/subscribeApprovals.ts:164-187` — \`routeWithPolicy<K extends 'bash' | 'plan' | 'proposal' | 'retry', P>\` constructs \`{ kind, payload }\` and casts via \`Extract<Parameters<typeof enqueueApproval>[0], { kind: K }>\`. Type-safe today because every member of \`enqueueApproval\`'s input union with those four kinds is shaped exactly \`{ kind: 'X'; payload: T }\`. The K-constraint is the load-bearing protection: \`'toolEdit'\` uses \`request\` (not \`payload\`), so it's excluded from K. **The pre-PR closure-builder would have caught a future mismatch at compile time; the new \`as Extract<...>\` cast won't.** If anyone later widens K to include \`'toolEdit'\`, the runtime payload would be missing \`request\`. Consider a short comment cross-referencing \`approvalQueue.ts:23-28\` so future-K-extenders look at the queue type first.
- \`packages/cli/src/chat/tui/state/cliState.ts:154-163\` — \`removeStream\`'s new single-pass cleanup is correct, including the case where \`streamId\` appears as both a child key AND a parent value (the loop hits both entries and deletes both). Iterating the original \`parents\` while mutating \`nextParents\` is safe — Map iteration is over a snapshot of the source map, not the one being mutated. Performance is a strict improvement.
- \`packages/cli/src/chat/tui/state/subscribeApprovals.ts:88-97\` — \`let decision\` reassignment style is fine here; \`const\` with a conditional would have required a ternary.
- \`packages/cli/src/chat/tui/state/subscribeApprovals.ts:281-300\` — \`userQuestionFeedback\` is declared after its only caller \`handleUserQuestion\`. Minor declaration-order nit (TS hoisting makes it work).
- \`packages/cli/src/chat/tui/state/subscribeRuntimeHost.ts:87-99\` — dropping the \`s.processOutput.size > 0\` guard is correct: empty-Map iteration is a no-op, \`pruned\` stays undefined, and \`patchStream\`'s \`next === slice\` reference-equality check prevents spurious signal writes.
- No \`vscode\` imports introduced; the state layer is host-neutral by zone.

### Verified

- Read \`subscribeApprovals.ts:88-107\` and \`:164-187\` post-change; confirmed the cast is type-safe for the constrained K union and \`APPROVAL_EVENT_SET\` is initialized once at module load.
- Read \`approvalQueue.ts:23-28\` — confirmed the four kinds \`'bash' | 'plan' | 'proposal' | 'retry'\` all use \`{ kind; payload }\`; \`'toolEdit'\` uses \`{ kind: 'toolEdit'; request }\`, hence excluded from K.
- Traced \`removeStream\` against the original on three cases: streamId as child key only, parent value only, and both — identical output.
- Confirmed inlined \`patchStream\` still short-circuits on \`next === slice\` before the map copy + signal write.
- Greped \`packages/cli/src/chat/tui/state/**\` for \`from 'vscode'\` — clean.

PR #4148, opened 2026-05-17
fix(cli): surface tool-use cycle errors in chat transcript

LionSR:
## Summary

The CLI chat TUI showed the user's prompts but neither model responses nor any error feedback, leaving the agent stuck in `idle` with an empty transcript. Two root causes plus three adjacent fixes:

**Root cause** — `ToolUseCycleNode.post` stored cycle failures in `shared.lastError` but never logged them. `ToolUseWaitNode.post` then cleared `lastError` when the user sent a follow-up, so a recurring failure (missing API key, auth error, etc.) was silently invisible. Fixed by writing the failure message into the StreamLogStore as a `MESSAGE_TYPES.ERROR` entry — the TUI already renders ERROR entries as red lines via the existing `TranscriptEntry` path.

**Adjacent fixes** uncovered while diagnosing:

- `subscribeStreamLog`: sync throttle dropped 200ms → 16ms, and `AgentLogger.flushPendingStreamUpdates()` is now called before reading the store. Short replies that finalized between sync ticks could otherwise stay buffered in AgentLogger's 50ms throttle and never reach the transcript.
- `runChatTui`: agent errors routed through `appendLocalErrorTranscript` instead of `writeTextStderr` (Ink owns stdout/stderr while mounted, so stderr writes are invisible). Force a `syncStreamLog(streamId)` after `executeAgent` resolves so the final assistant chunk lands before the `appendAssistantTranscriptIfMissing` fallback runs.
- `transcript`: factored shared `appendLocalTranscriptEntry` helper, added `appendLocalErrorTranscript`.
- `ConversationPane`: wrap the live region in `<Box minHeight={1}>` so toggling live presence between renders no longer shifts the input bar down by a row.

## Test plan

- [x] Build clean (`pnpm --filter @texra/cli build`)
- [x] `texra chat --model=<model>` with no API key configured → red error line appears in the transcript instead of silent idle.
- [x] `texra chat --model=<model>` with valid auth → model response renders, no layout shift on first/last token.
- [x] User submits a second prompt after a cycle failure → both prompts and the error remain visible.

---

> [!NOTE]
> **Medium Risk**
> Touches the CLI TUI transcript plumbing and stream-log syncing; while behavior is well-scoped, timing/throttling changes and forced flushes could affect ordering or duplication of transcript entries.
> 
> **Overview**
> Fixes missing/hidden feedback in `texra chat` by ensuring **tool-use cycle failures are logged as `MESSAGE_TYPES.ERROR`** and therefore rendered in the TUI transcript.
> 
> Improves transcript reliability by **syncing any buffered model-response chunks on agent completion** (`syncStreamLog` now forces `AgentLogger.flushPendingStreamUpdates()`), lowering stream-log sync throttling to ~1 frame, and routing uncaught runtime errors into the transcript via `appendLocalErrorTranscript` instead of `stderr`.
> 
> Stabilizes the conversation layout by reserving a persistent one-line live region (`minHeight={1}`) to prevent the input bar from shifting when streaming starts/stops, and adds/updates tests covering the new error logging and flush behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 161699e775ba8a5a00d7fded6802e970e92c52b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/tui/panes/ConversationPane.tsx
+++ b/packages/cli/src/chat/tui/panes/ConversationPane.tsx
@@ -56,7 +56,7 @@ function LiveTranscriptEntry({
   readonly entry: ConversationEntry;
 }): React.JSX.Element {
   return (
-    <Box marginBottom={1}>
+    <Box>
       <Text>{entry.text}</Text>
     </Box>
   );
@@ -82,7 +82,14 @@ export function ConversationPane(
           <TranscriptEntry key={entry.id} entry={entry} width={props.width} />
         )}
       </Static>
-      {live ? <LiveTranscriptEntry entry={live} /> : null}
+      {/* Reserve a single line for the live region even when nothing is
+        * streaming. Ink renders Static items into scrollback above the
+        * live area, so toggling live presence between renders made the
+        * input bar appear to "shift down" by a line. A stable footer
+        * keeps the layout pinned. */}
+      <Box minHeight={1}>
+        {live ? <LiveTranscriptEntry entry={live} /> : null}
+      </Box>
     </Box>
   );
 }
--- a/packages/cli/src/chat/tui/panes/ConversationPane.tsx
+++ b/packages/cli/src/chat/tui/panes/ConversationPane.tsx
@@ -83,10 +83,10 @@ export function ConversationPane(
         )}
       </Static>
       {/* Reserve a single line for the live region even when nothing is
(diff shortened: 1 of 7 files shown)


LionSR:
yes we should show error

PR #4160, opened 2026-05-17
fix(cli): cap live streaming region so scrollback isn't truncated

LionSR:
## Summary

Long assistant responses streaming into the chat TUI were truncating earlier turns from terminal scrollback. The culprit: Ink redraws the live region on every text delta by moving the cursor up and rewriting the rows. When the live region grew taller than the viewport (a multi-paragraph response easily does), the cursor-up overshoots and clobbers `<Static>` items that were already pushed into scrollback — so when the user scrolled back they saw missing/torn earlier history.

This PR caps `LiveTranscriptEntry` to the **last 12 lines** of the in-flight entry, with a dim hint showing how many earlier lines are hidden. The full content is never lost: the entry transitions into Ink's `<Static>` the moment it finalizes (which writes it to scrollback in one shot), so all the text reappears intact.

## Why not a deeper fix

Claude Code solves this with a custom Ink fork that diffs frames against terminal scrollback and emits full-reset sequences when the cursor would overshoot. We don't ship a fork. Bounding the live region is the smallest change that stops the regression without restructuring the renderer — a future PR can revisit (e.g. forklift the conversation onto a stdout-write path independent of Ink).

## Test plan

- [ ] `node packages/cli/dist/bin/texra.js --model=deepseekT` — ask something that returns a long response (e.g. "explain tensor networks in detail").
- [ ] During streaming: only the last ~12 lines of the response are visible above the input bar; a dim "⋯ N earlier lines hidden while streaming" line shows above them.
- [ ] After the response finalizes: the full body appears in scrollback (scroll up to see it).
- [ ] Submit several more long-response turns. Scroll back — all earlier turns are intact (no torn / clipped messages).
- [ ] Short responses (≤12 lines) render normally without the truncation hint.

---

> [!NOTE]
> **Low Risk**
> Low risk UI-only change to streaming rendering; main risk is minor regressions in live transcript wrapping/truncation behavior for edge-case terminal widths.
> 
> **Overview**
> Prevents long, streaming assistant responses from overwriting prior `<Static>` transcript rows by **capping the live (in-flight) render to the last 12 wrapped terminal rows** and showing a dim hint when earlier rows are hidden.
> 
> Adds width-aware wrapping via `wrapAnsiToWidth`, plus an incremental newline-count cache (cleared on `cliState` reset) to keep per-delta work bounded while estimating how many rows were truncated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d96aa67c2cd8e626c84a9a6be6e50ce1a453dd0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/tui/panes/ConversationPane.tsx
+++ b/packages/cli/src/chat/tui/panes/ConversationPane.tsx
@@ -55,14 +55,34 @@ function TranscriptEntry({
   );
 }
 
+// Cap the streaming live region to a short tail. Ink redraws the live
+// region on every text delta by moving the cursor up and rewriting the
+// rows; if the live region grows taller than the terminal viewport the
+// cursor-up overshoots and clobbers Static items that have already been
+// pushed into scrollback (truncating earlier turns when the user scrolls
+// back). The full content is preserved — the assistant entry transitions
+// into Ink's <Static> the moment it finalizes, which writes it to
+// scrollback in one shot. Keep this number small; it bounds the
+// scrollback damage during long streaming responses.
+const LIVE_TAIL_LINES = 12;
+
 function LiveTranscriptEntry({
   entry,
 }: {
   readonly entry: ConversationEntry;
 }): React.JSX.Element {
+  const lines = entry.text.split('\n');
+  const hiddenLines = Math.max(0, lines.length - LIVE_TAIL_LINES);
+  const tail = hiddenLines > 0 ? lines.slice(-LIVE_TAIL_LINES) : lines;
   return (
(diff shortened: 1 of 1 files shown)


LionSR:
Filed follow-up #4163 with a deeper-fix design — recommends restructuring the live region so Ink only owns input+status, and pushing finalized transcript entries plus the side panes through stdout. Tagged `ambitious` since it's a UX-affecting restructure, not just a cap.

PR #4421, opened 2026-05-22
fix(cli): render chat in primary buffer so history lands in native scrollback

LionSR:
## What this fixes

The CLI chat collapses history so fast that nothing older than a few rows is readable — and the terminal's own scroll-up is dead while the chat is running. Reported as a release blocker.

The root cause is two compounding properties:

1. **The chat enters the terminal alternate screen** — `terminalCleanup.ts` unconditionally writes `\x1b[?1049h` from `runChatTui.tsx` before `render()`. The alt screen disables the terminal's *own* scrollback by design (Ink's own docs say so).
2. **The conversation pane clips finalized entries to a viewport tail** — `selectFinalizedEntriesForViewport` keeps only the last entries that fit `maxRows` and replaces the rest with `⋯ N earlier entries hidden`, with no scroll offset and no key handler to walk backwards.

So neither native nor in-app history reach existed. #4160 partially helped by capping the live streaming region, but the underlying alt-screen-plus-clipped-viewport problem was untouched.

## The fix (native scrollback via Ink 7 `<Static>`)

Ink 7 made this the idiomatic path: `alternateScreen` is a first-class render option (default off), and `<Static>` is purpose-built to commit immutable items to the terminal's real scrollback. The intent was already in the codebase — `transcriptEntries.ts` and `ConversationPane.vitest.ts` both reference `<Static>` scrollback by name.

- **`state/scrollbackLog.ts` (new)** — append-only signal of finalized entries, merged across all streams. Dedup keyed by `entry.id` so `moveLocalTranscriptToStream` (re-parents an entry; keeps the id) doesn't double-commit. `Signal.subtle.Watcher` on `cliState.streams` triggers `syncScrollbackFromStreams`.
- **`App.tsx`** — renders `<Static items={scrollback}>` at the top of the tree; drops `height={rows}` from the root and the fixed-height side-column reservation so the live frame is content-sized.
- **`panes/ConversationPane.tsx`** — pending/live-only and content-sized; no clipped finalized viewport, no fixed-height box reserving blank rows. Exports `TranscriptEntry` so `<Static>` can render committed history with the same formatting. The pure `select*` / `estimate*` helpers are kept exported so `TuiStateAndFocus.vitest.mts` still has its guards.
- **`runChatTui.tsx`** — removes `enterTerminalFullScreen()`; adds `disposers.push(startScrollbackSync())`. The render now lives in the primary buffer.
- **`terminalCleanup.ts`** — drops the now-unused `enterTerminalFullScreen`; the reset string still resets mouse / kitty / bracketed-paste / cursor; `?1049l` kept defensively (no-op when no alt screen was entered).

Result for the user: two-finger trackpad swipe / mouse wheel scrolls through every past message, text selection and copy work, Cmd+F searches the history — exactly the way a normal shell session behaves.

## Verification

- **Type-check** (`packages/cli`): clean.
- **`check:architecture`**: clean.
- **TUI tests**: **54 / 54 pass** — including 7 new `ScrollbackLog.vitest.ts` cases that lock the data layer feeding `<Static>`:
  - commits only finalized entries, in stream order
  - append-only — committed items are never dropped or reordered (existing item identity is preserved)
  - exactly once across repeated sync calls
  - **no double-commit when an entry is re-parented to another stream (same id)**
  - cross-stream merge into one linear log
  - promotion when an entry transitions to `finalized: true`
  - clean reset on session reset
- **PTY smoke** (real `texra chat` under macOS `script`): zero `\x1b[?1049h` in the capture (alt screen is gone). A submitted user line lands once at the top of the capture — committed to scrollback — while the header / input / status live frame redraws below.

## Trade-offs

- **Header stays pinned in the live frame** (above the prompt, below the rule), instead of committing once to `<Static>` as a one-shot banner. Simpler, lower-risk, keeps session context persistently visible. Easy to flip later.
- **Multi-stream Tab switching** only drives the live region. Sub-agent finalized entries interleave into the single linear scrollback in commit order — deliberately no per-stream visual labels in this first cut. If interleaving feels noisy in practice, labeling is a tiny follow-up.
- **`select*` / `estimate*` helpers in `ConversationPane`** are kept exported but unused in production (still exercised by tests). Pruning them belongs in a separate cleanup PR.
- **No mouse-tracking sequences enabled.** Trackpad scroll works through the terminal directly — no `\x1b[?1000/?1006` plumbing, no broken text selection.

## Files

- `packages/cli/src/chat/tui/state/scrollbackLog.ts` (new)
- `packages/cli/src/chat/tui/App.tsx`
- `packages/cli/src/chat/tui/panes/ConversationPane.tsx`
- `packages/cli/src/chat/tui/panes/HeaderPane.tsx` (comment only)
- `packages/cli/src/chat/tui/runChatTui.tsx`
- `packages/cli/src/chat/tui/terminalCleanup.ts`
- `src/test-kernel/cli/ScrollbackLog.vitest.ts` (new)
- `CHANGELOG.md`

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes core CLI TUI rendering behavior (switching away from alternate screen and refactoring transcript rendering), which could affect layout, performance, or stream ordering during long/tool-heavy sessions.
> 
> **Overview**
> Enables full terminal-native scrollback for `texra chat` by rendering the Ink TUI in the primary buffer and committing finalized transcript entries via Ink 7’s `<Static>` instead of keeping history in a clipped in-app viewport.
> 
> This introduces a new append-only `scrollbackLog` store synced from finalized stream entries, updates `App` to render committed history above a pinned live frame, and refactors `ConversationPane` to render *live/pending only* (removing the `⋯ N earlier entries hidden` behavior). It also removes the explicit alternate-screen entry helper while keeping defensive terminal mode cleanup, adds sync wiring in `runChatTui`, and includes new unit tests covering scrollback append/dedup/reset behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa172e096a67254d4cb4474e7f45011ec2934400. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- **Scroll back through the full chat history** — the CLI chat now renders to your terminal's primary screen and commits finalized messages to native scrollback. Two-finger swipe (or mouse wheel) scrolls up through every past message; selection, copy, and Cmd+F all work. Long replies are no longer collapsed behind a `⋯ N earlier entries hidden` marker.
 - **CLI chat stays pinned during long output** — the chat header, input, and status rows now stay visible while long replies and tool output stream in the terminal.
 
 ## [0.37.9] - 2026-05-18
--- a/packages/cli/src/chat/tui/App.tsx
+++ b/packages/cli/src/chat/tui/App.tsx
@@ -1,14 +1,14 @@
 // Ink TUI root: conversation, optional side column, status, approval modal,
 // and input bar. Tab / Shift-Tab cycles focus across subagent streams.
 
-import { Box, Text, useInput, useWindowSize } from 'ink';
+import { Box, Static, Text, useInput, useWindowSize } from 'ink';
 import { useState } from 'react';
 
 import { SLASH_PALETTE_ROWS } from './commands/SlashPalette';
 import { REVERSE_SEARCH_ROWS } from './input/ReverseSearch';
 import { ApprovalModal } from './modals/ApprovalModal';
 import { ChildControlPicker } from './modals/ChildControlPicker';
-import { ConversationPane } from './panes/ConversationPane';
+import { ConversationPane, TranscriptEntry } from './panes/ConversationPane';
 import { HeaderPane } from './panes/HeaderPane';
 import { InputBar } from './panes/InputBar';
 import { StatusBar } from './panes/StatusBar';
@@ -24,6 +24,7 @@ import {
 } from './state/childControls';
 import { canShowSubagentControls, cliState } from './state/cliState';
 import { nextFocusBack, nextFocusForward } from './state/focusCycle';
+import { scrollbackLog } from './state/scrollbackLog';
 import { useSignal } from './state/useSignal';
 import type { InputHistory } from './history/inputHistory';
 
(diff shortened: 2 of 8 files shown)


LionSR:
need to rebase

PR #4479, opened 2026-05-24
fix(core): make model/auth error messages host-neutral

### LINKED ISSUE #4478: CLI: model/auth errors tell users to open VS Code UI that doesn't exist in the terminal
## Summary

While testing `texra chat` hands-on, the very first message in the default relay (included-access) configuration fails with:

> Model "deepseekT" is not available with your current subscription tier. **Switch to "Use My Own Keys" via the TeXRA Profile panel**, or select a model included in your tier.

The CLI has no "TeXRA Profile panel" — that's a VS Code webview. A terminal user has no way to act on that guidance. The CLI equivalent is the `/api` command (or `--api-mode personal`).

## Root cause

Several user-facing error strings live in the **host-agnostic core** (`src/agent`, `src/model`) yet hard-code VS Code-specific remediation:

- `src/agent/modelHandlers/ModelHandler.ts` — `"...via the TeXRA Profile panel..."`, `'...the "Set API Key" command.'`
- `src/model/computeModelOptions.ts` — `"...Set your OpenRouter key in the extension settings."`, `"...Set it in the extension settings..."`
- `src/model/apiProviders.ts` — `'...the "Set API Key" command...'`

This violates the platform-agnosticism rule in CLAUDE.md: core business logic should not embed host-specific UI references; it should return host-neutral results and let the host enrich at the command layer.

## Repro

1. `texra chat --api-mode included` with the built-in default model on an account whose tier/quota excludes that model.
2. Send any message → error references the nonexistent "Profile panel".

## Suggested fix

Reword the core strings to host-neutral guidance (e.g. "Switch to personal API keys", "Set the `<ENV>` environment variable"). Hosts can still layer richer actions on top. (PR incoming.)

## Related (not fixed here)

- `texra models list` reports availability by **personal API key**, not by relay tier — so a model shows "api key set" while relay rejects it. Worth a separate look.
- The error attributes failures to "subscription tier" even when the real cause is relay **quota exhaustion** ("0% remaining" in `/api`).

LionSR:
Closes #4478

## What

Reword six user-facing error strings in host-agnostic core so they no longer point at VS Code-only UI ("extension settings", "TeXRA Profile panel", the `"Set API Key"` command). The same messages now read correctly from both the extension and the CLI.

| File | Before | After |
|------|--------|-------|
| `ModelHandler.ts` (tier) | `Switch to "Use My Own Keys" via the TeXRA Profile panel, …` | `Switch to personal API keys, …` |
| `ModelHandler.ts` (OpenRouter / provider key) | `… the "Set API Key" command.` | `Set your … API key to continue.` |
| `apiProviders.ts` | `… the "Set API Key" command or <ENV> environment variable.` | `Set the <ENV> environment variable, or configure your … API key.` |
| `computeModelOptions.ts` (×2) | `… in the extension settings.` | host-neutral wording |

## Why

Found while driving `texra chat` hands-on: a relay user's first message fails and tells them to open a "Profile panel" that doesn't exist in a terminal. Embedding host-specific UI in core also violates the platform-agnosticism rule in CLAUDE.md (core returns host-neutral results; hosts enrich at the command layer).

## Scope notes

- Left `executeAgent.ts`'s `requestShowInstruction` untouched: it already passes structured `actions` for the host to render/interpret, so the host owns that UI.
- Pure string changes; no behavior change. CLI `build` (typecheck + architecture + react-compiler smoke + bundle) passes; verified the new text live in a relay `texra chat` session.

## Verification

Before/after confirmed by reproducing the error in `texra chat --api-mode included`:
```
Model "deepseekT" is not available with your current subscription tier. Switch to personal API keys, or select a model included in your tier.
```

### DIFF
--- a/src/agent/modelHandlers/ModelHandler.ts
+++ b/src/agent/modelHandlers/ModelHandler.ts
@@ -336,7 +336,7 @@ export abstract class ModelHandler<
       // No access token available - shouldUseServerSideKeys() returned true, meaning isEnabled()
       // returned true. Don't fall back to personal keys - throw an actionable error.
       throw new Error(
-        'Unable to authenticate with server. Please sign out and sign back in, or switch to "Use My Own Keys" mode.',
+        'Unable to authenticate with server. Please sign out and sign back in, or switch to personal API keys.',
       );
     }
 
--- a/src/agent/modelHandlers/ModelHandler.ts
+++ b/src/agent/modelHandlers/ModelHandler.ts
@@ -345,7 +345,7 @@ export abstract class ModelHandler<
     if (shouldUseOpenRouter(this.config)) {
       return this.fetchApiKeyOrThrow(
         'openRouter',
-        'Missing API key for OpenRouter. Please set it using the "Set API Key" command.',
+        'Missing API key for OpenRouter. Set your OpenRouter API key to continue.',
       );
     }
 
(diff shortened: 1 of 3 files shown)


LionSR:
Thanks — good catch. Folded the `getApiKey` server-auth failure (`ModelHandler.ts:339`) into this PR: `… or switch to "Use My Own Keys" mode.` → `… or switch to personal API keys.` Now the whole `getApiKey` path is host-neutral.

Agreed on leaving `executeAgent.ts:235` out of scope — it carries structured `actions` the host re-interprets, so the host owns that surface.

PR #4492, opened 2026-05-25
refactor(scripts): simplify build and verification scripts

LionSR:
Non-behavioral cleanup pass over `packages/cli/scripts/` and repo-root `scripts/`, driven by a parallel `/simplify` review (reuse / quality / efficiency).

## Changes
- **smoke-react-compiler.mjs** — replace the single-field `recorded` wrapper object and the `fakeBuildApi._handler` side-channel with plain closure variables (`loadHandler`, `contents`).
- **check-host-imports.mjs** — strict `!== null` for the `regex.exec` loop guard (`exec` returns `null`, never `undefined`).
- **aliases.mjs / aliasUtils.mjs** — `node:` specifier prefix for built-in imports, matching every other script in the repo.
- **bump-workspace-version.mjs** — drop the redundant `// Node.js imports` what-comment.
- **smoke-desktop-package-launch.mjs** — import the shared `readPositiveNumber` from `smoke-process-utils.mjs` instead of redefining an identical copy; drop the unused `false` return from `failIfFatalLog` (no caller reads it).
- **verify-desktop-package.mjs** — comment now states *why* `statFile` throws (path not in archive) rather than restating control flow.

No behavior changes; `node --check` and Prettier pass on all touched files.

## Reviewed but deliberately not applied
- Sharing the `react-devtools-core` stub path between `build-bundle.mjs` and `build-harness.mjs` — `build-harness.mjs` is untracked WIP on another branch, and exporting a stub path from the React-compiler plugin module leaks an unrelated concern.
- Guarding `stripPattern` with a match check in `check-host-imports.mjs` — premature optimization for a lint script over 3 small files.
- Removing the provably-no-op self-visit in `sync-remote-agents.mjs` `walkParents` — correctness depends on the inheritance callback's guards; left for a human call.

---

> [!NOTE]
> **Low Risk**
> Script-only hygiene with no production code or behavioral changes; smoke and verify paths stay equivalent.
> 
> **Overview**
> Non-behavioral cleanup across `packages/cli/scripts/` and repo-root `scripts/`: simpler smoke-test wiring, shared helpers, and consistent Node import style.
> 
> The React compiler smoke test drops the `recorded` object and `fakeBuildApi._handler` side channel in favor of plain `loadHandler` and `contents`. CLI host-import linting uses `!== null` for `RegExp#exec` loop guards. Alias scripts import built-ins via the `node:` prefix; the workspace version bump script loses a redundant comment.
> 
> Desktop launch smoke imports `readPositiveNumber` from `smoke-process-utils.mjs` instead of a local duplicate, and `failIfFatalLog` no longer returns an unused `false`. A comment in desktop package verification explains why `statFile` failures are retried with alternate ASAR paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7c9d77c212de1e4e592ff0ccc1760ab418d1ae6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/scripts/check-host-imports.mjs
+++ b/packages/cli/scripts/check-host-imports.mjs
@@ -73,7 +73,7 @@ function importSources(source) {
   ];
   for (const pattern of patterns) {
     let match;
-    while ((match = pattern.exec(source)) != null) {
+    while ((match = pattern.exec(source)) !== null) {
       sources.push(match[1]);
     }
   }
--- a/packages/cli/scripts/smoke-react-compiler.mjs
+++ b/packages/cli/scripts/smoke-react-compiler.mjs
@@ -21,40 +21,37 @@ import { reactCompilerPlugin } from './reactCompilerPlugin.mjs';
 const here = dirname(fileURLToPath(import.meta.url));
 const smokeFile = resolve(here, '../src/chat/tui/App.tsx');
 
-const recorded = {
-  contents: undefined,
-};
-
+let loadHandler;
 const fakeBuildApi = {
   onLoad(_filter, handler) {
-    fakeBuildApi._handler = handler;
+    loadHandler = handler;
   },
 };
 
 const plugin = reactCompilerPlugin();
 plugin.setup(fakeBuildApi);
 
-if (typeof fakeBuildApi._handler !== 'function') {
+if (typeof loadHandler !== 'function') {
   console.error('react-compiler smoke: plugin did not register onLoad hook');
   process.exit(1);
 }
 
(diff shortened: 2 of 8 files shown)


LionSR:
Fixed in b7c9d77c2 — `scripts/aliasUtils.mjs:11` now uses `node:path`. Confirmed with a grep that no bare built-in imports remain under `scripts/` or `packages/cli/scripts/`. Thanks for the catch. The other items in the review were confirmations, so nothing else to change.

PR #4513, opened 2026-05-27
feat(cli): detect newer npm release and offer to self-update

LionSR:
## Summary

Adds an auto-update detector to the `texra` CLI. On every interactive launch (`texra orchestrate` / `texra chat`, including bare `texra` in a TTY), the CLI checks the npm registry for a newer `@texra-ai/cli` release and, when found, offers to run the install for you.

- **Detects the install method** (npm / pnpm / yarn / bun) from the path the binary runs out of, so the suggested command matches how the user actually installed it (`npm install -g …`, `pnpm add -g …`, `yarn global add …`, `bun add -g …`).
- **Notify + prompt** — never installs without confirmation. Prints `current → latest`, asks `Update now? [Y/n]`, then runs the global install (stdio inherited so the user sees progress) and tells them to restart.
- **Stays out of the way** — fire-and-forget with a 2.5s registry timeout, fails silently when offline. Skipped in CI, with `--output-format ndjson`, or `--quiet`; only prompts on an interactive TTY. A once-per-process guard prevents the `orchestrate → chat` path from prompting twice.
- **Opt-out**: `TEXRA_NO_UPDATE_CHECK=1` disables the check entirely.

## Implementation

- New `packages/cli/src/runtime/updateChecker.ts` — pure helpers (`parseSemver`, `isNewerVersion`, `detectInstallMethod`, `buildUpdateCommand`, `fetchLatestCliVersion`) plus the `notifyCliUpdate(context)` orchestrator. Uses native `fetch` and `node:child_process` — no new dependencies. A lightweight semver compare (release ranks above its prerelease) avoids pulling in `semver`.
- Wired into `commands/orchestrate.ts` (after the headless guard) and `commands/chat.ts` (before the TUI launches).

## Testing

- `src/test-kernel/cli/UpdateChecker.vitest.mts` — 12 tests covering semver parsing/comparison (incl. prerelease ranking), install-method detection across pnpm/yarn/bun/npm layouts, command formatting per manager, and `fetchLatestCliVersion` happy path + non-ok / offline / missing-version fallbacks (injected `fetchImpl`).
- `npm run typecheck` ✅ · `eslint` on touched files ✅ · vitest ✅

## Verified

- `packages/cli/src/commands/orchestrate.ts:42` — `notifyCliUpdate` runs only after the interactive-TTY guard, before platform init.
- `packages/cli/src/commands/chat.ts` — checked before `runChat`; the once-per-process guard means orchestrate→chat does not double-prompt.
- `packages/cli/src/runtime/logSinks.ts` — reused existing `askCliQuestion` / `writeTextStderr` for the prompt and notices.

---

> [!NOTE]
> **Medium Risk**
> On user confirmation it runs a global package-manager install via shell spawn and fetches version metadata from the public npm registry; failures are silent but a mistaken affirmative could modify the user's global CLI install.
> 
> **Overview**
> Interactive **`texra chat`** and **`texra orchestrate`** now run a **once-per-process** npm registry check for a newer **`@texra-ai/cli`** release before the TUI starts. When a newer version is found, the CLI prints **`current → latest`**, infers **npm / pnpm / yarn / bun** from the install path, and—only on a full interactive TTY—asks whether to run the matching **global `@latest` install**; declining still prints the manual command.
> 
> The logic lives in new **`updateChecker`** helpers (semver compare, registry fetch with timeout, command building, optional **`spawn`** with inherited stdio). Checks are skipped for **CI**, **`TEXRA_NO_UPDATE_CHECK`**, **ndjson / quiet**, and non-TTY streams. **`semver`** is added as a CLI dependency, with **vitest** coverage for version comparison, install detection, commands, and registry fetch fallbacks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5da60370562c20bea0d694d7058e7f6308841167. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -70,6 +70,8 @@
     "p-queue": "^9.3.0",
     "picocolors": "^1.1.1",
     "react": "^19.2.0",
+    "semver": "^7.8.1",
+    "@types/semver": "^7.7.1",
     "string-width": "^8.0.0",
     "typescript": "^6.0.3",
     "wrap-ansi": "^10.0.0"
--- a/packages/cli/src/commands/chat.ts
+++ b/packages/cli/src/commands/chat.ts
@@ -20,6 +20,8 @@ export const chatCommand = defineCommand({
     rejectHeadlessOnlyFlags(ctx.rawArgs, 'chat');
     const modelOverride = assertExplicitModelKnown(optString(ctx.args.model));
     const context = await contextFromArgs(ctx.args);
+    const { notifyCliUpdate } = await import('../runtime/updateChecker');
+    await notifyCliUpdate(context);
     const { runChat } = await import('../chat/tui/runChatTui');
     const result = await runChat(context, {
       agentOverride: optString(ctx.args.agent),
--- a/packages/cli/src/commands/chat.ts
+++ b/packages/cli/src/commands/chat.ts
@@ -1,5 +1,7 @@
 import { defineCommand } from 'citty';
 
+import { notifyCliUpdate } from '../runtime/updateChecker';
+
 import { contextFromArgs } from './_helpers/context';
 import { setExitCode } from './_helpers/exitCode';
 import {
@@ -20,7 +22,6 @@ export const chatCommand = defineCommand({
     rejectHeadlessOnlyFlags(ctx.rawArgs, 'chat');
     const modelOverride = assertExplicitModelKnown(optString(ctx.args.model));
     const context = await contextFromArgs(ctx.args);
-    const { notifyCliUpdate } = await import('../runtime/updateChecker');
     await notifyCliUpdate(context);
     const { runChat } = await import('../chat/tui/runChatTui');
     const result = await runChat(context, {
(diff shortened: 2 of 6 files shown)


LionSR:
## Addressed Bugbot review items

| # | Issue | Status |
|---|-------|--------|
| P2 | Move update prompt after chat TTY validation | ✅ **Fixed** in `0b7db4d8a` — `notifyCliUpdate` now checks all three `ambient.*IsTty` flags (stdin, stdout, stderr) before prompting, so a stdout-redirected invocation is handled before any prompt appears |
| Medium | Spawn without shell breaks update on Windows | ✅ **Fixed** in `c3acf76c` — added `shell: true` to the `spawn` call in `runCliUpdate`, matching the codebase convention in `tools.ts::shellRun` |
| Low | Inconsistent import style | ✅ **Fixed** in `bea2b2460` — both `orchestrate.ts` and `chat.ts` now use static `import { notifyCliUpdate }` |
| Low | Exported function never imported/used | ✅ **Fixed** in `bea2b2460` — removed `resetUpdateNotifierForTests` and unused test-only exports |

PR #5012, opened 2026-06-01
Replace custom write locks with async-lock and write-file-atomic

LionSR:
## Summary

Replaces three custom write-lock implementations with the `async-lock` library, and replaces the manual atomic write implementation in `JsonStore` with the `write-file-atomic` library. This reduces code complexity and leverages well-tested, production-grade libraries for these critical operations.

### `src/platform/defaults/jsonStore.ts`
- Removed custom `writeChain` Promise-based write queue implementation
- Replaced manual atomic write logic (temp file + rename) with `write-file-atomic` library
- Simplified `flush()` method from ~20 lines to 4 lines
- Removed `enqueueFlush()` method entirely
- `write-file-atomic` handles fsync, temp file cleanup, and serialization internally

### `src/tools/inquiry/externalInquiryStorage.ts`
- Removed custom `threadLocks` Map and `withThreadLock()` function (~25 lines)
- Replaced with `AsyncLock` instance for per-thread write serialization
- Updated three call sites: `recordOpenQuestion()`, `recordAnswerForOpenTurn()`, `markDropped()`, and `persistOpenTurnDraft()`
- Changed from `withThreadLock(threadId, fn)` to `threadLock.acquire(threadId, fn)`

### `src/tools/lean/direct/lakeCommands.ts`
- Removed custom `workspaceMutexes` Map and manual Promise-chaining logic (~15 lines)
- Replaced with `AsyncLock` instance for per-workspace command serialization
- Simplified `runLakeCommand()` from manual Promise chain to single `workspaceLock.acquire()` call

### `package.json`
- Added `async-lock@^1.4.1` dependency
- Added `write-file-atomic@^7.0.1` dependency
- Added `@types/async-lock@^1.4.2` dev dependency
- Added `@types/write-file-atomic@^4.0.3` dev dependency

## Issue acceptance gates

N/A — this is a refactoring to improve code maintainability and reliability.

## Single source of truth

Lock management is now delegated to the `async-lock` library, which owns the invariant that concurrent operations on the same key are serialized. Atomic file writes are delegated to `write-file-atomic`, which owns the invariant that writes are crash-safe.

Removed ~60 lines of custom lock and atomic-write boilerplate across three modules. These implementations were error-prone (manual Promise chaining, temp file cleanup) and duplicated logic that is better handled by battle-tested libraries.

## Host impact

**Extension:** No user-facing changes. Internal write serialization for inquiry storage remains unchanged in behavior.

**Desktop:** No user-facing changes. Lake command serialization remains unchanged in behavior.

**CLI:** No impact (does not use these modules).

## Scheduled refactoring

None.

## Validation

- TypeScript compilation passes with new type definitions
- No logic changes to write serialization or atomicity guarantees
- Existing tests should pass without modification (behavior is identical)

https://claude.ai/code/session_0139mHcyBMB6sRwFy7E6nuQ9

---

> [!NOTE]
> **Medium Risk**
> Touches crash-safe JSON persistence and per-thread inquiry / per-workspace lake serialization; behavior should match prior logic but ordering and atomic-write semantics now depend on library behavior.
> 
> **Overview**
> This PR swaps hand-rolled concurrency and persistence helpers for **`async-lock`** and **`write-file-atomic`**, with matching type packages and lockfile updates.
> 
> **`JsonStore`** still chains flushes on an internal promise queue so `set()` snapshots land on disk in call order, but each flush now uses **`write-file-atomic`** instead of a manual temp file, rename, and cleanup path.
> 
> **External inquiry** thread mutations (`recordOpenQuestion`, answer/drop/draft paths) now serialize per `threadId` via a shared **`AsyncLock`** instead of a per-thread promise map and custom `withThreadLock`.
> 
> **Lean `lake` commands** with `serialize: true` use the same pattern: one **`AsyncLock`** keyed by resolved workspace root replaces promise-chained workspace mutexes in **`runLakeCommand`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41a06f971f5bd27c0f482222069f6a2846fccf6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "@electron/asar": "^4.2.0",
     "@eslint/js": "^9.39.4",
     "@stylistic/eslint-plugin": "^5.10.0",
+    "@types/async-lock": "^1.4.2",
     "@types/diff-match-patch": "^1.0.36",
     "@types/fs-extra": "^11.0.4",
     "@types/he": "^1.2.3",
@@ -79,6 +80,7 @@
     "@types/sortablejs": "^1.15.9",
     "@types/turndown": "^5.0.6",
     "@types/vscode": "^1.105.0",
+    "@types/write-file-atomic": "^4.0.3",
     "@typescript-eslint/eslint-plugin": "^8.60.0",
     "@typescript-eslint/parser": "^8.60.0",
     "@vscode/test-cli": "^0.0.12",
@@ -129,6 +131,7 @@
     "@xterm/addon-fit": "^0.11.0",
     "@xterm/xterm": "^6.0.0",
     "arxiv-client": "^0.0.9",
+    "async-lock": "^1.4.1",
     "axios": "^1.16.1",
     "bibtex": "^0.9.0",
     "bottleneck": "^2.19.5",
@@ -174,6 +177,7 @@
     "tar": "^7.5.15",
     "turndown": "^7.2.4",
     "turndown-plugin-gfm": "^1.0.2",
+    "write-file-atomic": "^7.0.1",
     "yaml": "^2.9.0",
     "zod": "^4.4.3"
   }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -71,6 +71,9 @@ importers:
       arxiv-client:
         specifier: ^0.0.9
         version: 0.0.9
+      async-lock:
+        specifier: ^1.4.1
+        version: 1.4.1
       axios:
         specifier: ^1.16.1
         version: 1.16.1
@@ -206,6 +209,9 @@ importers:
       turndown-plugin-gfm:
         specifier: ^1.0.2
         version: 1.0.2
+      write-file-atomic:
+        specifier: ^7.0.1
+        version: 7.0.1
       yaml:
         specifier: ^2.9.0
         version: 2.9.0
@@ -222,6 +228,9 @@ importers:
       '@stylistic/eslint-plugin':
         specifier: ^5.10.0
         version: 5.10.0(eslint@9.39.4(jiti@2.7.0))
+      '@types/async-lock':
+        specifier: ^1.4.2
+        version: 1.4.2
       '@types/diff-match-patch':
         specifier: ^1.0.36
         version: 1.0.36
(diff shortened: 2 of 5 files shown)


LionSR:
Rebased onto latest `main` (now at `41a06f971`). **No conflicts** — the rebase applied cleanly; the PR's three source files (`jsonStore.ts`, `externalInquiryStorage.ts`, `lakeCommands.ts`) are byte-identical to the pre-rebase versions. typecheck passes; the 3 relevant vitest suites (JsonStore, LakeCommandsMutex, InquiryStorage — 12 tests) all pass.

### Review comments

**Addressed earlier / now stale (no action needed):**
- Copilot's two findings on `jsonStore.ts` (out-of-order persistence + stale docstring) were already fixed by commit `fix(jsonStore): restore flush ordering queue` — `writeChain` is in place and the docstring (lines 24-34) now documents the call-order guarantee. Verified against current file state.

**Verified incorrect (no change):**
- *texra-review: `write-file-atomic@7.0.1` is ESM-only → `ERR_REQUIRE_ESM` risk.* The installed v7.0.1 is **CommonJS** — its `package.json` has no `"type": "module"` and `lib/index.js` is `'use strict'; module.exports = writeFile`. No ESM constraint exists, so no downgrade is needed.
- *texra-review/texra-review: remove `@types/write-file-atomic` because v7 bundles its own types.* v7.0.1 ships **no** `.d.ts` (no `types` field, no declaration files). The `@types/write-file-atomic@4.0.3` package is required, and its signature (`writeFile(filename, data, options?): Promise<void>`) exactly matches the usage. The version-number mismatch is cosmetic.

**Deferred (design decision, explicitly not a regression):**
- *texra-review: add `timeout` to the two `AsyncLock` instances.* The reviewer notes this is not a regression (the original hand-rolled mutexes had no timeout either). Adding a hard lock timeout could introduce new failure modes during legitimately long-running operations (e.g. a slow `lake build`), so this is a product/design call left for a maintainer rather than guessed at here.

LionSR:
**Decision — `AsyncLock` timeout: not adding one in this PR.**

This PR's goal is to swap the hand-rolled mutexes for `async-lock` while *preserving* behavior, and the originals had no timeout. A hard timeout is new behavior with real risk: `lake build` can legitimately run for minutes, and a merely-slow filesystem op shouldn't abort the serialized queue. If we want lock-hang protection it's better as a separate, deliberate change with a per-lock budget chosen on purpose — tracking it as a possible follow-up rather than blocking this PR.

The ESM-only / redundant-`@types` warnings remain false positives: `npm view write-file-atomic@7.0.1` shows no `types` field (so `@types/write-file-atomic` is needed), and `validate` is green (ESM→CJS handled by the bundler). No change.

PR #5019, opened 2026-06-01
fix(tools): inherit parent bash bypass on child subagent streams

LionSR:
## Context

Follow-up to #5014, addressing the substantive (non-blocking) concern raised in that PR's TeXRA review:

> **`enableYoloOnChildStream` does not mirror bash bypass.** After #5014, child streams spawned with YOLO enabled will have tool-edit bypass but not bash bypass. This is now an inconsistency with the extension shield behavior.

## Problem

`enableYoloOnChildStream` (in `toolEditApproval.ts`) only set the **tool-edit** bypass on a freshly resolved child subagent stream. So when a parent delegates under YOLO / delegated-task auto-approval (`DelegationTools.proposeAndExecute`), the child inherits edit auto-approval but **keeps prompting for every bash command** — the same gap #5014 fixed for the main shield, one level down.

## Fix

Mirror the parent's bash bypass onto the child, **independently** of the tool-edit flag:

```ts
export function enableYoloOnChildStream(
  childStreamId: StreamTabId,
  parentStreamId?: StreamTabId,
): void {
  toolEditApprovalController.setBypass(childStreamId, true);
  if (parentStreamId && bashApprovalController.isBypassed(parentStreamId)) {
    bashApprovalController.setBypass(childStreamId, true);
  }
}
```

Both `DelegationTools` call sites now pass the orchestrator (parent) stream id.

### Why parent-state inheritance (not unconditional)

- **Extension:** the YOLO shield keeps a stream's bash and tool-edit bypass symmetric (#5014), so the child inherits both — fixing the bug.
- **CLI:** `AUTO-BASH` and `AUTO-APPROVE` are independent, intentional grants. A parent with edits auto-approved but bash still gated propagates **only** the edit bypass to the child — bash stays gated, preserving the CLI's safety distinction.

Tool-edit remains unconditional (callers only invoke this when the parent is already auto-approving edits / delegated tasks), matching prior behavior.

## Testing

- New `ChildStreamYoloInheritance.vitest.ts` covers both directions:
  - parent bash bypassed → child bash bypassed
  - parent bash gated → child bash gated (tool-edit still on)
- `npx vitest run` on the new file — 2 passed
- `npm run typecheck` — passes
- `prettier --check` on all changed files — clean

(Concern B — Super-YOLO toast wording — left as-is per the review; "auto-approval" is generic and accurate. Concern C — no `ProgressViewMessageHandler` unit tests — is a pre-existing gap; this PR adds coverage for the core propagation it touches.)

---

> [!NOTE]
> **Medium Risk**
> Changes approval bypass propagation for delegated subagents (bash execution gating), which affects safety behavior but is narrowly scoped with targeted tests.
> 
> **Overview**
> Child subagent streams now **inherit the parent’s bash-approval bypass** when delegation resolves a child stream, via a new **`inheritBashBypassOnChildStream`** helper that only mirrors bash state (not tool-edit YOLO).
> 
> **`DelegationTools`** invokes it in both `onStreamResolved` paths (headless and async) **before** optional `enableYoloOnChildStream`, so a parent with **AUTO-BASH** but edits still gated propagates bash to the child without forcing edit auto-approval. **`enableYoloOnChildStream`** remains tool-edit-only.
> 
> New **`ChildStreamYoloInheritance.vitest.ts`** covers parent→child bash mirroring, gated parents, bash-only vs edit-YOLO independence, and edit-YOLO without bash; headless delegation tests mock the new export.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b333c81a4ffbd57e44fee33a269e0f4fb8c24f24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/agent/toolUse/ChildStreamYoloInheritance.vitest.ts
+++ b/src/test-kernel/agent/toolUse/ChildStreamYoloInheritance.vitest.ts
@@ -1,0 +1,47 @@
+// Third-party imports
+import { afterEach, describe, expect, it } from 'vitest';
+
+// Local imports
+import type { StreamTabId } from '@shared/schemas';
+import {
+  cleanupAllApprovals,
+  enableYoloOnChildStream,
+  isApprovalBypassedForStream,
+  isBashApprovalBypassedForStream,
+  setBashApprovalSessionBypass,
+} from '@tools/approval';
+
+import { createRecordingHost } from '../progressTestUtils';
+
+describe('enableYoloOnChildStream bash inheritance', () => {
+  afterEach(() => {
+    cleanupAllApprovals();
+  });
+
+  it('mirrors the parent bash bypass onto the child stream', () => {
+    const { host } = createRecordingHost();
+    const parent = 'stream:parent-yolo' as StreamTabId;
+    const child = 'stream:child-yolo' as StreamTabId;
+    setBashApprovalSessionBypass(parent, true, host, { silent: true });
+    // Sanity: the parent bypass round-trips through the public barrel.
+    expect(isBashApprovalBypassedForStream(parent)).toBe(true);
+
+    enableYoloOnChildStream(child, parent);
+
+    // Tool-edit bypass is always enabled for the child; bash follows the parent.
+    expect(isApprovalBypassedForStream(child)).toBe(true);
+    expect(isBashApprovalBypassedForStream(child)).toBe(true);
+  });
+
+  it('keeps child bash gated when the parent still prompts for bash', () => {
+    // Mirrors the CLI case where AUTO-APPROVE (tool edits) is on but AUTO-BASH
+    // is off: the child inherits exactly the parent's bash state.
+    const parent = 'stream:parent-edit-only' as StreamTabId;
+    const child = 'stream:child-edit-only' as StreamTabId;
+
+    enableYoloOnChildStream(child, parent);
+
+    expect(isApprovalBypassedForStream(child)).toBe(true);
+    expect(isBashApprovalBypassedForStream(child)).toBe(false);
+  });
+});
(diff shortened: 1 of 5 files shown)


LionSR:
@address

LionSR:
Addressed both review findings in `b333c81` — they shared one root cause.

**Codex (P2, `DelegationTools.ts:489`) — bash mirroring was gated behind edit-YOLO.** The `enableYoloOnChildStream` call only fired when `options.enableYoloOnChild` was true, and that flag is derived solely from the parent's *tool-edit* bypass (`isApprovalBypassedForStream` → `toolEditApprovalController.isBypassed`). So a parent with **AUTO-BASH but not AUTO-APPROVE** (the CLI case where the two are independent) skipped the call entirely and the child kept prompting for bash — contradicting this PR's own "mirror the parent's bash bypass **independently of the tool-edit flag**" intent.

Fix — split the two concerns:
- `inheritBashBypassOnChildStream(child, parent?)` — mirrors the parent's bash bypass, called **unconditionally** on child stream resolution.
- `enableYoloOnChildStream(child)` — sets **only** the tool-edit bypass, still gated by `enableYoloOnChild`.

Both `onStreamResolved` sites now always mirror bash, then conditionally enable edit-YOLO.

**Copilot (`toolEditApproval.ts:423`) — docstring inaccuracy.** Same root cause: the old comment claimed `SYNC_STREAM_CONTENT` "carries the bypass state" and the child "inherits exactly the parent's bash state." Rewritten accurately — SYNC carries the tool-edit/super-YOLO bypass; bash is applied directly here and only when the parent has it.

**Tests:** `ChildStreamYoloInheritance` now covers the regression (bash mirrored with edit-YOLO **off**) plus the gated and edit-only cases; the `DelegationHeadless` mock gains `inheritBashBypassOnChildStream`. 7/7 pass; `tsc` (root + test-kernel) and `prettier --check` clean.

(Concern B — Super-YOLO toast wording — and Concern C — `ProgressViewMessageHandler` tests — remain as-is per the original PR.)

Issue #5451, opened 2026-06-06
Unify the progress-view backend across extension + desktop (relocate to shared core)


LionSR:
## Summary

The Electron desktop's `DesktopProgressBridge` (`packages/desktop/src/main/desktopAgentExecution.ts`, ~1400 lines) is a **parallel reimplementation** of the VS Code extension's progress-view backend — the event→message mapping (`ProgressEventHandler` + `WebviewUpdater`) *and* the inbound-command dispatch (`ProgressViewMessageHandler`). Both produce the identical `ProgressViewOutboundMessage` protocol for the **same** Lit frontend (`@progressView/frontend`), and both handle the same `PROGRESS_VIEW_COMMANDS`. Maintaining two implementations of the same command/status/event logic is a standing drift hazard (agent status, stream commands, and message shapes must be kept in lockstep by hand).

This was previously deferred (the plan's "DROP the god-bridge" call) because the two hosts owned state differently. **That premise is now gone:** per-stream state is consolidated behind one host-neutral `StreamSnapshotStore` + `ProgressViewState`, so the remaining divergence is *incidental* (transport + a few host callbacks), which is exactly the case where extraction pays rather than net-adds.

## Goal

One host-neutral `ProgressBackend` (bus subscriber + `ProgressViewState`/`StreamSnapshotStore` + outbound-message builder + inbound-command dispatch) that **both** the extension and desktop construct, injecting only:
- `send: (msg: ProgressViewOutboundMessage) => void` — the single transport seam (extension: `webview.postMessage`; desktop: IPC `postToRenderer`)
- host callbacks for genuinely host-specific actions (file open / dialogs / `vscode.commands`)

## Why it's feasible now (audit findings)

- `ProgressViewState` (`packages/extension/src/progressView/state/ProgressViewState.ts`) is **vscode-free** except the *default* storage — its constructor already accepts an injected `storage?: MementoStorage`, so desktop can pass its own backend or run ephemeral.
- `WebviewUpdater` (`managers/WebviewUpdater.ts`) — its message builders are pure; the **only** vscode coupling is `webview.postMessage` (one seam).
- `WebviewBridge` (`managers/WebviewBridge.ts`) — type-only vscode import; cursor/log-range logic is pure.
- `ProgressEventHandler` (`events/ProgressEventHandler.ts`) — host-neutral except a `vscode.Disposable` wrapper around `setupEventListeners()` (inject a `dispose()` callback instead).
[...]

PR #5473, opened 2026-06-07
fix: retry desktop ghost display restore

### LINKED ISSUE #5472: Desktop ghost stream restore can be permanently suppressed after async read miss
## Problem

After #5469/#5471, desktop restores prior-launch ghost streams from two sources:

- `streams.json` for the rail entry
- shared `streamData/{streamId}/*` for durable display state (todos, plan, usage, outputs)

`DesktopProgressBridge.sendRestoredDisplay()` currently adds the stream id to `restoredDisplaySent` before `durableSnapshots.read(streamId)` resolves. If the read rejects, or if the user switches to another stream before the read resolves, the restore is skipped but the stream remains marked as already restored. Switching back later will not retry, so the renderer never receives the persisted display state for that ghost stream.

## Expected

A ghost stream should be marked restored only after its durable display messages have actually been sent for the currently active stream. Failed reads and stale async reads should remain retryable.

## Scope

Low-risk P5 hardening for #5451. This is desktop-only restore timing and should not add another backend abstraction or pass-through layer.

LionSR:
## Summary
- fix desktop ghost-stream restore so failed or stale async durable reads stay retryable
- mark a restored display only after the active stream still matches and the persisted snapshot is actually sent
- add focused desktop bridge coverage for read failures and stale async reads

Fixes #5472.
Advances #5451.

## Validation
- `corepack pnpm vitest run src/test-kernel/desktop/DesktopAgentExecution.vitest.mts`
- `npm run typecheck -- --pretty false`
- `npm run lint` (passes; existing import-order warnings only)
- `npm run compile:fast`
- `git diff --check`

---

> [!NOTE]
> **Low Risk**
> Scoped to desktop ghost-stream display hydration in `sendRestoredDisplay`; behavior fix with focused tests and no auth or data-model changes.
> 
> **Overview**
> **Ghost stream UI restore** on the desktop progress bridge no longer treats a durable read as “done” when the async `StreamSnapshotStore.read` starts. A new **`restoredDisplayInFlight`** guard blocks overlapping restores; **`restoredDisplaySent`** is set only after todos/plan/usage/files are actually posted and the stream is still active and still a ghost.
> 
> Failed reads and **stale completions** (user switched away before the read finished) stay **retryable** on the next activation. **`restoredDisplayInFlight`** is cleared on stream delete, delete-all, and dispose.
> 
> Vitest adds coverage for read failure retry, duplicate in-flight deduplication, and stale async read after switching ghost streams.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3a956d6cb81d74d8b01a916b02adc90f17be5583. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/desktopAgentExecution.ts
+++ b/packages/desktop/src/main/desktopAgentExecution.ts
@@ -201,6 +201,8 @@ export class DesktopProgressBridge {
   private readonly durableSnapshots = new StreamSnapshotStore();
   /** Ghost streams whose persisted display has already been restored this session. */
   private readonly restoredDisplaySent = new Set<StreamTabId>();
+  /** Ghost streams with an async persisted-display restore already pending. */
+  private readonly restoredDisplayInFlight = new Set<StreamTabId>();
 
   readonly runtimeHost: AgentRuntimeHost;
 
@@ -489,6 +491,7 @@ export class DesktopProgressBridge {
     this.agentProposals.clear();
     this.restoredStreams.clear();
     this.restoredDisplaySent.clear();
+    this.restoredDisplayInFlight.clear();
   }
 
   private send(message: ProgressViewOutboundMessage): void {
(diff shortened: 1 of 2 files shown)


LionSR:
Addressed the overlapping-restore review: the desktop bridge now tracks restored display reads that are in flight, clears that state on failure/stale completion/delete, and includes a regression test showing two overlapping activations emit one restore payload.

PR #5509, opened 2026-06-07
fix(webview): slim inline banners and correct sign-in model list

LionSR:
## Summary

The "Researcher Access Program" sign-in banner was too tall. Root cause: WebAwesome's `wa-callout` `:host` hardcodes `padding: 1em` and **never reads a `--padding` custom property**, so every inline banner that set `--padding` was silently stuck at the 1em default. The fix sets the real `padding` property instead (outer-scope rules win over the component's `:host` via shadow-DOM encapsulation).

This was systemic — the same dead `--padding`-on-`wa-callout` pattern existed in three places, all fixed here:
- shared `bannerStyles.ts` → Login, ApiKey, GettingStarted, Dependency, AgentConfig banners
- standalone `WorkflowHintBanner` (progress view)
- LaTeX-tab prerequisite hint (settings)

(`SettingsApp.ts` also sets `--padding`, but on `wa-tab-panel`, which *does* read it — left unchanged.)

## Changes

- **`bannerStyles.ts`** — dead `--padding` → real `padding: var(--wa-space-xs) var(--wa-space-s)` (8/12px) for the five shared banners. This is a deliberately chosen compact value (the previous dead value was `2xs/xs` = 4/8px, which was too tight for the icon + multi-line + buttons login layout). Also zeroed the message-part `padding-block` to drop the prior padding *doubling*; the 8px callout padding now supplies the vertical spacing.
- **`LoginBanner.ts`** — split the description into a value **lead** and a lighter, smaller privacy **fineprint** for a clear two-tier hierarchy; tightened line-height to normal; top-aligned the sparkle icon so it pairs with the title.
- **`promoNotice.ts`** — updated the free-model list to "GPT-5, Claude, Gemini, DeepSeek, and more." (added DeepSeek, dropped the pinned Claude version so it won't go stale); restructured `PROMO_NOTICE_SHORT` into `{ lead, fineprint }` (single consumer: `LoginBanner`; `PROMO_NOTICE_LONG` untouched).
- **`WorkflowHintBanner.ts`**, **`LaTeXTab.ts`** — same dead-`--padding` fix, keeping each of *these two* components' originally-intended compact value (they are simpler single-line hints).
- **`README.md`** — relabeled the npm badge as **CLI downloads** instead of the generic "downloads".

## Notes
- CHANGELOG.md was intentionally **not** touched (a 0.38.6 release-notes draft is already in progress locally).
- Heads-up: the four sibling banners and the two extra callouts will render more compact now too — that's the intended consistency fix, but worth an eyeball.
- Spacing is real: `--wa-space-scale: 1` in every webview bundle, so `--wa-space-xs` = 8px (not a hairline).

## Verification
- `npm run compile:fast` — built clean
- `npm run typecheck` — clean
- code-simplifier reviewed the diff (zero changes; flagged the WorkflowHintBanner instance, now folded in)
- Both review bots green (claude: "no issues"; TeXRA/deepseek: non-blocking notes, dispositioned in a comment)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![Open VSX Version](https://img.shields.io/open-vsx/v/texra-ai/texra)](https://open-vsx.org/extension/texra-ai/texra)
 [![Open VSX Downloads](https://img.shields.io/open-vsx/dt/texra-ai/texra)](https://open-vsx.org/extension/texra-ai/texra)
 [![npm version](https://img.shields.io/npm/v/@texra-ai/cli?label=%40texra-ai%2Fcli)](https://www.npmjs.com/package/@texra-ai/cli)
-[![npm downloads](https://img.shields.io/npm/dm/@texra-ai/cli)](https://www.npmjs.com/package/@texra-ai/cli)
+[![CLI downloads](https://img.shields.io/npm/dm/@texra-ai/cli?label=CLI%20downloads)](https://www.npmjs.com/package/@texra-ai/cli)
 
 A LaTeX research assistant for VS Code and the terminal. Multi-agent
 workflows for writing, reviewing, formalizing, and rendering academic
--- a/packages/extension/src/progressView/frontend/components/WorkflowHintBanner.ts
+++ b/packages/extension/src/progressView/frontend/components/WorkflowHintBanner.ts
@@ -26,7 +26,8 @@ export class WorkflowHintBanner extends LitElement {
     }
     wa-callout {
       margin: 0 0 var(--wa-space-2xs) 0;
-      --padding: var(--wa-space-2xs) var(--wa-space-xs);
+      /* wa-callout ignores --padding; set the real property (its :host hardcodes 1em). */
+      padding: var(--wa-space-2xs) var(--wa-space-xs);
     }
     .banner-row {
       display: flex;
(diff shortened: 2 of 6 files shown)


LionSR:
## Review disposition

Thanks both. No blocking findings; no code changes required. Dispositions:

**1. `bannerStyles.ts` padding value shifted (deepseek #1).** Correct, and deliberate. The shared banners use `var(--wa-space-xs) var(--wa-space-s)` (8/12px), which is a chosen compact value, *not* the dead-intended `2xs/xs` (4/8px). I picked it because the login banner carries an icon, a two-line text block, and buttons, and 4px vertical felt too tight for that one. Only `WorkflowHintBanner` and `LaTeXTab` kept their originally-intended values (those are simpler single-line hints). Clarified in the PR description so it can't be misread as "values preserved everywhere."

**2. `::part(message) { padding-block: 0 }` could feel cramped (deepseek #2).** Checked this rather than assumed: `--wa-space-scale: 1` in every webview bundle (progressView/settingsView/webview), so `--wa-space-xs` resolves to a real **8px** top/bottom from the callout-level `padding`. The message-part `padding-block` was zeroed only to remove the previous *doubling* (callout padding stacked on top of message padding); the callout padding now supplies the breathing room, so effective vertical padding is 8px, not 0. Not cramped at the shipped scale. The cross-banner visual pass is still the author's call, as noted.

**3. `PROMO_NOTICE_SHORT` shape change (deepseek #3, claude).** Breaking but fully contained: `LoginBanner` is the only consumer and is updated; `PROMO_NOTICE_LONG` (used by `ApiAccessSection`) is untouched. TypeScript would catch any missed consumer.

**4. `SettingsApp.ts` `wa-tab-panel --padding` (deepseek #4).** Left unchanged intentionally; `wa-tab-panel` does read `--padding` (it has `padding: var(--padding)`), unlike `wa-callout`.

PR #5510, opened 2026-06-07
fix(auth): stop Codespaces/web OAuth failing with bad_oauth_state

LionSR:
## Summary

Signing in from Codespaces / vscode.dev (web/tunnel environments) failed the OAuth callback with **`bad_oauth_state`** ("OAuth state not found or expired"). Two compounding bugs in the VS Code web auth redirect:

1. **Tunnel state clobbered GoTrue's OAuth state.** The VS Code callback's tunnel routing token (`?state=TUNNEL`) was passed to GoTrue as `queryParams.state` on `/authorize`, which overwrote GoTrue's *own* OAuth state. Now the tunnel token rides on `redirect_to` instead, by using the full callback URL as `redirectTo`. `fullUrl === baseUrl` when there is no tunnel state, so plain web is unaffected.
2. **Double-encoded routing token.** The callback URL was serialized with default encoding, so `auth-js` then percent-encoded the already-encoded tunnel token a *second* time, corrupting it and silently dropping the callback (timeout). Serialize with `skipEncoding` (`externalUri.toString(true)`) to avoid the double-encode.

The two changes work together: `extension.ts` now produces a correctly-encoded `fullUrl`, and `SupabaseAuthProvider.ts` uses that `fullUrl` as `redirectTo`.

## Changes
- `packages/extension/src/extension.ts` — serialize the external callback URI with `toString(true)` (skipEncoding).
- `src/auth/SupabaseAuthProvider.ts` — return `{ redirectTo: callbackInfo.fullUrl }` instead of putting the tunnel token in `queryParams.state`.

## Notes
- Scope: implicit flow (tokens in the URL fragment). A code comment flags that switching the client to PKCE would require the callback parser to exchange `?code=` instead.
- Desktop / standard `vscode://` redirect path is unchanged (still uses `getAuthCallbackUri`).

## Verification
- `npm run typecheck` — clean (0 errors)
- Confirmed the change merges cleanly on top of current `main` (sits between the unrelated upstream `toolAvailability` edits in `extension.ts`).

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth redirect and encoding on web/tunnel hosts only; desktop path is unchanged, but auth callback misconfiguration could still break sign-in in Codespaces.
> 
> **Overview**
> Fixes **Codespaces / vscode.dev** Supabase OAuth sign-in by changing how the web callback URL is built and passed to GoTrue.
> 
> **`extension.ts`** now returns only `fullUrl` from the external auth resolver, serialized with `externalUri.toString(true)` so VS Code’s tunnel `?state=` token is not **double-encoded** when `auth-js` encodes `redirectTo` (which previously caused silent callback timeouts).
> 
> **`SupabaseAuthProvider`** uses that **`fullUrl` as `redirectTo`** on web instead of splitting the URL and sending the tunnel token as `queryParams.state`, which had **overwritten GoTrue’s OAuth state** and produced `bad_oauth_state`.
> 
> **`ExternalAuthCallbackInfo`** is simplified to a single `fullUrl` field; tests and fallbacks are updated accordingly. **Desktop** OAuth still uses `getAuthCallbackUri` unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 849e9ed7ed7cf12cbfef442ba6b6bb7f77c4f4fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/src/extension.ts
+++ b/packages/extension/src/extension.ts
@@ -289,7 +289,11 @@ export async function activate(context: vscode.ExtensionContext) {
       const vscodeState = new URLSearchParams(externalUri.query).get('state');
       const baseUrl = `${externalUri.scheme}://${externalUri.authority}${externalUri.path}`;
 
-      return { baseUrl, vscodeState, fullUrl: externalUri.toString() };
+      // skipEncoding (toString(true)) so auth-js's encodeURIComponent over the
+      // redirectTo does not double-encode the already percent-encoded tunnel
+      // ?state= token; double-encoding corrupts the routing token and the
+      // callback never makes it back to the editor (silent timeout).
+      return { baseUrl, vscodeState, fullUrl: externalUri.toString(true) };
     });
 
     if (!isSupabaseConfigured()) {
--- a/packages/extension/src/extension.ts
+++ b/packages/extension/src/extension.ts
@@ -286,14 +286,14 @@ export async function activate(context: vscode.ExtensionContext) {
         getAuthCallbackUri(vscode.env.uriScheme),
       );
       const externalUri = await vscode.env.asExternalUri(baseCallbackUri);
-      const vscodeState = new URLSearchParams(externalUri.query).get('state');
-      const baseUrl = `${externalUri.scheme}://${externalUri.authority}${externalUri.path}`;
 
-      // skipEncoding (toString(true)) so auth-js's encodeURIComponent over the
-      // redirectTo does not double-encode the already percent-encoded tunnel
-      // ?state= token; double-encoding corrupts the routing token and the
-      // callback never makes it back to the editor (silent timeout).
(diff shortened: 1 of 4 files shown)


LionSR:
Addressed the review in `849e9ed7e`.

Applied the non-blocking cleanup from @claude's review: `buildOAuthOptions` now reads only `callbackInfo.fullUrl`, so `baseUrl` and `vscodeState` had no remaining production consumer. Collapsed `ExternalAuthCallbackInfo` to `{ fullUrl }` and updated the resolver (`extension.ts`), the fallback (`config.ts`), and the test (`config.test.ts`). `npm run typecheck` is clean.

No correctness changes — the fix itself is unchanged; this just removes the now-dead fields the review flagged.

PR #5539, opened 2026-06-07
refactor(cli): centralize active stream ancestor traversal

### LINKED ISSUE #5538: CLI active stream ancestor fallback should have one owner
## Problem

CLI TUI code currently has two separate loops that walk `parentStream` with their own visited-set cycle guards:

- `packages/cli/src/chat/tui/state/childControls.ts` chooses the nearest ancestor that owns child-control rows.
- `packages/cli/src/chat/tui/panes/StatusBar.tsx` chooses the nearest ancestor that owns a visible/stoppable status.

The predicates are different and should stay local, but the parent-edge traversal itself is the same ownership rule. Repeating it makes future subagent focus behavior easier to diverge.

## Desired direction

Keep `StreamSlice` plus the child-to-parent edge map as the authoritative state, and let the stream view/state layer own ancestor traversal. Callers should provide the predicate they own, without rebuilding the walk/cycle guard themselves.

## Scope

- CLI/TUI only.
- Refactor only; no product behavior change intended.
- Add or adjust focused tests that prove the nearest-ancestor behavior remains intact for child-control fallback and status-bar stoppability.

LionSR:
## Summary

- add `nearestActiveStreamAncestor` beside the active stream scope helpers so `streamViews` owns parent-edge traversal and cycle protection
- route child-control fallback and status-bar ancestor lookup through that single traversal owner while keeping their local predicates unchanged
- cover nearest ancestor matching and parent-cycle termination in the stream view tests

Closes #5538.

## Validation

- `corepack pnpm vitest run src/test-kernel/cli/StreamTabsStrip.vitest.mts src/test-kernel/cli/ChildControls.vitest.mts src/test-kernel/cli/StatusBar.vitest.mts`
- `corepack pnpm exec prettier --check packages/cli/src/chat/tui/state/streamViews.ts packages/cli/src/chat/tui/state/childControls.ts packages/cli/src/chat/tui/panes/StatusBar.tsx src/test-kernel/cli/StreamTabsStrip.vitest.mts`
- `git diff --check`
- `npm run typecheck -- --pretty false`
- `npm run lint` (passes with existing import-order warnings outside touched files)
- `npm run compile:fast`

## Design note

The callers still own the meaning of a usable stream: child controls look for rows, and the status bar looks for status. The shared helper owns only the child-to-parent traversal so future focused-subagent behavior does not need to reimplement the same fallback walk.

### DIFF
--- a/packages/cli/src/chat/tui/panes/StatusBar.tsx
+++ b/packages/cli/src/chat/tui/panes/StatusBar.tsx
@@ -28,7 +28,10 @@ import {
   type BypassState,
   type StreamSlice,
 } from '../state/cliState';
-import { activeStreamScope } from '../state/streamViews';
+import {
+  activeStreamScope,
+  nearestActiveStreamAncestor,
+} from '../state/streamViews';
 import { useLiveNowMs } from '../state/useLiveNowMs';
 import { useSignal } from '../state/useSignal';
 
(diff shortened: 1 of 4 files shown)


LionSR:
Review note: the reported behavioral change in `resolveChildControlStreamTarget` is not new in this PR. `origin/main` already has the final `return { streamId: activeStreamId, slice: activeSlice };` after the ancestor loop, and this PR preserves that fallback while replacing only the duplicated parent traversal. So the PR remains refactor-only; no behavior change is intended.

PR #5691, opened 2026-06-09
Extract shared agent-CLI session loop into reusable module

LionSR:
## Summary

Refactors the duplicated agent-CLI child-session loop shared by `claudeAgent.ts` and `codex.ts` into `src/tools/agentCliSessionLoop.ts`.

The shared runner owns the common lifecycle once: follow-up queue drain, per-turn abort controller, child stream state, report persistence, parent `subagent_result` delivery, terminal status, interrupt unregister, queue release, and finalization. `claudeAgent.ts` and `codex.ts` now provide provider-specific strategy hooks for turn execution, result formatting, thread/session registration, and usage publishing.

## Single Source Of Truth

- `src/tools/agentCliSessionLoop.ts` owns the external-agent child-session loop choreography.
- `src/tools/agentCliShared.ts` owns shared usage/status/summary helpers used by both providers.
- Provider files keep only provider-specific SDK/thread behavior.

## Review Fixes

- Kept `AgentCliSession` private to the loop module.
- Documented the strategy call order and error path contract.
- Ordered cleanup by owner: the shared loop unregisters interrupt state and releases the follow-up queue before provider registry cleanup, preventing a same-thread Codex call from slipping into the teardown window.

## Validation

Local:
- `corepack pnpm vitest run --config vitest.config.mjs src/test-kernel/agent/toolUse src/test-kernel/tools src/test/tools` — 45 files, 210 tests passed.
- `npm run typecheck -- --pretty false` — passed.
- `npm run lint -- --quiet` — passed.

Remote:
- GitHub CI validation — passed.
- Format, label, Cursor Bugbot, Claude reviews, and TeXRA review — passed.

### DIFF
--- a/src/test-kernel/agent/toolUse/CodexProgressEvents.vitest.ts
+++ b/src/test-kernel/agent/toolUse/CodexProgressEvents.vitest.ts
@@ -19,11 +19,8 @@ import type {
 } from '@shared/schemas';
 
 // Local imports - tools
-import {
-  publishCodexStreamUsage,
-  publishCodexTodos,
-  runStreamedTurn,
-} from '@tools/codex';
+import { publishAgentCliStreamUsage } from '@tools/agentCliShared';
+import { publishCodexTodos, runStreamedTurn } from '@tools/codex';
 
 // Local imports - test
 import { createRecordingHost } from '../progressTestUtils';
@@ -65,7 +62,7 @@ describe('codex progress events', () => {
     const active = createRecordingHost();
 
     publishCodexTodos(streamId, todos, active.host);
-    publishCodexStreamUsage(streamId, executionId, usage, active.host);
+    publishAgentCliStreamUsage(streamId, executionId, usage, active.host);
 
     expect(active.events).toEqual([
       {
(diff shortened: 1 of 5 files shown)


LionSR:
Thanks for the review. Addressed the actionable points in `398a5f4` (a follow-up quality pass over the shared loop):

- **`AgentCliSession` visibility (loop:50):** un-exported — it's an implementation detail of `runAgentCliSession`, matching the old private per-tool session classes. No external importers.
- **Strategy contract docs (loop:84):** added a doc-comment to `AgentCliSessionStrategy` spelling out the per-turn call order and that `onTurnError` fires only for a non-throwing turn that reports an app-level error (a thrown turn goes through the loop's catch and is formatted via `formatError(null, …)`).
- **Cleanup ordering (loop:231):** verified safe, no change. `AgentCliSessionRegistry.release()` is a pure `this.sessions.delete(sessionId)` with no dependency on the interrupt registry, so running `onSessionCleanup()` before `interruptRegistry.unregister()` is behaviorally inert. The two original implementations actually differed on this ordering (Claude released before unregister, Codex after), which is only possible *because* they're independent.
- **`onSessionStart` `isActive` guard (codex):** intentional — it unifies the start-path and turn-success registration through one `registerThread` helper; at session start the thread is never already active, so it's a no-op there and not an observable change.

Same pass also removed two redundant per-turn strategy-hook evaluations and hoisted the duplicated `updateStreamUsage` emit into a shared `publishAgentCliStreamUsage` helper. Full workspace typecheck + the 210 toolUse/tools vitest suites pass.

---

PR #5703, opened 2026-06-09
fix(cli): keep TUI header above transcript rows

LionSR:
## Summary
- render the session header and finalized transcript rows through one ordered Ink Static item list
- remove the split live-header/static-entry path that let submitted user rows print above the TeXRA banner
- add an order-sensitive TUI harness assertion for the approval frame

## Validation
- corepack pnpm vitest run --config vitest.config.mjs src/test-kernel/cli/ConversationPane.vitest.mts src/test-kernel/cli/StaticBandResize.vitest.mts
- corepack pnpm --filter @texra-ai/cli typecheck
- npm run lint -- --quiet
- corepack pnpm --filter @texra-ai/cli validate:tui -- --snapshot-dir /private/tmp/tui-static-header-final bash-approval
- real PTY run: HARNESS_ENTRIES=4 HARNESS_BASH_APPROVAL=1 node packages/cli/dist/bin/tui-harness.js
- ad hoc PTY resize probe: 80→120 columns, headerCount=1, headerBeforeEntry=true

---

> [!NOTE]
> **Medium Risk**
> Touches core CLI scrollback/render behavior and resize semantics; wrong Static ordering or keys could duplicate or mis-order transcript output, though targeted harness and unit tests mitigate regressions.
> 
> **Overview**
> Fixes the CLI TUI so the **TeXRA session banner stays above finalized transcript rows** instead of user history printing above it (notably on bash-approval frames after resize).
> 
> **`StaticConversationTranscript`** now feeds the session header and finalized entries through **one ordered Ink `<Static>` list** keyed only by `transcript:${ownerKey}` (no width-keyed remount on entries). When a header becomes eligible after compact layout, **`appendStaticTranscriptItems` splices it before existing entry items**. Comments in **`TranscriptEntry`** clarify that static rows are print-once scrollback and do not reflow on resize.
> 
> **`validate-tui.mjs`** gains scenario **`resizes`**, **`ordered`** text-order checks, and stricter **`bash-approval`** expectations (120-col resize, single `{ T } TeXRA`, header before history and approval). Unit tests cover header insertion order; **`StaticBandResize`** docs note the main transcript intentionally avoids remounting historical static rows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 24184775222a03d3c0fc1b05d94c643c64fda707. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/scripts/validate-tui.mjs
+++ b/packages/cli/scripts/validate-tui.mjs
@@ -851,7 +851,9 @@ const SCENARIOS = [
     name: 'bash-approval',
     env: { HARNESS_ENTRIES: '4', HARNESS_BASH_APPROVAL: '1' },
     bootExpect: '[Ctrl-C]',
+    resizes: [{ cols: 120 }],
     expect: [
+      'agent: chat · model: harness-model',
       'Run bash command?',
       '$ npm run compile:safe',
       'y approve',
@@ -859,6 +861,17 @@ const SCENARIOS = [
       'Use foreground panel shortcuts',
     ],
     unexpect: ['[Alt-p]tasks', '[Option-p]tasks', '[/model]models'],
+    maxOccurrences: [{ text: '{ T } TeXRA', max: 1 }],
+    ordered: [
+      {
+        before: 'agent: chat · model: harness-model',
+        after: '› entry-1 chat history line',
+      },
+      {
+        before: 'agent: chat · model: harness-model',
+        after: 'Run bash command?',
+      },
+    ],
     maxBlankLinesBetween: [
       { from: 'entry-4 chat history line', to: 'Run bash command?', max: 3 },
     ],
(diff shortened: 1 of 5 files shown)


LionSR:
don't do band aids over band aids . This has regressed a few times. Solve the root cause

LionSR:
Root cause here is the split ownership of the same visual scrollback band: the header was rendered as live Ink content while finalized user/tool rows were rendered through Ink `<Static>`. Ink writes Static rows into terminal scrollback above live content, so the prompt/approval transcript could appear before the TeXRA header.

This patch removes that dual path. `appendStaticTranscriptItems()` already owned the ordered list (`header`, then entries); the renderer now renders that list through one `<Static>`. The harness assertion checks this ordering directly, and I also verified a real PTY frame plus an 80→120 column PTY resize with `headerCount=1` and `headerBeforeEntry=true`.

So this is not an extra fallback or modal-level workaround; it collapses the two output owners into the single existing transcript owner.

PR #5745, opened 2026-06-10
fix(relay): enforce server-side request limits

### LINKED ISSUE #5724: Add server-side per-user rate/concurrency limits and free-tier max_tokens clamp
## Context

Follow-up from #5720 (fix(relay): gate free tier on output price; default free to gemini-3.5-flash).

The PR fixed the immediate free-tier abuse vector (expensive-output flagships leaking in on input price alone) but left the $10/month client-reported spending cap in place. That cap is racy and only a volume brake — real enforcement needs server-side limits.

## Details

- Add server-side per-user rate limiting and concurrency limiting to the relay
- Add a free-tier `max_tokens` clamp so free requests can't consume unbounded output
- Once server-side limits are in place, retire the client-reported dollar counter

## References

- PR: #5720

LionSR:
## What

Closes #5724.

- Add a single tier-policy source for per-user relay request gates in `supabase/functions/relay/models.ts`.
- Add a DB-backed `relay_request_limits` table plus `relay_request_gate` / `relay_request_release` RPCs so rate/concurrency limits work across edge-function isolates instead of relying on per-process memory.
- Acquire a request slot immediately before proxying upstream and release it when the upstream response stream closes or is canceled.
- Clamp free-tier max output server-side to `8192` tokens across OpenAI-compatible, Responses API, Anthropic-style, and Google `generationConfig.maxOutputTokens` request bodies.
- Extend the existing relay request-limit tests for max-output clamping and stream-slot release behavior.

## Design Notes

The relay already owned tier/model/spend policy, so the new request limits live beside those constants rather than adding a client-side mirror. The edge function mutates the parsed request body once, after model validation and GPT-5 reasoning caps, then forwards that same body.

## Verification

- `corepack pnpm vitest run src/test-kernel/supabase/RelayRequestLimits.vitest.ts`
- `npm run typecheck`
- `corepack pnpm eslint supabase/functions/relay/index.ts supabase/functions/relay/models.ts supabase/functions/relay/requestLimits.ts supabase/functions/relay/requestGate.ts src/test-kernel/supabase/RelayRequestLimits.vitest.ts --quiet`
- `npm run lint -- --quiet`
- `git diff --check`
- `npx esbuild supabase/functions/relay/index.ts --bundle --platform=neutral --format=esm --outfile=/tmp/texra-relay.bundle.js '--external:jsr:*' '--external:npm:*'`

Supabase CLI check attempted: `supabase db lint --local --schema public --fail-on error` could not run because no local Supabase Postgres was listening on `127.0.0.1:54322`, and Docker is not running in this environment.

---

> [!NOTE]
> **High Risk**
> Changes the relay proxy path for all tiers (429/503 behavior, streaming wrapper) and depends on a new migration/RPCs; misconfiguration or lease bugs could block or over-admit relay traffic.
> 
> **Overview**
> Adds **server-side per-user relay request gates** (per-minute rate and in-flight concurrency by tier) backed by Postgres RPCs, so limits work across edge isolates. Slots are acquired after local validation and released when the upstream stream ends, is canceled, or fails; long streams **refresh** leases periodically, with stale slots expiring after ~10 minutes.
> 
> For **free tier**, the relay now requires a JSON object on model endpoints, **clamps max output to 8192 tokens** (OpenAI chat/Responses fields, Google `generationConfig.maxOutputTokens`), and returns **429** with structured metadata on rate/concurrency denials.
> 
> Tests expand `RelayRequestLimits.vitest.ts` for clamping and slot lifecycle; `tsconfig` excludes `src/test-kernel` from the main project and enables `.ts` imports in `tsconfig.test-kernel.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit febdb6b8a5ea66fd1a2ae620087238657ed6cbfd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/supabase/RelayRequestLimits.vitest.ts
+++ b/src/test-kernel/supabase/RelayRequestLimits.vitest.ts
@@ -375,9 +375,10 @@ describe('relay free-tier request limits', () => {
     }
   });
 
-  it('releases request slots when stream lease refresh keeps failing', async () => {
+  it('keeps streams open through transient lease refresh failures', async () => {
     vi.useFakeTimers();
     try {
+      let refreshes = 0;
       let releases = 0;
       const wrapped = await releaseWhenStreamCloses(
         byteStream([new Uint8Array([1])]),
@@ -385,6 +386,7 @@ describe('relay free-tier request limits', () => {
           releases += 1;
         },
         async () => {
+          refreshes += 1;
           throw new Error('refresh unavailable');
         },
         10,
@@ -392,13 +394,78 @@ describe('relay free-tier request limits', () => {
       if (wrapped === null) assert.fail('expected wrapped stream');
 
       await vi.advanceTimersByTimeAsync(30);
-      assert.equal(releases, 1);
+      assert.equal(refreshes, 3);
+      assert.equal(releases, 0);
 
       const reader = wrapped.getReader();
-      await assert.rejects(reader.read(), /refresh unavailable/);
+      assert.deepEqual(await reader.read(), {
+        done: false,
+        value: new Uint8Array([1]),
+      });
+      assert.deepEqual(await reader.read(), {
+        done: true,
+        value: undefined,
+      });
+      assert.equal(releases, 1);
 
       await vi.advanceTimersByTimeAsync(30);
       assert.equal(releases, 1);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it('releases request slots when stream lease refresh loses the slot', async () => {
+    vi.useFakeTimers();
+    try {
+      const calls: string[] = [];
+      const client = {
+        async rpc(name: string, args: Record<string, unknown>) {
+          calls.push(name);
+          return {
+            data:
+              name === 'relay_request_gate'
+                ? {
+                    allowed: true,
+                    slotId: args.p_slot_id,
+                    activeRequests: 1,
+                    concurrencyLimit: 2,
+                    requestsThisMinute: 1,
+                    rateLimitPerMinute: 20,
+                  }
+                : name === 'relay_request_refresh'
+                  ? { refreshed: false }
+                  : {},
+            error: null,
+          };
+        },
+      };
+      const slot = await acquireRelayRequestSlot(client, crypto.randomUUID(), {
+        ratePerMinute: 20,
+        concurrent: 2,
+      });
+      if (!slot.allowed) assert.fail('expected slot to be allowed');
+
+      const wrapped = await releaseWhenStreamCloses(
+        byteStream([new Uint8Array([1])]),
+        slot.release,
+        slot.refresh,
+        10,
+      );
+      if (wrapped === null) assert.fail('expected wrapped stream');
+
+      await vi.advanceTimersByTimeAsync(10);
+      assert.deepEqual(calls, [
+        'relay_request_gate',
+        'relay_request_refresh',
+        'relay_request_release',
+      ]);
+
+      const reader = wrapped.getReader();
+      await assert.rejects(
+        reader.read(),
+        /relay_request_refresh did not find request slot/,
+      );
     } finally {
       vi.useRealTimers();
     }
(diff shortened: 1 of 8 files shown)


LionSR:
Review follow-up notes:

- Stale slot recovery now uses `active_started_at`, separate from `updated_at`, so successful requests no longer keep leaked active slots alive.
- Free-tier output clamping now clamps all present OpenAI-compatible token fields and also adds the route/model default field, covering decoy fields such as `max_output_tokens` on chat routes and `max_tokens` on Responses routes.
- GPT-5/o-series chat requests without an explicit cap now default to `max_completion_tokens`, with tests for that path and for `max_completion_tokens`-only clamping.
- `once()` now flips the release flag synchronously and returns the release promise directly.
- For Google, replacing a non-object `generationConfig` with `{ maxOutputTokens: 8192 }` is intentional: this is the server-side free-tier policy boundary, and preserving `null`/invalid config would leave the cap unenforced.
- The local `isJsonRecord` helper in `requestLimits.ts` is intentionally left local for this PR. Importing the `.ts` Deno edge helper into this file makes the Node test-kernel TypeScript config fail with `TS5097`; adding a new shared module solely for this guard would add more indirection than it removes.

LionSR:
Addressed the stale-reset concurrency review in `27c05db7e` by replacing the aggregate active-request counter with explicit per-request lease rows:

- `relay_request_limits` now owns only the fixed-window rate counter.
- `relay_request_slots` is the single source of truth for active concurrency.
- `relay_request_gate` counts active lease rows and deletes only expired, unrefreshed leases.
- active streams refresh their own lease via `relay_request_refresh`; release deletes the exact `slot_id`.
- focused tests now cover slot-id continuity, heartbeat refresh, release-on-close/cancel/null-body, and the output-token clamps.

This avoids zeroing a shared active counter while a legitimate long-lived stream is still alive.

LionSR:
Addressed the free-tier unclampable-body review in `14ab20682`.

Free-tier model endpoints now fail closed unless the parsed request body is a JSON object. That preserves path-derived model validation for providers like Google, but prevents forwarding a raw/array/string body where `clampFreeTierMaxOutputTokens` cannot enforce the 8192 output-token cap.

LionSR:
Review follow-up notes for the remaining relay threads:

- `PRRT_kwDOL6NX9s6IW4Eg` (`requestLimits.ts` Deno import): fixed in `6810eb92` by using `./reasoning.ts` and moving `src/test-kernel/**/*.ts` under `tsconfig.test-kernel.json` with `allowImportingTsExtensions`, so typecheck and Deno-style Edge imports agree.
- `PRRT_kwDOL6NX9s6IW7o_` (refresh failure can stale an active stream): fixed in `6810eb92`; refresh RPC errors now reject, the timer logs failures, and after repeated failures the wrapper errors/cancels the stream and releases the slot. Added a regression test for this path.
- `PRRT_kwDOL6NX9s6IW-oo` (unread streams): leaving code as-is intentionally. A returned response body that remains open and is not canceled is still an active relay request, so it should keep its concurrency slot. Expiring that slot while the live isolate is still proxying would undercount in-flight work. The release paths remain stream close/cancel, upstream fetch failure, null body release, and repeated lease-refresh failure; stale DB cleanup is for abandoned slots where the isolate can no longer refresh/release.

LionSR:
Follow-up for `PRRT_kwDOL6NX9s6IXB3s`: addressed in `77cc6d071986446febf58d434946b9d8993d1169`. `relay_request_refresh` now treats `refreshed: false` as a refresh failure, so a stream cannot keep proxying after its DB slot has disappeared. Added a regression test for the direct slot refresh path and reran focused relay tests, full typecheck, full lint, esbuild bundle, and `git diff --check`.

LionSR:
Addressed the lease-refresh review in `febdb6b8a`:

- Transient `relay_request_refresh` RPC errors are now logged and retried; they no longer truncate an otherwise healthy upstream response.
- A definitive missing-slot response (`refreshed: false`) is represented separately and remains fatal for the stream, because that means the proxy no longer owns a concurrency slot.
- The stream race is guarded by storing `leaseError` and checking it before `controller.close()`, so a lease-loss cancel cannot close an already errored stream.

Verified locally with focused relay tests, `npm run typecheck`, `npm run lint -- --quiet`, the relay esbuild bundle smoke check, and `git diff --check`.

PR #5769, opened 2026-06-10
refactor: eliminate ten pass-through thin layers

LionSR:
Inline wrappers that only forwarded to another function without adding
logic, per the flattening-abstraction-layers guidance in CLAUDE.md:

- buildCodexWorkspaceOptions / CodexWorkspaceOptions: call
  buildAgentWorkspaceOptions directly from codex.ts; move the behavior
  doc to the source function and drop the duplicate test block already
  covered by AgentWorkspaceOptions.vitest.ts
- interruptAllCodexSessions / interruptAllClaudeAgentSessions: call
  the session registries' interruptAll() directly in agentShutdown
- createDefaultTotals: inline as RunUsageTotalsSchema.prefault({})
- readAttachedMemoryMisses: export the tolerant schema and parse at
  the single call site
- formatCompileLogTail: inline into readLogTail; move the trailing-
  newline regression coverage to a new splitContentLines test
- getWorkflowAgents / getToolUseAgents: export getAgentsByCategory
  instead and drop the never-used includeInternal flag; category-
  dispatch ternaries at three call sites collapse to a single call
- getListSettings (desktop): inline loadFileListSettings(getConfig)
- resolveLocalCommandEntry (desktop): inline into
  getDesktopCommandMenuEntries (the isDesktopLocalCommandId type guard
  stays — the catalog lookup below needs its narrowing)
- formatCliHistoryLine / formatCliHistoryFile: inline as map lambdas
- defaultUnsupportedCommand (desktop): inline as the ?? fallback

https://claude.ai/code/session_01ED31pwY6ponqBWtDNvYmYY

---

> [!NOTE]
> **Low Risk**
> Mechanical inlining and a narrower public agent-registry surface (`getAgentsByCategory` without `includeInternal`); behavior should match prior defaults unless something depended on listing internal agents.
> 
> **Overview**
> This PR **flattens pass-through helpers** so callers hit the real implementation directly, aligned with the repo’s “no thin abstraction layer” guidance.
> 
> **Agent catalog API:** `getWorkflowAgents` / `getToolUseAgents` are removed in favor of a single exported **`getAgentsByCategory`**. CLI multi-agent planning, agent listing, desktop/settings agent controllers, the TUI harness, `@texra/core`, and extension test commands all switch to that API. The old **`includeInternal`** optional is dropped—category listings always exclude internal agents.
> 
> **Runtime / tools:** Codex thread setup now calls **`buildAgentWorkspaceOptions`** instead of a Codex-specific wrapper (docs live on the shared helper). Host shutdown calls **`codexThreads.interruptAll()`** and **`claudeAgentSessions.interruptAll()`** directly. Attached-memory misses use exported **`AttachedMemoryMissesSchema`** at the launch site. Usage JSON defaults use **`RunUsageTotalsSchema.prefault({})`** without a separate factory. LaTeX compile log tail formatting is inlined; **`splitContentLines`** gets the regression test that used to cover the removed helper.
> 
> **Desktop / CLI polish:** Several one-liner desktop helpers (file list settings, local command menu resolution, progress IPC unsupported-command default) and CLI history formatters are inlined into their call sites. Vitest mocks and expectations are updated for **`getAgentsByCategory`**; duplicate Codex workspace-option tests are removed where **`AgentWorkspaceOptions`** already covers behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8bd5749a3d744638f71661dea4c32509840dac73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Review dispositions after rebase:

- `getWorkflowAgents` / `getToolUseAgents` removal is intentional. These were internal pass-throughs with all in-repo callers using the default non-internal path; keeping compatibility aliases would preserve the thin layer this PR is deleting. The canonical registry API is `getAgentsByCategory(category)`.
- `AttachedMemoryMissesSchema` being tolerant is intentional at this boundary: attached-memory miss state is persisted/runtime-loaded data, and the single caller wants graceful degradation for corrupt saved state.
- `RunUsageTotalsSchema.prefault({})` was verified by the reviewer and remains the single default source; no extra helper needed.

PR #5860, opened 2026-06-12
refactor(runtime): extract AgentLaunchContext→RunContext mapping; document overlap and staleness

LionSR:
## Summary

**Problem identified:** `AgentLaunchContext` and `RunContext` both carry ambient per-run fields — `runtimeHost`, `streamId`, `executionId`, `workingDirectory`, `delegationDepth`, `approvalPromptsUnavailable`, `stopAfterCycle`, etc. The translation between the two was an inline 11-field copy block inside `withExecutionRunContext`, invisible to code reviewers and silently error-prone to extend.

**Concrete risk:** When a new runtime flag is added (e.g. `stopAfterCycle`, `approvalPromptsUnavailable`), engineers must update `AgentLaunchContext`, `RunContext`, *and* the inline copy block — with no compiler error if the copy step is forgotten. The copy block also hides three field renames (`logger`→`trace`, `config.agent`→`agentName`, `config.model`→`model`) that are otherwise invisible.

**Additional issue discovered:** `RunContext.model` is snapshotted at run start and is not updated when `onModelChanged` fires mid-session (which updates `AgentLaunchContext.config.model` but not the frozen `RunContext`). `DelegationTools` reads `parentContext.model` to inherit the parent model for subagents — so after a model switch, delegated subagents receive the old parent model. This was undocumented.

## Changes

- **Extract `agentContextToRunContext(ctx: AgentLaunchContext): CreateRunContextOptions`** — a single named function that owns the full `AgentLaunchContext → RunContext` projection. Adding a new runtime flag now has one obvious place.

- **Simplify `withExecutionRunContext`** to a one-liner calling `withRunContext(createRunContext(agentContextToRunContext(ctx)), fn)`.

- **Document the overlap** on both interfaces:
  - Which fields collide and why (`logger`/`trace`, `config.agent`/`agentName`, `config.model`/`model`)
  - The `RunContext.model` staleness limitation and what it affects (`delegate_agent` inheriting the wrong parent model after a mid-session switch)

- **Update the `withRunContext` comment** to accurately describe the relationship (replace the vague "migration bridge" note with a concrete description of who populates it and who reads it).

## What this is not

This PR does not fix the `model` staleness bug (that requires either making `RunContext` mutable for that field, or threading a getter closure). It documents the issue and its scope so a follow-up can address it with full context.

## Test plan

- [x] `corepack pnpm run typecheck` — passes (exit 0)
- [x] `corepack pnpm test` — 439 test files, 2994 tests pass
- [x] No behavior change: `agentContextToRunContext` is a pure projection; `withExecutionRunContext` produces an identical `RunContext` to before

https://claude.ai/code/session_01NEBXoWyMJPtfYfx86pgJEY

### DIFF
--- a/src/agent/runtime/AgentLaunchContext.ts
+++ b/src/agent/runtime/AgentLaunchContext.ts
@@ -109,7 +109,7 @@ const STATUS_MESSAGES: Record<string, string> = {
  *
  * Centralises the mapping so new per-run flags (e.g. `stopAfterCycle`,
  * `approvalPromptsUnavailable`) live in one place and are never silently
- * dropped. Two field renames are unavoidable given the interfaces were
+ * dropped. Three field renames are unavoidable given the interfaces were
  * designed independently:
  *  - `AgentCore.logger`   → `RunContext.trace`
  *  - `AgentConfig.agent`  → `RunContext.agentName`
--- a/src/agent/runtime/AgentLaunchContext.ts
+++ b/src/agent/runtime/AgentLaunchContext.ts
@@ -47,6 +47,7 @@ import { RetryRequestCoordinatorImpl } from './RetryRequestCoordinator';
 import {
   createRunContext,
   withRunContext,
+  type CreateRunContextOptions,
   type RunCoordinators,
 } from './RunContext';
 import {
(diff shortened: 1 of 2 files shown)


LionSR:
Indirection that they are called differently

PR #5963, opened 2026-06-13
SDK 7d residues: scope desktop approval delete-all; document subagent-delivery / binder-queue

LionSR:
Recovered on top of current `main` after #5962 landed, keeping only the residue cleanup commits plus the review fix for session-owned coordinator cleanup.

## Summary

- scope desktop `deleteAllStreams` approval cleanup to this window's streams instead of using the process-wide approval reset
- reject rare unscoped pending approvals after the per-stream sweep so prompts without a stream context do not hang
- clear the owning desktop window's session coordinator bridge on delete-all, covering child/subagent retry/plan/proposal requests that are not in the visible stream list
- document why `subagentDeliveryRegistry` intentionally remains process-wide and executionId-keyed
- document the invariant that `ExecutionSubscriptionBinder` depends on the streamId-keyed `ToolUseFollowUpQueue`

## Verification

- `corepack pnpm exec prettier --write packages/desktop/src/main/desktopAgentExecution.ts src/test-kernel/desktop/DesktopAgentExecution.vitest.mts src/agent/runtime/SessionHandle.ts src/tools/approval/index.ts src/tools/subagentDeliveryState.ts src/test-kernel/tools/ApprovalCleanupScope.vitest.ts`
- `corepack pnpm exec eslint packages/desktop/src/main/desktopAgentExecution.ts src/agent/runtime/SessionHandle.ts src/tools/approval/index.ts src/tools/subagentDeliveryState.ts src/test-kernel/tools/ApprovalCleanupScope.vitest.ts src/test-kernel/desktop/DesktopAgentExecution.vitest.mts`
- `corepack pnpm vitest run src/test-kernel/desktop/DesktopAgentExecution.vitest.mts src/test-kernel/tools/ApprovalCleanupScope.vitest.ts src/test-kernel/agent/runtime/SessionHandle.vitest.ts`

---

> [!NOTE]
> **Medium Risk**
> Changes process-global approval cleanup paths used on stream deletion; incorrect scoping could leave hung prompts or still affect sibling windows, though new tests target the multi-window case.
> 
> **Overview**
> Fixes **multi-window desktop** behavior where one window’s **delete all streams** used the process-wide `cleanupAllApprovals` reset and could reject pending tool/bash/user-question prompts (and related state) for **other** windows.
> 
[...]

PR #5960, opened 2026-06-13
SDK Step 7d: per-session SessionHandle + terminal result event (consolidated)

LionSR:
## Summary

Consolidates the seven-PR **SDK Step 7d** stack (#5948, #5951, #5952, #5955, #5957, #5958, #5959) into a single PR against `main`, with the post-stack audit fixes folded in. Supersedes the stacked PRs (they can be closed in favor of this one).

Step 7d composes the landed runtime owners — `InterruptRegistry`, `ExecutionRegistry`, `RunCoordinatorBridge`, `ExecutionSubscriptionBinder` — plus the trace-flusher set and an optional host channel into one per-session **`SessionHandle`** (a composition record, not a facade), threads it through the run via the frozen `RunContext`, scopes sweeps/flushers to the session, gives each desktop window its own session, and adds the terminal **`result`** `AgentEvent` that hosts consume to present run outcomes. Single-session hosts (extension, CLI) stay byte-identical: `defaultSession` wraps the process singletons by identity.

Design doc: `docs/proposals/session-handle-7d-design.md`.

## Audit fixes folded in

A multi-agent audit of the stack against the design doc surfaced (and adversarial verification confirmed) these, all included here:

- **Blocking** — desktop follow-ups resolved the process-default session instead of the per-window session, so a live tool-use follow-up was dropped as `no_session`. `sendFollowUp` gains an optional `session` (defaults to `currentSession()`); the desktop bridge passes `this.session`.
- **Ownership** — `SessionHandle.dispose()` now unregisters its flusher set (no per-window `Set` leak into the process-wide drain), drains first, is throw-safe (`liveSessions` delete in `finally`), and clears result listeners.
- **Correctness** — the terminal `result` event is guaranteed exactly-once: a `resultEmitted` guard plus a guarded success-arm cleanup stop a post-completion listener throw from re-entering the catch arm and publishing a second `failed`.
- **Cleanups** — three host toast consumers consolidated behind a shared `presentTerminalResult` dispatcher + `attachTerminalResultToast(session, host)`; catch-arm outcome via one mapper; `AgentLaunchContext.session` made required; exhaustiveness guard on the toast mapper; per-listener guard in `attachRunTrace`; `@texra/core` re-exports `AgentErrorKind` / `RunUsageTotals`.

## Deferred by design (F-1)

[...]

PR #5964, opened 2026-06-14
SDK 7d multi-window: preserve delete-protection for orphaned runs on window close

LionSR:
## Summary

**SDK Step 7d follow-on: multi-window window-close policy.** Stacked on #5963 (residues) → ... → #5960 (7d).

On macOS the process outlives a closed window, so its agent runs keep going. Policy (maintainer-decided): **keep them running** — they're snapshot-backed and resumable, so killing them would lose the resume path — but **restore the history delete-protection** that `session.dispose()` was dropping (a regression from main's never-disposed global registry, where a reopened window could delete a still-running orphan's history).

- `SessionHandle.dispose()` records its still-active execution ids in a process-global `orphanedActiveExecutions` set.
- `getAllActiveExecutionIds()` reads that set in addition to the live sessions.
- Orphaned ids linger until process exit (an untracked orphan's completion isn't observable); the over-protection is **safe-leaning** — it only ever refuses an *unsafe* delete. A resumed run is re-tracked live in its new session, so it's also covered.
- Full **rebind-on-reopen** (re-binding the live run to the new window's renderer) remains future work; this PR closes only the concrete delete-protection regression.

## Tests / verification

- `SessionScope.vitest.ts`: new orphan-protection case (active-at-dispose stays protected); the "drops on dispose" case corrected to untrack (finish) before dispose.
- root + test-kernel + extension + CLI + desktop typecheck clean; full suite **451 files / 3124 passing**; eslint + prettier clean.

---

> [!NOTE]
> **Medium Risk**
> Touches cross-session execution tracking and process-global approval controllers used by multi-window desktop; behavior is safe-leaning (over-protection on delete) with new tests covering orphan and approval scope cases.
> 
> **Overview**
> **Multi-window orphan runs:** When a desktop window closes while agent runs are still active (headless on macOS), `SessionHandle.dispose()` now records those execution ids in a process-global `orphanedActiveExecutions` set so `getAllActiveExecutionIds()` still blocks history delete from a reopened window. Resumed runs reconcile stale orphan markers when re-tracked live.
> 
> **Scoped approval cleanup:** Desktop `deleteAllStreams` loops `cleanupApprovalsForStream` per window stream plus session-scoped coordinator cleanup instead of process-wide `cleanupAllApprovals`, so one window cannot wipe another's pending approvals.
> 
> **Smaller fixes:** `onRun` lifecycle callbacks swallow async rejections; child stream handles attach run trace and derive terminal outcome from `finalStatus`; documentation for global vs per-session approval and registry invariants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c14fb40f602caffbac5f793ff4e4b0e64a46f57a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
this is too complicated

PR #6167, opened 2026-06-17
refactor: move host-neutral latexdiff core out of the VS Code command layer

LionSR:
## What

Moves the host-neutral **latexdiff output-discovery and diff-operation-building** logic out of the VS Code command layer into the latex core:

```
packages/extension/src/commands/latex/latexdiff/   →   src/latex/latexdiff/
  outputDiscovery.ts   diffOperations.ts   service.ts   types.ts
```

`latexdiffCommands.ts` stays in the command layer and keeps every genuinely host-bound piece (QuickPick math-markup selection, `vscode.window.withProgress`, the result `showMessage`s); it now imports the moved core from `@latex/latexdiff/*`.

## Why this was a layering violation

These ~617 lines are pure domain logic — recursive `.tex` walking, run-storage scanning, execution-metadata matching against `StreamSnapshotStore`, and round / between-round `DiffOperation[]` construction. **Every dependency is already VS Code-free** (`@transcript`, `@agent/storage`, `@agent/output/workflowOutputLayout`, `@latex/latexdiff/*`, `@utils/files`, `@shared/schemas`), and the sibling diff modules (`diffCommandExecutor`, `diffFileProcessor`, …) already live in `src/latex/latexdiff/`.

The **only** thing tying this code to VS Code was the FS triad:

```ts
// before — in commands/latex/latexdiff/
import * as vscode from 'vscode';
entries = await vscode.workspace.fs.readDirectory(vscode.Uri.file(dir));
const isFile = (type & vscode.FileType.File) !== 0;
```

```ts
// after — in src/latex/latexdiff/
import { platform } from '@platform/platform';
import { isFile, isDirectory } from '@utils/files/fsEntryType';
const entries = await platform().fs.readDirectory(dir); // [string, number][]
if (isFile(type)) { … } else if (isDirectory(type)) { … }
```

This is exactly the swap CLAUDE.md prescribes (`platform().fs.readDirectory` + `isFile()/isDirectory()` instead of `vscode.workspace.fs` + `vscode.FileType`). Because the logic lived in `commands/`, the **CLI has no latexdiff capability at all** and **desktop reimplements a degraded single-file diff** (`desktopProgressFileActions.ts` → one round-unaware `service.runDiff`). Both can now reuse the full round-aware core.

## Behavior-preserving

- `platform().fs.readDirectory` returns the same `[name, type]` tuple shape, and the platform `FileType` enum is **bitmask-compatible with `vscode.FileType`** (`File=1`, `Directory=2`, `SymbolicLink=64`).
- **Symlink parity (caught in review, fixed in follow-up commit):** four sites used strict equality (`=== File` / `!== Directory`), which excluded symlinks — both `vscode.workspace.fs` and `platform().fs` report a symlink as `SymbolicLink | targetType` (65/66). The `isFile()`/`isDirectory()` bitmask helpers *include* symlinks, so those four sites now guard with `isSymlink()` to keep skipping symlinked outputs/round dirs — restoring exact parity and matching the documented intent (symlinks are mirrored dependency copies, not revised outputs).
- The `node:fs` `lstat` symlink guard in `collectTexFiles` is **retained** and unchanged (it already used a bitmask + `lstat` guard, so its swap was equivalent).
- The `vscode.Progress<…>` parameter is replaced by a host-neutral `DiffProgressReporter` interface, which `vscode.Progress` structurally satisfies — the command layer passes its real progress object unchanged.

No behavior change for the extension; net ≈ 0 LOC (the small `+` is the `DiffProgressReporter` interface).

## Scope

This PR is **only the move + FS-API swap**. Wiring the CLI/desktop onto the now-shared core is intentionally left as follow-up.

## Verification

- `tsc --noEmit` (root) and the extension `tsc --noEmit` both pass with **zero errors** in the moved files and the caller.
- `prettier --check` and `eslint` clean (imports reordered into the aliased group).
- No tests referenced the old paths (only `latexdiffCommands.ts` imported these modules).

---
🤖 Found by an abstraction-layer audit (multi-lens find → adversarial verify against documented exceptions → prioritize). The other four lenses — render-time workarounds, wrapper/factory indirection, infra-in-domain, TUI viewport-ownership — came back clean.


LionSR:
Thanks — good catch on the symlink semantics. You're right that the swap was **not** behavior-identical: both `vscode.workspace.fs.readDirectory` and `platform().fs.readDirectory` report a symlink as `SymbolicLink | targetType` (65/66), so the old strict-equality checks (`!== File`, `=== File`, `!== Directory`) excluded symlinks while the plain `isFile()` / `isDirectory()` bitmask helpers include them.

Fixed in `1723aec79`: the four strict-equality sites now guard with `isSymlink()`, so the workspace and run-storage scans keep skipping symlinked outputs and round dirs — restoring exact parity and matching the documented intent (symlinks are mirrored dependency copies, not revised outputs — the same reason `collectTexFiles` skips them via its `lstat` guard, which is unchanged).

- `diffOperations.ts:234` (workspace `.tex` scan) → `!isFile || isSymlink || …`
- `diffOperations.ts:256` (r{round} dir scan) → `!isDirectory || isSymlink`
- `diffOperations.ts:276` (mid-era file match) → `isFile && !isSymlink`
- `outputDiscovery.ts` (run-storage round dir scan) → `!isDirectory || isSymlink`

Also corrected the "type checks are identical" wording in the PR description.

PR #6177, opened 2026-06-17
refactor(runtime): drop dead/redundant data from the launch→flow projection

LionSR:
Two independent, behavior-preserving net-negative cleanups on the busiest call boundary in the runtime (`executeAgent` → `runToolUseFlow`). Found by the deep architectural audit (god-context / dead-carrier lenses).

## 1. Redundant flag re-passes (dual-write at the call boundary)

`executeAgent` assigns three per-run flags onto `ctx`:

```ts
ctx.approvalPromptsUnavailable = options.approvalPromptsUnavailable;
ctx.runtimeUnavailableTools   = options.runtimeUnavailableTools;
ctx.stopAfterCycle            = options.stopAfterCycle;
```

…then builds the flow input as `{ ...ctx, … }` **and** re-passes the same three from `options` — byte-identical, no intervening reassignment. The tell that this is accidental: `delegationDepth` is set the same way but is **not** re-passed, and flows correctly through the spread alone. Removed the redundant explicit fields at both the fresh and resume call sites; the spread carries them.

## 2. Dead `RunContext.trace` projection chain

`RunContext.trace` was set from `AgentLaunchContext` (`logger → trace`), defaulted to `noopTrace`, and **read in zero production sites** — every `tryUseRunContext()`/`useRunContext()` consumer reads `executionId`/`workingDirectory`/`streamId`/`coordinators`/`session`/`runtimeHost`/`model`/`delegationDepth`, never `.trace`. The only readers were two self-validating test assertions. The field's own JSDoc conceded it "exists for test helpers and future SDK consumers."

The public SDK trace surface — `AgentRunHandle.trace`, a distinct constructor-injected handle field — is **not** this projection and is untouched. Removed:
- `RunContext.trace` + `CreateRunContextOptions.trace` + the `noopTrace` default and now-unused `@agent/trace` import
- the `trace: ctx.logger` projection line + stale JSDoc (and the "three fields → two fields" count)
- the trace-default test; retargeted the freeze-immutability poke from `trace` to `runtimeHost`

Trace still reaches run-scoped code via explicit `AgentTrace` parameter threading / `createRunTrace()` — unchanged.

## Verification
- `tsc --noEmit` (root + test-kernel) clean for all touched files; prettier + eslint clean.
- `RunContext.vitest` passes (5/5).
- No production reader of `RunContext.trace`; no `createRunContext` caller passes `trace:`.

Net −31 LOC, no behavior change.

---
🤖 Round-3–10 deep audit (run against current origin/main).

---

[...]

PR #6178, opened 2026-06-18
refactor(cli): use core formatCompactDuration for the run-line elapsed

LionSR:
`runProgressRenderer` hand-rolled a private `formatElapsed` (minutes/seconds split with zero-padded seconds), duplicating `@utils/core` `formatCompactDuration` — the canonical ms→human duration primitive the rest of the CLI already uses (`StatusBar`, `childControls`). The same run rendered its elapsed time through **two** formatters.

```ts
// before — private, in runProgressRenderer.ts
function formatElapsed(ms){ s=floor(ms/1000); m=floor(s/60); r=s%60;
  return m ? `${m}m ${pad(r)}s` : `${r}s`; }
parts.push(formatElapsed(now - this.startedAt));

// after
import { formatCompactDuration } from '@utils/core';
parts.push(formatCompactDuration(now - this.startedAt));
```

Folds the run status line onto the shared primitive (split-ownership of a deliberately-centralized helper).

**Output delta:** sub-minute is identical (`0s`/`1s`/`2s` — the only values `RunProgressRenderer.vitest` asserts). For ≥1min it drops the cosmetic zero-pad (`1m 5s` vs `1m 05s`); for ≥1h it uses `1h 15m` instead of `75m 00s` — both of which **match the rest of the CLI** rather than diverge from it.

## Verification
- CLI `tsc --noEmit`, prettier, eslint clean.
- `RunProgressRenderer.vitest` assertions are all sub-minute and unchanged under `formatCompactDuration` (CI runs the full suite; the local sandbox can't load it due to an unrelated missing-dep).

Net −6 LOC.

---
🤖 Round-3–10 deep audit (reimplemented-primitive lens).

---

> [!NOTE]
> **Low Risk**
> Cosmetic formatting consolidation in CLI progress output only; sub-minute behavior stays the same per existing tests.
> 
> **Overview**
> The run progress status line no longer uses a **private `formatElapsed` helper** in `runProgressRenderer.ts`. It now imports **`formatCompactDuration` from `@utils/core`**, the same duration formatter already used by the TUI status bar and child controls.
> 
> That removes duplicated minutes/seconds formatting logic (~8 lines) and **aligns long-run elapsed display** with the rest of the CLI (e.g. no zero-padded seconds after the minute; hour-scale durations use hours instead of very large minute counts).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 165fc1942c53341af0ffe9b71f5c16c0bfc1fd1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/runtime/runProgressRenderer.ts
+++ b/packages/cli/src/runtime/runProgressRenderer.ts
@@ -10,6 +10,7 @@ import {
   type ExecutionStatus,
   type StreamStatus,
 } from '@shared/schemas';
+import { formatCompactDuration } from '@utils/core';
 
 // Local imports - CLI runtime
 import { writeRawStderr } from './logSinks';
@@ -288,7 +289,7 @@ class DefaultRunProgressRenderer implements RunProgressRenderer {
     ) {
       parts.push(`tools: ${this.state.toolCallCount}`);
     }
-    parts.push(formatElapsed(now - this.startedAt));
+    parts.push(formatCompactDuration(now - this.startedAt));
     return parts.join(' · ');
   }
 }
@@ -328,12 +329,4 @@ function formatActiveChildren(
   const suffix =
     namedChildren.length > 1 ? ` +${namedChildren.length - 1}` : '';
   return `${label}: ${first}${suffix}`;
-}
-
-function formatElapsed(ms: number): string {
-  const seconds = Math.max(0, Math.floor(ms / 1000));
-  const minutes = Math.floor(seconds / 60);
-  const remainingSeconds = seconds % 60;
-  if (minutes === 0) return `${remainingSeconds}s`;
-  return `${minutes}m ${remainingSeconds.toString().padStart(2, '0')}s`;
 }


LionSR:
Closing — the review (thanks @chatgpt-codex-connector) surfaced that this "duplication" is actually **justified divergence**, not a redundant reimplementation.

The run-progress line is a live, per-second heartbeat, and the bespoke `formatElapsed` had two properties *together* that neither core formatter provides:

1. **`0` renders as `0s`** (the at-start display). `formatDuration` returns `1s` for sub-second durations, so it would change the initial line and break the `"0s"` assertions.
2. **Seconds always visible** (liveness). `formatCompactDuration` caps at 2 units, so once a run passes 1h (`1h 1m`) it drops seconds and the heartbeat looks frozen for up to a minute — exactly the regression flagged.

So the centralization would trade a real liveness property for consistency. Not worth it; the local formatter stays. (The other deep-audit dedups — #6177, #6179, #6183 — stand on their own.)

PR #6179, opened 2026-06-18
fix(fs): add appendFile to the platform FS port so output appends use one authority

LionSR:
## The split

`BaseFS` (the platform-routed FS facade) routes `write`/`read`/`stat`/`delete` through `platform().fs` — `vscode.workspace.fs` in the extension, a **virtual, remote/SSH/web-aware** filesystem. But `appendFile` called `fs.promises.appendFile` directly — the **local node disk**:

```ts
BaseFS.write       -> platform().fs.writeFile        // virtual FS
BaseFS.appendFile  -> fs.promises.appendFile          // LOCAL disk (bypass)
```

The `FileSystemProvider` port had no append surface, so the bypass was **forced by an incomplete port**, not a host-allowed exception.

**Live symptom:** `ResponseCycleFlow` writes an output file's first chunk via the port and appends every later chunk via the bypass — *one file, one flow, two different filesystems* in a remote/web workspace.

## The fix — complete the port

- Add `appendFile(path, content): Promise<void>` to `FileSystemProvider`.
- Implement in all three providers:
  - `nodeFilesystem` → `fs.promises.appendFile`
  - `VscodeFileSystem` → read-modify-write through `vscode.workspace.fs` (`FileNotFound` → empty, so append-creates like node)
  - `FakePlatform` (test fake) → in-memory concat
- Repoint `BaseFS.appendFile` → `platform().fs.appendFile`.

Every append caller (`flexibleFS` → `ResponseCycleFlow`, CLI `inputHistory`) now appends through the same authority it writes through.

## Behavior

- **Node / CLI / desktop: byte-identical** — still `fs.promises.appendFile` via the node provider.
- **Extension:** appends now hit the virtual FS that `writeFile` already used. In a local workspace that *is* the local disk (equivalent); in a remote/web workspace it fixes the split.

## Scope

`appendFile` only. The finding also flagged two `runReflectionFlow` `existsSync` resume-compat checks — but those sit in a **synchronous** path resolver (`(round) => AgentFileLocation`), so converting them to the async port-backed `exists()` needs a signature change and is intentionally **not** in this PR.

## Verification
- `tsc --noEmit` (root + test-kernel + extension) clean — no `FileSystemProvider` implementer left without `appendFile`.
- prettier + eslint clean.
- `InputHistory` (3/3), `ResponseCycleTools` (5/5), `ProcessOutputPoller` (4/4), `ToolEditApprovalGate` (5/5) pass.

---
🤖 Round-3–10 deep audit (leaky-port / incomplete-abstraction lens) — the highest-value finding.

---

> [!NOTE]
> **Medium Risk**
> Changes how appended output is persisted in the VS Code extension (including remote/web workspaces); the VS Code path relies on read-modify-write rather than true append, with per-path serialization to limit races.
> 
> **Overview**
> **Adds `appendFile` to the platform `FileSystemProvider` and routes `BaseFS.appendFile` through `platform().fs` instead of `fs.promises.appendFile`.** That closes a leak where writes used the host FS (e.g. `vscode.workspace.fs` in remote/web workspaces) while appends always hit the local Node disk—so flows like streaming output could mix authorities on one file.
> 
> **Node** still uses `fs.promises.appendFile`; **tests** use in-memory concat on the fake provider. **VS Code** emulates append with read-modify-write on `vscode.workspace.fs`, treats missing files as empty (append-create), and **serializes concurrent appends per path** so parallel RMW cannot clobber each other.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb4d49d6a28afa2983af8bbb00a174a4b2cf7013. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/src/frontend/vscode/vscodeFileSystem.ts
+++ b/packages/extension/src/frontend/vscode/vscodeFileSystem.ts
@@ -6,6 +6,7 @@
 import * as fs from 'node:fs';
 import * as vscode from 'vscode';
 
+import { isFileNotFoundError } from '@common/errors';
 import type {
   FileSystemProvider,
   FileStat,
@@ -47,6 +48,24 @@ export class VscodeFileSystem implements FileSystemProvider {
 
   async writeFile(target: string, content: Uint8Array): Promise<void> {
     await vscode.workspace.fs.writeFile(vscode.Uri.file(target), content);
+  }
+
+  async appendFile(target: string, content: Uint8Array): Promise<void> {
+    // vscode.workspace.fs has no native append, so read-modify-write through
+    // the same virtual FS keeps appends on the one authority writeFile uses
+    // (a remote/web workspace is not the local node disk).
+    const uri = vscode.Uri.file(target);
+    let existing: Uint8Array;
+    try {
+      existing = await vscode.workspace.fs.readFile(uri);
+    } catch (error) {
+      if (!isFileNotFoundError(error)) throw error;
+      existing = new Uint8Array(0);
+    }
+    const merged = new Uint8Array(existing.length + content.length);
+    merged.set(existing, 0);
+    merged.set(content, existing.length);
+    await vscode.workspace.fs.writeFile(uri, merged);
   }
 
   async delete(
--- a/packages/extension/src/frontend/vscode/vscodeFileSystem.ts
+++ b/packages/extension/src/frontend/vscode/vscodeFileSystem.ts
@@ -13,6 +13,14 @@ import type {
 } from '@platform/interfaces/filesystem';
 
 export class VscodeFileSystem implements FileSystemProvider {
+  /**
+   * Per-path append chains. vscode.workspace.fs has no atomic append, so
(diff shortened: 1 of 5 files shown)


LionSR:
Thanks all — addressed the non-atomic read-modify-write in `eb4d49d6a`.

`VscodeFileSystem.appendFile` now **serializes appends per path**: each call chains after the previous append to the same path, so two concurrent appends can no longer both read the same bytes and clobber each other. The per-path entry is dropped once that path's chain drains (no unbounded map growth).

On the remaining (acknowledged, non-blocking) point — reading the whole file into memory per append: that's inherent to `vscode.workspace.fs` having no native append, and as @claude noted it's bounded by the **per-cycle** (not per-token) append cadence in `ResponseCycleFlow`, so it's an acceptable cost for routing appends onto the same virtual FS `writeFile` already uses. If a future hot path needs large streaming appends in a remote workspace, an atomic/streaming append on the port would be the follow-up — but that's out of scope for this correctness fix.

PR #6186, opened 2026-06-18
fix: make clickable file/ref links keyboard-activatable in progress view

LionSR:
## Summary

Clickable file/ref links in the progress view were **mouse-only**: rendered as `<span>` with a click handler but no `role`/`tabindex`/keydown, so they weren't in the tab order and Enter/Space did nothing. Keyboard-only and screen-reader users couldn't open a proposed file (agent-proposal approval — a primary, frequent flow), a generated file, or a latexdiff result. Tellingly, the shared `.file-link:focus-visible` style was **dead** because the spans could never receive focus — focusability was intended but never wired.

## Fix

Make each clickable span a real activatable control — `role="button"` + `tabindex="0"` + an Enter/Space `keydown` that runs the **same action as click** (with `preventDefault`) — following the existing `GoalTab.handleRowKey` pattern:

- **PermissionCard** (agent-proposal approval): `@keydown` parity on the `.file-list` container; shared `openFileFromTarget`.
- **LatexdiffResults**: `role`/`tabindex` + keydown on the file link.
- **FileList** (generated files): `role`/`tabindex` on the row span; `@keydown` on the container mirroring the existing composed-path click delegation (`getComposedPathElement`, which already accepts any `Event`).

## Scope / verification
- **Additive** — mouse behavior is unchanged; this only adds keyboard/SR reachability and restores the intended focus-visible styling.
- Changed files typecheck clean; CI `validate` confirms the full build.
- Found by the focus/keyboard sweep, verified against the `GoalTab` precedent.

### Not included
The `LogList` transcript-link case (`file-link` / `latex-ref` / `proposal-restore-link` produced via HTML-string formatter builders across ~9 render sites, plus the `MouseEvent`-gated delegation owner) is a larger, separate change — left as a follow-up rather than bundled here.

---

> [!NOTE]
> **Low Risk**
> Additive accessibility wiring in progress-view UI; mouse behavior unchanged and a targeted vitest guards FileList keyboard delegation.
> 
> **Overview**
> Makes **mouse-only** file links in the progress view keyboard- and screen-reader reachable by treating clickable spans as controls (`role="button"`, `tabindex="0"`) and firing the **same open-file behavior** on **Enter/Space** (with `preventDefault`), so existing `.file-link:focus-visible` styling can apply.
> 
> **PermissionCard** adds container `@keydown` and **`openFileFromTarget`** for proposal file lists. **LatexdiffResults** wires keydown on each file link. **FileList** refactors click handling into **`dispatchFileActionFrom` / `dispatchFileAction`** and limits keyboard activation to **`.file-path`** targets so **Space** on native **`wa-button`** action buttons is not intercepted.
> 
> Adds **`FileList.vitest.mts`** to assert Space on a file path dispatches **`OPEN_FILE`** and does not double-fire when Space hits a **`wa-button`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0bcf829c10e75e1dca05bf4582be458a3622dbf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/src/progressView/frontend/components/FileList.ts
+++ b/packages/extension/src/progressView/frontend/components/FileList.ts
@@ -300,6 +300,7 @@ export class FileList extends LitElement {
           id=${ELEMENT_IDS.GENERATED_FILES}
           class="files-container"
           @click=${this.handleFileClick}
+          @keydown=${this.handleFileKey}
         >
           ${repeat(
             this.sortedRounds,
@@ -452,6 +453,8 @@ export class FileList extends LitElement {
             title=${tooltipPath}
             data-command=${PROGRESS_VIEW_COMMANDS.OPEN_FILE}
             data-file=${filePath}
+            role="button"
+            tabindex="0"
           >
             <span class="file-dir">${dir}</span>
             <span class="file-basename">${basename}</span>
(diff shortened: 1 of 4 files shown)


LionSR:
Addressed the requested FileList fix: keyboard activation now resolves only `.file-path[data-command]`, so native `wa-button` action buttons keep their own Enter/Space behavior and are not double-dispatched by the container key handler. I also reused the resolved file-path element directly instead of doing a second composed-path lookup.

Added a focused regression test that Space opens the file-path span but does not dispatch from a native action button.

Verified locally:

```sh
corepack pnpm vitest run --config vitest.config.mjs src/test-kernel/progressView/FileList.vitest.mts
```

Result: 1 file passed, 1 test passed.

PR #6194, opened 2026-06-18
fix: truncate by code points so truncateWithEllipsis never splits a surrogate pair

LionSR:
## Summary

`truncateWithEllipsis` truncated with `text.slice(0, maxLen - 1)`, which operates on **UTF-16 code units**. When the cut index landed inside a surrogate pair (any astral character — emoji, CJK extension, math alphanumerics), it left a dangling **lone high surrogate** immediately before the `…`, rendering as a broken replacement glyph (�).

It feeds CLI tool-error previews (`MAX_ERROR_PREVIEW`) and other one-line summaries, so a preview whose text has an emoji near the boundary showed a corrupted character.

## Fix
Slice by **code points** (`[...text]`) instead of code units — matching the code-point-safe iteration the CLI's own `clipToWidth` (`render/terminalText.ts`) already uses.

```ts
const codePoints = [...text];
if (codePoints.length <= maxLen) return text;
return `${codePoints.slice(0, maxLen - 1).join('')}…`;
```

## Verification
- **ASCII output is byte-identical** to the old behavior (verified across the existing examples + boundary cases); astral characters at the cut are now kept whole instead of split.
- `maxLen` now counts code points rather than UTF-16 units — the intended "characters" semantics (the doc comment is updated accordingly).
- Found by the parsing/escaping sweep; CI `validate` + the `WolframRunSummary` test (which exercises this util) confirm.

---

> [!NOTE]
> **Low Risk**
> Localized display-string utility change with no auth, persistence, or API surface impact; only `maxLen` semantics differ for strings containing astral characters.
> 
> **Overview**
> **`truncateWithEllipsis` and `tailWithEllipsis`** no longer cut on UTF-16 indices via `slice`; they spread the string into code points, enforce `maxLen` on that count, and join before adding `…`. That stops astral characters (emoji, etc.) at the boundary from becoming lone surrogates and broken `` glyphs in one-line previews (tool errors, session labels, summaries via `truncateSummary`).
> 
> Docs now state that **`maxLen` is code points**, not UTF-16 units. ASCII truncation behavior stays the same.
> 
> Vitest adds a **Unicode-safe ellipsis** block with surrogate-pair cases for both helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c399f7297f04d2a54f170009d35e5257ed688a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/utils/text/StringUtils.vitest.ts
+++ b/src/test-kernel/utils/text/StringUtils.vitest.ts
@@ -6,6 +6,8 @@ import {
   pluralize,
   splitContentLines,
   splitOutputLines,
+  tailWithEllipsis,
+  truncateWithEllipsis,
 } from '@utils/text/stringUtils';
 
 describe('splitContentLines', () => {
@@ -57,5 +59,15 @@ describe('pluralize', () => {
 
   it('honors explicit plural overrides', () => {
     expect(pluralize(2, 'person', 'people')).toBe('people');
+  });
+});
+
+describe('Unicode-safe ellipsis helpers', () => {
+  it('does not split a surrogate pair at the trailing truncation boundary', () => {
+    expect(truncateWithEllipsis('abc🍕def', 5)).toBe('abc🍕…');
+  });
+
+  it('does not split a surrogate pair at the leading truncation boundary', () => {
+    expect(tailWithEllipsis('abc🍕def', 5)).toBe('…🍕def');
   });
 });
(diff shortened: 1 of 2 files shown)


LionSR:
Addressed the review note: `tailWithEllipsis()` now truncates by Unicode code points as well, and `StringUtils.vitest.ts` covers both leading and trailing surrogate-pair boundaries.

Verified locally:

```sh
corepack pnpm vitest run --config vitest.config.mjs src/test-kernel/utils/text/StringUtils.vitest.ts
```

Result: 1 file passed, 13 tests passed.

PR #6199, opened 2026-06-18
fix(github): read the GitHub token via platform secrets (works in CLI/desktop)

LionSR:
## The dead-in-two-hosts bug

`getGitHubToken()` read from a module-scope singleton in `githubAuth.ts`:

```ts
let provider: GitHubTokenProvider = () => undefined;
export function getGitHubToken() { return provider()?... }
```

…and **only the extension ever filled it** (`extension.ts` cached `SecretManager.getGitHubToken()` and pushed it in via `setGitHubTokenProvider`). The **CLI and desktop never call the setter**, so `getGitHubToken()` returned `undefined` there and every GitHub tool/probe was silently dead in two of three hosts — a parallel owner of a fact `platform().secrets` already owns and all three hosts already wire.

## Fix — read through the secrets port

```ts
export async function getGitHubToken(): Promise<string | undefined> {
  const stored = await platform().secrets.get('github.token');
  const token = stored ?? process.env.GITHUB_TOKEN;
  return token && token.length > 0 ? token : undefined;
}
```

- The storage key `github.token` differs from the conventional `GITHUB_TOKEN` env var, so the env fallback is **explicit** (the per-host secret providers only probe `process.env[key]`, never `GITHUB_TOKEN`) — this is the subtlety the original `SecretManager.getGitHubToken` handled, now preserved host-neutrally.
- The three readers — `ghGet`, `getGitHubPRPrerequisites`, `requireToken` — already run in async contexts; `requireToken` becomes `async` and its two callers `await` it.
- Deleted: the `githubAuth` singleton + `setGitHubTokenProvider`, the `@tools/github` barrel re-export, the extension caching dance, and the now-dead `SecretManager.getGitHubToken` (its logic lives once now). The SecretStorage key is owned once by `githubAuth` (`GITHUB_TOKEN_STORAGE_KEY`) and referenced by `SecretManager`.
- The `secrets.onDidChange` listener still re-probes tool availability; `getGitHubToken()` now reads SecretStorage live (no cache to refresh).

## Behavior
- **Extension: preserved** — reads `context.secrets['github.token']` + `GITHUB_TOKEN` env, same as before.
- **CLI / desktop: fixed** — GitHub tools now resolve the token from their wired secret store + `GITHUB_TOKEN` env (previously always dead).

## Verification
- `tsc --noEmit` (root + extension) clean — no remaining `setGitHubTokenProvider`/`getGitHubToken` sync callers.
- prettier + eslint clean; `GitHubSubscriptionProgressEvents.vitest` passes (5/5).

> Note: the deep-audit finding proposed the **wrong** storage key (`GITHUB_TOKEN` instead of `github.token`) and mis-stated all readers as already-async; this PR uses the verified key + explicit env fallback and makes `requireToken` async.

---
🤖 Round-3–10 deep audit (parallel-singleton / incomplete-port lens), corrected.

---

> [!NOTE]
> **Medium Risk**
> Changes how GitHub credentials are resolved across extension, CLI, and desktop; behavior should match the extension path but any host with incomplete `platform().secrets` wiring could still fail silently.
> 
> **Overview**
> **GitHub PAT resolution is host-neutral** — `getGitHubToken()` is now async and reads `github.token` from `platform().secrets` (via `tryPlatform()`), with an explicit `GITHUB_TOKEN` env fallback. The extension-only injectable provider (`setGitHubTokenProvider`), in-memory token cache, and `SecretManager.getGitHubToken()` are removed.
> 
> **Call sites follow the async API** — `ghGet`, GitHub tool availability probes, and `requireToken` in the subscription tool `await` the lookup. The VS Code extension still re-probes tool availability when the GitHub secret changes; it no longer refreshes a cached token first.
> 
> **Single storage key** — `GITHUB_TOKEN_STORAGE_KEY` lives in `githubAuth` and is reused by `SecretManager.GITHUB_TOKEN_KEY`. Vitest covers env-only (pre-platform) and secret-over-env when the platform is initialized.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84e61ed33a43e4cc509d7c6b17dcb290309bd7bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/src/extension.ts
+++ b/packages/extension/src/extension.ts
@@ -37,8 +37,8 @@ import {
 } from '@auth/config';
 import { getAuthStatus } from '@auth/authCommands';
 import { AUTH_PROVIDER_ID } from '@auth/constants';
-import { tryResumeFromSnapshot } from '@commands/agent/resumeFromSnapshot';
 import { hasAnyUsableSetupCredential } from '@commands/setup';
+import { tryResumeFromSnapshot } from '@commands/agent/resumeFromSnapshot';
 import { createSampleProjectWithoutWorkspace } from '@commands/system/sampleProjectCommands';
 import { toErrorMessage } from '@common/errors';
 import { SIDEBAR_VIEWS, setActiveSidebarView } from '@common/webview';
@@ -86,7 +86,6 @@ import { setOpenPdfOpener } from '@tools/OpenPdfTool';
 import { refreshToolAvailability } from '@tools/toolAvailability';
 import { setSetupPlatform } from '@tools/setup';
 import {
-  setGitHubTokenProvider,
   prPollingSource,
   repoPollingSource,
   issuePollingSource,
@@ -523,15 +522,6 @@ export async function activate(context: vscode.ExtensionContext) {
       runCommand: (args) => runTerminalCommand(args),
     },
   });
-  // GitHub token lives in SecretStorage (managed via the Git settings tab).
-  // The tool layer only supports a synchronous token lookup, so we cache here
-  // and refresh on secret changes.
-  let cachedGitHubToken: string | undefined;
-  const refreshGitHubToken = async () => {
-    cachedGitHubToken = await SecretManager.getGitHubToken();
-  };
-  setGitHubTokenProvider(() => cachedGitHubToken);
-  void refreshGitHubToken();
   // VS Code's event emitters don't await async listeners, so we funnel
   // fire-and-forget async work through this helper to log rejections
   // instead of letting them become unhandled promise rejections.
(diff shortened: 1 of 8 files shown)


LionSR:
tests are failing

LionSR:
Addressed the CI failure.

Root cause: `getGitHubToken()` now read through `platform().secrets`, but `PRPollingSource` unit tests can subscribe before `initPlatform()`. That made the immediate poll throw `Platform not initialized`, which put the subscription into backoff; then the annotation-drain regression test skipped delivery because `skipPollUntilMs` was in the future.

Fix: use the existing `tryPlatform()` facade in `getGitHubToken()`. When the platform is initialized, the platform secret remains the source of truth; before initialization, the conventional `GITHUB_TOKEN` fallback still works and missing tokens return `undefined` instead of throwing.

Added focused coverage for both lookup paths.

Verified locally:

```sh
corepack pnpm vitest run --config vitest.config.mjs src/test-kernel/tools/GitHubAuth.vitest.ts src/test-kernel/tools/PRPollingSource.vitest.ts
corepack pnpm exec tsc --noEmit --pretty false
git diff --check
```

Results: focused Vitest 2 files passed / 6 tests passed; typecheck and diff-check clean.

PR #6237, opened 2026-06-18
fix(security): bind desktop OAuth callbacks to a per-attempt nonce

LionSR:
> ⚠️ **Draft pending a live desktop sign-in smoke-test.** The nonce logic, `texra://` URL parsing, and the test suite are all verified, but I could not run the live Supabase OAuth flow to confirm the redirect-URL allowlist accepts a `?app_nonce=` query param on the `texra://` callback. The existing Codespaces `?state=` routing token (documented in `src/auth/config.ts`) relies on the *same* redirect-param-preservation mechanism, so this is very likely fine — please confirm with one real sign-in, then un-draft.

## The vulnerability (login-CSRF / session fixation)

The desktop `texra://` OAuth callback was accepted on `hasPendingSignIn()` **alone**. `DesktopPendingOAuthState` carried only `{ createdAt }` — **no client secret** tying the inbound callback to the sign-in attempt *this machine* started.

The flow is **implicit** (tokens in the URL fragment) and GoTrue owns the OAuth `state` (a custom `queryParams.state` would overwrite it → `bad_oauth_state`), so there was no CSRF binding. An attacker who completes a real OAuth flow for **their own** account gets a valid token deeplink; delivered to a victim who is mid-sign-in (10-min TTL), the desktop app accepts it and **signs the victim into the attacker's account**.

## The fix

Bind each attempt to a one-time nonce that rides the callback URL:

1. `signIn` mints `randomBytes(16).toString('hex')` and sets `redirectTo = …/auth-callback?app_nonce=<nonce>`.
2. The nonce is persisted in the pending state (`DesktopPendingOAuthState.nonce`).
3. The callback handler extracts `app_nonce` from `callback.query` and **rejects** anything that doesn't match the pending nonce *before* a session is created.

Supabase preserves `redirect_to` query params through to the callback — the exact mechanism the **Codespaces `?state=` routing token** already relies on (`config.ts`) — so the nonce returns in the `texra://` callback query. A foreign deeplink (wrong nonce, or none) is dropped with a `warn`.

## Verification
- `tsc --noEmit` (desktop + `tsconfig.test-kernel.json`), prettier, eslint clean.
- `DesktopSupabaseAuth.vitest` **14/14**, including two new tests: a mismatched-nonce foreign callback and a no-nonce callback are both rejected (no `createSessionFromCallback`, no `storeSession`).
- Verified `new URL('texra://…?app_nonce=x').searchParams` and `URLSearchParams(callback.query)` parse the nonce correctly for the custom scheme.

Old persisted `{ createdAt }`-only state fails the stricter schema → treated as no-pending (a sign-in spanning the upgrade is dropped, not mis-trusted).

---
🤖 Security/integrity audit (auth-callback-trust lens). Re-verified against live code.

---

> [!NOTE]
> **Medium Risk**
> Changes authentication callback acceptance in the desktop app; incorrect nonce handling or Supabase not preserving `redirect_to` query params could break sign-in, but the fix is narrowly scoped and well-tested.
> 
> **Overview**
> Closes a **login-CSRF** gap where any valid `texra://` token deeplink could complete sign-in while the desktop app merely had a pending OAuth attempt, potentially signing the user into an attacker’s account.
> 
> Each `signIn` now generates a 32-hex `app_nonce`, appends it to the OAuth `redirectTo` callback URL, and stores it in persisted pending state. Incoming protocol callbacks must carry a matching `app_nonce` in the query string before session creation; mismatches or missing nonces are rejected with a warning. Pending state schema now requires `nonce`; legacy `{ createdAt }` blobs fail validation and are treated as no pending attempt.
> 
> Tests were updated to expect nonce-bearing redirects and add coverage for foreign callbacks, missing nonces, persist-before-OAuth ordering, and expired nonce matching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a791c529f86db8d0658eadcc2092fd41145e115. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/desktopSupabaseAuth.ts
+++ b/packages/desktop/src/main/desktopSupabaseAuth.ts
@@ -1,3 +1,5 @@
+import { randomBytes } from 'node:crypto';
+
 import { z } from 'zod';
 
 import { platform } from '@platform/platform';
@@ -30,6 +32,7 @@ const DESKTOP_PENDING_OAUTH_STATE_TTL_MS = 10 * 60 * 1000;
 
 const DesktopPendingOAuthStateSchema = z.object({
   createdAt: z.number(),
+  nonce: z.string(),
 });
 type DesktopPendingOAuthState = z.infer<typeof DesktopPendingOAuthStateSchema>;
 
@@ -41,7 +44,13 @@ export interface DesktopSupabaseAuth {
 
 export interface DesktopAuthCallbackState {
   hasPendingSignIn(): boolean;
-  beginAuthAttempt(): Promise<void>;
+  beginAuthAttempt(nonce: string): Promise<void>;
+  /**
+   * True only when a sign-in is pending AND its stored nonce equals `nonce`.
+   * Binds an inbound callback to the attempt THIS client started, so a verified
+   * but foreign-account token deeplink can't complete a sign-in (login-CSRF).
+   */
+  matchesPendingNonce(nonce: string | undefined): boolean;
   clearAwaitingCallback(): Promise<void>;
 }
 
@@ -107,9 +116,14 @@ export function createDesktopAuthCallbackState(
       }
       return true;
     },
-    async beginAuthAttempt() {
-      pendingState = { createdAt: Date.now() };
+    async beginAuthAttempt(nonce: string) {
+      pendingState = { createdAt: Date.now(), nonce };
       await persistPendingState(pendingState);
+    },
+    matchesPendingNonce: (nonce: string | undefined) => {
+      if (!nonce || !pendingState) return false;
+      if (isPendingOAuthStateExpired(pendingState)) return false;
+      return pendingState.nonce === nonce;
     },
     async clearAwaitingCallback() {
       pendingState = null;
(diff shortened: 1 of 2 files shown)


LionSR:
Addressed in af7062947 — now picks `&` vs `?` based on whether the callback URI already has a query string, so a future query param can't produce a malformed `redirect_to`. No behavior change today (the URI is bare). Tests still 14/14.

LionSR:
Live Supabase redirect probe for the draft gate:

- Started a real GoTrue OAuth authorize flow against `https://remote.texra.ai` using the production publishable key and `redirect_to=texra://texra-ai.texra/auth-callback?app_nonce=0123456789abcdef0123456789abcdef`.
- Supabase returned `302` to GitHub with `redirect_to=texra://texra-ai.texra/auth-callback?app_nonce=0123456789abcdef0123456789abcdef` preserved in the provider URL.
- Using the issued OAuth `state`, simulated the provider denial callback to `/auth/v1/callback` with redirects disabled. GoTrue returned `302` to:

```text
texra://texra-ai.texra/auth-callback?app_nonce=0123456789abcdef0123456789abcdef&error=access_denied&error_description=codex+smoke+denied#error=access_denied&error_description=codex+smoke+denied&sb=
```

This is not a browser-account success login, but it exercises the live Supabase authorize + callback state path and proves the merge-gating behavior we needed to know: the custom-scheme `redirect_to` query is accepted and `app_nonce` survives the GoTrue round trip back to the `texra://` callback.

PR #6273, opened 2026-06-19
refactor: extract roundMapSchema helper for round-keyed stream data

LionSR:
## Summary

A data-structure review flagged duplicated transformation code in the per-stream sidecar schemas. The three round-keyed schemas in `src/shared/schemas/streamData.ts` — `OutputFilesDataSchema`, `MissingOutputsDataSchema`, and `CompileFailuresDataSchema` — each repeated the same construction verbatim:

```ts
z.record(z.string(), <itemList>)
  .transform(recordToRoundMap)
  .catch(new Map()) as z.ZodType<Map<number, T[]>>
```

That's three copies of the same pipeline plus three copies of the same `as z.ZodType<…>` cast (needed because Zod widens the `.catch(new Map())` output type).

This PR extracts a generic `roundMapSchema<T>(itemList)` helper so each schema becomes a single declarative call and the cast lives in exactly one place.

## What changed

- Added `roundMapSchema<T>` next to `recordToRoundMap` with a doc comment explaining the key coercion, drop-empty-rounds, and malformed-fallback semantics.
- Rewrote the three schemas as one-line calls to the helper.
- Dropped the now-unused `OutputFileInfo` / `CompileFailure` type-only imports (they were only referenced by the removed casts).

Behavior is unchanged: identical key coercion (`RoundKeySchema`), identical drop-empty-rounds rule, identical malformed-input fallback to an empty `Map`.

## Verification

- `npm run typecheck` — passes
- `npx eslint src/shared/schemas/streamData.ts` — clean
- `npx vitest run src/test-kernel/transcript/ src/test-kernel/schemas/` — 38 tests pass (covers `StreamSnapshotStore` and `StreamDataUsage`, which exercise these schemas through the disk-read path)

## Scope note

The broader review surfaced larger structural items (the four parallel `Map<StreamTabId, Map<…>>` accumulators in `StreamSnapshotStore`, and the 11-optional-field `SyncStreamContentMessageSchema`). Those are higher-risk refactors that touch the webview contract and were intentionally left out of this focused, behavior-preserving PR.

https://claude.ai/code/session_011w4bNQ3Gagyyznz8nAD5hd

---
_Generated by [Claude Code](https://claude.ai/code/session_011w4bNQ3Gagyyznz8nAD5hd)_

---

> [!NOTE]
> **Low Risk**
> Localized schema refactor with a defensive parsing fix and new unit tests; no API or webview contract changes beyond safer empty-map fallbacks on malformed disk data.
> 
> **Overview**
> **Refactors** the three round-keyed sidecar parsers (`OutputFilesDataSchema`, `MissingOutputsDataSchema`, `CompileFailuresDataSchema`) in `streamData.ts` through a shared **`roundMapSchema`** helper so the `record` → `recordToRoundMap` → typed `Map` pipeline and its cast live in one place instead of three copies.
> 
> **Changes** malformed-parse fallbacks from `.catch(new Map())` to **`.catch(() => new Map())`** on those schemas and on **`UsageDataSchema`**, so Zod does not reuse one empty `Map` instance when sidecar JSON fails validation—important because **`StreamSnapshotStore`** keeps parsed maps by reference and mutates them.
> 
> Adds **`StreamDataRoundMaps.vitest.ts`** to lock in string round-key coercion, dropping empty rounds, empty-map fallback, and distinct fallback map instances per failed parse.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c0eb8a9a46db48d2a6a6af38843c323913f4daa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/shared/schemas/streamData.ts
+++ b/src/shared/schemas/streamData.ts
@@ -179,8 +179,14 @@ function recordToRoundMap<T>(record: Record<string, T[]>): Map<number, T[]> {
  * the schema itself does not fall back to insertion order.
  *
  * The single cast here replaces the per-schema casts that the
- * `.transform(recordToRoundMap)` → `.catch(new Map())` pipeline would
- * otherwise require at each call site.
+ * `.transform(recordToRoundMap)` → `.catch(…)` pipeline would otherwise
+ * require at each call site.
+ *
+ * The fallback uses a `() => new Map()` factory rather than a literal
+ * `new Map()`: Zod stores a literal catch value once and returns that same
+ * instance on every failure, and consumers (e.g. `StreamSnapshotStore`) hold
+ * the parsed map by reference and mutate it via `.set()`, so a shared instance
+ * would leak rounds across streams whose sidecar JSON is malformed.
  */
 function roundMapSchema<T>(
   itemList: z.ZodType<T[]>,
@@ -188,7 +194,7 @@ function roundMapSchema<T>(
   return z
     .record(z.string(), itemList)
     .transform(recordToRoundMap)
-    .catch(new Map()) as z.ZodType<Map<number, T[]>>;
+    .catch(() => new Map()) as z.ZodType<Map<number, T[]>>;
 }
 
 function warnDroppedItem(
@@ -203,7 +209,8 @@ function warnDroppedItem(
 }
 
 // ============================================================================
-// Output files: { round: OutputFileInfo[] }
+// Output files: { round: OutputFileInfo[] } (caller pre-flattens legacy input,
+// threading the activeRunId hint in via flattenLegacyRuns).
 // ============================================================================
 
 const OutputFileListSchema = z
@@ -218,13 +225,6 @@ const OutputFileListSchema = z
   )
   .catch([]);
 
-/**
- * Parses a flat round-keyed record to `Map<number, OutputFileInfo[]>`.
- * Legacy nested records (`{ runId: { round: items[] } }`) must be
- * pre-flattened by the caller (see `flattenLegacyRuns`) so the
- * activeRunId hint can be threaded in — the schema itself does not fall
- * back to insertion order.
- */
 export const OutputFilesDataSchema = roundMapSchema(OutputFileListSchema);
 
 // ============================================================================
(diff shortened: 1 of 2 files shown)


LionSR:
Addressed both reviews in `4ec25f3`:

1. **Doc-comment trim** (claude[bot] suggestion) — folded the redundant `OutputFilesDataSchema` JSDoc into the section banner so it matches the one-line style of its siblings.

2. **Lazy catch factory** (texra-review residual risk #1) — switched the helper's fallback from `.catch(new Map())` to `.catch(() => new Map())`. I verified this is a real latent bug rather than theoretical: with `.catch()` present, `safeParse` always succeeds, so on malformed sidecar JSON `readStreamData` returns the *shared* caught `Map`; `StreamSnapshotStore.applyStreamData` then stores it by reference (`this.outputFiles.set(stream, data.outputFiles)`) and `applyOutputFilesPatch` mutates it via `.set()`. Two streams with malformed `outputFiles.json` (or `missingOutputs.json` / `compileFailures.json`) would therefore share one map and leak rounds across streams. The `usage` map is unaffected because `applyStreamData` copies it. This pre-existed the refactor, but consolidating the three pipelines into one helper made it a single-line fix.

3. **Coverage gap** (texra-review note #2) — added `src/test-kernel/schemas/StreamDataRoundMaps.vitest.ts` with direct schema tests for key coercion, empty-round dropping, malformed-input fallback, and per-failure map isolation (the last would fail under the old shared-instance behavior).

Local verification: `npm run typecheck`, eslint, and the transcript + schema suites (41 tests now) all pass.

---

LionSR:
Applied the `UsageDataSchema` suggestion in `6c0eb8a` — switched its `.catch(new Map())` to `.catch(() => new Map())` so the file's catch pattern is uniform. As noted, this is defensive only: `StreamSnapshotStore.applyStreamData` copies the usage map (`new Map([...data.usage]…)`), so the shared literal was already safe in practice; the factory just guards a future by-reference reader. typecheck, eslint, and the transcript + schema suites (41 tests) pass.

---

PR #6276, opened 2026-06-19
refactor(auth): back TierService cache with lru-cache

LionSR:
## Summary

Replaces `TierService`'s bespoke single-slot promise cache with the already-present `lru-cache@11` (no new dependency). This came out of a "reinvented wheels" audit; after checking both audit candidates deeply, this is the one that's a genuine improvement.

The old cache hand-rolled five interacting pieces — `cache` + `cacheTimestamp` + `fetchPromise` + `fetchGeneration` + `cachedWithAuth` + `isCacheValid()` — to get TTL, concurrent-fetch dedup, auth-state invalidation, and out-of-order-resolution protection. lru-cache provides the first three; keying by auth state (`'auth'` / `'anon'`) replaces the fourth.

## Why this is cleaner, not just different

- **TTL, concurrent-fetch dedup, and in-flight abort** move to a maintained library instead of hand-rolled timestamp/promise plumbing.
- The **"a late anonymous response must never clobber the authenticated spend snapshot"** guarantee becomes **structural** (separate cache slots) rather than **timing-dependent** (the old `fetchGeneration` counter that compared generations after each await).
- **Sign-out** (`clearCache`) now aborts any in-flight fetch via its `AbortSignal`, so a response that lands after invalidation can't repopulate the snapshots.
- `fetchFromServer` is now **pure** with respect to instance state and signal-aware; `fetchMethod` owns the snapshot updates, so invalidation lives in one place.

## Behavior parity

Preserved for every reachable flow:
- The **sole** `getConfig` caller (`ServerSideKeyService.canUseServerSideKeys`) derives the token from real auth state, so the `'auth'`/`'anon'` key always tracks reality.
- Sign-out routes through `clearCache()` (wired via the `onCacheCleared` subscription), which resets the snapshots.
- Synchronous accessors (`getConfigSync`, `getProviders`, `getSpendingStatus`, …) keep their persist-past-TTL semantics via the `configSnapshot` / `userStatus` / `spendingStatus` fields.
- Failure handling matches the old "reset on failure, no negative caching" behavior: `fetchMethod` throws on transport failure / missing config / post-sign-out abort, so lru-cache drops the entry and the next call retries.

One deliberate, safer difference: an anonymous fetch no longer clears an authenticated snapshot on its own (only an explicit `clearCache` does) — which also aligns with the `preserveTierCache` intent.

## Not changed (and why)

The audit also flagged `logUtils.getTimestamp` for `Intl.DateTimeFormat`. On a deeper look that's **not** a clean win: the format is a fixed `YYYY-MM-DD HH:mm:ss.mmm` in **local** time. `Intl.DateTimeFormat` is locale-dependent and can't reproduce it without `formatToParts` reassembly that's longer than the current 6 lines; `toISOString()` would silently switch every log line to UTC; and no date library is a dependency. The explicit version is the simplest correct option, so it's left as-is.

## Testing

- `tsc --noEmit` on root + test-kernel projects: clean
- ESLint on changed files: clean
- `src/test-kernel/auth` (9 files, 59 tests): pass
- Added tests: concurrent-fetch dedup, `clearCache` snapshot reset. The existing "stale anonymous fetch can't clear authenticated spend" test stays green.
- Full suite: 3355/3356 relevant tests pass; the lone failure (`execUtils.vitest.ts` process-group kill) reproduces identically on the clean baseline — it's an environment/container process-kill timing issue, untouched by this change.

---
_Generated by [Claude Code](https://claude.ai/code/session_0114wfchJHdpqE4SWZmx98RL)_

---

> [!NOTE]
> **Medium Risk**
> Changes relay tier-config caching and quota/access snapshot timing on sign-in/out, but intent is behavioral parity with stronger isolation between auth and anonymous fetches.
> 
> **Overview**
> **`TierService`** drops its hand-rolled TTL/promise/`fetchGeneration` cache in favor of **`lru-cache`**, with separate **`'auth'`** and **`'anon'`** slots so anonymous `/tier-config` responses cannot overwrite authenticated spend or access snapshots.
> 
> **`getConfig`** now routes through `configCache.fetch`; **`fetchFromServer`** is stateless, passes **`AbortSignal`** into `fetch`, and throws on transport errors so failed entries are not negatively cached. Snapshot fields (`configSnapshot`, `userStatus`, `spendingStatus`) are updated only in the cache **`fetchMethod`**, and **`clearCache`** clears the LRU (aborting in-flight work) and resets those snapshots. Sync helpers read **`configSnapshot`** instead of the old single cache field.
> 
> **`ServerSideKeyService`** only clarifies why enabling included access does not pre-fetch tier config anonymously (that would fill the wrong cache slot).
> 
> Tests add coverage for concurrent dedup, post-**`clearCache`** abort behavior, and snapshot reset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf5b14cddc0a93846ea0ac32728cfef025a76d3d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/auth/serverKeys/ServerSideKeyService.ts
+++ b/src/auth/serverKeys/ServerSideKeyService.ts
@@ -239,8 +239,8 @@ export class ServerSideKeyService {
       this.clearAllCaches(cacheOptions);
       // No pre-fetch on enable — the next canUseServerSideKeys() does
       // its own auth'd fetch in parallel with fetchAccessStatus(), and
-      // an anonymous pre-fetch here would just be discarded by the
-      // cachedWithAuth mismatch in TierService.getConfig.
+      // an anonymous pre-fetch here would populate the 'anon' cache slot
+      // rather than the 'auth' slot needed for the access check.
       this._onDidChangeModelAccess.fire(value);
     }
   }
--- a/src/auth/tier/TierService.ts
+++ b/src/auth/tier/TierService.ts
@@ -13,6 +13,8 @@
  * - Ultra: All models above $3/M input
  */
 
+import { LRUCache } from 'lru-cache';
+
 import { toErrorMessage } from '@common/errors/errorMessage';
 import {
   SpendingStatusSchema,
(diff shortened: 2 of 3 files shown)


LionSR:
@claude fix

LionSR:
@claude fix

PR #6426, opened 2026-06-21
feat: experimental ChatGPT-subscription login for Codex models

LionSR:
## Summary

Implements `docs/proposals/chatgpt-subscription-codex-auth.md`: let a user sign in with their **own ChatGPT Plus/Pro/Team subscription** and drive OpenAI's Codex models from TeXRA instead of paying per-token against an API key.

Rather than a separate provider or duplicate picker entries, this is an opt-in **"prefer ChatGPT subscription" switch**: when it's on *and* you're signed in, the Codex-eligible OpenAI models you already see route through your subscription; everything else is untouched. So there's nothing new to "switch to" — sign in once, flip one setting, keep using the same model picker.

> Experimental / personal use. It rides an **unofficial** OpenAI endpoint (`chatgpt.com/backend-api/codex`) and the Codex CLI's OAuth client id, so it can break or be revoked without notice and is never presented as an OpenAI-sanctioned integration. Off by default; uses only the signed-in user's own session; tokens live in the OS keychain (file on the CLI), never logged.

## What's included

- **OAuth coordinator** (`src/auth/codex/`, host-neutral, `vscode`-free): PKCE (S256) loopback on the registration-fixed ports 1455/1457, headless **device-code** flow, token storage + single-flight proactive refresh (5-min buffer, fatal-vs-transient classification) via `platform().secrets`, and JWT account-id extraction (`id_token` preferred, three claim locations).
- **Codex Responses handler** (`modelHandlerCodex.ts`): subclasses the OpenAI Responses handler; the OAuth token is the SDK credential, targets the Codex backend with `chatgpt-account-id` / `originator: texra` / beta headers, and a custom `fetch` adapts the request to the backend's quirks — required non-empty `instructions`, `stream: true` (with SSE→JSON for non-streaming callers), `store: false`, no `max_output_tokens`; token-counting and compaction are disabled (no such endpoints).
- **Routing + availability**: a single shared predicate routes only when `preferSubscription && signed-in && eligible(model)`. New `subscription-access` availability kind runs in **both** API modes (it's your own credential), with a clear sign-in-required state otherwise.
- **The switch**: `texra.chatgptCodex.preferSubscription` (off by default, experimental), in the core settings schema + VS Code contribution.
- **CLI**: `texra auth chatgpt login | logout | status` (browser loopback; device-code default on non-TTY / `--no-input` / piped output), with JSON/NDJSON parity.

## Tested

- Unit: 24 new tests (PKCE, JWT claim extraction, refresh/single-flight/fatal-handling) + a ModelFactory routing guard. Full suite green (2128 tests), typecheck + lint clean.
- **Live, end-to-end against the real endpoint** via `texra-local`: browser loopback sign-in succeeded; a real agent run on `gpt-5.5` completed through the subscription (verified the request hits `…/codex/responses` over OAuth, not an API key). Debugging the live path is what surfaced the `instructions`-required and `stream: true`-required backend rules now handled here.

## How to use

1. `texra auth chatgpt login` (or, in the extension, set the switch in VS Code settings).
2. Enable `texra.chatgptCodex.preferSubscription` (VS Code settings, or `.texra/config.json` for the CLI).
3. Pick a Codex-eligible model (e.g. `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`) — it now runs on your subscription.

## Out of scope (v0)

In-extension "Sign in with ChatGPT" button in Settings → Models (the CLI + the config switch cover sign-in today), dynamic model discovery, and shared/server deployments.

---

> [!NOTE]
> **High Risk**
> Touches authentication, OAuth token storage/refresh, and a new unofficial third-party API path for model inference; breakage or policy changes on OpenAI’s side could affect users who enable the feature.
> 
> **Overview**
> Adds an **experimental, opt-in path** to run Codex-eligible OpenAI models through the user’s **ChatGPT subscription** instead of an API key or TeXRA relay.
> 
> **CLI & config:** New `texra auth chatgpt login | logout | status` (browser loopback or device code on headless/`--no-input`). Setting `chatgptCodex.preferSubscription` routes eligible models when signed in; if the switch is on but you’re not signed in, models still fall back to keys/relay.
> 
> **Auth stack (`@auth/codex`):** OAuth with PKCE, loopback on fixed localhost ports, device flow, tokens in platform secrets, single-flight refresh, and shared routing via `shouldUseCodexSubscription`.
> 
> **Runtime:** `ModelFactory` picks `ModelHandlerCodex` when preference + sign-in + eligibility match; handler uses OAuth against the unofficial Codex Responses backend (request shaping, streaming-only SSE collapse, no chaining/compaction/token-count APIs). `subscription-access` is a new model availability kind in both included and personal API modes. Base OpenAI Responses handler gains capability flags so Codex can disable chaining, compaction, and file upload without forking the whole path.
> 
> **Also:** CHANGELOG entry, shell completion for `auth chatgpt`, and broad unit tests for auth, routing, and handler behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2dd7d0a8aa13c9b8b1f4b516bdc44630d935f8b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
all reviews must be addressed
Also should run code simplifier

LionSR:
Check also ownership

LionSR:
Check also code spaghetti

LionSR:
Latest commit tops up the review fixes (the substantive items were already handled across the prior fix commits). Status of the remaining review threads:

**Addressed in code (threads not auto-marked resolved):**
- Auto-compaction is disabled for Codex — `shouldCompact()` now gates on `supportsManualCompaction` (which the Codex handler overrides to `false`), so the unsupported `/compact` endpoint is never called.
- System-prompt duplication is prevented by the `.includes()` dedup when hoisting system items into `instructions`.
- `accountId` is included in the login ndjson output (alongside `email`).
- Response chaining is disabled (`store:false` can't be chained) via `supportsResponseChaining=false`; inline file uploads are disabled for the Codex backend.

**Fixed in this commit:**
- The Codex request-adapter `catch` now logs a warning instead of silently forwarding an unparseable body.
- The wrapped auth error preserves the original `CodexAuthError` as `cause`.
- Unexpected failures reading ChatGPT session status are logged (still reported as signed-out).

**Known / inherited (non-blocking):**
- IPv6 loopback: the callback binds `127.0.0.1` while the borrowed client id fixes the redirect host to `localhost`. This mirrors the upstream codex-rs client and cannot use a different host/port; on an IPv6-first dual-stack box the timeout message points users to `--device`.
- In-extension Settings "Sign in with ChatGPT" button is intentionally out of scope for v0 (sign-in is via `texra auth chatgpt login` + the config switch).

Verified live end-to-end against the real endpoint: loopback sign-in, a single-turn `gpt-5.5` run, and a multi-turn chat (confirming chaining-disabled) all succeed through the subscription.

PR #6503, opened 2026-06-22
fix(agent): attach structured ProviderError to flow-failure rethrows

### LINKED ISSUE #6010: Attach structured ProviderError to flow-failure rethrows (full T2-2)
## Context

Follow-up from #5965 (SDK 7d T2-2/T3-2: consolidate result.error onto the structured ProviderErrorPartial shape).

The scoped PR landed the type consolidation (`ResultEvent.error` → `{ kind: AgentErrorKind } & ProviderErrorPartial`) and best-effort enrichment via `buildResultError`. Flow failures still rethrow a bare `Error` at `runReflectionFlow.ts:297` and `runToolUseFlow.ts`, so for those only `message` + the always-set `userRetryable`/`isRelayError` are present. The full T2-2 is needed to make flow failures carry the full structured shape.

## Details

1. **Widen the persisted `RetryErrorInfo`** to include the structured `ProviderErrorPartial` fields (`statusCode`, `provider`, `requestId`, etc.) so the retry infrastructure can surface them instead of sniffing `message`.
2. **Reconstruct/attach the structured shape at the two flow-exit rethrows** — `runReflectionFlow.ts:297` and `runToolUseFlow.ts` — so flow failures carry the full provider error metadata to `buildResultError`.
3. **Gate the duplicate ERROR rows** that appear when the flow and the lifecycle boundary both emit error trace events.

The `PersistedFlow` structuredClones shared state is noted as trap-dense (design doc), making this medium-risk work best suited for a focused, standalone pass.

## References

- PR: #5965
- Deferred scope documented in PR body "Scope (honest)" section
- Tracking issue: #4737

LionSR:
## Summary

Preserves structured provider-error metadata across flow-failure rethrows, so downstream logging/formatting recovers more than just `message` + `userRetryable`.

- Widen `RetryErrorInfoSchema` (`src/shared/schemas/errors.ts`) to carry optional structured `ProviderError` fields; add `toRetryErrorInfo` / `toProviderErrorFromRetry`.
- Widen `InvocationResult['failed']` / fallback handling in `src/agent/core/flows/RetryState.ts`, and propagate the full `RetryErrorInfo` through the cycle nodes (`ResponseCycleNode`, `ToolUseCycleNode`) into `shared.lastError`.
- Attach the structured shape at the rethrows in `runReflectionFlow.ts` and `runToolUseFlow.ts` via `attachProviderError` (`src/common/errors/`); recover it downstream by walking the `cause` chain in `normalizeProviderError`, which `getSdkErrorMessage` / `buildErrorLogData` now consume.
- `normalizeProviderError` no longer caches incidental formatting — it only migrates an *explicitly-attached* `ProviderError`. This avoids a stale-cache bug that lost late-attached `streamDiagnostics` / `partialText` (Anthropic streams) and kept status-stripped wrappers (OpenAI background polling) correctly user-retryable.

Schema changes are backward-compatible (new fields optional; `isRelayError` left `undefined` when unknown rather than forced to `false`).

> Note: an earlier draft also proposed a `markErrorLogged` / `isErrorLogged` dedup guard in `AgentRunLifecycle.ts`; that was deliberately dropped because the reflection flow emits no structured ERROR event, so the guard would have suppressed the only ERROR row for those failures.

Closes #6010.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm test` (error-pipeline + flow suites)

---

> [!NOTE]
> **Medium Risk**
> Touches shared error normalization and persisted retry-state shape; behavior changes how failures are logged and surfaced, but new schema fields are optional and tests lock in the caching and rethrow paths.
>
> **Overview**
> **Structured provider errors now survive agent run failures end-to-end.** `RetryErrorInfo` is widened (optional `statusCode`, `provider`, relay/credential flags, `requestId`, stream diagnostics, partial text) with `toRetryErrorInfo` / `toProviderErrorFromRetry`; retry `failed` results and cycle nodes propagate that shape into `shared.lastError`.
>
> **Reflection and tool-use flows** rethrow with `attachProviderError(toProviderErrorFromRetry(shared.lastError))` so lifecycle logging sees the same fields the retry layer already classified.
>
> **`normalizeProviderError`** only reuses errors explicitly attached at boundaries (including walking `cause`); incidental formatting no longer caches on the thrown value, so late `streamDiagnostics` / `partialText` and status-stripped wrappers (e.g. OpenAI background polling) stay correct. `isRelayError` on retry records is optional when unknown.
>
> Tests cover round-trips, wrapped causes, and the no-stale-cache stream path.

### DIFF
--- a/src/agent/core/flows/RetryState.ts
+++ b/src/agent/core/flows/RetryState.ts
@@ -309,9 +309,7 @@ export abstract class RetryableInvocationNode<
 
   protected getFallbackResult(
     error: Error,
-  ):
-    | { kind: 'cancelled' }
-    | ({ kind: 'failed' } & RetryErrorInfo) {
+  ): { kind: 'cancelled' } | ({ kind: 'failed' } & RetryErrorInfo) {
     if (this._userCancelled || isUserAbort(error)) {
       return { kind: 'cancelled' };
     }
--- a/src/agent/core/flows/RetryState.ts
+++ b/src/agent/core/flows/RetryState.ts
@@ -15,7 +15,11 @@ import {
   toErrorMessage,
 } from '@common/errors';
 import { isUserAbort } from '@common/errors/sdkErrorUtils';
-import { STREAM_STATUS, type RetryErrorInfo } from '@shared/schemas';
+import {
+  STREAM_STATUS,
+  toRetryErrorInfo,
+  type RetryErrorInfo,
+} from '@shared/schemas';
 import { getConfig } from '@utils/config/configUtils';
 
 const BACKGROUND_MODE_MIN_RETRIES = 3;
@@ -43,7 +47,7 @@ interface ManualRetryPromptResult {
 /** success: model response | failed: retries exhausted | cancelled: user cancelled | skipped: shouldStop was true */
 export type InvocationResult<TSuccess> =
   | ({ kind: 'success' } & TSuccess)
-  | { kind: 'failed'; message: string; userRetryable?: boolean }
+  | ({ kind: 'failed' } & RetryErrorInfo)
   | { kind: 'cancelled' }
   | { kind: 'skipped' };
 
@@ -307,7 +311,7 @@ export abstract class RetryableInvocationNode<
     error: Error,
   ):
     | { kind: 'cancelled' }
-    | { kind: 'failed'; message: string; userRetryable?: boolean } {
+    | ({ kind: 'failed' } & RetryErrorInfo) {
     if (this._userCancelled || isUserAbort(error)) {
       return { kind: 'cancelled' };
     }
@@ -322,8 +326,7 @@ export abstract class RetryableInvocationNode<
     }
     return {
       kind: 'failed',
-      message: formatted.message,
-      userRetryable: formatted.userRetryable,
+      ...toRetryErrorInfo(formatted),
     };
   }
 }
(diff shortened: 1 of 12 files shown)


LionSR:
@claude The dedup-guard regression is now fixed on this branch. There's a remaining follow-up that multiple reviewers (Cursor Bugbot + Codex) flagged: the structured `ProviderError` attachment added by this PR is **inert on the common failure path**, because the cycle nodes narrow `lastError` to `{ message, userRetryable }` before the flow-level rethrow. Please plumb the full structured error through, on this branch (`claude/issue-6010-20260622-1209`):

1. In `src/agent/implementations/flows/reflection/nodes/ResponseCycleNode.ts` (~L110-115, L151) and `src/agent/implementations/flows/tooluse/nodes/ToolUseCycleNode.ts` (~L115-120), write the **full `RetryErrorInfo`** (the widened shape, carrying `statusCode`, `provider`, `requestId`, `streamDiagnostics`, `partialText`, the relay/credit flags, etc.) into `shared.lastError` instead of just `message`/`userRetryable`. The round-level shared state already holds the richer record after `handleInvocationResult`; carry it out rather than reconstructing a bare `Error`.
2. Fix `toProviderErrorFromRetry` (`src/shared/schemas/errors.ts`) so it does **not** default missing `isRelayError` to `false` — leave it `undefined` when absent so `normalizeProviderError` doesn't read a wrong relay verdict from the cached shape.
3. Add kernel coverage that a tool-use and a reflection model-failure surface a reconstructed `ProviderError` with `statusCode`/`provider` populated end-to-end.

Leave the dead `detectProviderError` export and the `?? false` in `RetryState.ts` as-is unless trivial. Run `npm run typecheck`, `npm run lint`, and `npm test`.

---

PR #6593, opened 2026-06-24
feat(cli): add round-aware latexdiff command

LionSR:
## Summary

Addresses the **CLI half** of #6191. Wires the `texra` CLI to the host-neutral latexdiff core so terminal users get the same round-aware diffs the VS Code extension produces. (#6167 moved the core into the VS Code-free `src/latex/latexdiff/` zone but only re-pointed the VS Code host; the CLI had no latexdiff capability at all.)

The **desktop half** of #6191 is tracked separately in #6529, so this PR does not auto-close the issue.

## Single source of truth

The output-resolution + dispatch policy — prefer caller-supplied metadata → run-id-scoped run-dir scan → agent/model/input auto-discovery, then choose the metadata-driven vs workspace-scan engine — is extracted into `runLatexdiffForExecution` (`src/latex/latexdiff/runLatexdiff.ts`).

- The VS Code command now calls it instead of inlining ~80 lines of the same precedence logic.
- The new `texra latexdiff` command calls the same function — no duplicated policy across hosts.

## CLI command

```
texra latexdiff <agent> -m <model> -i <file> [-o <file>…] [--math-markup off|whole|coarse|fine] [--between-rounds] [--run-id <id>]
```

- `text` / `json` / `ndjson` output via `emitCliResult`.
- Platform init runs **before** the `latexdiff` tool probe so its debug logs never leak onto stdout and corrupt `--output-format json|ndjson` (headless parity).
- Exits non-zero only when a diff operation actually failed; "nothing to diff" is a success (exit 0).

## Tests

- `RunLatexdiff.vitest.mts` unit-covers the extracted resolution precedence (the correctness core), including:
  - a pinned `--run-id` whose scan misses does **not** silently auto-discover a different run;
  - an invalid `--run-id` neither scans nor auto-discovers.
- Completion snapshot updated for the new subcommand.

## Verification

- `npm run typecheck` — clean
- `npx vitest run src/test-kernel/cli src/test-kernel/latex` — 1488 passed
- `eslint` + `prettier` on changed files — clean
- Built `texra-local` and smoke-tested: `--help` renders clig-style usage + examples; missing `--model`/`--input` → exit 2; bad `--math-markup` → exit 2; `--output-format json` emits clean JSON (no log leakage); text/json/ndjson all correct.

https://claude.ai/code/session_01LWZ68B1fnWJovfgo9wa7Hk

---

> [!NOTE]
> **Low Risk**
> New CLI surface and refactored shared discovery logic, but behavior is covered by unit tests and mirrors the prior extension flow; no auth, billing, or data-model changes.
> 
> **Overview**
> Adds **`texra latexdiff`** so terminal users get the same round-aware LaTeX diffs as the VS Code extension: required agent name plus `--model`, `--input`, optional repeatable `--output`, `--math-markup`, `--between-rounds`, and `--run-id`, with **text / json / ndjson** output, progress on stderr, and exit codes that treat “nothing to diff” as success.
> 
> **Shared orchestration** moves output discovery and engine choice into **`runLatexdiffForExecution`** in `src/latex/latexdiff/runLatexdiff.ts` (caller metadata → run-dir scan when `--run-id` is set → auto-discovery only when unpinned → metadata vs workspace-scan engines). The extension latexdiff command delegates to this helper instead of inlining the same policy; `runLatexdiffFromMetadata` no longer takes `TaskRunFileService`.
> 
> Shell **completion snapshots** and **CHANGELOG** document the new command. **`RunLatexdiff.vitest.mts`** locks in precedence, including that a pinned run id whose scan misses does not fall back to a newer matching execution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 288d684172325d959481c59e5e17986b9a784c78. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Addressed both review suggestions in 288d68417:
- **`--quiet` gating** — the CLI progress reporter now also checks `!context.quietLogs`, matching sibling commands.
- **Dead `fileService`** — removed the unused param from `runLatexdiffFromMetadata`, its `TaskRunFileService` import in `diffOperations.ts`, and the now-pointless construction in `runLatexdiffForExecution`.

PR #6597, opened 2026-06-24
Extract OpenAI prefill logic and add schema path validation

LionSR:
## Summary

This PR extracts the OpenAI-compatible output initialization and prefill logic into a reusable support module, and adds compile-time validation to ensure `CORE_SETTING_PATHS` stays synchronized with the `CoreSettings` schema.

### Changes

1. **Refactored `initializeOutputAndPrefill` in `ModelHandlerOpenAIResponse`**
   - Extracted the 75-line implementation into a new `initializeOpenAiCompatibleOutputAndPrefill` support function
   - Replaced direct file operations and message manipulation with dependency-injected callbacks (`appendPseudoPrefill`, `pushAssistantText`, `addContinue`)
   - This enables code reuse across OpenAI-compatible providers and improves testability

2. **Added compile-time schema path validation in `CoreSettingsSchema`**
   - Introduced TypeScript type-level helpers (`LeafPaths`, `IsRecord`, `AssertNever`) to enumerate all dotted leaf paths in the settings tree
   - Added `satisfies readonly LeafPaths<CoreSettings>[]` constraint to `CORE_SETTING_PATHS` to catch typos or renamed paths at build time
   - Added `_AssertCorePathsExhaustive` type assertion that fails the build if any schema leaf is missing from the paths list
   - Updated documentation to explain the dual compile-time guards and their purpose

## Issue acceptance gates

N/A (refactoring and internal validation improvement)

## Single source of truth

- **OpenAI prefill logic**: Now owned by `initializeOpenAiCompatibleOutputAndPrefill` in `src/agent/modelHandlers/support/openAiCompatiblePrefill.ts`
- **Core settings paths**: Validated against `CoreSettings` schema via type-level assertions in `src/shared/schemas/coreSettings.ts`

- **Removed**: 75 lines of duplicated prefill/output initialization logic from `ModelHandlerOpenAIResponse`
- **Added abstraction**: `initializeOpenAiCompatibleOutputAndPrefill` with dependency injection for message operations, enabling reuse across OpenAI-compatible providers without pass-through layers
- **Improved validation**: Eliminated silent failures in host typo-detection by adding build-time guards that ensure `CORE_SETTING_PATHS` stays in lockstep with schema changes

## Host impact

Extension: No functional change; refactored internal logic with same behavior.

Desktop: No functional change; refactored internal logic with same behavior.

CLI: No functional change; refactored internal logic with same behavior.

## Scheduled refactoring

None

## Validation

- Refactoring preserves existing behavior: same prefill logic, same message handling, same file operations
- Build-time type assertions ensure `CORE_SETTING_PATHS` cannot drift from schema without explicit update
- Existing tests for `initializeOutputAndPrefill` continue to pass (behavior unchanged)

https://claude.ai/code/session_01MXTjnJqbnyuVFvjGL449rC

---

> [!NOTE]
> **Low Risk**
> Behavior-preserving refactor plus schema-path typing; no auth, data, or API contract changes.
> 
> **Overview**
> **OpenAI Responses prefill** is delegated to `initializeOpenAiCompatibleOutputAndPrefill` in `openAiCompatiblePrefill.ts`. `ModelHandlerOpenAIResponse.initializeOutputAndPrefill` now only wires provider-specific callbacks (pseudo-prefill append, assistant message push, continuation prompt) instead of inlining file checks and message shaping. When resuming from an existing output file without an end tag, that path logs at **debug** instead of **warn**.
> 
> **Core settings** gains type-level `LeafPaths` helpers and two compile-time guards on `CORE_SETTING_PATHS`: `satisfies readonly LeafPaths<CoreSettings>[]` catches invalid or renamed paths, and `_AssertCorePathsExhaustive` fails the build if a new schema leaf is missing from the list—so host typo-detection key sets cannot drift silently.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f65a4e6f8e0d8dfce314f0ee2660aa9e6d5fe35. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts
+++ b/src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts
@@ -44,10 +44,9 @@ import {
   getWebSocketEnabled,
   getUseOpenRouter,
 } from '@utils/config/providerConfig';
-import { flexibleFS } from '@utils/files/flexibleFS';
 import { toOpenAIReasoningEffort } from '../support/reasoningEffort';
 import { normalizeUsage } from '../support/UsageNormalizer';
-import { prepareExistingOutputContent } from '../utils/fileContentUtils';
+import { initializeOpenAiCompatibleOutputAndPrefill } from '../support/openAiCompatiblePrefill';
 import { tagOpenAISdkError } from './openAISdkError';
 import { withSdkErrorTag } from '../support/sdkErrorTagging';
 import {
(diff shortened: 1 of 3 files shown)


LionSR:
Thanks for the reviews. Consolidated response:

**1. `LeafPaths` robustness (optional groups) — fixed in f8663fc.**
Good catch on the soundness gap: `T[K] extends object` would let a future group declared `.optional()` without a default (`Group | undefined`) silently collapse to a single leaf key instead of failing the build. I hardened `LeafPaths` to apply `NonNullable` before the `extends object` test (with a comment documenting the assumption). It's a no-op for the current schema — every group uses `.prefault()` — so the derived path union is unchanged and the exhaustiveness guard still fires (verified: a planted typo still produces TS2820 + TS2344).

**2. Resume-path log level (`debug` → `warn`) — keeping `warn`, intentionally.**
This is a real delta, but the proposed fix (change `openAiCompatiblePrefill.ts:81` to `debug`) would do more than restore the old OpenAI-Responses behavior: that line lives in the **shared** helper that `modelHandlerOpenAI.ts` and `modelHandlerOpenRouterNative.ts` have used since before this PR, so both already `warn` on resume-from-partial. Switching it to `debug` would change *those two* handlers' behavior too — a new, broader change rather than a scoped revert. Routing OpenAI-Responses through the helper makes all three handlers consistent (which is the point of the dedup), so I've left it at `warn`.

The "preserves existing behavior" wording in the description should be read as *control-flow* preserving; the one observable delta is this handler's resume log moving `debug`→`warn` to match its siblings. If the team would rather this path be `debug` across **all** handlers, that's a deliberate one-line change to the shared helper — happy to do it as a follow-up if you'd prefer that direction; just say the word.

**3. Description nit:** as noted, `openAiCompatiblePrefill.ts` already existed (added by the prior OpenAI/OpenRouter dedup); this PR adds the **third** caller rather than creating it new.

**Typecheck:** root + extension + CLI `tsc --noEmit` are green locally, and the full model-handler vitest suite (216 tests) passes; CI's typecheck will confirm the guards compile.

---

LionSR:
Reversing my earlier call on the log severity — applied the fix in 6f65a4e.

On reflection the reviewers are right, and my "keep `warn` for consistency" reasoning had it backwards: the resume-from-partial path is a routine continuation, not an error, so `debug` is the correct level and consistency-*at-debug* beats consistency-at-warn. Two deciding facts:

- The original inline OpenAI-Responses code logged this at `debug` (confirmed at the pre-refactor `modelHandlerOpenAIResponse.ts`), so lowering the shared helper restores that severity and makes the PR's "preserves existing behavior" claim accurate for this handler.
- A `warn` on a normal resume is exactly the kind of thing that trips warning-based monitoring — a real downside, with no upside for a routine path.

So I changed `openAiCompatiblePrefill.ts:81` `warn` → `debug`. As the latest comment noted, this is in the shared helper, so it also corrects the level for the OpenAI Chat and OpenRouter callers. Verified: lint clean, and the prefill/Responses/OpenRouter vitest suites pass. This resolves the log-severity thread.

---

PR #6632, opened 2026-06-26
refactor: consolidate duplicate logic and use shared utilities

LionSR:
## Summary

Four behavior-preserving consolidations identified by an automated scan, re-based onto current `main`. No logic changes — only structural cleanup that removes duplicate field/block declarations.

### F9 — `historyViewMessages.ts`: make `HistoryIdMessageSchema` load-bearing

`HistoryIdMessageSchema` was exported as a base schema but none of its six siblings composed from it — they re-declared `historyId: z.string().min(1)` inline. All six now use `.extend()`, so the `historyId` contract has a single source of truth.

### F7 — `progressView/`: `StreamScopedBaseSchema` eliminates ~30 inline `stream` repeats

Defines `StreamScopedBaseSchema = z.object({ stream: StreamTabIdSchema })` once in `progressView/data.ts` and composes the qualifying inbound (21) and outbound (15) message schemas from it via `.extend()`, instead of copying the `stream` field verbatim. Schemas with special variants (`stream` as `.nullish()`, a `z.union([…])`, or a differently-named field such as `streamId`/`activeStream`) are deliberately left untouched. Re-targeted to the per-domain modules introduced by the schema-monolith split (`6ce0f4de8`).

### F6 — `applyStringReasoningToWorkspaceState`: remove a duplicated guard block

The identical guard-and-assign that sets thinking state was duplicated across the OpenAI and OpenRouter handlers. Extracted to a protected method on `ModelHandler`; both handlers call it.

### F2 — `shouldContinueOnLengthStop`: share OpenAI-family continuation logic

The identical continuation body was copied verbatim into both `ModelHandlerOpenAI` and `ModelHandlerOpenRouterNative`. Extracted to `shouldContinueOnLengthStop()` in `stopReasonUtils.ts` (which already owns related stop-reason predicates); both handlers delegate to it.

### Dropped during rebase — F10 (`WorkflowBaseSchema`)

The original scan also proposed an F10 `WorkflowBaseSchema` to share the housekeeping command base. `main` already implements that consolidation — and better — via `packages/extension/src/commands/housekeeping/fileOpSchemas.ts` (`FileOpParamsSchema` / `fileOpConfigFields`), which `cleanCommands.ts` and `packCommands.ts` already import. Adding `WorkflowBaseSchema` would have created a duplicate identical schema, so F10 was dropped; its intent is fully covered by the existing code.

---

Typecheck clean (only the known-preexisting `@openrouter/sdk` / `vscode-jsonrpc` module-resolution errors remain); 272 schema/handler tests pass.

### DIFF
--- a/src/agent/modelHandlers/ModelHandler.ts
+++ b/src/agent/modelHandlers/ModelHandler.ts
@@ -977,6 +977,20 @@ export abstract class ModelHandler<
     workspaceState?: AgentWorkspaceState,
   ): string | null;
 
+  /** Applies a single-string reasoning value to workspace thinking state.
+   *  No-op when workspaceState is absent or thinking was already recorded. */
+  protected applyStringReasoningToWorkspaceState(
+    reasoning: string,
+    workspaceState?: AgentWorkspaceState,
+  ): void {
+    if (workspaceState && !workspaceState.reasoning.thinkingAdded) {
+      workspaceState.reasoning.thinkingBlocks = [
+        { type: 'thinking', thinking: reasoning },
+      ];
+      workspaceState.reasoning.thinkingAdded = true;
+    }
+  }
+
   /**
    * Extracts tool-use information from provider responses.
    * @param responseObject The raw response object from the model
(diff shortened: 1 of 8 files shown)


LionSR:
Both bot reviews verified the four delivered consolidations (F9/F7/F6/F2) as behavior-preserving. The shared finding — F10 listed but absent — was correct: F10 was intentionally dropped during the rebase because `main` already implements that consolidation via `fileOpSchemas.ts` (`FileOpParamsSchema`). Updated the PR description to list only the four delivered changes and document why F10 was dropped.

Issue #6644, opened 2026-06-26
ModelFactory: single-source the handler routing precedence (compat-key vs createModelHandler drift)


LionSR:
## Problem

`src/agent/runtime/ModelFactory.ts` encodes the model-handler **routing precedence** in two functions that must be kept in lockstep:

- `modelHandlerCompatibilityKey()` (~L260-286) — a **pure, synchronous** predicate returning only the compatibility *key*. Used by exception-free callers (history restore, `modelSwitchDisabledReason`), so it must never throw and must not do async work.
- `createModelHandler()` (~L334-432) — the **async** live path that actually instantiates a handler.

Both walk the same ordered precedence (Validation → Responses → Google Interactions → OpenRouter → provider default). If a route is added or reordered in one, the other can silently drift, mismatching the compatibility key from the handler actually created.

## Why it isn't a trivial extract

The two are deliberately **not** identical, which is what makes a naive "share one decision function" risky:

- `createModelHandler()` has an extra **Codex-subscription** branch (`shouldUseCodexSubscription` + `await isCodexSignedIn()`) that `modelHandlerCompatibilityKey()` lacks — but it maps to the **same** key (`ModelHandlerOpenAIResponse`), which is precisely why the pure key function can omit it.
- `createModelHandler()` also performs config transforms (`codexBackendModelId`, `openrouterFullName`), applies `withReasoningOverride`, and calls the throwing side effect `assertGoogleInteractionsRoutable`. The pure key function must do none of these.

So the shared part is really just the **ordered key-precedence predicate** (`shouldUseResponsesAPI` → `shouldUseGoogleInteractionsAPI` → `openRouterOnly || useOpenRouter` → provider default).

## Proposed direction (for discussion)

Make the ordered precedence the single source of truth — e.g. have `createModelHandler()` compute the base key via `modelHandlerCompatibilityKey(config, useOpenRouter)` and `switch` on it for instantiation, keeping the async Codex special-case as an explicit override (key unchanged) before the switch. The pure key function stays the owner of precedence; the live path owns only instantiation + transforms + side effects.

## Risk

Touches the **live model-handler creation path** and model-handler territory is sensitive. Requires the full `ModelFactoryRouting` suite plus manual sanity across providers (OpenAI Responses, Google Interactions, OpenRouter, Codex subscription) before merge. Lower priority than the already-landed small SSOT cleanups (#6643); file here so the drift risk is tracked rather than rushed.

[...]

PR #6652, opened 2026-06-26
test(model-handlers): lock the compat-key ↔ createModelHandler routing invariant (#6644)

LionSR:
## Summary

Resolves the drift risk in #6644 — the safe way.

The handler-routing precedence is encoded twice in `ModelFactory.ts`:

- `modelHandlerCompatibilityKey()` — a **pure, synchronous** predicate that returns only the compatibility *key*. Used by exception-free callers (history restore, `modelSwitchDisabledReason`).
- `createModelHandler()` — the **async** live path that instantiates the handler and tags it with a key.

They walk the same precedence (Validation → Responses → Google Interactions → OpenRouter → provider default) but deliberately diverge in shape (the live path adds an async Codex-subscription override, config transforms, and a throwing `assertGoogleInteractionsRoutable` side-channel the pure predicate omits). If the two orderings drift, a restored session could key on a handler different from the one actually built.

## Why a guard test, not a live-path restructure

#6644 proposed consolidating the precedence into one owner. Having looked closely, restructuring `createModelHandler` to dispatch off `modelHandlerCompatibilityKey`'s key is **high blast radius**: a subtle mistake would route a model to the wrong provider API for end users, and the consolidation has to re-introduce the Interactions-under-OpenRouter throw and the Codex override by hand. The drift risk #6644 actually worries about is fully addressable without touching the live path.

This PR adds a **parametric invariant test**: for every one of the 128 registered models, the key tagged onto the created handler must equal `modelHandlerCompatibilityKey`'s prediction; a model with no route (predicted `undefined`, e.g. Copilot) must make `createModelHandler` reject. It **passes on current `main`** (the two encodings agree today) and **fails fast** on any future reorder/addition that desyncs them — turning a silent latent bug into a CI failure.

## Verification

- `npx vitest run …/ModelFactoryRouting.vitest.ts` — **27 passed** (the new guard sweeps all 128 models in ~2s).
- `npm run typecheck` clean; prettier clean.

The full structural consolidation can still be done later with maintainer awareness; this removes the urgency by making drift impossible to merge unnoticed. Test-only — no changelog entry.

---

> [!NOTE]
> **Low Risk**
> Adds regression tests only; no runtime or routing logic is modified.
> 
> **Overview**
> Adds a **parametric CI guard** so `modelHandlerCompatibilityKey` and `createModelHandler` cannot drift on the compatibility key used for history restore and model-switch gating.
> 
> The new test in `ModelFactoryRouting.vitest.ts` initializes a fake platform, re-imports `ModelFactory` (to survive `vi.resetModules()` elsewhere in the file), and sweeps **all** entries in `MODEL_CONFIGS`. For each model it asserts the key on the created handler matches the pure predicate’s prediction, and that models with no route (`undefined` key, e.g. Copilot) cause `createModelHandler` to **reject** instead of building a mismatched handler.
> 
> **Test-only** — no production routing changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e06a194fda55cc4029c862924453adb399f6596. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/agent/modelHandlers/ModelFactoryRouting.vitest.ts
+++ b/src/test-kernel/agent/modelHandlers/ModelFactoryRouting.vitest.ts
@@ -515,30 +515,53 @@ describe('routing precedence: compat-key ↔ createModelHandler invariant', () =
   // The handler-routing precedence is encoded twice: once in the pure
   // `modelHandlerCompatibilityKey` predicate (used exception-free by
   // history-restore and model-switch gating) and once in the live
-  // `createModelHandler` dispatch. The two must never drift — otherwise the
-  // key a restored session keys on could disagree with the handler actually
-  // built. For every registered model the key tagged onto the created handler
-  // must equal the predicted key, and a model with no handler route (predicted
-  // `undefined`, e.g. Copilot) must make `createModelHandler` reject rather
-  // than silently build a mismatched handler. Default routing (no OpenRouter
-  // proxy, signed out) is the shared baseline both paths observe, so they are
-  // compared under identical inputs.
+  // `createModelHandler` dispatch. The two must never drift on the *key* they
+  // produce — otherwise the key a restored session keys on could disagree with
+  // the handler actually built. For every registered model the key tagged onto
+  // the created handler must equal the predicted key, and a model with no
+  // handler route (predicted `undefined`, e.g. Copilot) must make
+  // `createModelHandler` reject rather than silently build a mismatched
+  // handler. Default routing (no OpenRouter proxy) is the shared baseline both
+  // paths observe. The Codex-subscription override in `createModelHandler` is
+  // intentionally absent from the predicate, but it is key-neutral: a signed-in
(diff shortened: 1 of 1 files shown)


LionSR:
Addressed in 1e06a19:
1. **Platform init (blocking)**: the test now does its own `initPlatform(createFakePlatform())` and re-imports the factory afterward (mirroring the Google Interactions block), so it survives the earlier `vi.resetModules()` and no longer free-rides on a prior test. Verified it passes under an isolated `-t` run.
2. **Codex coverage (suggestion)**: clarified the comment — the Codex override is *key-neutral* (`ModelHandlerCodex` is tagged `ModelHandlerOpenAIResponse`, the same key the predicate returns), so the key invariant holds in both states and the signed-out sweep is sufficient. Added handler `dispose()` hygiene too.

PR #6688, opened 2026-06-27
fix(latexdiff): discover headless run outputs via on-disk run-dir scan

### LINKED ISSUE #6687: latexdiff: `texra latexdiff -m <model> -i <file>` finds no outputs for headless `texra run` executions
## Summary

The documented `texra latexdiff <agent> -m <model> -i <file>` invocation returns `No LaTeX diff operations available for this run.` for any run created by a headless `texra run`, even though the run produced per-round outputs on disk and they are diffable via `--run-id`.

## Reproduce

```sh
texra run polish -i messy.tex -m deepseek --api-mode=byok --output polished.tex --no-input
# run completes, writes execution outputs to run storage

texra latexdiff polish -m deepseek -i messy.tex --output-format json
# => { "source": "workspace-scan", "executionId": null, "totalOperations": 0, ... }
# CLI prints: "No LaTeX diff operations available for this run."

# But scoping to the same execution works:
texra latexdiff polish -m deepseek -i messy.tex --run-id <execId> --output-format json
# => { "source": "run-dir-scan", "executionId": "<execId>", "totalOperations": 2, "successCount": 2 }
```

## Root cause

`discoverLatestExecutionOutputs` (`src/latex/latexdiff/outputDiscovery.ts`) matches the execution by agent/model/input, but retrieves its outputs **only** from the `StreamSnapshotStore` (a progress-view artifact). Headless `texra run` executions never persist those stream-tab snapshots, so discovery returns `null` and the orchestration drops to the cwd-based workspace scan, which only sees files copied back into the workspace — not the per-round execution outputs. The `--run-id` path works because it uses `scanRunDirForOutputs` (a direct on-disk scan of the run directory).

## Fix

Fall back to `scanRunDirForOutputs` (the same on-disk scan the `--run-id` path uses) when the stream-tab snapshot is empty, so both discovery paths share the same on-disk source of truth.

Found via hands-on CLI testing. PR incoming.

LionSR:
## Problem

`texra latexdiff <agent> -m <model> -i <file>` — the documented primary invocation — returns `No LaTeX diff operations available for this run.` for any run created by a headless `texra run`, even though the run's per-round outputs are on disk and `--run-id` diffs them fine.

## Root cause

`discoverLatestExecutionOutputs` matches the execution by agent/model/input but reads its outputs **only** from the `StreamSnapshotStore` (a progress-view artifact). Headless runs never persist those stream-tab snapshots, so discovery returns `null` and the orchestration falls through to the cwd-based workspace scan — which only sees files copied back into the workspace, not the per-round execution outputs. `--run-id` works because it uses `scanRunDirForOutputs` (a direct on-disk run-dir scan).

## Fix

When the stream-tab snapshot is empty, fall back to `scanRunDirForOutputs` (the same on-disk scan the `--run-id` path uses), so a headless run's outputs are discovered from its run directory. Both discovery paths now share the same on-disk source of truth. The change is additive and reuses already-proven code.

## Validation

- New regression test `src/test-kernel/latex/OutputDiscovery.vitest.mts` reproduces the headless case (matching execution, empty snapshot, on-disk `r0`/`r1` outputs) and asserts the fallback returns the outputs; plus a null case when neither source has outputs.
- Verified end-to-end against a real headless `polish` run: before, `latexdiff` reported 0 operations (`executionId: null`); the matched execution's outputs scan to 2 operations via `--run-id`, which this fix now reaches automatically.
- Existing `RunLatexdiff.vitest.mts` suite still green.

Fixes #6687

---

> [!NOTE]
> **Low Risk**
> Additive fallback in latexdiff output discovery reusing existing scanRunDirForOutputs; covered by new unit tests with no auth or data-model changes.
> 
> **Overview**
> Fixes **`texra latexdiff`** without `--run-id` reporting no diff operations for headless **`texra run`** executions, even when per-round outputs exist on disk.
> 
> **`discoverLatestExecutionOutputs`** still prefers stream-tab snapshots, but when that store is empty it now falls back to **`scanRunDirForOutputs`** on the matched execution’s run directory—the same on-disk path **`--run-id`** uses. Extra configured input files from the execution are passed as diff bases so multi-file runs don’t all compare against the single **`-i`** file.
> 
> Adds **`OutputDiscovery.vitest.mts`** covering the empty-snapshot + on-disk **r0/r1** case and the null case when neither source has outputs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a61433510fa76de4ef8e22121a37dbf15f44ab7b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/latex/latexdiff/outputDiscovery.ts
+++ b/src/latex/latexdiff/outputDiscovery.ts
@@ -250,6 +250,16 @@ export async function discoverLatestExecutionOutputs(query: {
       if (rounds && rounds.size > 0) {
         return { executionId: candidate.id, rounds };
       }
+      // Stream-tab snapshots are a progress-view artifact that headless
+      // `texra run` executions never persist. Fall back to scanning the matched
+      // execution's run directory on disk — the same source the `--run-id` path
+      // uses — so a headless run's outputs are still discovered instead of
+      // dropping through to the cwd-based workspace scan (which only sees files
+      // copied back into the workspace, not the per-round execution outputs).
+      const scanned = await scanRunDirForOutputs(candidate.id, query.inputFile);
+      if (scanned && scanned.size > 0) {
+        return { executionId: candidate.id, rounds: scanned };
+      }
     }
   } catch (error) {
     logger.debug(
--- a/src/latex/latexdiff/outputDiscovery.ts
+++ b/src/latex/latexdiff/outputDiscovery.ts
@@ -256,7 +256,15 @@ export async function discoverLatestExecutionOutputs(query: {
       // uses — so a headless run's outputs are still discovered instead of
       // dropping through to the cwd-based workspace scan (which only sees files
       // copied back into the workspace, not the per-round execution outputs).
-      const scanned = await scanRunDirForOutputs(candidate.id, query.inputFile);
+      // Mirror the `--run-id` path's base set: the matched execution's other
+      // configured input files become extra diff bases, so a multi-file run's
+      // secondary outputs diff against their own originals rather than all
+      // collapsing onto the single `-i` file.
+      const scanned = await scanRunDirForOutputs(
+        candidate.id,
+        query.inputFile,
+        candidate.agentConfig?.inputFiles?.slice(1),
+      );
       if (scanned && scanned.size > 0) {
         return { executionId: candidate.id, rounds: scanned };
       }
--- a/src/test-kernel/latex/OutputDiscovery.vitest.mts
+++ b/src/test-kernel/latex/OutputDiscovery.vitest.mts
@@ -1,0 +1,123 @@
+// Standard library imports
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
+import * as os from 'node:os';
+import * as path from 'node:path';
+
+// Third-party imports
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// `discoverLatestExecutionOutputs` matches a run by agent/model/input and then
+// reads its per-round outputs. Headless `texra run` executions persist those
+// outputs on disk but never write the progress-view stream-tab snapshot, so the
+// snapshot read comes back empty. These mocks reproduce that case: a matching
+// execution whose stream-tab snapshot is empty while its run directory holds
+// r0/r1 outputs on disk (the same source the `--run-id` path scans).
+const mocks = vi.hoisted(() => ({
+  listExecutions: vi.fn(),
+  resolveRunDir: vi.fn(),
+  readOutputFiles: vi.fn(),
+}));
+
+vi.mock('@agent/storage', async (importActual) => ({
+  ...(await importActual<typeof import('@agent/storage')>()),
+  listExecutions: mocks.listExecutions,
+}));
+
+vi.mock('@utils/files', async (importActual) => ({
+  ...(await importActual<typeof import('@utils/files')>()),
+  resolveRunDir: mocks.resolveRunDir,
+}));
+
+vi.mock('@transcript', async (importActual) => {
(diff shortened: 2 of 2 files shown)


LionSR:
What is this operation? Why does CLi need this?

LionSR:
## Replies

**@LionSR — "What is this operation? Why does CLI need this?"**

`texra latexdiff` is an existing CLI command (`packages/cli/src/commands/latex.ts`, with `--help` examples) — I didn't add it. It generates a round-aware, marked-up diff PDF showing exactly what an edit workflow changed (via `latexdiff` + the `merge` workflow), i.e. the CLI surface of the same diff feature the extension/desktop expose. CLI users run edit workflows headlessly (`texra run correct|polish|revise -i paper.tex …`) and want to see/share the changes as a diff. The run already auto-produces these diff PDFs; `texra latexdiff` (re)generates them on demand (e.g. a different `--math-markup` mode, or `--between-rounds`).

This PR doesn't add capability — it fixes a real defect: the documented `texra latexdiff <agent> -m <model> -i <file>` returned *"No LaTeX diff operations available"* for any **headless** `texra run`, because auto-discovery read outputs only from the progress-view stream-tab snapshot (which headless runs never persist). Only the `--run-id` workaround worked. If you'd prefer this stay extension-only and not be fixed for the CLI, happy to close — but as-is the listed command silently fails for headless users.

**Bot review — `extraBaseFiles` (Cursor Bugbot / claude / codex P2):** fixed in a61433510. The matched execution's `agentConfig.inputFiles` carries the full input list, so the fallback now passes the remaining entries as `extraBaseFiles` — multi-file secondary outputs diff against their own originals instead of collapsing onto the single `-i`. (This is the bit claude[bot] flagged as "not fixable here" — it is, using the matched execution's config rather than the CLI query.)

**Codex P2 — copy-back base (`-i paper.tex --output paper.tex`):** out of scope here. That's a pre-existing property of `scanRunDirForOutputs`, which the `--run-id` path already uses identically — both compute the base from the workspace file, not the execution's stored `original/` snapshot. This PR only routes headless auto-discovery onto that same proven path; changing the base source would affect both paths and belongs in a separate change.

LionSR:
## Re: the latest P2 suggestions (codex) + re-posted Bugbot note

Verified these are **pre-existing**, not introduced here. In `runLatexdiff.ts`, the `--run-id` path (L82–100) and this auto-discovery path (L107–122) both set `outputsByRound` and **converge on the same `runLatexdiffFromMetadata` call (L124–131)**. So:

- **Between-round grouping with scanned locations** — `runLatexdiffFromMetadata` receives the same `scanRunDirForOutputs` locations (round dir in `relativePath`) whether they came from `--run-id` or auto-discovery; identical behavior on both.
- **Output-file (`-o`) filtering** — `outputFiles` is never applied as a post-filter on the metadata path; it's passed to `scanRunDirForOutputs` (as base files) and to the workspace scan. Neither `--run-id` nor auto-discovery filters metadata rounds by `-o`. Pre-existing, shared.
- **Windows copy-as-dependency `.tex`** — a `collectTexFiles` property shared by every caller of `scanRunDirForOutputs`, including `--run-id`.
- **Re-posted extraBaseFiles (Bugbot, same bug id)** — addressed in a61433510: the fallback now passes the matched execution's `agentConfig.inputFiles.slice(1)` as extra bases (the Bugbot CI check passes; the inline note is a stale re-post on the synchronize commit).

My change only routes headless auto-discovery onto the same proven path `--run-id` already uses. Fixing the above would alter `--run-id` behavior too and belongs in a separate, scoped change rather than this bug fix.

PR is approved and green. @LionSR — holding merge for your go-ahead given your question above; happy to merge or close per your call.

PR #6731, opened 2026-06-28
fix(model-handlers): unify output cleanup, dedupe end-tag, finalize streams on error

LionSR:
## Summary

A model-handler tech-debt audit surfaced two real bugs and one duplication cleanup, all in `src/agent/modelHandlers/`. This PR fixes them together because they touch the same `extractResponse` / streaming code paths.

1. **Cross-provider output cleanup (bug).** `extractResponse` ran `replacementEngine.applyAll` (the LaTeX/text cleanup pass) for the Anthropic and Google handlers but **not** for OpenAI, OpenAI-Responses, or OpenRouter. The reflection flow hid this because it re-applies cleanup at `ResponseCycleFlow`, but the **tool-use flow** (`ToolUseProcessNode`) consumes the handler text directly — so the final assistant response from OpenAI/OpenRouter/xAI bypassed cleanup that Anthropic/Google output received. Now applied symmetrically in all handlers (xAI inherits via OpenAI).

2. **Un-finalized progress streams on mid-stream error (bug).** Only the OpenRouter handler finalized its thinking/output streams when a stream failed mid-flight. OpenAI, OpenAI-Responses, GoogleGenAI, and Google Interactions left them open, which can leave the progress view stuck in a loading state after a mid-stream failure. Each now finalizes on the error path.

3. **End-tag dedupe (refactor).** The duplicated "append the agent end tag on a natural stop" block (7 copies) is extracted into a shared `ModelHandler.appendEndTagIfNeeded` helper.

## Issue acceptance gates

No tracked issue — findings came from an internal audit. Acceptance: tool-use final responses from every provider receive the same text cleanup; a mid-stream error finalizes the progress streams for every provider; the end-tag logic has a single source.

## Single source of truth

- **End-tag appending** now lives in `ModelHandler.appendEndTagIfNeeded` (base class), replacing 5 inline copies.
- **Error-path stream finalization** for the OpenAI-Responses handler is owned by the new `ResponseStreamProcessor.abort()`, alongside the existing `finalize()`.

- Removed 5 copies of the end-tag append block in favor of one base-class helper.
- `appendEndTagIfNeeded` owns the invariant "append `endTag` only on a natural stop when not already present (`includes` check)"; callers still supply their own provider-specific natural-stop predicate, so no provider's stop vocabulary leaks into the base.
- `ResponseStreamProcessor.abort()` owns "finalize streams without a completed response"; it reuses the existing idempotent `StreamHandle.finalize`, adding no pass-through layer.
- Bug #1 is fixed by applying the existing `replacementEngine` symmetrically rather than adding indirection. Hoisting `applyAll` into the shared dispatcher was considered but rejected: two handlers route their streamed **display** text through `extractResponse`, so hoisting would have regressed the progress-view display to raw text. Left as a documented follow-up.

## Host impact

Extension: Indirect only — affects agent model output (cleaner OpenAI/OpenRouter tool-use responses; progress view no longer hangs on mid-stream errors). No API/UI changes.

Desktop: Same indirect behavior via the shared core. No host-specific changes.

CLI: Same indirect behavior via the shared core. No host-specific changes.

## Scheduled refactoring

Follow-ups identified by the audit, intentionally **not** included here to keep this PR focused and low-risk:
- Hoist `replacementEngine.applyAll` into the shared `extractModelResponse` dispatcher and fold display-text cleanup, removing the per-handler `applyAll` copies.
- Extract a `BackgroundPoller` collaborator for the duplicated OpenAI/Google background-mode poll loops.
- Extract a base-class `runStreamingResponse` template unifying the open/accumulate/finalize/tail-attach scaffold across handlers.
- Share a `createSettingsViewCommandHandlers` factory across the extension + desktop settings dispatch (80 duplicated command keys), mirroring the existing progress-view factory.

## Validation

- `npm run typecheck` — passes.
- `npm test` — all model-handler and flow suites pass (3970 passing). One unrelated failure, `src/test-kernel/utils/system/execUtils.vitest.ts` (POSIX process-group kill), reproduces on the clean base in this sandbox and is environmental, not caused by this change.
- `npm run lint` — changed files are clean (0 errors); pre-existing warnings elsewhere untouched.
- Added behavior is covered by the existing streaming/extract suites, including the Google Interactions streaming test that asserts the success-path finalize content is preserved (guarded so the error-path finalize does not overwrite it).

https://claude.ai/code/session_01XsAhWnjErzCEbASXYv9pV8

---
_Generated by [Claude Code](https://claude.ai/code/session_01XsAhWnjErzCEbASXYv9pV8)_

---

> [!NOTE]
> **Medium Risk**
> Changes span all major model handlers and shared response-cycle output paths; behavior fixes are targeted but affect final assistant text and progress UI on errors across providers.
> 
> **Overview**
> **Model output cleanup is applied once per provider** via each handler’s `extractResponse` (`replacementEngine.applyAll`), including OpenAI, OpenAI-Responses, OpenRouter, and the validation handler. **ResponseCycleFlow no longer re-applies cleanup**, avoiding double runs of non-idempotent replacements.
> 
> **End-tag appending on natural stop** is centralized in `ModelHandler.appendEndTagIfNeeded`, replacing duplicated inline logic in Anthropic/OpenAI handlers.
> 
> **Mid-stream failures** now finalize thinking/output progress streams (and OpenAI-Responses adds `ResponseStreamProcessor.abort()`) across Google GenAI, Google Interactions, OpenAI chat streaming, OpenAI-Responses HTTP/WebSocket, and OpenRouter—preserving partial streamed text instead of leaving the UI loading or overwriting with empty final text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e839525e15bbbf274dec8a4a372275167b4fe0b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/core/flows/ResponseCycleFlow.ts
+++ b/src/agent/core/flows/ResponseCycleFlow.ts
@@ -21,7 +21,6 @@ import { isTokenLimitStopReason } from '@agent/modelHandlers/utils/stopReasonUti
 import { K_SLICE } from '@agent/core/constants';
 import { toErrorMessage } from '@common/errors';
 import type { ToolDefinition } from '@model';
-import replacementEngine from '@replacement/engine';
 import { MESSAGE_TYPES, AgentFileLocationSchema } from '@shared/schemas';
 import { isApprovalGatedToolName } from '@tools/approvalGatedTools';
 import { AbsoluteFS, flexibleFS } from '@utils/files';
@@ -297,7 +296,10 @@ class ResponseProcessNode<C> extends BaseNode<
       let updatedAccumulatedOutput: string | undefined;
 
       if (newResponse) {
-        processedResponse = replacementEngine.applyAll(newResponse);
+        // Text cleanup (replacementEngine.applyAll) is applied once inside each
+        // handler's extractResponse, so newResponse is already cleaned here.
+        // Re-applying would run non-idempotent custom replacements twice.
+        processedResponse = newResponse;
 
         const connector = await this.services.bestConnectionMethod(
           prepRes.lastResponse.slice(-K_SLICE),
--- a/src/agent/modelHandlers/ModelHandler.ts
+++ b/src/agent/modelHandlers/ModelHandler.ts
@@ -652,6 +652,23 @@ export abstract class ModelHandler<
     return { endTurn, shouldStop };
   }
 
+  /**
+   * Append the agent's end tag to a natural-stop response when the model did
+   * not already emit it. Shared by the provider handlers that use an
+   * `includes` presence check; each caller supplies its own "natural stop"
+   * predicate because provider stop-reason vocabularies differ.
+   */
+  protected appendEndTagIfNeeded(
+    text: string,
+    endTag: string,
+    isNaturalStop: boolean,
+  ): string {
+    if (isNaturalStop && endTag && !text.includes(endTag)) {
+      return `${text}\n${endTag}`;
+    }
+    return text;
+  }
+
   protected containCutOffMessage(
     content: Array<{ type: string; text?: string }> | string,
   ): boolean {
--- a/src/agent/modelHandlers/anthropic/modelHandlerAnthropic.ts
+++ b/src/agent/modelHandlers/anthropic/modelHandlerAnthropic.ts
@@ -966,12 +966,11 @@ export class ModelHandlerAnthropic extends ModelHandler<
       .join('');
 
     // Add end tag if needed
-    if (
-      stopReason === ANTHROPIC_STOP.STOP_SEQUENCE &&
-      !newResponse.includes(endTag)
-    ) {
-      newResponse += `\n${endTag}`;
-    }
+    newResponse = this.appendEndTagIfNeeded(
+      newResponse,
+      endTag,
+      stopReason === ANTHROPIC_STOP.STOP_SEQUENCE,
+    );
 
     newResponse = replacementEngine.applyAll(newResponse);
 
--- a/src/agent/modelHandlers/google/modelHandlerGoogleGenAI.ts
+++ b/src/agent/modelHandlers/google/modelHandlerGoogleGenAI.ts
@@ -574,9 +574,11 @@ export class ModelHandlerGoogleGenAI extends ModelHandler<
     } catch (error) {
       // Finalize the progress streams on error so the view does not hang in a
       // loading state. Undefined on the non-streaming path; `finalize` is
-      // idempotent so a re-finalize after the success path is a no-op.
+      // idempotent so a re-finalize after the success path is a no-op. No
+      // explicit final text so any chunks already streamed are preserved
+      // (passing `''` would overwrite the visible partial output).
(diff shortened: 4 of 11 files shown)


LionSR:
Both Codex P2 findings were valid and are addressed in `3ac8c92`:

1. **WebSocket transport stream finalization** — the `OpenAIResponseWebSocketTransport` owns its own `ResponseStreamProcessor` and only returns it on `resolve`, so every reject path (`response.failed`, socket error/close, abort, synchronous send error) left its streams open — the same hang the HTTP-path fix targeted. Now finalized via `processor.abort()` on all reject paths (`rejectWithPartial`, with the send-catch routed through it).

2. **Preserving partial output on error** — finalizing the output stream with `''` emits `stream.end` with `finalText: ''`, which overwrites the partial output already visible in the progress view. Switched the error-path finalizes to `output.finalize()` (no argument), which preserves the appended chunks. Applied to the new `abort()`, the OpenAI/GoogleGenAI/GoogleInteractions error paths, and the pre-existing OpenRouter catch for consistency. (`thinking.finalize(undefined)` already preserved.)

Validation: `npm run typecheck` clean, model-handler suites pass (287), lint clean on changed files.

---

LionSR:
Good catch on the Codex P2 (double cleanup in reflection flows) — valid, fixed in `4187e64`.

Since this PR makes **every** handler's `extractResponse` apply `replacementEngine.applyAll`, the re-application at `ResponseCycleFlow.ts:300` was a redundant second pass on already-cleaned text, which corrupts output for non-idempotent custom replacements (e.g. `a -> aa`). I took the "remove the duplicate pass in the shared response cycle" option: dropped the `applyAll` at line 300 (and the now-unused import) so cleanup runs exactly once per flow — in `extractResponse`, which also keeps the progress-view display cleaned for the handlers that finalize via `extractResponse`. The tool-use flow gets its single pass from the same `extractResponse` change.

Note this also fixes the same latent double-apply that pre-existed for the Anthropic/Google handlers (their `extractResponse` already cleaned before this PR).

Validation: `npm run typecheck` clean, full `npm test` green (3970 pass; the one unrelated `execUtils` process-group failure reproduces on the clean base in this sandbox), lint clean.

---

LionSR:
Cursor Bugbot's "Validation handler loses text cleanup" finding (and the matching note in the latest claude[bot] review) is valid — fixed in `8e83952`.

`ModelHandlerValidation.extractResponse` was the one handler returning raw text; its cleanup previously came only from the response cycle's pass, which this PR removed. Added `replacementEngine.applyAll` to its `extractResponse`, completing the "every handler cleans in extractResponse" invariant the response cycle now relies on. Verified the invariant holds for all other handlers, including DeepSeek/GLM/Kimi/MiniMax (inherit `ModelHandlerOpenAI`) and xAI (`super.extractResponse`).

Validation: typecheck clean, 687 agent tests pass, lint clean.

---

PR #6738, opened 2026-06-28
refactor(settings): route state-backed reads through the catalog schema (#6609)

### LINKED ISSUE #6609: Unify the settings access layer through settingsAccess
Follow-up to #6602.

The catalog is the SSOT for keys/defaults/descriptions, but **reads/writes are still split**: `settingsAccess` is used by the CLI `/config` panel, while the extension/desktop read the same state-backed keys directly (`src/latex/texFormatter.ts`, `src/agent/modelHandlers/ModelHandler.ts`, …) with hand-passed defaults. (`gitAuthorSettings.ts` already references the shared default constant.)

Route those getters through `settingsAccess.readSetting(...)` so the schema `.prefault()` is the **single** default source.

**Low urgency:** defaults already converge via shared constants (`LATEX_CONFIG_DEFAULTS`, `DEFAULT_GIT_*`), so there's no live drift today — this just closes the "unified at the metadata layer, not the read/write layer" gap noted in review.

**Acceptance:** the state-backed read sites resolve their default from the catalog schema, not a separately-passed constant.

LionSR:
## What

The settings catalog (`stateSettings.ts`) is the SSOT for keys/defaults, but **reads were still split**: the extension/desktop/agent read state-backed keys directly via `platform().<store>.get(key, handPassedDefault)`, duplicating the catalog default as a second argument. This routes those reads through a single accessor so the default comes only from the entry's schema `.prefault()`.

Resolves #6609. Subsumes #6605 (same read/write-layer unification — closing that one as a duplicate).

## How

- **New `readPlatformSetting(key)`** (`src/utils/config/platformSettings.ts`): looks up the catalog entry, assembles the platform stores, and delegates to the existing `settingsAccess.readSetting` pipe. Graceful (returns the schema default) when the platform isn't initialized — matching the prior `tryWorkspaceState()?.get(...) ?? default` sites.
- **Rerouted 8 read sites**, dropping their hand-passed default constants:
  - `latex/texFormatter.ts`, `latex/formatter/indentDirectory.ts`, `latex/latexdiff.ts`, `latex/latexdiff/diffCommandExecutor.ts`
  - `agent/output/compileCheck.ts`, `agent/output/LatexDiffManager.ts`, `agent/implementations/flows/reflection/runReflectionFlow.ts`
  - `utils/config/providerConfig.ts` (the `WEBSOCKET_OPENAI` read — the one catalog-modeled key there; the other provider toggles aren't in the catalog and stay on the local `read`)

## Behavior

Defaults are **unchanged**: every rerouted key's `.prefault()` already equals the constant it replaced (verified against `LATEX_CONFIG_DEFAULTS` / `false`). The only behavior delta is the documented improvement from `readSetting`'s `safeParse`: a stale/invalid stored value now snaps back to the catalog default instead of propagating. The existing `Math.max(...)` range guards on the timeout reads are kept as-is.

## Verify

- `tsc --noEmit` (root + test-kernel + extension + cli): clean
- `eslint` on all changed files: clean
- New `PlatformSettings.vitest.ts` (default-from-catalog, stored-value, invalid-snaps-to-default, no-entry-throws) + `stateSettings` / `settingsConfiguration` / `LatexConfigPersistenceController` / `XmlOutputManager` suites: 40 passed
- Disjoint from the in-flight `codex/decouple-ui-agent-core` branch (no shared files).

---

> [!NOTE]
> **Low Risk**
> Mechanical read-path refactor with unchanged intended defaults; only behavioral delta is stricter validation of corrupt stored values.
> 
> **Overview**
> Introduces **`readPlatformSetting(key)`** so catalog-modeled workspace/global settings are read through `settingsAccess.readSetting` instead of `platform().<store>.get(key, handPassedDefault)`. Defaults now come only from each entry’s schema `.prefault()`; invalid stored values snap back via `safeParse`.
> 
> **Eight call sites** in LaTeX formatting/diff, workflow compile checks, reflection flow output policy, and `getWebSocketEnabled()` are switched over and no longer pass duplicated `LATEX_CONFIG_DEFAULTS` (or hardcoded `false`). Pre-`initPlatform()` behavior is preserved by returning the schema default when `tryPlatform()` is unset—replacing prior `tryWorkspaceState()?.get(...) ?? default` patterns.
> 
> Adds **`PlatformSettings.vitest.ts`** covering unset defaults, valid stored values, invalid snap-back, and missing catalog keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8eca0714518cae7545558171b9cd4b1ba064ec93. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/implementations/flows/reflection/runReflectionFlow.ts
+++ b/src/agent/implementations/flows/reflection/runReflectionFlow.ts
@@ -2,7 +2,6 @@
 import * as path from 'node:path';
 
 // Local imports
-import { platform } from '@platform/platform';
 import { getExecutionStore } from '@agent/storage';
 import type { StageHandle } from '@agent/trace';
 import {
@@ -37,7 +36,6 @@ import {
   type StorageKey,
   type WorkspaceFileLocation,
 } from '@shared/schemas';
-import { LATEX_CONFIG_DEFAULTS } from '@shared/constants/latex';
 import { WorkspaceStateKey } from '@shared/state/stateKeys';
 import {
   AbsoluteFS,
@@ -46,6 +44,7 @@ import {
   createWorkspaceLocation,
 } from '@utils/files';
 import { PromptBuilder } from '@utils/prompt';
+import { readPlatformSetting } from '@utils/config/platformSettings';
 
 // Local imports - flow nodes and state
 import { TeXCountNode } from './nodes/TeXCountNode';
@@ -352,14 +351,10 @@ export async function runReflectionFlow<C = unknown>(
 function createWorkspaceStateWorkflowOutputPolicy(): WorkflowOutputPolicy {
   return {
     shouldAutoOpenPdfOrLog: () =>
-      platform().workspaceState.get<boolean>(
-        WorkspaceStateKey.WORKFLOW_AUTO_OPEN_PDF,
-        LATEX_CONFIG_DEFAULTS.workflowAutoOpenPdf,
-      ),
+      readPlatformSetting<boolean>(WorkspaceStateKey.WORKFLOW_AUTO_OPEN_PDF),
     shouldRejectOnCompileFailure: () =>
-      platform().workspaceState.get<boolean>(
+      readPlatformSetting<boolean>(
         WorkspaceStateKey.WORKFLOW_REJECT_ON_COMPILE_FAILURE,
-        LATEX_CONFIG_DEFAULTS.workflowRejectOnCompileFailure,
       ),
   };
 }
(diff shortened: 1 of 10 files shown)


LionSR:
Thanks for the reviews. Addressed the one actionable suggestion in 8eca071: updated the `getLatexdiffTimeout` JSDoc (`latexdiff.ts:67-75`) to describe the `readPlatformSetting`/`tryPlatform()` graceful-default path instead of the old throwing `platform().workspaceState` wording.

Leaving the `<T>` note as-is: `readPlatformSetting`'s `as T` is the same assertion pattern as `settingDefault`/`readSetting` in this layer (both return `unknown`), and properly type-checking it would require threading a per-entry value type through the whole catalog — out of scope for this consolidation, as the review notes.

PR #6747, opened 2026-06-29
fix(output): recover percent-header workflow outputs

### LINKED ISSUE #6716: Add % filename extraction fallback for raw model output
## Context

Follow-up from #6715 (fix(output): warn when a non-empty response extracts zero files).

PR #6715 added a warning when the model returns content but nothing is extracted into output files — a safety net for the case where the model writes files using `% filename.tex` headers (a common reasoning-model pattern) instead of wrapping them in `<documentTag name="…">` tags. However, it only warns; it does not actually recover the files.

## Details

Implement an extraction fallback that parses `% <filename>` headers (e.g., `% chunk.tex`) from the raw model output and splits the content into the corresponding files, so that even when the model uses this alternate output format, the round produces usable `.tex` files rather than just a warning.

The implementation should:
- Recognize `% <filename>` lines as file boundaries in the raw output
- Extract content between such headers into the named output files
- Integrate into `OutputFileProcessor` or `XmlOutputManager` as a last-resort fallback (only when the primary `<documentTag>` extraction yields zero results)
- Respect existing file-naming and output-path conventions

## References

- PR: #6715
- PR body: "A `% <filename>` extraction fallback (to actually recover files in this format) can be a follow-up."

LionSR:
## Summary

This PR adds a last-resort workflow-output extraction fallback for raw model responses that use `% filename.ext` headers instead of XML document blocks.

The fallback now:

- preserves named `<document name="...">` regex recovery before percent-header scanning, so ordinary percent comments cannot preempt named document extraction;
- strips Markdown fence delimiters such as ```latex, ``` latex, and ~~~tex when they wrap the whole response, a recovered percent-header block, or a prose-prefaced fenced response;
- stops fenced percent blocks at their closing fence and ignores trailing explanatory prose until the next header;
- carries fence state across repeated empty headers and clears complete pre-header example fences before the real output;
- preserves fence-looking lines that are part of the recovered LaTeX document body;
- accepts `./file.tex`, `./dir/file.tex`, `_macros.tex`, `_generated/main.tex`, and backslash-separated relative paths, relying on the existing safe-path normalization;
- avoids treating `% file.tex` comments inside a LaTeX document preamble or open `document` environment as new file boundaries, including multiple filename-shaped preamble comments and comments before the first percent header;
- preserves likely LaTeX content that appears before the first percent header by routing it to the single input file when there is exactly one input;
- continues named percent-header recovery for multi-input runs when unroutable leading LaTeX content appears before the first header;
- allows subsequent real percent headers after documentclass-only or truncated blocks that never enter `\begin{document}`;
- ignores commented LaTeX begin/end markers when deciding whether a `% file.tex` line is a real boundary;
- ignores filename-looking lines inside `<scratchpad>...</scratchpad>` before percent-header recovery;
- keeps duplicate/colliding output names deterministic by reserving final output paths before writing.

Closes #6716.

## Verification

- `corepack pnpm exec vitest run src/test-kernel/agent/output/XmlOutputManager.vitest.ts`
- `corepack pnpm run lint` (passes with existing warnings)
- `corepack pnpm run test`

---

> [!NOTE]
> **Medium Risk**
> Changes agent output routing and on-disk file naming when fallbacks trigger; wrong boundary heuristics could mis-split or overwrite workflow artifacts, though existing XML/regex paths run first and tests are broad.
> 
> **Overview**
> **`XmlOutputManager`** gains a **last-resort fallback** that splits workflow outputs when models emit **`% path/to/file.ext` headers** instead of XML `<document>` blocks. It runs **after** XML parse and **named-document regex** recovery (on **raw** scratchpad output, with the thinking tag stripped).
> 
> The scanner treats filename-shaped `%` lines as file boundaries only when they are **not** LaTeX preamble/body comments, handles **markdown fences** (wrap, prefaced, and per-block closers), **ignores prose** after fenced blocks until the next header, and can **synthesize** leading LaTeX into the **single input file** when routing is unambiguous. Output names go through **safe-path normalization** and **collision avoidance** on final extracted paths.
> 
> Extensive **`XmlOutputManager.vitest.ts`** coverage documents precedence and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c30070c423437f8d2b5d3d8fcdb74f6f52d0dd51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/output/XmlOutputManager.ts
+++ b/src/agent/output/XmlOutputManager.ts
@@ -67,10 +67,25 @@ const EXTRACTION_METHOD_MESSAGES: Record<string, string> = {
 };
 
 const PERCENT_FILENAME_HEADER_REGEX =
-  /^%\s+([A-Za-z0-9][A-Za-z0-9._/-]*\.[A-Za-z0-9]+)\s*$/;
+  /^%\s+((?:\.\/)*[A-Za-z0-9][A-Za-z0-9._/-]*\.[A-Za-z0-9]+)\s*$/;
+const LATEX_DOCUMENT_BEGIN_REGEX = /\\begin\s*\{\s*document\s*\}/;
+const LATEX_DOCUMENT_END_REGEX = /\\end\s*\{\s*document\s*\}/;
 
 function isMarkdownFenceDelimiter(line: string): boolean {
-  return /^(?:`{3,}|~{3,})(?:\s+\S.*)?\s*$/.test(line.trim());
+  return /^(?:`{3,}|~{3,})(?:\S.*)?\s*$/.test(line.trim());
+}
+
+function isInsideLatexDocument(lines: readonly string[]): boolean {
+  let depth = 0;
+  for (const line of lines) {
+    if (LATEX_DOCUMENT_BEGIN_REGEX.test(line)) {
+      depth += 1;
+    }
+    if (LATEX_DOCUMENT_END_REGEX.test(line) && depth > 0) {
+      depth -= 1;
+    }
+  }
+  return depth > 0;
 }
 
 export class XmlOutputManager {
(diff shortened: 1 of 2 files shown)


LionSR:
Need to continue

PR #6758, opened 2026-06-29
fix(latex): resolve \input and \bibliography for subfolder documents

LionSR:
## Problem

When the main `.tex` file lives in a **workspace subfolder** (e.g. `Draft/LeanMPSPaper/Draft3.tex`), both the round-output compile check and the latexdiff compile fail:

```
! LaTeX Error: File `preamble.tex' not found.
l.2 \input{preamble}
```

and, for the diff, `\input{figures/...}` and `\bibliography{library,additional_references}` are unresolved too.

### Root cause

`compileLatex2Pdf` runs `latexmk`/`pdflatex` with the main file as an absolute path and `-output-directory=<flattened build dir>`, but never sets the subprocess cwd to the document's directory; `executeCommand` defaults cwd to the workspace **root**. TeX resolves relative `\input{...}` / `\bibliography{...}` against cwd + `TEXINPUTS`, **not** the main file's location. `TEXINPUTS` only contained `.` (= root), the workspace root, and the TikZ dir — none of which is `Draft/LeanMPSPaper/`. Flat projects work by accident (cwd happens to equal the file's dir).

Run-storage diff documents make this worse: the diff `.tex` is written to `diff/r{round}/` while its `\input`/`\bibliography` targets (and their mirrors) live elsewhere, so the relative paths never resolve.

The previous workaround was an `\input@path{{Draft/LeanMPSPaper/}...}` block in the user's **source** preamble — which a normal document does not have.

## Fix

Put the document's own directory and the original source directory on the kpathsea search path, so `\input`/`\bibliography` resolve exactly as if compiling from the document's folder ("from both locations"):

- **`compileLatex2Pdf`** prepends the main file's own directory, then caller-supplied `extraInputDirs`, then the workspace root, onto `TEXINPUTS`/`BIBINPUTS`/`BSTINPUTS`. File dir comes first so a **revised** sibling in run storage beats the original-source fallback.
- **`compileCheck`** passes the output's original workspace source directory (derived from the workspace-relative path).
- **`LatexDiffManager`** passes the base document's realpath-resolved source directory.

No source-file hack, works at any subfolder depth. Search-path composition is extracted into `buildLatexSearchParts`, a pure seam alongside the existing `buildLatexInputEnv`.

## Verification

- New unit tests for `buildLatexSearchParts` (ordering, document-dir-only, blank tikz).
- `npx vitest run src/test-kernel/latex` — 11 files / 52 tests pass.
- eslint/prettier clean on changed files.

---

> [!NOTE]
> **Medium Risk**
> Changes global LaTeX search-path ordering for all compiles (document dir before cwd), which fixes subfolder projects but could alter resolution if same-named files exist at workspace root vs document folder; latexdiff workspace mapping relies on stripping `r\d+/` prefixes.
> 
> **Overview**
> Fixes failed **workflow compile checks** and **latexdiff PDF builds** when the main `.tex` lives in a workspace subfolder or when run-storage copies compile from a separate build directory.
> 
> **`compileLatex2Pdf`** now builds kpathsea paths via new **`buildLatexSearchParts`**: the main file’s directory and optional **`extraInputDirs`** are prepended ahead of `.` (workspace-root cwd) and the workspace root on `TEXINPUTS` / `BIBINPUTS` / `BSTINPUTS`, so relative `\input{…}` and `\bibliography{…}` resolve like a compile from the document folder. **`LaTeXCompileOptionsSchema`** gains **`extraInputDirs`**.
> 
> **`compileCheck`** passes the original workspace source directory derived from the output’s relative path. **`LatexDiffManager`** adds **`resolveWorkspaceSourceDir`** (strips run-storage `r<N>/` prefix when mapping to the live workspace tree) and passes that directory when compiling diff `.tex` under `diff/r{round}/`.
> 
> Unit tests cover **`buildLatexSearchParts`** ordering and edge cases. Supabase edge functions bump **Hono** and **supabase-js** lockfiles only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d6fe871e1eb8280c5bc268f6fa30b8b8f08ce383. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/output/LatexDiffManager.ts
+++ b/src/agent/output/LatexDiffManager.ts
@@ -22,6 +22,7 @@ import {
   createWorkspaceLocation,
   flexibleFS,
   TaskRunFileService,
+  WorkspaceFS,
 } from '@utils/files';
 import { checkToolInstalled } from '@utils/system';
 import { getComparablePath } from '@utils/files/taskRunStorage';
@@ -59,6 +60,22 @@ export class LatexDiffManager {
       .fs.realPath(location.absolutePath)
       .catch(() => location.absolutePath);
     return path.dirname(resolved);
+  }
+
+  /**
+   * Directory of the document's live workspace source. `resolveBaseFilesForDiff`
+   * rewrites the base to its `executions/<id>/original/<rel>` snapshot, which
+   * holds only the snapshotted `.tex` files — not the `figures/` and `.bib`
+   * dependencies the diff still `\input`s. Map the workspace-relative path back
+   * onto the workspace root so the diff compile finds them; fall back to the
+   * file's own directory for external bases or when no workspace is open.
+   */
+  private resolveWorkspaceSourceDir(location: FileLocation): string {
+    const workspaceRoot = WorkspaceFS.getPath();
+    if (workspaceRoot && location.kind !== 'external') {
+      return path.join(workspaceRoot, path.dirname(location.relativePath));
+    }
+    return path.dirname(location.absolutePath);
   }
 
   private logLatexdiffResult(
@@ -376,10 +393,10 @@ export class LatexDiffManager {
       ),
     );
     // The diff `.tex` is written to `diff/r{round}/` rather than alongside the
-    // document's `\input`/`\bibliography` targets, so add the original source
-    // directory to the search path; otherwise `\input{figures/…}` and
+    // document's `\input`/`\bibliography` targets, so add the live workspace
+    // source directory to the search path; otherwise `\input{figures/…}` and
     // `\bibliography{…}` fail to resolve when the source lives in a subfolder.
-    const sourceDir = await this.getWorkingDirectory(referenceLocation);
+    const sourceDir = this.resolveWorkspaceSourceDir(referenceLocation);
     const ok = await compileLatex2Pdf(diffLocation, {
       channel: this.streamId,
       outputDirectory: buildDir,
--- a/src/agent/output/LatexDiffManager.ts
+++ b/src/agent/output/LatexDiffManager.ts
@@ -63,17 +63,21 @@ export class LatexDiffManager {
   }
 
   /**
-   * Directory of the document's live workspace source. `resolveBaseFilesForDiff`
-   * rewrites the base to its `executions/<id>/original/<rel>` snapshot, which
-   * holds only the snapshotted `.tex` files — not the `figures/` and `.bib`
-   * dependencies the diff still `\input`s. Map the workspace-relative path back
-   * onto the workspace root so the diff compile finds them; fall back to the
-   * file's own directory for external bases or when no workspace is open.
+   * Directory of the document's live workspace source — the one place that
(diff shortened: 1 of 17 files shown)


LionSR:
Thanks for the reviews — addressed the substantive findings in e48401936:

**Diff base resolved to the wrong directory (Codex P2 on `LatexDiffManager`, TeXRA review).** Correct and important: `resolveBaseFilesForDiff` rewrites the diff base to its `executions/<id>/original/<rel>` snapshot, which holds only the snapshotted `.tex` files — not the `figures/` and `.bib` the diff still `\input`s. Deriving the search dir via `realpath` landed on that snapshot, so figures/bib would have stayed unresolved. Now `resolveWorkspaceSourceDir` maps the workspace-relative path back onto the workspace root (falling back to the file's own dir for external bases).

**`.` ahead of the document dir (Codex P2 on `texTools`).** Moved the document + source dirs ahead of `.` in `TEXINPUTS`, so a subfolder document's relative `\input` resolves against its own tree rather than a same-named file at the workspace root (cwd).

**`compileCheck` derivation.** Confirmed sound for the in-place case: the output's run-storage `relativePath` is `r<N>/<workspace-rel>`; the `r<N>/` prefix is stripped, and the remainder is joined onto the workspace root, landing on the live source dir (with `figures/`+`.bib`).

Consciously **out of scope** for this minimal fix (pre-existing, multi-file-output only, not the single-main-file bug this targets): deriving paths from `outputFile.source` when an extracted output's basename differs from its source path, and threading revised round siblings into multi-file diff compiles. Happy to file a follow-up for those.

LionSR:
Good catch on the between-round case (Bugbot, medium) — fixed in af7e065d.

For between-round diffs the base is the prior round's output, whose `relativePath` is `r<N>/<rel>`, so joining onto the workspace root produced a nonexistent `<workspace>/r<N>/<rel>`. `resolveWorkspaceSourceDir` now strips a leading `r<N>/` segment, so both the round-diff snapshot base (`<rel>`) and the between-round prior-round base (`r<N>/<rel>`) map onto the live workspace source tree — the one place carrying `figures/` and `.bib`.

LionSR:
Addressed both nits in d6fe871e (comment-only, no behavior change):

- **Over-firing strip (Claude, low-pri/awareness):** documented the assumption in the JSDoc — a leading `r<digits>/` is treated as a run-storage round prefix, so a document whose live source genuinely sits under a top-level `r<digits>/` folder would be mis-stripped. As you noted, that's unavoidable with a relativePath-only heuristic and rare (it collides with round-dir naming), so I kept the heuristic and noted the assumption rather than adding FS probing.
- **`resolveBaseFilesForDiff` reference (texra-review):** it does exist — `src/agent/output/snapshotResolution.ts` — and I pinned the file in the JSDoc so it's not misleading.

PR #6771, opened 2026-06-29
fix(progress): make Approve/diff split-button carets read as one control

LionSR:
## Summary

This PR makes the approval and diff split buttons read as one segmented control. The main action half and caret half now share the same hover, focus-within, and open-menu border treatment, with the caret styled as part of the button rather than as a detached icon.

The review follow-up adds the missing border state on the main half for keyboard focus and open dropdown states in both the Approve split button and the shared diff-action styles.

## Verification

- `corepack pnpm exec prettier --write packages/extension/src/progressView/frontend/components/ApproveSplitButton.ts src/shared/styles/requestPanelStyles.ts`
- `git diff --check`
- `corepack pnpm run lint` (passes with existing warnings)

### DIFF
--- a/packages/extension/src/progressView/frontend/components/ApproveSplitButton.ts
+++ b/packages/extension/src/progressView/frontend/components/ApproveSplitButton.ts
@@ -79,30 +79,64 @@ export class ApproveSplitButton extends LitElement {
         min-width: 0;
       }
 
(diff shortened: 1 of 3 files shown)


LionSR:
a bit too complicated somehow feels

PR #6778, opened 2026-06-29
fix(cli): dedupe final transcript fallback

### LINKED ISSUE #6777: CLI chat duplicates final assistant transcript after streamed answer
## Summary

A real `texra chat` session can show the final assistant answer twice after a streamed response finishes. The stored stream log has only one `modelResponse`; the duplication happens in the TUI transcript projection when the final fallback answer is appended after the stream log has already materialized the same answer with slightly different Markdown punctuation.

## Reproduction

From a clean temporary workspace with included access available:

```sh
mkdir -p /tmp/texra-cli-real
corepack pnpm --filter @texra-ai/cli exec node dist/bin/texra.js chat \
  --cwd /tmp/texra-cli-real \
  --agent review \
  --model deepseekproT \
  --api-mode included \
  --approval-policy=never
```

In the TUI, submit:

```text
Solve this without tools: find all integer pairs (x,y) with 0 < y < 30 and x^2 - 2 y^2 = 1. Give a short recurrence argument.
```

After the answer finishes, the visible idle transcript repeats the answer: once from the stream-log `modelResponse`, and once from the final fallback response. In the observed run, the persisted stream log contained one `modelResponse`, while `flow_*.json` and `conversation.json` contained the final response text with small Markdown punctuation differences such as display-math periods and `\\[2pt]` spacing.

## Expected Behavior

If the stream log has already produced the final assistant answer for the current turn, the final fallback should not add a duplicate transcript row. The fallback should remain available only for cases where the stream log failed to materialize the assistant block.

## Notes

The ownership boundary is `packages/cli/src/chat/tui/state/transcript.ts`: fallback insertion should compare transcript content using a terminal-visible canonical form rather than strict byte equality. The runtime and chat-session controller do not need to change.

LionSR:
## Summary

- Gate final assistant fallbacks on stream-log ownership: after projection, the stream log suppresses a synthetic fallback only when the last non-synthetic row in the current turn is an assistant.
- If a tool follows an assistant, keep a different final fallback visible, but suppress a fallback that exactly matches earlier assistant text in the same turn.
- Keep same-anchor synthetic fallback dedupe, including the existing `\checkmark`/`✓` alias, and remove the rendered-text canonicalization helper.
- Fold coverage into the existing CLI TUI state suite and delete the standalone heuristic test file.

Closes #6777.

## Validation

- npm test -- --run src/test-kernel/cli/TuiStateAndFocus.vitest.mts src/test-kernel/cli/ConversationPane.vitest.mts
- npm run typecheck
- pre-commit npm run format

---

> [!NOTE]
> **Low Risk**
> CLI transcript display logic only; behavior is narrowly scoped to synthetic fallback insertion and covered by focused vitest cases.
> 
> **Overview**
> Refines when the CLI TUI adds **synthetic final assistant fallbacks** in `appendAssistantTranscriptIfMissing` so stream-log rows and fallbacks don’t fight each other.
> 
> **Stream-log ownership** now only blocks a fallback when the **last non-synthetic row in the current turn** is already an assistant—not when an earlier assistant message sits before a tool row. That restores a distinct **post-tool final answer** while still skipping the fallback when the stream log already ends the turn on assistant text.
> 
> **Deduping** is split: exact same-turn assistant text (with `\checkmark` / `✓` alias via `syntheticFallbackDedupeText`) still suppresses a matching fallback; repeated synthetic finals at the same stream anchor still dedupe; unrelated synthetic texts in the same turn are kept.
> 
> Tests in `TuiStateAndFocus.vitest.mts` were updated and extended for these cases (pre-tool vs post-tool, alias dedupe, distinct fallbacks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d606217e88ab61f7c8b3ea14e23eda7ee1f01438. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Looks sus

Solve the root cause not to put more band aids

PR #6812, opened 2026-06-30
refactor: simplify and consolidate code changed since v0.38.9 (round 1)

LionSR:
## Summary

Round 1 of a comprehensive, behavior-preserving simplification of the code that changed between `v0.38.9` and `main` (668 commits). This pass works at the file/module level: modern TypeScript, deduplication, and removing thin pass-throughs. A follow-up round will tackle the macroscopic/architectural consolidations surfaced here.

## Method

The changed surface (774 production source files) was split into **90 module-coherent batches** and simplified in parallel, each batch told to apply the full range of techniques but to edit only its own files and defer anything cross-module. **268 files were improved; the rest needed no change** (no churn where there was nothing to simplify).

## What changed

- **Pass-throughs removed** (~90): inlined single-use wrapper/factory functions, deleted re-export shims and identity factories, collapsed needless indirection layers.
- **Consolidations** (~164): unified duplicated run-host wiring, shared `try/catch` error-surfacing, and other near-identical branches behind helpers used 2+ times.
- **Modern TypeScript**: precise types over `any`, narrowing, optional chaining / nullish coalescing, `satisfies`, removed dead imports and redundant annotations, de-nested ternaries.
- **Zod v4 idioms** where applicable.
- Platform boundaries preserved: no `vscode` in agnostic zones, host services via `platform()`, native SDK types in model handlers.

## Verification

- Full workspace typecheck green across all four passes (root, `tsconfig.test-kernel.json`, extension, CLI).
- Four type regressions introduced by over-eager simplifications were caught by the typecheck gate and fixed by hand (union-narrowing in `ToolUseProcessNode`, `false`-vs-nullish in `texTools`, a Lit weak-type in `RecordingButtonController`, and an over-eager de-export of a test-covered function in `agentRegistry`).

## Reviewer notes

- A few batches consolidated a small public API and updated the matching test (11 `test-kernel` files). These touch test coverage; they are called out and verified in the review comment.
- The 74 cross-module consolidations the agents flagged but could not safely make in parallel (shared helpers spanning modules, duplicated subsystems) are deferred to the macroscopic Round 2.

---

> [!NOTE]
> **Low Risk**
> Refactor-only with no intended auth or data-model changes; risk is regression from wide touch surface, mitigated by full workspace typecheck per the PR description.
> 
> **Overview**
> Round 1 is a **behavior-preserving cleanup** across ~268 touched files (CLI TUI, chat session controller, desktop shell, VS Code progress view, and assorted commands/runtime helpers). The intent is less indirection and duplicated branches, not new product behavior.
> 
> **CLI chat & TUI:** `setupRunHost` centralizes runtime-host wrapping, approval installation, and `.finally` teardown for both **start** and **resume** runs. Slash commands route async work through `runGuardedSlashCommand` so failures consistently land in the transcript. Small TUI extractions (`foregroundMaxRowsForKind`, `entryCols`, `errorCornerLines`, harness stream bootstrap helpers) replace repeated inline logic.
> 
> **Extension progress view:** Stream state/log updates move to module-level `setStreamStateForId` / `setStreamLogsForId` in `progressState` (usable from desktop without duplicating `ProgressApp` private methods). Shared patterns appear for `wa-select` reads, agent-review QuickPicks, latexdiff “run and open,” and main-view option refresh. LaTeX/web formatters and diff-tab closing get similar deduplication.
> 
> **Desktop & commands:** Desktop command dispatch uses `requiredAction` / `optionalAction` helpers; diff opening and bootstrap recovery are factored. CLI command handlers often build one **payload** object reused for JSON and ndjson output.
> 
> **Types & hygiene:** Narrower types, removed dead imports/re-exports, and collapsed ternaries; platform boundaries (no `vscode` in agnostic zones) are unchanged per the PR description.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc9b5f3952e00b441874963061f695388b374f48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
## Review — round 1 simplification

Adversarial review of the full diff (274 files), run as 51 area reviewers hunting silent behavior changes plus the TeXRA platform/Zod/render-time rules, with a refute pass on every blocker/should-fix finding. The diff is behavior-preserving; the two issues surfaced are below, both resolved in `9618162d8`.

### Issues found and addressed
- `src/controllers/settingsView/SettingsProfileController.ts:88` — replacing the `.find()` lookup with `new Map(entries)` flipped duplicate-key resolution from first-match to last-match. The key `texra.model.useBackgroundResponses` is defined under two providers (`src/shared/constants/providers.ts:203` and `:237`), so `.get()` began resolving to the google def instead of the openai one. Harmless today (only `globalStateKey` is consumed, undefined for both), but a latent trap if a future duplicate carries differing backing stores. Fixed to keep the first def per key while retaining the Map's O(1) lookup.

### Notes (no change needed)
- `src/agent/export/normalizeConversation.ts:149` — collapsing the `inlineData`/`fileData` branches into `inlineData ?? fileData` differs from the original `&&`/typeof guards only when a part field is falsy-non-nullish or truthy-non-object. Both are typed `Blob | undefined` / `FileData | undefined` in the GenAI SDK (always objects when present), so the divergence is unreachable with real data. Restoring exact semantics would re-introduce the nested ternary the pass removed; left as-is.

### Completeness check
Because two reviewers independently flagged the `&&`→`??`/`?.` truthiness class, I swept the entire diff for it: all 82 added `??`/`?.` sites are legitimate default-value or object-safe-navigation uses — no primitive (numeric/boolean/length) receivers that could falsy-flip. The class is fully covered by the two findings above.

## Verified
- Ran `git diff origin/main...HEAD` across all 274 files in 51 module-coherent batches; behavior preservation checked hunk by hunk.
- Confirmed the duplicate provider key at `src/shared/constants/providers.ts:203` and `:237`; verified the fix preserves first-match.
- Read `src/agent/export/normalizeConversation.ts:120-175`; confirmed the `?? ` divergence is unreachable under the GenAI `Part` types.
- Full workspace typecheck green across all four passes (root, `tsconfig.test-kernel.json`, extension, CLI) after the fix.
- `deno check` on `supabase/functions/relay` (run from the function dir with its own `deno.json`) is clean — no new errors vs main.
- Platform boundaries: no `vscode` import introduced in any agnostic zone; no new `bus.emit`/subscribe surfaces.
- eslint: the 33 import-order warnings the parallel passes perturbed are restored to the repo baseline; remaining warnings are pre-existing on main.

Issue #6828, opened 2026-06-30
Converge the two parallel run/follow-up state machines (CLI PQueue vs extension/desktop imperative StreamStatusService)


LionSR:
## Problem

The hosts run two parallel run/follow-up state machines for the same conceptual lifecycle:

- The **extension and desktop** drive run lifecycle imperatively through `StreamStatusService` + `ToolUseFollowUpQueue` + a `resumeAttempts` Set.
- The **CLI** models the same conceptual run/follow-up queue with `PQueue` (`packages/cli/src/chat/chatSessionController.ts` imports `p-queue`, `readonly followUpQueue: PQueue`) plus its own `sessionResume.ts`.

The per-area simplification passes repeatedly carved the CLI out ("different model, out of scope") in the resume/queue findings, but never named the existence of two parallel run state machines as its own concern. This is the single largest cross-host runtime-coupling gap.

## Why it matters

Every resume / follow-up / stop / kill behavior has to be implemented and kept correct twice, in two different concurrency models. Several smaller findings (shared resume-queue helper, resume-resolution discriminated union, `AgentResumePort` orchestrator — see related issues) are really symptoms of this split; they reduce the duplicated surface but do not unify the underlying model.

## Proposed direction

Converge on one host-neutral run/follow-up state machine in shared core that both the imperative hosts and the CLI adapt to, rather than two independent engines. Scope and sequencing TBD — likely after the smaller resume/queue consolidations land and de-risk the seam.

## Notes

High value, high risk, its own project. Surfaced by the completeness critic in the macroscopic cross-host coupling analysis during the post-0.38.9 simplification effort.

LionSR:
Related symptoms of this split, scoped smaller: #6826 (resume orchestration) and #6827 (progress-event reducer). Those reduce the duplicated surface; this issue tracks unifying the underlying run/follow-up state machine.

LionSR:
better to be declarative

LionSR:
i like p-queue

LionSR:
## Deferring the core convergence (scope note after scouting)

A macroscopic scout + implementation pass on the related resume/queue issues confirms this one should stay its own project, gated on the smaller consolidations. Status:

[...]

Issue #6827, opened 2026-06-30
Share the progress-event → per-stream-meta reducer (CLI forks the webview's streamMetaSlice)


LionSR:
## Problem

Two independent reducers turn the same progress-fact stream into per-stream view state: the webview frontend `streamMetaSlice` (reused by the extension and desktop renderers) and the CLI's `applyToState`. They duplicate bounded process-output tail capping, stale-output pruning, and the active-subagents / active-processes / conversation-progress / todos / plan / usage reductions. The CLI comments explicitly say it "mirrors" the webview. Inputs differ (raw `ProgressEventPayloads` vs IPC `PROGRESS_VIEW_COMMANDS`) and output containers differ (Ink signal map vs Immer Map store), so the logic was forked rather than shared.

## Evidence

- `packages/cli/src/chat/tui/state/subscribeRuntimeHost.ts:77-269` — `applyToState` reduces setActiveStream / setTaskState / updateStreamUsage / updateConversationProgress / updateStreamDescription / updateActiveSubagents / updateActiveProcesses / updateProcessOutput / updateTodos / updatePlan
- same file `:164-165` comment "mirrors pruneStaleOutputs in the webview's streamMetaSlice"; `:90` "mirrors the extension progress view contract"; `:197-198` process-output tail cap
- `packages/extension/src/progressView/frontend/slices/streamMetaSlice.ts:47` `pruneStaleOutputs`, `:40-44` `capOutput`, `:73-171` `streamMetaHandlers` reducing the same facts
- `packages/desktop/src/main/desktopProgressEventBridge.ts` — translates `ProgressEventPayloads` into the `PROGRESS_VIEW_COMMANDS` the same frontend slice consumes, so desktop reuses `streamMetaSlice` while the CLI forks it

## Proposed direction

Extract the pure per-stream reduction (status, conversation, subagent/process badges, bounded process-output tail with stale pruning, todos, plan, usage) into a host-neutral reducer in shared core keyed off `ProgressEventPayloads`, returning a normalized `StreamMeta` value object. The webview slice and CLI `applyToState` become thin adapters copying normalized fields into their own stores.

## Notes

Large surface, high blast radius — architectural direction, not a single safe PR. Surfaced by the macroscopic duplication analysis during the post-0.38.9 simplification effort.

Issue #6826, opened 2026-06-30
Unify cross-host resume orchestration (AgentResumePort reimplemented in extension + desktop)


LionSR:
## Problem

`AgentResumePort.tryResumeStream` has two real implementations (VS Code extension and Electron desktop) that duplicate the same resume-orchestration skeleton: guard on `isActiveOrResuming`, gather `taskState` + `executionId`, call the shared `retrieveSessionResumeData`, then branch on `type === 'toolUse'` vs `'workflow'`. Only the leaf `retrieveSessionResumeData` is shared today; the surrounding orchestration is forked, and the desktop additionally reimplements run/resume-new logic that the extension routes through `ProgressWorkflowActionsController` (desktop `runNewStream` is effectively byte-identical to the controller's `runNew`).

## Evidence

- `src/platform/interfaces/agentResume.ts:10` — `AgentResumePort.tryResumeStream(streamId): Promise<boolean>`
- `packages/extension/src/commands/agent/resumeFromSnapshot.ts:51` — `retrieveSessionResumeData(...)`; `:71` branch `if (resumeData.type === 'toolUse')` → dispatches `texra.resumeAgent` / `texra.execute`, reads state from the `ProgressViewProvider` singleton
- `packages/desktop/src/main/desktopAgentExecution.ts:884` — `retrieveSessionResumeData(...)`; `:896` the same branch inlined; `:1081` `runNewStream` duplicates the controller
- `src/controllers/progressView/ProgressWorkflowActionsController.ts:70` — `runNew` with the identical body
- `src/agent/runtime/SessionResumeRetrieval.ts:111` — the only currently-shared leaf

## Proposed direction

Promote a host-neutral resume orchestrator (e.g. `resolveAndResumeStream`) into `src/controllers/progressView` that owns the guard + state resolution + retrieve + branch, taking injected ports for "resume tool-use snapshot" and "execute workflow config". The extension stops dispatching through `texra.resumeAgent`/`texra.execute`, the desktop stops inlining the branch, and both become thin adapters; point desktop `runNewStream` at `ProgressWorkflowActionsController.runNew`.

## Notes

Medium blast radius (command-dispatch indirection, `ProgressViewProvider` singleton coupling, desktop persisted-meta hydration). Depends on the shared resume-queue helper landing first (done in the round-2 simplification PR). Surfaced by the macroscopic cross-host coupling analysis during the post-0.38.9 simplification effort.

PR #6836, opened 2026-06-30
refactor: consolidate cross-host runtime coupling (macro round 2)

LionSR:
## Summary

Macroscopic round 2 of the post-0.38.9 refactor: consolidate **cross-host duplication** and prune dead surface, from a deep architecture audit. The VS Code extension, Electron desktop, and CLI were each reimplementing the same runtime/UI flows; this lifts shared logic into host-neutral helpers so the hosts become thinner adapters. Net ~250 lines removed.

## Cross-host coupling consolidated

- **Tool-use resume queue dance** -> new `src/agent/runtime/resumeQueuedToolUse.ts` owns acquire -> RESUMING -> drain+emit -> resume-with-append -> catch re-enqueue(force) -> finally WAITING. Extension and desktop keep only their own log/toast surfaces. The error callback is guarded so a broken host notification cannot turn a handled resume failure into an unhandled rejection.
- **ApprovalRequestHandler** relocated to `src/shared/progressView/backend` + new `createProgressBackendUiConfig` so both hosts build the permission UI callbacks one way; desktop drops its hand-rolled pending-permission Maps.
- **`detachSubagentsOnStop()`** was copy-pasted across five host sites; now one helper.
- **`selectAutoOpenFinalOutput()`** makes desktop match the extension's completed-output selection policy: highest `round`, not array tail.
- **EXTERNAL_INQUIRY** handling moved into the shared progress-view command factory.
- **Node-host platform composition** is split cleanly: `src/platform/defaults/nodeHost.ts` builds the shared Node platform shape and one-shot runtime registrations, while the CLI and desktop composition roots still own the actual `initPlatform(...)` call. This also fixes real drift: the CLI now calls `registerAgentFeatures`, so memory/goal tools are injected there too. Goal prompt paths remain refreshable per active resources bundle.

## Logic consolidations & dead code

`buildProviderKeyStatuses` lifted (desktop IPC + extension controller deduped); ~40 dead barrel re-exports pruned; pass-throughs/stubs removed (`registerLatex/ApiKeyCommands` no-op chain, `_selectedValue`, Google SDK usage type, `SupabaseSession` shim, `FlowParams`/`CycleParams` alias).

## Deferred / out of scope

The full `AgentResumePort` orchestrator unification, the CLI↔webview progress-reducer dedup, and the two parallel run state machines are tracked in #6826 / #6827 / #6828. The `SetupPlatform` composition-root dedup and larger folder moves are deferred to separate PRs.

## Verification

- `npm run lint -- --quiet`
- `npm run typecheck`
- `npm run format`
- `corepack pnpm exec vitest run src/test-kernel/cli/CliInitPlatform.vitest.mts src/test-kernel/desktop/ElectronCompositionRoot.vitest.mts src/test-kernel/desktop/DesktopAgentExecution.vitest.mts`
- Earlier branch verification: full workspace typecheck and `pnpm run test` passed before the final CI/review fix.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -62,7 +62,6 @@ const COMPOSITION_ROOT_FILES = new Set([
   path.join(__dirname, 'packages/extension/src/extension.ts'),
   path.join(__dirname, 'packages/desktop/src/main/platform/index.ts'),
   path.join(__dirname, 'packages/cli/src/runtime/initPlatform.ts'),
-  path.join(__dirname, 'src/platform/defaults/nodeHost.ts'),
 ]);
 
 const extensionPackageDir = path.join(__dirname, 'packages', 'extension');
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -62,6 +62,7 @@ const COMPOSITION_ROOT_FILES = new Set([
   path.join(__dirname, 'packages/extension/src/extension.ts'),
   path.join(__dirname, 'packages/desktop/src/main/platform/index.ts'),
   path.join(__dirname, 'packages/cli/src/runtime/initPlatform.ts'),
+  path.join(__dirname, 'src/platform/defaults/nodeHost.ts'),
 ]);
 
 const extensionPackageDir = path.join(__dirname, 'packages', 'extension');
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -7,7 +7,6 @@
 
 import PQueue from 'p-queue';
 
-import { tryPlatform } from '@platform/platform';
 import { getDefaultStreamLogStore, StreamSnapshotStore } from '@transcript';
 import { registerExecution, writeTerminalStatus } from '@agent/storage';
 import {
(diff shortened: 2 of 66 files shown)


LionSR:
Thanks for the careful review. Dispositions:

1. **`resumeQueuedToolUse.ts` — `onError` outside try/catch.** Fair tightening. It matches the pre-existing host behavior (both old hosts let log/toast throw), so it's not a regression; I'll fold a defensive `.catch(() => {})` around the host callback into a follow-up rather than re-spin CI here.
2. **`selectAutoOpenFinalOutput.ts` — `.at(-1)` → highest round.** Intentional unification (desktop now matches the extension). Outputs are appended in ascending round order in practice, so the selected file is the same; the helper documents "highest round wins" as the canonical final revision. Behavioral note acknowledged.
3. **`resumeQueuedToolUse.ts:84` — WAITING reset before `onError`.** Agreed, and intended — the old extension path could leave a stream in RESUMING behind a blocking dialog; the new ordering fixes that.
4. **`initPlatform.ts` — goal-prompt init now first-init-only.** Intended. `registerAgentFeatures`/Lean must be once-only, and the CLI's `resourcesPath` is the packaged resources dir (process-stable), so goal-prompt path-pinning has no practical effect. If re-init with a changed `resourcesPath` ever becomes real, goal prompts can be re-pointed there.

PR #6859, opened 2026-07-01
fix(codex): separate subscription usage accounting

LionSR:
## Summary
- Routes ChatGPT-subscription usage rows from the `log-usage` edge function into a new Supabase table, `chatgpt_subscription_usage_logs`, with its matching service-role upsert RPC, RLS policy, explicit grants, and per-stream aggregation.
- Keeps the Edge Function routing declarative through destination metadata, so paid and ChatGPT-subscription usage share the same validation, dedupe, row-mapping, and upsert flow.
- Retries failed usage batches with their original `batchId` while keeping later queued entries separate, so a partial paid/subscription write can be retried without double-counting the side that already committed.
- Relaxes the Codex subscription fallback guard so requests do not fail solely because an output budget that is stripped before reaching the Codex backend would exceed the local estimate.

## Verification
- `deno check --config supabase/functions/log-usage/deno.json supabase/functions/log-usage/index.ts`
- `CI=true corepack pnpm vitest run src/test-kernel/telemetry/UsageLogService.vitest.ts src/test-kernel/agent/modelHandlers/CodexSubscriptionFallback.vitest.ts`
- `CI=true corepack pnpm run typecheck`

## Notes
- `supabase db lint --local` was attempted earlier, but this worktree has no running local Postgres instance for the Supabase CLI to connect to.
- The migration includes explicit grants because new public tables should not rely on default Data API exposure.

### DIFF
--- a/src/agent/modelHandlers/openai/modelHandlerCodex.ts
+++ b/src/agent/modelHandlers/openai/modelHandlerCodex.ts
@@ -308,8 +308,13 @@ export class ModelHandlerCodex extends ModelHandlerOpenAIResponse {
     return this.usingSubscription() ? false : super.supportsTokenCounting;
   }
 
-  protected override shouldFailWhenFallbackOutputBudgetIsReduced(): boolean {
-    return this.usingSubscription();
+  protected override shouldFailWhenFallbackOutputBudgetIsReduced(
+    inputEstimate: number,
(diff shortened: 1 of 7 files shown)


LionSR:
Consolidate 

No way this needs this much code

PR #6860, opened 2026-07-01
fix(openai): fall back to polling when the SDK throws on an unrecognized stream event

LionSR:
## Summary
- `openai@6.45.0` can throw `Unhandled response stream event: ...` from its Responses stream accumulator when a long-running stream contains an event type that the SDK has not typed yet. TeXRA now catches only that SDK failure mode after a `response.created` id has been seen.
- Healthy streams continue through `stream.finalResponse()`. The polling fallback is used only after the guarded SDK accumulator error, and stateless backends such as ChatGPT subscription Codex (`store: false`) rethrow instead of attempting a doomed `responses.retrieve()`.
- Fallback `responses.retrieve()` and background polling preserve the original request `include` fields, so recovered responses retain web-search sources and replayable reasoning fields where requested.
- The WebSocket transport path remains independent of the SDK stream accumulator; TeXRA's `ResponseStreamProcessor.process()` already ignores unrecognized event types there.
- After rebasing over the dependency PRs, this PR no longer changes package manifests or the lockfile; it is now only the OpenAI streaming fallback and its regression tests.

## Follow-up
- #6868 tracks replacing the string-prefix guard with a stable OpenAI SDK discriminator if the SDK exposes one later.

## Verification
- `CI=true corepack pnpm vitest run src/test-kernel/agent/modelHandlers/ModelHandlerOpenAIResponse.vitest.ts`
- `CI=true corepack pnpm run typecheck`

---

> [!NOTE]
> **Medium Risk**
> Changes the hot path for OpenAI Responses streaming and background resume, but behavior is gated (SDK message shape, response id, server-side store) and heavily covered by new tests.
> 
> **Overview**
> When the OpenAI Responses **HTTP stream** hits an SDK accumulator error for an event type outside its typed union (e.g. `response.keepalive`), the handler now **recovers by id** instead of failing the turn—**only** after `response.created` has provided a response id and only for the narrow `OpenAIError` message prefix `Unhandled response stream event`.
> 
> Healthy streams are unchanged (`finalResponse()` as before). **Stateless** backends (`store: false`, e.g. Codex subscription) still **rethrow** because `responses.retrieve()` is not viable.
> 
> Fallback **`responses.retrieve()`** and **background polling** now pass through the original request **`include`** params (web-search sources, encrypted reasoning, etc.), and pending-response resume state stores those params so retries resume the same id with the same shape. Pending background resume is attempted **earlier** in `createResponseImpl` so compaction state is not cleared before a resume path runs.
> 
> Regression tests cover happy path, fallback, pre-`response.created` failures, non-OpenAI errors, stateless behavior, `include` preservation, and retry after retrieve/poll failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 160179a599384c8d38ffc9fce8de93f81cc36e16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts
+++ b/src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts
@@ -1,5 +1,5 @@
 // Third-party imports
-import OpenAI from 'openai';
+import OpenAI, { OpenAIError } from 'openai';
 // Exported by openai@6.x via the package's ./lib/* subpath; keep typecheck
 // coverage around SDK upgrades because this is not a top-level public helper.
 import { addOutputText } from 'openai/lib/ResponsesParser';
@@ -117,6 +117,7 @@ import type {
   ResponseCreateParamsBase,
   ResponseCreateParamsNonStreaming,
   ResponseReasoningItem,
+  ResponseRetrieveParamsNonStreaming,
   ResponseFunctionToolCall,
   ResponseInputItem,
   ResponseInputContent,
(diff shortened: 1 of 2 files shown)


LionSR:
strange solution 
Do not merge yet

PR #6861, opened 2026-07-01
fix(anthropic): stop mislabeling post-message_start stream failures

LionSR:
## Summary
- The Anthropic streaming error-wrap guard checked `stream.currentMessage`, which the SDK does not keep populated once `finalMessage()` has resolved/rejected — this caused genuinely informative errors thrown *after* `message_start` (most notably the message_stop guard's "proxy idle timeout during extended thinking" message) to be silently overwritten with a generic, misleading "Stream closed before message_start / connection dropped before the API responded" message, even when diagnostics confirmed `message_start` had in fact been received.
- Gate the wrap on `diagnostics.messageStartReceived` instead — tracked directly from the `message_start` SSE event, unaffected by the SDK's internal stream-state bookkeeping.
- User-visible effect: long extended-thinking streams that get cut off mid-generation (e.g. by a proxy/VPN idle timeout) now surface the accurate "stream ended without message_stop, likely proxy idle timeout" diagnostic instead of the misleading "connection dropped before the API responded" message.

## Test plan
- [x] Added a regression test simulating `message_start` received + a subsequent generic stream failure (the message_stop-guard scenario) — asserts the original, more specific error message survives unwrapped.
- [x] Existing "preserves AnthropicUserAbortError thrown before message_start" test still passes unchanged.
- [x] `npx vitest run src/test-kernel/agent/modelHandlers/ModelHandlerAnthropic.vitest.ts` — 33/33 passing.
- [x] `npx tsc --noEmit -p tsconfig.json` — clean.
- [x] `npx prettier --check` — clean.

---

> [!NOTE]
> **Low Risk**
> Narrow change to streaming error enrichment in the Anthropic handler plus tests; no auth, billing, or API contract changes.
> 
> **Overview**
> Fixes **misleading Anthropic streaming error messages** when a stream fails after `message_start` (e.g. truncated extended-thinking runs).
> 
> The catch-path guard that wraps generic failures as *"Stream closed before message_start"* now keys off **`diagnostics.messageStartReceived`** (set from the real `message_start` SSE event) instead of **`stream.currentMessage`**, which the SDK clears after `finalMessage()` settles. That mismatch used to **overwrite** more specific errors—especially the **`message_stop` guard** text about proxy idle timeout—with the generic pre-start message.
> 
> Adds a **regression test** that emits `message_start` then fails, asserting the original error message is preserved; the stream stub can optionally fire `message_start` via a captured `streamEvent` handler.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 029bd75bead69a824c29ffb08b49504f59b82aed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/modelHandlers/anthropic/modelHandlerAnthropic.ts
+++ b/src/agent/modelHandlers/anthropic/modelHandlerAnthropic.ts
@@ -666,14 +666,20 @@ export class ModelHandlerAnthropic extends ModelHandler<
         );
         const requestId = stream.request_id;
 
-        // Wrap only non-APIError, non-abort stream failures. APIError
-        // subclasses carry status/headers/requestID/type needed for retry
-        // classification; aborts are identified through the metadata tag
-        // attached above (isUserAbort), which wrapping would hide.
+        // Wrap only non-APIError, non-abort stream failures that happened
+        // before message_start. APIError subclasses carry status/headers/
+        // requestID/type needed for retry classification; aborts are
+        // identified through the metadata tag attached above (isUserAbort),
+        // which wrapping would hide. Gate on diagnostics.messageStartReceived
+        // (tracked from the actual message_start SSE event) rather than
+        // stream.currentMessage, which the SDK does not keep populated once
+        // finalMessage() has settled — using it here previously mislabeled
+        // post-message_start failures (e.g. the message_stop guard above) as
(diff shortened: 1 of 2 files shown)


LionSR:
this is critical we need to check if this is true

Issue #7058, opened 2026-07-04
history show --export html should render the real Progress View (Lit) — chat-bubble export doesn't look like the stream tab


LionSR:
## Problem

`texra history show <execId> --export html` (new; PR pending) reuses the Lit-SSR chat exporter (`formatChatAsHtml`). The output is a simplified chat-bubble document — it does not look like the Progress View stream tab at all: no round grouping, no scratchpad/thinking presentation, no tool-call I/O panels, no status/usage chrome. For sharing or auditing a run as a web page, users expect the page to look like what the extension shows.

## Proposal

A static **trace-viewer bundle** built from the *actual* Progress View Lit components, so exported traces are pixel-faithful by construction:

- Components: `packages/extension/src/progressView/frontend/` + `src/shared/progressView` — already proven to run outside VS Code by the **desktop app** (vite aliases + `themeTokens.css` + no-op host bridge; that recipe is the template).
- Feed them from persisted data (`streamLogs` + `streamData` + config) — a `trace.json` — instead of live webview IPC.
- Output: `viewer.html` + one JS bundle; `--export html` (or a new `--export trace`) emits it next to the data.

Depends on: #7057 (headless runs must persist streamLogs first — without it there is nothing faithful to render for CLI runs).

The current chat-bubble export remains useful as `--export md` / lightweight provenance; this issue is about the faithful trace surface.

_Requested for QITPaperBench, where every benchmark run's trace should be reviewable on the web exactly as the extension displays it._

LionSR:
Broken into phases, tracked separately:

- #7057 → headless streamLogs persistence — fix in review: PR #7133
- Phase 0 (prototype/spike) — done: mounted `<stream-conversation>`/`<stream-tabs>` outside VS Code, fed by a hand-dumped `trace.json`, replayed through the real `dispatchMessage`. Confirmed round grouping, thinking banners, tool I/O panels, and usage chrome all render correctly with zero component changes; ~2ms replay cost even on an 89-entry/3MB trace; zero fontawesome CDN calls. Also surfaced two real `file://` output-format gaps for #7136 to solve (crossorigin module scripts, and `fetch()` of a sidecar JSON both fail under `file://`).
- #7135 — trace.json assembler
- #7136 — productionize the trace-viewer package (build format, read-only mode, offline icon fix)
- #7137 — CLI wiring (`--export trace`)

Sequencing: #7057 → #7135 → #7136 → #7137.

LionSR:
[...]

PR #7133, opened 2026-07-05
fix(cli): persist streamLogs for headless CLI runs

### LINKED ISSUE #7057: CLI headless runs don't persist streamLogs — Progress View timeline is lost for headless executions
## Problem

Headless CLI runs (`texra run --print ...`) never persist the `streamLogs/` timeline. `getDefaultStreamLogStore().load()` is only called by interactive hosts (e.g. `runChatTui.tsx:391`); the headless path in `packages/cli/src/runtime/runExecution.ts` skips it.

Consequence: for a headless execution, the `StreamLogEntry[]` record that the Progress View renders (thinking blocks, tool calls with I/O, per-event timestamps, round structure) does not exist on disk. Only `conversation.json`/`config.json`/`meta.json` and the usage sidecars survive, which is not enough to replay the run in the Progress View or export a faithful trace. NDJSON output doesn't cover it either — progress records carry status/usage, not message content (the LOG_DELTA channel is webview IPC only).

## Proposal

Call `getDefaultStreamLogStore().load()` on the headless CLI run path the same way the TUI host does, so every execution — interactive or headless — stores the same UI timeline. Small change; the store and its persistence already exist.

## Why now

Benchmark/automation runs are headless by definition; their traces are the artifact we want to audit and publish. This is the prerequisite for any faithful trace export (see companion issue on `--export html`).

_Found while building QITPaperBench on texra-cli 0.39.3._

LionSR:
## Summary
- Headless CLI runs (`texra run`, `agents run`, `multi-agent run`) never called `StreamLogStore.load()`, so `save()`/`flush()` silently no-op for the whole run — every headless execution's `streamLogs` timeline (thinking blocks, tool calls with I/O, round structure) was appended in memory and then lost entirely on process exit.
- `executeCliRequest` now loads the store before the run and flushes it (alongside `flushPendingRunTraces()` + the snapshot store) on exit, mirroring the interactive TUI path already used in `runChatTui.tsx`.
- The `.load()` call is placed immediately before `invoke()` (not earlier) so it can't delay the synchronous shutdown-hook registration a few lines above — an earlier draft placed it right after runtime-host setup and broke the existing "marks owned executions interrupted during platform shutdown" test.

Fixes #7057. This is the prerequisite for #7058 (faithful trace-viewer export), which needs a real `streamLogs` timeline to render for headless runs.

## Test plan
- [x] Extended `src/test-kernel/cli/RunExecution.vitest.mts` with two tests: load happens before the run and flush happens after (in order), and flush still happens when the run throws.
- [x] Full existing `RunExecution.vitest.mts` suite passes (17/17).
- [x] `src/test-kernel/cli` + `src/test-kernel/transcript` suites pass (128 files, 1574 tests).
- [x] `CI=true npm run typecheck` clean across all 5 workspace projects.

---

> [!NOTE]
> **Low Risk**
> Localized to headless CLI teardown/persistence; mirrors an existing TUI pattern with graceful load failure handling.
> 
> **Overview**
> Headless CLI execution (`texra run`, `agents run`, etc.) was dropping the full **streamLogs** timeline on exit because **`StreamLogStore.save()`/`flush()` no-op until `load()` has run once** — the interactive TUI already loads the store; headless did not.
> 
> **`executeCliRequest`** now **best-effort `load()`s** the default stream log store **immediately before** `invoke()` (after shutdown-hook registration), and on teardown **`flushPendingRunTraces()`** plus **`streamLogStore.flush()`** run **in parallel** with the existing snapshot flush. Failed loads keep in-memory-only behavior.
> 
> Tests assert **load → run → flush** ordering and that **flush still runs when the run throws**; **`vi.restoreAllMocks()`** was added in **`afterEach`** so store spies do not leak across cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c77b5c7a15095fe17e84ddc285d925a6c70a9e96. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/runtime/runExecution.ts
+++ b/packages/cli/src/runtime/runExecution.ts
@@ -1,6 +1,10 @@
 import { SHUTDOWN_PHASE } from '@platform/interfaces/lifecycle';
 import { tryPlatform } from '@platform/platform';
-import { StreamSnapshotStore } from '@transcript';
+import {
+  flushPendingRunTraces,
+  getDefaultStreamLogStore,
+  StreamSnapshotStore,
+} from '@transcript';
 import { writeTerminalStatus } from '@agent/storage';
 import type { AgentConfigPayload } from '@agent/core/definition/AgentConfig';
 import { AgentCategory } from '@agent/core/definition/AgentDataclass';
@@ -210,6 +214,20 @@ export async function executeCliRequest(
       ],
     });
 
+  // Headless runs append to StreamLogStore the same as the interactive TUI
+  // (via createRunTrace's transcript recorder), but StreamLogStore.save()/
+  // flush() silently no-op until .load() has run once — without this, every
+  // headless run's streamLogs are appended in memory and then lost entirely
+  // when the process exits. Mirrors runChatTui.tsx's best-effort load. Done
+  // last (right before the run starts) so it can't delay the synchronous
+  // shutdown-hook registration above.
+  const streamLogStore = getDefaultStreamLogStore();
+  try {
+    await streamLogStore.load();
+  } catch {
+    // Persistence stays disabled; the run still executes in-memory as before.
+  }
+
   let result: ExecuteAgentResult;
   try {
     result = await (options.wrap ? options.wrap(invoke) : invoke());
@@ -228,7 +246,8 @@ export async function executeCliRequest(
     detachResultToast();
     uninstallApprovalHandlers();
     try {
-      await snapshotStore.flush();
+      flushPendingRunTraces();
+      await Promise.all([streamLogStore.flush(), snapshotStore.flush()]);
     } finally {
       await runtimeHost.close();
     }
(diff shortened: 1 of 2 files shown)


LionSR:
Rebased onto main (resolved a conflict with the `@texra/core` package removal — `runExecution.ts`'s imports now come straight from `@agent/*`/`@platform/*`, matching that refactor; this PR's addition just carried forward onto the new import shape).

Also addressed the reviews' cosmetic suggestion: destructured `getDefaultStreamLogStore()` into a single `streamLogStore` local instead of calling it twice.

Re-verified after rebase: full `RunExecution.vitest.mts` suite + `src/test-kernel/cli`/`src/test-kernel/transcript` (128 files, 1574 tests) pass, and `CI=true npm run typecheck` is clean across all 4 workspace projects.

PR #7179, opened 2026-07-05
Derive hardcoded model defaults from the live registry

### LINKED ISSUE #7081: Registry-derive the remaining hardcoded model defaults with hard-failure modes
## Context

From a model-handler friction investigation (evidence pinned to `37d1211`). #6953 ruled the C3f stringly-routing family "as-touched" — this issue collects only the subset with **hard failure modes on the next model churn**, each a small data-flag change:

| Gate | Failure mode | Fix |
|---|---|---|
| `setupModelDefaults.ts:6-17` — 10 hardcoded per-provider setup models | When e.g. `opus48T` retires, the setup assistant probes a retired model → key verification fails for a **valid** key | Derive from `MODEL_CONFIGS` (first non-retired per provider) or an llm-zoo `flagship` flag |
| `modelHandlerKimi.ts:11-16,52` — exact-name gates for fixed temperature / thinking-disable | A `kimi-k3` misses the pin → Moonshot API 400 or degraded sampling | llm-zoo `fixedTemperature?: number`, `thinkingDefaultOn?: boolean` |
| `modelHandlerOpenAI.ts:97,321-328` — `deepseek-chat` → `max_tokens=8192` | Renamed successor misses the clamp → 400 or silent truncation; wrong if DeepSeek raises the cap | per-route `maxOutputTokens` in llm-zoo (official vs OpenRouter) |
| `modelOptionsBasic.ts:14-27` — `DEFAULT_MODELS` + hand-bumped `MODEL_LIST_VERSION=21` | New models invisible to existing users until someone remembers the bump; retired defaults dangle | `featured` flag + hash-based reconciliation |
| `providerCapabilities.ts:44-56` — `CODEX_SUBSCRIPTION_MODEL_FULLNAMES` (self-documented "WILL go stale") | A non-"codex"-named subscription model misses the `/codex/i` catch-all → picker silently offers paid API path instead of subscription | llm-zoo `codexSubscriptionEligible` flag, or probe the Codex models endpoint at refresh |

## Proposed

One small PR per row (or one batch PR); each is data plumbing plus a test that a registry-only model addition behaves correctly. Rows needing llm-zoo flags coordinate with the same release as the anthropicThinking issue.

LionSR:
Closes #7081

Three hard-failure-mode hardcoded model defaults, all fixed by deriving from the live `llm-zoo` `MODEL_CONFIGS` registry instead of a pinned literal:

1. **`src/model/setupModelDefaults.ts`** — 10 hardcoded per-provider setup models. The curated preference table stays (llm-zoo has no "flagship" capability flag to derive it from), but each pick is now validated against `MODEL_CONFIGS` at module load and swapped for a still-usable model of the same provider when the preferred pick has been retired or is OpenRouter-only. Concretely: `xai: 'grok4'` is *already retired* in the live registry today, so before this fix the setup assistant would silently probe a dead model whenever xai was the only usable credential.

2. **`src/agent/modelHandlers/openai/modelHandlerKimi.ts`** — exact-fullName gates (`isKimiK25`, `isKimiK27Code`, and an inline `fullName === 'kimi-k2.5'` check) for fixed-temperature and thinking-disable behavior. Replaced with:
   - A thinking-disable set computed from the registry itself: any Moonshot `fullName` shared by both a `supportsReasoning: true` and `supportsReasoning: false` registry entry needs the non-reasoning entry to explicitly disable thinking (Moonshot's API can't otherwise tell the two apart and defaults to thinking-on). This also fixes `kimi26`/`kimi26T`, which share the same ambiguous-fullName pattern as `kimi25`/`kimi25T` but weren't covered by the old hardcoded check.
   - A small explicit fixed-temperature table (genuine per-model API knowledge with no llm-zoo capability-flag equivalent), so adding a new fixed-temperature Kimi model is one new table entry instead of a new string-equality branch.

3. **`src/agent/modelHandlers/openai/modelHandlerOpenAI.ts`** — the DeepSeek official-API `max_tokens` override matched the exact fullName `'deepseek-chat'`, which no longer exists as a live (non-retired) registry entry. Now gates on `provider === DEEPSEEK && !supportsReasoning && maxOutputTokens <= DEEPSEEK_OFFICIAL_API_MAX_TOKENS`, so any current or future small-output non-reasoning DeepSeek model gets the correct cap automatically.

`llm-zoo` is consumed as an external published npm package (not a workspace member of this repo), so it can't be patched directly — capability flags that would ideally live upstream (`fixedTemperature`, `thinkingDefaultOn`) are instead expressed as small local lookup tables/derivations keyed off registry data already available on `ModelConfig` (`provider`, `capabilities.supportsReasoning`, `maxOutputTokens`, `retired`, `openRouterOnly`), which is the closest available mechanism per `src/model/providerCapabilities.ts`'s existing local-capability-resolver pattern.

## Testing
- New `src/test-kernel/model/SetupModelDefaults.vitest.ts`: no provider resolves to a retired/OpenRouter-only model, covers all direct-key providers, and exercises the real `xai/grok4` retirement as a concrete regression case.
- Extended `OpenAICompatibleProviderParams.vitest.ts` with Kimi K2.5/K2.6 reasoning and non-reasoning cases.
- Extended `ModelHandlerDeepSeek.vitest.ts` with registry-derived max_tokens cases (low-budget non-reasoning override applies regardless of fullName; large-output and reasoning entries are unaffected).
- `npm run typecheck` (root, test-kernel, extension, CLI) passes.
- Targeted `eslint` over all changed files passes.
- All touched vitest suites (`src/test-kernel/model`, `src/test-kernel/agent/modelHandlers`, `src/test-kernel/controllers/SetupLaunch.vitest.ts`) pass, including pre-existing tests (no behavior change for currently-live models beyond the intended kimi-k2.6 fix).

---

> [!NOTE]
> **Medium Risk**
> Touches credential setup model selection and live LLM request parameters (thinking, temperature, max_tokens); behavior is mostly equivalence for current models plus targeted fixes (xai setup, Kimi K2.6), but wrong fallback or gating logic could break onboarding or API calls.
> 
> **Overview**
> Replaces **pinned model literals** with **registry-driven rules** so onboarding and API calls stay correct when `llm-zoo` entries retire or rename.
> 
> **Setup onboarding** (`setupModelDefaults.ts`): curated per-provider probe models are still preferences, but `SETUP_MODEL_BY_PROVIDER` is resolved at module load via `resolveSetupModel`, which rejects retired/OpenRouter-only picks and substitutes another usable model for that provider (e.g. **xai** no longer stuck on retired `grok4`). OpenAI probes remain **Codex-eligible**.
> 
> **Kimi / Moonshot** (`modelHandlerKimi.ts`): thinking-off for non-reasoning variants is derived from Moonshot `fullName`s that map to **both** reasoning and non-reasoning registry entries (covers **K2.6** as well as K2.5). Fixed temperatures and compaction-summary behavior move into a small **`FIXED_TEMPERATURE_BY_FULLNAME`** table instead of scattered `fullName ===` branches.
> 
> **DeepSeek official API** (`modelHandlerOpenAI.ts`): the legacy `max_tokens` bypass applies to **DeepSeek, non-reasoning, `maxOutputTokens` ≤ 8192** (any `fullName`), sending **`config.maxOutputTokens`** unreduced in tool-use mode—not only the old `deepseek-chat` name.
> 
> New/extended vitest coverage for setup resolution, Kimi K2.5/K2.6 params, and DeepSeek `max_tokens` behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac053eaef61e9cb6ec72d8622946124eda2a2b15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/modelHandlers/openai/modelHandlerKimi.ts
+++ b/src/agent/modelHandlers/openai/modelHandlerKimi.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { MODEL_CONFIGS, ModelProvider } from 'llm-zoo';
 
 // Local imports - agent
 import type { NormalizedUsage } from '@agent/types/NormalizedUsage';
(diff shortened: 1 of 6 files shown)


LionSR:
Addressed all three review findings (return-type honesty in `resolveSetupModel`, inaccurate "newest" fallback comment, test/production Codex-eligibility parity) in ac053eaef. Resolving the corresponding threads.

PR #7237, opened 2026-07-05
fix(trace-viewer): validate loaded trace data against a schema before replay

### LINKED ISSUE #7187: Validate trace-viewer's loaded trace data against a schema before replay
## Context

Follow-up from #7142 (feat(trace-viewer): productionize the trace-viewer package).

`packages/trace-viewer/src/main.ts`'s `loadTrace()` casts fetched (`trace.json`, selectable via the `?trace=` query string) or inline (`window.__TEXRA_TRACE__`) data straight to the trace type with no runtime validation, before feeding it into `replayTrace()` → `dispatchMessage()`. `messageDispatcher.ts` deliberately skips Zod validation on its hot path (comment: "TypeScript enforces the shape at compile time"), so a malformed or stale trace file (e.g. exported by an older/newer TeXRA version after a schema change) has no check catching it anywhere in this pipeline — it fails deep inside the dispatcher with no clear error pointing at "the trace file doesn't match what this viewer expects." Reviewers flagged this repeatedly across the PR's review rounds, with one explicitly calling it "worth a follow-up before this format needs to stay stable across releases."

## Details

- Add a schema (reusing/deriving from the relevant shared schemas, e.g. `StreamSnapshotSchema` and the trace document's fields) and `safeParse` the loaded trace in `loadTrace()` before returning it.
- On validation failure, surface a clear error (e.g. via the existing `console.error('[trace-viewer] ...')` path) identifying that the trace file doesn't match the expected shape, rather than letting a cryptic error surface from deep inside `dispatchMessage`/`replayTrace`.
- This follows CLAUDE.md's "validate at system boundaries" guidance — `loadTrace()` is the external boundary (fetched or externally-authored inline data).

## References

- PR: #7142
- Parent tracking issue: #7185

LionSR:
## Summary

Adds runtime validation at the trace-viewer load boundary for fetched `trace.json` and inline `window.__TEXRA_TRACE__` data before replay.

The CI fix keeps that schema browser-safe: `packages/trace-viewer/src/traceDataSchema.ts` reuses shared browser-safe schemas and mirrors the agent config / execution meta shapes with type-only compatibility assertions, instead of importing `@agent/*` runtime modules into the Vite browser bundle.

## Issue acceptance gates

#7187: loaded trace data is schema-validated before `replayTrace()`; malformed or incompatible traces throw a clear trace-shape error through the existing trace-viewer error path.

## Single source of truth

`packages/trace-viewer/src/traceDataSchema.ts` is the trace-viewer validation boundary for externally loaded trace payloads. `TraceDocument` remains the source type for the replay document shape, enforced through compile-time compatibility assertions.

No pass-through layer was added. The trace-viewer keeps a small browser-safe mirror for `AgentConfig` and `ExecutionMeta` because importing the source `@agent/*` schemas pulls Node-only dependencies into the browser build; type-only assertions keep the mirror honest.

## Host impact

Extension: trace-viewer assets build again, and malformed trace JSON is rejected before replay.

Desktop: no runtime behavior change.

CLI: exported trace-viewer HTML validates inline trace data before replay; CLI command output contracts are unchanged.

## Scheduled refactoring

None. No legacy-retirement ledger row is needed because this PR adds no shim, fallback, dual-write, alias, or migration path.

## Validation

- `corepack pnpm exec prettier --write packages/trace-viewer/src/traceDataSchema.ts src/agent/storage/ExecutionKVStore.ts src/agent/storage/index.ts src/test-kernel/transcript/traceViewerSchema.vitest.ts packages/trace-viewer/src/main.ts`
- `corepack pnpm --filter @texra/trace-viewer run typecheck`
- `corepack pnpm --filter @texra/trace-viewer run build`
- `corepack pnpm exec vitest run src/test-kernel/transcript/traceViewerSchema.vitest.ts`
- `npm run typecheck`
- `npm run compile:fast`
- `npm run lint` (0 errors; existing unrelated import-order warning in `packages/extension/src/progressView/frontend/formatters/logFormatters/codexToolTemplates.ts`)
- `npm test`
- `git diff --check`

Closes #7187

---

> [!NOTE]
> **Low Risk**
> Changes are confined to trace-viewer load path and reject bad input earlier; no auth, persistence, or live execution behavior changes.
> 
> **Overview**
> Adds a **browser-safe Zod validation boundary** at trace load time so inline `window.__TEXRA_TRACE__` and fetched `trace.json` are checked before `replayTrace()`.
> 
> `loadTrace()` in `main.ts` now passes raw payloads through `parseTraceData()` instead of trusting JSON shape; failures surface as a clear “incompatible/malformed trace” error rather than breaking deep in message dispatch. The new `traceDataSchema.ts` mirrors `TraceDocument` using shared `@shared/schemas` plus local mirrors for agent config and execution meta (avoiding `@agent/*` runtime imports in the Vite bundle), with compile-time assertions that exported traces still align with `assembleTrace` output.
> 
> Vitest coverage confirms real `assembleTrace` documents pass validation and common malformed payloads are rejected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfbcf6033b41f48f39c38e32e982ba619028856b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/trace-viewer/src/main.ts
+++ b/packages/trace-viewer/src/main.ts
@@ -19,6 +19,7 @@ import {
 } from '@progressView/frontend/eventHandlers';
 
 import { replayTrace } from './replayTrace';
+import { parseTraceData } from './traceDataSchema';
 import type { TraceDocument } from '@transcript';
 
 const root = document.querySelector<HTMLElement>('#app');
@@ -50,19 +51,21 @@ conversationView.addEventListener(
  * sidecar JSON — `fetch()` of a local file fails entirely under `file://`,
  * regardless of bundling), or a fetched `trace.json` next to the page (the
  * default when served over http/https, e.g. a static site hosting many
- * traces that shouldn't duplicate the bundle per page). Both paths feed the
- * exact same `replayTrace`.
+ * traces that shouldn't duplicate the bundle per page). Both paths are
+ * externally-authored (exported by whatever TeXRA version produced them), so
+ * both get validated through the same `parseTraceData` boundary before
+ * feeding the exact same `replayTrace` — a stale/malformed trace file throws
+ * a clear error here instead of failing deep inside `dispatchMessage`.
  */
 async function loadTrace(): Promise<TraceDocument> {
-  const inline = (window as { __TEXRA_TRACE__?: TraceDocument })
-    .__TEXRA_TRACE__;
-  if (inline) return inline;
+  const inline = (window as { __TEXRA_TRACE__?: unknown }).__TEXRA_TRACE__;
+  if (inline) return parseTraceData(inline);
 
   const params = new URLSearchParams(window.location.search);
   const file = params.get('trace') ?? 'trace.json';
   const res = await fetch(file);
   if (!res.ok) throw new Error(`Failed to fetch ${file}: ${res.status}`);
-  return res.json();
+  return parseTraceData(await res.json());
 }
 
 loadTrace()
--- a/packages/trace-viewer/src/traceDataSchema.ts
+++ b/packages/trace-viewer/src/traceDataSchema.ts
@@ -1,0 +1,67 @@
+/**
+ * Runtime validation for trace data loaded by `main.ts`'s `loadTrace()` —
+ * either fetched (`trace.json`) or inlined (`window.__TEXRA_TRACE__`).
+ *
+ * This is the trace-viewer's external boundary: the data was authored by
+ * whatever TeXRA version produced the export (CLI, extension, or desktop),
+ * and `main.ts` has no static guarantee it matches this build's expected
+ * shape. Mirrors `TraceDocument` (`@transcript/traceAssembler`) field-for-
+ * field, reusing the same shared schemas that already define its parts
+ * (`StreamSnapshotSchema`, `StreamLogEntrySchema`, `AgentConfigSchema`,
+ * `ExecutionMetaSchema`, …) rather than re-deriving them, so this schema
+ * tracks the real document shape instead of drifting from it.
+ */
+import { z } from 'zod';
+
+import { ExecutionMetaSchema } from '@agent/storage';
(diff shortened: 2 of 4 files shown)


LionSR:
Follow-up on the review threads flagged here (compile-time assertion no-op, and `schemaVersion` `.catch()` leniency): this PR was merged before this task ran, and both defects were independently fixed afterward in separate PRs against `main`:

- #7260 — replaces the `? true : never` assertions with a real `_AssertTrue<_IsExact<...>>` bidirectional-equality check (the standard TS type-testing idiom). Verified empirically: a minimal repro of the old pattern silently compiles under a deliberate field divergence (`tsc` exit 0), while the new pattern fails with `TS2344: Type 'false' does not satisfy the constraint 'true'.` under the same divergence.
- #7255 — replaces the reused `StreamSnapshotSchema` (whose `schemaVersion` is `z.literal(...).catch(...)`) with a trace-viewer-local `TraceStreamSnapshotSchema` using `.literal(...).prefault(...)` instead, so an incompatible `schemaVersion` is rejected rather than silently coerced. Ships with a regression test (`rejects a trace snapshot stamped with an incompatible schema version`).

Re-ran `pnpm --filter @texra/trace-viewer run typecheck` and `vitest run src/test-kernel/transcript/traceViewerSchema.vitest.ts` against current `main` (8cfff2e67): typecheck clean, 6/6 tests pass, including the new schemaVersion-rejection case.

No further code changes needed on this PR's branch (already merged and deleted); flagging here so the threads can be closed with a link to the fixing commits.

LionSR:
Resolving the remaining threads: both confirmed defects (the no-op `? true : never` type assertion, and the overly-permissive reused `StreamSnapshotSchema.catch()`) were independently fixed on main after this PR merged, via #7260 and #7255 respectively. Verified both are present on current main.

PR #7271, opened 2026-07-06
refactor: simplify data-structure designs flagged by complexity audit

LionSR:
## Summary

Follows up on a codebase-wide audit of data-structure complexity (nested maps, kitchen-sink optional schemas, duplicated transforms, scattered fallback logic). Addresses all 14 confirmed findings, grouped by severity:

**High**
- `StreamSnapshotStore`: extract shared `applyRoundKeyedPatch` for the round-keyed accumulators (missing outputs / compile failures); dedupe the `executionId` legacy-fallback derivation into one `executionIdFromMeta` helper (was repeated 3x in the same file)
- `bash`/`codex`/`claudeAgent`/`plan` tools now route through the existing `requireRunStream`/`requireStreamId` helpers instead of each hand-rolling its own missing-context guard with divergent behavior (throw vs. warn-and-degrade vs. silently pass `undefined`)
- `ProviderErrorObjectSchema`: replace 3 independent exhaustion booleans (`isCredentialExhausted`, `isUpstreamCreditDepleted`, `isChatGptSubscriptionLimited`) with one `exhaustionReason` discriminant, with a legacy on-disk migration for old trace/transcript data
- `StreamStatusService`/`WebviewUpdater`: single `getAllStreamStates()` map instead of two independently-rebuilt phase/substate maps that could silently diverge in stream coverage
- `IModelHandler.createBatchedToolUseFollowUpMessages`: one array of `{call, result, attachments}` records instead of three positionally-zipped parallel arrays — removes the runtime "mismatched length" guards this fragility required

**Medium/Low**
- Unified the stored-conversation formatter shared by `ExecutionsTool` and the CLI `history` command into `src/agent/storage/conversationFormat.ts`
- Dropped the redundant `streamId` echoed into `ModelOutputBackup` values (the map key, already in scope, is now used directly)
- `ContextManagementDataSchema`, `ActiveChildInfoSchema`, `StreamTabInfoSchema`: converted to Zod discriminated unions instead of always-optional "kitchen sink" bags
- `getEffectiveDiffBase`: one helper for the `diffBase ?? original` fallback, reused by `output.ts`, latexdiff, and the progress-view file list (previously duplicated 3x with the same rationale comment)
- Narrowed `ResultEvent.error` per `AgentErrorKind` instead of `Partial<...>` of the full provider-error shape
- Use the existing `computeUtilizationPercent` helper in the transcript recorder and CLI status bar instead of re-deriving the formula inline

One finding (agentCategory fallback sites) was investigated and found to have no further applicable call sites beyond the two already using the canonical helper — no change was needed there.

## Issue acceptance gates

Internal code-review follow-up (not tied to a filed issue): implement the confirmed findings from a data-structure complexity audit of this repo. All 14 confirmed findings are addressed; each is verified by grep/type-check, not just described.

## Single source of truth

- `executionIdFromMeta()` (`StreamSnapshotStore.ts`) is now the one place that resolves a stream's execution id from meta.
- `requireRunStream()`/`requireStreamId()` (`contextHelpers.ts`) are the one place tools assert an active run context.
- `getAllStreamStates()` (`StreamStatusService.ts`) is the one accessor for combined phase+substate state.
- `getEffectiveDiffBase()` (`output.ts`) is the one place that resolves a file's diff base.
- `computeUtilizationPercent()` (`contextUtilization.ts`) is the one formula for context-window utilization.

Removed: 3x `executionId` fallback, 3x `diffBase ?? original` fallback, 2 independently-rebuilt phase/substate maps, 3 positional-array parameters, 2 parsed-conversation formatters, duplicated tool-context guard logic across 4 tool files, duplicated builder functions (`buildSubagentItem`/`buildProcessItem` → one `buildChildControlItem`). No new abstraction layers were introduced beyond the discriminated-union schema shapes themselves and the small helpers named above, each of which owns exactly one invariant (e.g. `requireStreamId` owns "no runtime host required," distinct from `requireRunStream`, which does require one).

## Host impact

Extension: progress-view frontend components (`StreamTabs`, `StreamConversation`, `ProcessStreamContent`, `WorkflowStreamContent`, `FileList`, `BackgroundTasksPanel`, `RetryRequestPanel`) updated to narrow on the new `kind` discriminants and `exhaustionReason`.

Desktop: `desktopAgentExecution.ts` and `desktopCommandPalette.ts` updated for the unified stream-state accessor and `kind` discriminant.

CLI: `StatusBar.tsx`, `childControls.ts`/`childExecutions.ts`, `cliState` slices, `subscribeApprovals.ts`, `approvalPolicy.ts`, and the CLI conversation-history formatter updated for the same schema/helper changes.

## Scheduled refactoring

None — all confirmed findings from the audit are addressed in this PR.

## Validation

- `npm run typecheck` (root, test-kernel, extension, cli, trace-viewer projects) — clean, 0 errors.
- `npm run lint` — 0 errors (2 pre-existing, unrelated warnings left untouched; auto-fixed the import-order warnings introduced by this change).
- `npx vitest run` across `src/test-kernel/{agent,progressView,common,cli}` and `src/tools` — 2845 passed, 6 skipped, 0 failed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WXiz3MetAcTLcMS2RPG9VB)_

---

> [!NOTE]
> **Medium Risk**
> Wide cross-host refactor touching retry/credential classification and persisted JSON migrations; behavior is heavily test-covered but regressions could affect approvals, stream tabs, and auto-retry.
> 
> **Overview**
> Follow-up to a data-structure complexity audit: **tightens schemas and removes duplicated logic** across agent runtime, progress UI, and CLI.
> 
> **Provider / retry errors** replace three independent exhaustion booleans with a single **`exhaustionReason`** (`relay-limit`, `upstream-credit`, `chatgpt-subscription`), plus **`normalizeLegacyProviderErrorFields`** on read for old persisted logs and cached errors. Call sites use **`isCredentialExhausted`**, **`isUpstreamCreditDepletedError`**, and related helpers instead of raw flags.
> 
> **Stream and child metadata** use Zod discriminated unions: **`StreamTabInfo`** (`kind: 'agent' | 'process'`), **`ActiveChildInfo`** (`subagent` vs `process`, with legacy `kind` backfill from `childStreamId`), and **`ContextManagementData`** (action-specific fields). Progress hosts switch rendering on **`kind`** rather than guessing from optional `model`/`command`.
> 
> **Stream status** exposes **`getAllStreamStates()`** (phase + substate together); **`WebviewUpdater`** / desktop / extension take one map so phase and substate coverage cannot diverge.
> 
> **Model handlers** change batched parallel tool follow-ups to one **`entries: { call, result, attachments }[]`** API (dropping positional zip + **`checkBatchedToolCalls`**).
> 
> **Shared helpers**: stored-conversation text formatting moves to **`@agent/storage/conversationFormat`** (CLI history delegates); **`getEffectiveDiffBase`** centralizes diff-base resolution; CLI status bar uses **`computeUtilizationPercent`**. **`ResultEvent.error`** is narrowed by **`AgentErrorKind`** so abort/disk-full payloads do not claim full provider-error fields.
> 
> CLI child pickers merge **`buildSubagentItem` / `buildProcessItem`** into **`buildChildControlItem`** keyed on **`child.kind`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d239f4386641cc097f23ff335f8b4ce0612853ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/scripts/tui-harness.tsx
+++ b/packages/cli/scripts/tui-harness.tsx
@@ -784,7 +784,7 @@ function makeRetryApprovalPayload(): RetryPermission {
     operation: 'Tool-use call',
     errorMessage: 'HTTP 429 Too Many Requests',
     errorDetails: {
-      isCredentialExhausted: true,
+      exhaustionReason: 'relay-limit',
       isRelayError: true,
       statusCode: 429,
     },
@@ -996,6 +996,7 @@ if (SHOW_CHILDREN) {
   const startedAt = Date.now() - 74_000;
   const nestedStartedAt = startedAt + 24_000;
   const nestedStrategyChild = {
+    kind: 'subagent' as const,
     executionId: 'harness-nested-local-checker',
     agentName: 'localChecker',
     childStreamId: 'harness-nested-local-checker-stream',
@@ -1003,6 +1004,7 @@ if (SHOW_CHILDREN) {
     startedAt: nestedStartedAt,
   };
   const nestedStrategyProcess = {
+    kind: 'process' as const,
     executionId: 'harness-nested-proof-audit',
     agentName: 'proof audit',
     toolName: 'bash',
@@ -1011,6 +1013,7 @@ if (SHOW_CHILDREN) {
   };
   const childStreams = [
     {
+      kind: 'subagent' as const,
       executionId: 'harness-child-strategy',
       agentName: 'strategy',
       childStreamId: 'harness-child-strategy-stream',
@@ -1018,6 +1021,7 @@ if (SHOW_CHILDREN) {
       startedAt,
     },
     {
+      kind: 'subagent' as const,
       executionId: 'harness-child-lean',
       agentName: 'leanSolver',
       childStreamId: 'harness-child-lean-stream',
@@ -1025,6 +1029,7 @@ if (SHOW_CHILDREN) {
       elapsed: '2m 3s',
     },
     {
+      kind: 'subagent' as const,
       executionId: 'harness-child-review',
       agentName: 'reviewer',
       childStreamId: 'harness-child-review-stream',
@@ -1052,6 +1057,7 @@ if (SHOW_CHILDREN) {
     childStreams,
     activeProcesses: [
       {
+        kind: 'process' as const,
         executionId: 'harness-process-latexmk',
         agentName: 'latex build',
         toolName: 'bash',
--- a/packages/cli/src/chat/tui/panes/StatusBar.tsx
+++ b/packages/cli/src/chat/tui/panes/StatusBar.tsx
@@ -4,6 +4,7 @@ import { MODEL_CONFIGS } from 'llm-zoo';
 import { useEffect, useState } from 'react';
 import stringWidth from 'string-width';
 
+import { computeUtilizationPercent } from '@agent/modelHandlers/support/contextUtilization';
 import { isCodexSubscriptionActive } from '@auth/codex';
 import { shortCliApiMode } from '@cli/runtime/apiAccessMode';
 import {
@@ -157,7 +158,10 @@ function formatUsage(
   }
 
   const ratio = used / contextWindow;
-  const percent = Math.max(1, Math.round(ratio * 100));
+  const percent = Math.max(
+    1,
+    Math.round(computeUtilizationPercent(used, contextWindow)),
+  );
   let color: StatusBarColor;
   if (ratio >= 0.9) color = 'red';
   else if (ratio >= 0.6) color = 'yellow';
--- a/packages/cli/src/chat/tui/state/childControls.ts
+++ b/packages/cli/src/chat/tui/state/childControls.ts
@@ -7,6 +7,7 @@ import {
   STREAM_STATUS,
   type ActiveChildInfo,
   type StreamTabId,
+  type SubagentChildInfo,
 } from '@shared/schemas';
 import { formatStreamStatusLabel } from '@shared/streams/streamStatusDisplay';
 import { formatCompactDuration } from '@utils/core';
@@ -132,12 +133,10 @@ export function childElapsed(
 }
 
 function streamDescription(
-  child: Pick<ActiveChildInfo, 'childStreamId'>,
+  child: SubagentChildInfo,
   streamsById: ReadonlyMap<StreamTabId, Pick<StreamSlice, 'description'>>,
 ): string | undefined {
-  return child.childStreamId
-    ? streamsById.get(child.childStreamId)?.description
-    : undefined;
+  return streamsById.get(child.childStreamId)?.description;
 }
 
 const TASK_DETAIL_TRANSCRIPT_COLUMNS = 120;
@@ -153,10 +152,9 @@ function streamEntryTailLines(entry: ConversationEntry): readonly string[] {
 }
 
 function streamTranscriptLines(
-  child: Pick<ActiveChildInfo, 'childStreamId'>,
+  child: SubagentChildInfo,
   streamsById: ReadonlyMap<StreamTabId, Pick<StreamSlice, 'entries'>>,
 ): readonly string[] {
-  if (!child.childStreamId) return [];
   const stream = streamsById.get(child.childStreamId);
   if (!stream) return [];
   return stream.entries.flatMap((entry) =>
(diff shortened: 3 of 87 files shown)


LionSR:
Addressed the review feedback in 3fcf300:

- Fixed the real leftover: `ProgressWorkflowFileActionsController.ts`'s `backupModelOutput` was still writing `{ content, streamId }` even though `ModelOutputBackup` dropped `streamId` — now just `{ content }`, matching the rest of the change.
- Applied the `contextManagement.ts:31` suggestion: `TokensFreedDataSchema.action` now derives from `ContextManagementAction.exclude(['max_tokens_reduced'])` instead of duplicating the four literals.

Left the `contextManagementFormatters.ts:94` suggestion (removing `action !== 'max_tokens_reduced'` as "redundant" alongside `TOKENS_FREED_ACTIONS.has(action)`) unapplied — verified by removing it and running `tsc`: it's load-bearing for the discriminated-union narrowing (`Set.has()` is a runtime check that doesn't narrow `data`'s type), so removing it breaks the build with a real type error on `data.tokensAfter`.

Typecheck, lint, and the progressView/compaction test suites are all clean after these changes.

---

LionSR:
Removed the dead code flagged in the latest review (a428fc5): `checkBatchedToolCalls` in `openAIMessageUtils.ts` lost both production callers when H5's refactor removed the runtime alignment guards, leaving it exercised only by its own test — deleted the function and its dedicated `describe` block.

---

LionSR:
Good catch, fixed in 7624ecc: `replayTrace` was hard-coding `kind: 'agent'` regardless of the traced agent, so a replayed process-agent trace (`bash`) would render as a tool-use agent instead of the process/terminal layout `StreamConversation` now picks strictly by `kind`. It now derives `kind` the same way `buildStreamTabInfo` does (`isProcessAgent(trace.config.agent)`), carrying `command` instead of `model` for process traces. Typecheck (including the `trace-viewer` package) is clean.

(The earlier "validate" CI failure notification was just an in-flight run getting cancelled by this follow-up push, not a real code issue.)

---

LionSR:
Reviewed Cursor Bugbot's 3 findings:

**Fixed (b7df88c)** — "Strict context events drop logs" (`contextManagement.ts`): real gap. Unlike `ActiveChildInfo`, `ContextManagementData` genuinely persists into the stream transcript (`TexraTranscriptRecorder` → `MESSAGE_TYPES.CONTEXT_MANAGEMENT`), so an older log entry written under the previous `.optional()` shape would fail to parse under the new discriminated union and silently disappear from the progress view on replay/history view. Added a `z.preprocess` step defaulting `tokensAfter`/`utilizationAfter` from the before-values for legacy entries, plus a regression test in `SchemaMigrations.vitest.ts`.

**Not applicable — verified, no change:**
- "Legacy child rows lack migration" (`ActiveChildInfo`): `activeSubagents`/`activeProcesses` are explicitly documented as ephemeral, backend-owned runtime state (`ProgressViewState.ts`'s `StreamExecutionState`, always reset to `[]` on construction) — never persisted to disk or read back from an old trace. Every construction site is either a fresh reset or `executionRegistry.ts`'s live `collectChildSummary()`, which already sets `kind` correctly. There's no code path that could hand this schema an old `childStreamId`-only payload.
- "Cached errors skip exhaustion migration" (`normalizeProviderError`): traced the full chain — `providerErrorMetadata` is a `Symbol.for()`-keyed attachment directly on the live `Error` object (can't survive JSON serialization), and the only producers that feed it (`formatProviderHttpError`, `toProviderErrorFromRetry(shared.lastError)`) build `RetryErrorInfo`/`ProviderError` fresh via `normalizeProviderError` within the same process — never by deserializing old persisted JSON without going through the schema's migration first. So a "cached" object here can't be legacy-shaped.

Typecheck and the progressView/compaction test suites are clean.

---

LionSR:
Addressed the four open review threads (commit d239f4386):

1. **contextManagementFormatters.ts**: Verified — the `action !== 'max_tokens_reduced'` check is *not* actually redundant. Removing it (confirmed by temporarily deleting it and running `tsc`) breaks the build: it's what lets TypeScript narrow `data` to the tokens-freed branch of the discriminated union, since `TOKENS_FREED_ACTIONS.has(action)` alone doesn't narrow. Kept both checks and added a comment explaining why, so a future cleanup pass doesn't remove it.

2. **streamState.ts** (`ActiveChildInfo`): Added a `z.preprocess` migration (`fillLegacyActiveChildKind`) that backfills `kind` from `childStreamId` presence for pre-discriminant persisted/replayed entries, following the same pattern as `ContextManagementDataSchema`'s existing legacy migration. Added coverage in `SchemaMigrations.vitest.ts`.

3. **providerErrorFormat.ts**: `normalizeProviderError` now runs `normalizeLegacyProviderErrorFields` (now exported) on cached errors before returning/re-caching them, so a cached error attached before migration (e.g. a resumed tool-use flow's raw persisted `lastError`, which bypasses the schema-level migration via `migrateSharedState`'s direct cast) gets normalized too. Also fixed `normalizeProviderErrorRetryFlag` to actually drop the legacy `retryable` key (previously relied on an enclosing Zod object schema to strip it, which isn't present when this migration is called directly). Added a regression test in `SdkErrorUtils.vitest.mts`.

4. **conversationFormat.ts** (`asText` duplication): No code change — per the review comment's own conclusion ("Keeping them separate is fine; just noting for awareness"), this wasn't a request to extract a shared export for a one-line helper.

Full repo typecheck (`npm run typecheck`, covering extension/CLI/trace-viewer/test-kernel) and the touched vitest suites pass; ESLint clean on touched files.

LionSR:
Reviewed and verified d239f43 independently — this is correct, and it catches a real gap in my earlier dismissal of the "cached errors skip exhaustion migration" finding.

I had traced `attachProviderError`'s callers (`toProviderErrorFromRetry(shared.lastError)`) and concluded `shared.lastError` was always freshly computed in-process. What I missed: `migrateSharedState` (`src/agent/implementations/flows/tooluse/nodes/types.ts`) casts a resumed session's persisted shared state directly (`obj as ToolUseRunShared`) rather than validating the full shape through Zod — so a resumed tool-use flow's `lastError`, if persisted before the `exhaustionReason` migration existed, really could reach `attachProviderError` still in the legacy boolean shape. The fix (running `normalizeLegacyProviderErrorFields` on cached errors too, now exported) closes that gap correctly, and the accompanying `normalizeProviderErrorRetryFlag` fix (actually dropping the stale `retryable` key instead of relying on an enclosing schema to strip it) is a real correctness improvement for that same direct-call path.

Confirmed clean: full `npm run typecheck` and the two affected test suites (`SdkErrorUtils.vitest.mts`, `SchemaMigrations.vitest.ts`, 56 tests) all pass on d239f43.

---

Issue #7246, opened 2026-07-05
Stage 3c decision: completed-run archive owner and flow-node replay policy


LionSR:
Part of #6966 / #6951. This is a maintainer-decision issue, not an implementation PR.

## Status 2026-07-06

Decision 2 (flow-node replay policy) has been implemented by #7282 (`c13e74ed4`): schema-v2 `FlowRecord.cursor` is the replay SSOT, and `nodes[]` is audit data for cursor-bearing records. Cursor-less v1 replay compatibility is tracked in #6981 for D3 retirement. This issue remains open for Decision 1: completed-run display/export ownership after `conversation.json` / `todos.json` projection deletion.

## Why this is blocked

After #7244 and #7245, the remaining Stage 3c work is blocked on two policy choices that change persisted/read semantics. Per the migration rules, do not guess on on-disk format changes or behavior changes not explicitly flagged.

## Decision 1: completed-run display/archive owner after deleting `conversation.json` / `todos.json`

Stage 3c task bullet 2 says to delete the `conversation.json` / `todos.json` KV projections because the flow record is the resume SSOT. Current code still depends on those projections for completed-run surfaces:

- Tool-use still writes the projections on every persisted-flow step: `src/agent/implementations/flows/tooluse/runToolUseFlow.ts:386-389`.
- Normal completed/cancelled tool-use teardown deletes the flow record: `src/agent/implementations/flows/tooluse/runToolUseFlow.ts:443-447`.
- `ExecutionKVStore.readConversation()` / `readTodos()` read only `conversation.json` / `todos.json`: `src/agent/storage/ExecutionKVStore.ts:271-278`.
- CLI history details read the conversation projection for previews/full transcript/workspace-file extraction: `packages/cli/src/runtime/history.ts:143-166`.
- Chat export reads config + conversation via the projection path: `src/agent/export/loadChatExportInput.ts:66-76`.
- `/executions/{id}` summary still reads todos from the projection path: `src/tools/ExecutionsTool.ts:443-448`.

Deleting the projections now would drop completed-run history/export/todo display unless another durable archive owner is selected first.

### Recommended answer

Keep flow records resume-only; do not retain completed flow records just to replace display/export archival data. Make completed-run display/export read through a session storage/display facade backed by StreamLogStore + StreamSnapshotStore sidecars keyed by execution/stream mapping. Keep `conversation.json` / `todos.json` only as a legacy read fallback during migration, with a #6981 ledger row and an explicit removal trigger.

[...]

Issue #6981, opened 2026-07-03
Tracking: legacy retirement ledger — every shim gets a removal trigger


LionSR:
## Purpose

This is the index for TeXRA compatibility retirement. A temporary reader, alias, migration, projection, or dual writer is not accepted unless it has:

1. the old and current writers identified;
2. an introduction or first-release date;
3. an explicit retirement date or event;
4. the protected released/public surface stated;
5. one canonical downstream representation.

When the trigger fires, delete the schemas, transforms, branches, comments, fixtures, and compatibility-only tests together. Desktop-only state does not receive migration machinery because the desktop application has not had a public release.

## Authoritative active trackers

| Tracker | Remaining scope |
|---|---|
| #9422 | Four persisted readers: AgentWorkspace legacy snapshot after 2026-08-14; legacy usage accumulator after 2026-09-22; disk `meta.taskState` after 2026-10-05; pre-`modelId` shared-state recovery after 2026-11-01. |
| #9627 | Small dated readers from August through November 2026, including onboarding, model-name, team-icon, roster, task-group, banner-setting, resume, and Copilot cohorts. |

These trackers are the current source of truth. Completed historical rows are intentionally not copied here.

## Recently completed

- #9646 removed fifteen compatibility arms and their compatibility-only tests.
- #9657, #9661, #9663, #9665, and #9667 completed Stages 2–6 of #9590.
- #9755 executed #9590 Stage 7 — removed legacy identity scans, suffix decoding, and the sidecar `description` fallback. #9590 is now fully resolved.
- #9674 made remote-agent categories a database invariant. Its migration is applied in production: `agent_category` is non-nullable, defaults to `workflow`, and all 23 current rows are non-null.
- Desktop legacy import and data-root migration machinery is absent; desktop reads and writes only the current formats.
- CLI/TUI workflow checkpoints, workflow-agent rosters, and workflow-call traces use only their current protocols.
- Native settings have one persisted spelling and begin from current defaults rather than importing retired VS Code settings.

## Permanent boundaries, not retirement rows

Some tolerant parsing remains because the input is external or public rather than an internal historical format. In particular, provider/model error normalization is a current external parse boundary, not a dated compatibility shim. Such a boundary must be documented as current input validation and must not be described as a migration.

## Review rule

A pull request that adds temporary compatibility must update the appropriate active tracker. "Backward compatibility" alone is not a justification. If a current public contract or released user-data rule requires longer retention, state that surface and its retirement condition explicitly.

[...]

Issue #6976, opened 2026-07-03
F6 — Unify child-agent run drivers behind one session loop


LionSR:
Part of #6953. Audit ref: `tech-debt-audit-2026-07.md` §C1b. **Best after #6951 stages 2–3** so the unified driver lands on `session.runs` rather than being rewired twice.

## Context

Native subagents (`src/tools/delegation/subagentExecution.ts`, 542 LoC) and external CLI agents (`src/tools/agentCliSessionLoop.ts`, 351) hand-implement the same choreography twice: launch child → drive turns → deliver each result to the parent follow-up queue → persist report best-effort → derive outcome → write terminal status → finalize. Twin delivery-envelope formatters too: `subagentResults.ts` (469) vs `agentCliShared.ts:60-95`, both emitting the same `<…-result>`/`<…-error>` XML. The codex/claude pair already proves the fix: both tools are ~30-line strategies over one loop (`AgentCliSessionStrategy`, `agentCliSessionLoop.ts:91-133`).

## Task

- Promote the `AgentCliSessionStrategy` loop into the single child-run driver (owned by `session.runs` post-#6951, or a shared module pre-migration); native delegation (`proposalFlow`/`subagentExecution`) becomes a strategy over it.
- One delivery-envelope builder parameterized by tag/id-attr/extra-lines, replacing the twin formatters.
- Preserve the choreography subtleties both copies handle: interrupt cascade, `subagentDeliveryRegistry` dedup between `onBeforeWaiting`/`onCompleted`, late-delivery after orchestrator disposal, resume via existing execution id.

## Acceptance gates

- [x] One loop, N strategies (codex, claude, native workflow, native toolUse); the twin delivery/error formatters are deleted.
- [x] Delegation E2E fixtures (delegate → interrupt mid-run; delegate → complete → follow-up delivery; late result after parent stop) pass identically for native and CLI-agent children.
- [x] Net LoC reduction reported (net −213, below the forecast −500 to −800 magnitude; see #7536 body for detailed accounting).

## Follow-ups

- [x] #7543 - Fix cross-process restart recovery for delegated native subagents

Size: L.

LionSR:
Partial F6 slice landed in PR #7275, merge commit `7853d5009`.

Scope completed:
- Added `src/tools/childRunDelivery.ts` so agent-CLI children and native delegation share best-effort report persistence plus explicit-session parent follow-up wake/release mechanics.
[...]

PR #7282, opened 2026-07-06
refactor: add persisted flow replay cursor

LionSR:
## Summary

- Add a versioned `FlowRecord.cursor` that stores the next node to execute, making it the authoritative replay state for `PersistedFlow`.
- Keep `nodes[]` as an audit log rather than the replay source, while preserving cursor-less v1 records through the existing node-path replay on their next step.
- Add regression coverage showing cursor replay works even when `nodes[]` is empty, and legacy node-path records migrate to cursor replay after one step.

## Why this is the right next slice

This implements the replay-policy half of #7246 without touching completed-run display/archive ownership. It also removes a concrete blocker for #6976: a future native tool-use WAITING suspension boundary can preserve flow state without depending on an unbounded `nodes[]` replay path.

## Single Source Of Truth

`FlowRecord.cursor.nextNodeId` is now the replay fact. `nodes[]` is no longer consulted for records that carry a cursor; it remains only an append-only audit log and a v1 compatibility source for records written before this PR.

## Scheduled Refactoring

A #6981 ledger row will be added for the cursor-less v1 flow-record replay compatibility path introduced here. Removal trigger: D3 age-based expiry per persisted-run retention, after schema-v2 cursor records have aged past the supported retention window.

## Validation

- `npm test -- --run src/test-kernel/agent/PersistedFlow.vitest.ts src/test-kernel/agent/storage/Resumability.vitest.ts src/test-kernel/agent/SessionResumeRetrieval.vitest.ts src/test-kernel/tools/ExecutionsToolResumability.vitest.ts`
- `npm test -- --run src/test-kernel/agent/followUp/ToolUseWaitNode.vitest.ts src/test-kernel/agent/runtime/resumeToolUseSnapshot.vitest.ts src/test-kernel/desktop/DesktopAgentExecution.vitest.mts`
- `npm run typecheck`
- `npm run compile:fast`
- `npm run lint` (passes with the existing unrelated `codexToolTemplates.ts` warning)
- `npm test`
- `git diff --check`

Part of #7246. Prepares #6976.

---

> [!NOTE]
> **Medium Risk**
> Changes core agent execution resume semantics and persisted record shape (v2); incorrect cursor or node indexing could mis-route resumed runs, though legacy replay and new tests mitigate regressions.
> 
> **Overview**
> **Persisted flow resume** no longer derives the next node by replaying the full `nodes[]` action history. New and updated records use **schema v2** with an authoritative **`FlowRecord.cursor`** (`nextNodeId`, optional `lastAction`); `nodes[]` is append-only audit data (with optional `nodeId` per step).
> 
> **`PersistedFlow`** builds stable graph-local node IDs (from `start` plus encoded action paths via new **`BaseNode.successorEntries()`**), resolves the executable node from `cursor` when present, and falls back to **legacy node-path replay** only for cursor-less records. Each step updates the cursor after execution; **`resetNodeHistory`** clears the audit log and resets the cursor to the start node. **`run()`** / terminal steps prefer **`cursor.lastAction`** over the last audit entry.
> 
> **Tests** cover cursor-based resume with an empty `nodes[]`, legacy records migrating to cursor replay on the next step, and new records stamping v2 cursor state.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf9f080bf09e68838bac2cb96c9b75eb7ef14b80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/node/index.ts
+++ b/src/agent/node/index.ts
@@ -97,6 +97,10 @@ class BaseNode<
     }
     return next;
   }
+  /** Successor table used by persisted flows to build graph-local replay cursors. */
+  successorEntries(): readonly [Action, BaseNode][] {
+    return [...this._successors.entries()];
+  }
   clone(): this {
     const clonedNode = Object.create(Object.getPrototypeOf(this));
     Object.assign(clonedNode, this);
--- a/src/agent/node/persistedFlow.ts
+++ b/src/agent/node/persistedFlow.ts
@@ -16,10 +16,19 @@ export function flowKey(runId: string): string {
 const CHANNEL = 'PersistedFlow';
 logger.initialize(CHANNEL);
 
-export const FLOW_RECORD_SCHEMA_VERSION = 1;
+export const FLOW_RECORD_SCHEMA_VERSION = 2;
+const START_NODE_ID = 'start';
 
 interface NodeRecord {
   action?: string;
+  nodeId?: string;
+}
+
+export interface FlowCursor {
+  /** The next graph-local node path, or null when the flow has reached a terminal edge. */
+  nextNodeId: string | null;
+  /** Last action emitted by the previous node, for diagnostics and legacy parity. */
+  lastAction?: string;
 }
 
 export interface FlowRecord {
@@ -28,6 +37,11 @@ export interface FlowRecord {
   params: Record<string, unknown>;
   shared: unknown;
   createdAt: string;
+  /**
+   * Authoritative replay cursor. `nodes[]` is now only an audit log: it can be
+   * capped or moved later without changing resume semantics.
+   */
+  cursor?: FlowCursor;
   nodes: NodeRecord[];
 }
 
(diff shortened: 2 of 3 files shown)


LionSR:
Amended in `cf9f080bf`: cursor IDs are now graph-local action paths (`start/...`) rather than BFS-order numbers, successor traversal is sorted by action label, and cursor writes preserve empty-string actions with an explicit `!== undefined` check. This addresses the stability concern and the minor `lastAction` robustness note while keeping cursor-less v1 replay as the ledgered compatibility path.

LionSR:
On the remaining throw-vs-fallback point: I am keeping `resolveCursor()` loud when a schema-v2 cursor cannot be resolved. Once `cursor` is present, it is the replay fact; falling back to `nodes[]` after a graph topology change would make the audit log a second replay owner again. Cursor-less v1 records still use the ledgered legacy path, but invalid v2 cursor records should fail by name rather than silently choose another replay rule.

PR #7456, opened 2026-07-07
Remove the max-delegation-depth guard

LionSR:
## Summary

Removes the "Max delegation depth" cap (Settings → Multi-Agent) that limited how many levels deep agents could delegate to subagents (default: 1 — only the root orchestrator could delegate at all).

Cost roll-up already propagates through the whole delegation tree regardless of depth, so the depth cap wasn't buying any real safety — it was just refusing legitimate decomposition (an orchestrator spawning workers that need to spawn their own helpers).

## What changed

- **Runtime gate removed**: `depthGateError()` in `subagentExecution.ts` and its call sites in `executeSubagent`/`DelegationTools.resumeAgent` (resume_agent) are gone. `delegate_agent`/`delegate_workflow`/`resume_agent` are never blocked by depth anymore.
- **Tool-list filtering removed**: `agentToolResolution.ts`/`runToolUseFlow.ts` no longer strip delegation tools out of a subagent's toolset based on depth.
- **Prompt injection removed**: the "delegation depth reached, don't try to delegate further" notice in `PromptBuilder.ts` is gone (nothing to tell the LLM anymore).
- **Settings UI/schema/IPC removed**: the "Max delegation depth" slider in `MultiAgentTab.ts`, its signal in `settingsState.ts`, the `SET_NESTED_DELEGATION_MAX_DEPTH` command/schema, and the extension + desktop host wiring that persisted it.
- **`src/shared/constants/delegationPolicy.ts` deleted** entirely (range constants, clamp, pure gate evaluator) — nothing left to gate.

## What's kept

Delegation depth is still tracked and displayed for observability (CLI history "Delegation depth: N", trace metadata, `isSubagent` detection on resume) — only the cap that *blocked* delegation is removed. `computeDelegationDepthFromStorage` (`agent/runtime/delegationPolicy.ts`) is simplified: its elaborate corruption/cycle fail-closed handling existed solely to keep a broken-meta resume from bypassing the gate, so with the gate gone it now just falls back to depth 0 on any unreadable/cyclic lineage instead of a sentinel.

## Verification

- `npm run typecheck`: clean across the full monorepo (root, test-kernel, cli, extension, trace-viewer).
- `eslint` on all 27 changed files: clean.
- Targeted vitest: `DelegationTools`, `DelegationHeadless`, `DelegationAgentAvailability`, `DelegationWorktreeAvailability`, `ToolUseToolResolution`, `HistoryStatus`, `Orchestration` — 88/88 passing.
- Broader sweep: `src/test-kernel/agent/`, `src/test-kernel/tools/`, `src/test-kernel/progressView/`, `src/test-kernel/settings/`, `src/test-kernel/desktop/` — 2112/2116 passing (4 pre-existing skips, 0 failures).
- Deleted the now-fully-obsolete `src/test-kernel/shared/delegationPolicy.vitest.ts` (tested only the deleted gate-evaluation file); updated the 5 other test files that referenced the removed `delegationBlocked`/`delegationTrimmed`/`depthGateError` surface.

Net: -489 LOC across 27 files.

---

> [!NOTE]
> **Medium Risk**
> Unbounded nested delegation can increase cost and runaway subagent fan-out; changes touch core delegation and tool resolution paths, though approval and other gates remain.
> 
> **Overview**
> Removes the **Max delegation depth** setting and every mechanism that blocked or hid delegation past a configured limit (default depth 1).
> 
> **Runtime:** `depthGateError` and gate checks in `subagentExecution` / `DelegationTools.resumeAgent` are gone; `resolveAgentTools` no longer strips delegation tools via `delegationBlocked` or exposes `delegationTrimmed`; `runToolUseFlow` no longer evaluates a delegation gate. Tool-use prompts no longer inject a “you cannot delegate further” suffix when tools were trimmed.
> 
> **Settings / IPC:** The Multi-Agent tab numeric control, `nestedDelegationMaxDepth` state, `SET_NESTED_DELEGATION_MAX_DEPTH`, `WorkspaceStateKey.NESTED_DELEGATION_MAX_DEPTH`, and `setNestedDelegationMaxDepth` are removed from extension and desktop hosts; Super YOLO outbound messages no longer carry depth.
> 
> **Policy module:** `src/shared/constants/delegationPolicy.ts` is deleted (clamp, gate evaluator, `UNKNOWN_DELEGATION_DEPTH`). `computeDelegationDepthFromStorage` is retained but documented as **observability-only** and returns best-effort depths instead of fail-closed unknown depth for corrupted lineage.
> 
> **Tests:** Gate/delegation-block tests removed or updated; new `DelegationDepthRecovery.vitest.ts` covers depth recovery behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0b79d92fb3fae060bec88a27991a70cec96f6ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
we should do more things like this

LionSR:
Both findings are correct — thanks. Fixed in 521ccd671: once `rootMeta.parentExecutionId` proves an execution is nested, a broken link further up the walk (cycle, unreadable ancestor, or MAX_WALK exceeded) now returns the deepest depth confirmed so far instead of silently promoting it to 0. Added regression coverage in `src/test-kernel/agent/runtime/DelegationDepthRecovery.vitest.ts` (verified the new cases fail against the old `return 0` behavior and pass with the fix).

LionSR:
Correct catch — real off-by-one. Fixed in 14927fc7a: the cycle and MAX_WALK-exceeded branches now return `depth - 1` (the last node actually confirmed to exist), since `depth` had already been pre-incremented for a next hop that either loops back (cycle) or was never reached (MAX_WALK). Traced through both branches by hand with a concrete a→b→a example and a 40-node chain, updated the regression tests to assert the exact corrected values (2 instead of >0 for the cycle case; added a new MAX_WALK case expecting 31), and clarified the JSDoc to explain why the unreadable-ancestor branch doesn't need the same adjustment (a real parent meta already asserted that link before the read of its target failed, so no pre-increment happened past an unconfirmed node).

PR #7472, opened 2026-07-07
fix(agent-runtime): guard waiting-kill on genuine WAITING stream status

LionSR:
## Summary

Follow-up to #7324 (already merged), addressing 4 review threads (cursor Bugbot + 2 GitHub Copilot comments + a claude bot comment) posted against that PR's final commits after it had already been squash-merged to main. Since the PR was closed, this delta lands as its own PR rather than reopening the original.

`NativeSubagentStrategy.onBeforeWaiting` registers `registerWaitingCleanup(() => this.abandon())` unconditionally on every delivered turn — including one where `ToolUseWaitNode` finds a follow-up already queued and keeps the flow running in-process instead of genuinely suspending via `{ kind: 'waiting' }`. A kill landing in the narrow window between that turn's interrupt-unregister and its handle's normal untrack could previously mistake this stale registered cleanup for a real suspension and incorrectly cancel a run that was still executing or had already completed successfully.

## What changed

- **`executionRegistry.ts`**: `terminateWaitingHandle` now requires `streamStatus.get(handle.childStreamId) === STREAM_PHASE.WAITING` before treating a registered cleanup as safe to run. This is an independent, authoritative confirmation on top of `clearWaitingCleanup`'s own best-effort clearing on non-WAITING terminal paths (already landed in #7324) — belt and suspenders against any future non-waiting exit that forgets to call it.
- **`AgentRunLifecycle.ts`**: the per-suspension parent transcript stage is now closed by writing its `GROUP_END` entry directly via `ctx.session.transcripts.update(...)`, instead of calling `ctx.parentStage.end()`. The latter is a silent no-op by the time a later kill's closure runs — `disposeTrace()` already unsubscribed the transcript recorder from that stage's trace the instant the WAITING branch returned.
- **`nativeSubagentStrategy.ts`**: doc comment update explaining the belt-and-suspenders relationship between the two guards.
- Updated `ExecutionRegistry.vitest.ts` (seeded WAITING status in the two pre-existing tests that now require it; added a new regression test reproducing the non-suspend-path stale-cleanup scenario) and `AgentRunLifecycle.vitest.ts` (seeded WAITING status for the fake-runner test; replaced the now-stale `ctx.parentStage.end()` spy assertion with one on `ctx.session.transcripts.update`).

## Verification

- `npm run typecheck`: clean.
- `eslint` on all 5 changed files: clean.
- `vitest run` on `ExecutionRegistry.vitest.ts`, `NativeSubagentStrategy.vitest.ts`, `AgentRunLifecycle.vitest.ts`, `src/test-kernel/agent/followUp/`: 153/153 passing.
- Confirmed this branch's base diff against current `origin/main` is empty (i.e., this is a pure delta on top of what #7324 already merged) before opening this PR.

---

> [!NOTE]
> **Medium Risk**
> Changes agent-runtime stop/kill and transcript teardown for suspended subagents; behavior is narrow but user-visible (incorrect cancel vs ignored stop) and covered by new regression tests.
> 
> **Overview**
> Follow-up to #7324: hardens **stop/kill of suspended native subagents** so speculative waiting-cleanup registrations cannot cancel runs that never actually parked.
> 
> **`terminateWaitingHandle`** now runs waiting-cleanup only when `streamStatus` is **`WAITING`** or the stream is **`RESUMING`** (resume window before a new interrupt is registered). A registered cleanup alone is no longer enough—this blocks incorrect abandons when `onBeforeWaiting` registered cleanup but the flow kept running because a follow-up was already queued.
> 
> **WAITING kill cleanup** no longer calls `ctx.parentStage.end()` (a no-op after `disposeTrace()` on suspend return). It writes the run transcript **`GROUP_END`** directly via `ctx.session.transcripts.update`, so a killed suspended subagent’s “Run: …” group does not stay **`running`** forever.
> 
> Tests seed **`WAITING`**/`RESUMING` where required and add regressions for stale-cleanup no-op and RESUMING-window kill. **`nativeSubagentStrategy`** docs note the belt-and-suspenders relationship with the registry guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ce2411aae6aa668d35eb55e5caa3740efe13762. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
All 3 findings addressed in 9d556316b:

1. **Stop ignored during resume transition** (cursor) — confirmed the race: `resumeQueuedToolUseSnapshot` flips `streamStatus` to RUNNING+RESUMING *before* the resumed run installs its own interrupt context. `terminateWaitingHandle`'s guard now accepts either a genuine WAITING phase or a RESUMING substate (mirroring `getToolUseFollowUpTarget`'s existing analogous treatment of RESUMING-as-WAITING for follow-up admission), so a kill landing in that window still tears the stalled resume down instead of silently no-opping. Added a regression test that seeds RESUMING and asserts `kill()` now returns true and runs the cleanup — verified it fails without the fix (reverted locally, confirmed the failure, restored).
2. **Swallowed rejection needs a comment** (copilot) — added an inline comment on the `.catch(() => {})` explaining why: this closure only fires from a later kill with no live caller left to propagate to, and a failed delete just leaves a stale flow-record on disk, not a correctness gap.
3. **Stale doc comment** (copilot) — updated `terminateWaitingHandle`'s doc comment to describe the WAITING-or-RESUMING guard instead of implying WAITING is the only accepted phase.

Verification: `npm run typecheck` clean, `eslint` clean on all 3 changed files, targeted vitest (ExecutionRegistry/AgentRunLifecycle/NativeSubagentStrategy/followUp) 154/154 passing.

LionSR:
Sequencing (see the work-queue comment on #6951): this overlaps #7474 (Sweep 1) on `src/agent/runtime/AgentRunLifecycle.ts`. #7474 is P0 and merges first — please hold this and rebase on main right after it lands rather than racing the seam (that exact race cost three rebases today). The fix itself is welcome P0-tail: after rebase, re-verify the guard against the sweep's final `terminateWaitingHandle` shape.

LionSR:
#7474 merged — rebase now and re-verify against the final `terminateWaitingHandle` shape on main.

Issue #7447, opened 2026-07-07
Align extension/desktop HostInteractions resolve/cancel semantics; share the result mappers


LionSR:
From the 2026-07-07 campaign audit (adversarially verified). The per-host HostInteractions implementations are intended Stage-4 architecture, but within one week they drifted in observable, untracked ways that break the one-vocabulary rule:

- **Kind check on resolve**: extension enforces a result-kind match (`extensionHostInteractions.ts:85`) while desktop ignores `result.kind` for non-inquiry kinds (`desktopHostInteractions.ts:162-206`).
- **`toProposalResult` duplicated with divergent fallbacks**: desktop passes through any object containing `action` (`desktopHostInteractions.ts:272-275`); extension whitelists approve/setup/reject (`extensionHostInteractions.ts:319-322`).
- **Cancellation cause dropped by extension**: `cancelForStream` ignores the interface-declared `cause` (`extensionHostInteractions.ts:277` vs `HostInteractions.ts:126`) and settles bash as bare `{accepted:false}`, while desktop forwards cause as agent-visible reject feedback (`desktopHostInteractions.ts:208-217`). The same core cancellation ('Stream resources released.' from `src/tools/approval/index.ts:38`) yields different agent-visible feedback per host. #7333 extended the divergence (`void cause` vs forwarding).

**Scope**: align cause-propagation and kind-check semantics, and share the single resolution-mapping vocabulary (toProposalResult / plan mapping). A wholesale shared pending-map extraction is a separate, larger decision (see the fewer-elements doc's PendingInteractionTable recommendation) and must clear the abstraction-cost guardrails.

Relates to #6967, #6968.

LionSR:
Implementation PR opened: #7495. This covers the shared host-interaction result mapper, extension cancellation cause propagation, and desktop wrong-kind resolve guard.

LionSR:
Coordination update: draft #7495 is superseded by the non-draft implementation PR #7494 for the same HostInteractions resolve/cancel alignment.

PR #7482, opened 2026-07-07
fix(cli): stop masking non-ENOENT read failures on .gitignore and config reads

### LINKED ISSUE #7470: Non-ENOENT read-default batch: .gitignore overwrite on read failure; corrupt user config silently drops chat defaults
Medium (M2/M6), from the 2026-07-07 audit (CONFIRMED):

- `ensureTexraGitignored` treats ANY read failure as file-absent, then overwrites `.gitignore`, destroying user content on e.g. a transient EACCES.
- Corrupt user-level `config.json` silently drops the user's chat defaults (asymmetric with the workspace loader, which surfaces the failure).
- Same class: `getValidatedConfig` swallows invalid user settings via `schema.catch(default)`.

Fix per checklist §15: `isFileNotFoundError` predicate + warn on everything else; the .gitignore write must not proceed past a failed read. Small, host-path only, no Stage-5 collision.

LionSR:
## Root cause

Three related read-default sites (2026-07-07 audit, CONFIRMED, checklist §15) treated *every* read failure the same as "file absent," which either destroys user data or silently drops a setting instead of surfacing the problem:

1. **`ensureTexraGitignored`** (`packages/cli/src/runtime/initConfig.ts`) caught any `readFile` error — not just `ENOENT` — and fell through to the write path. A transient `EACCES` (or any other non-missing-file error) on an *existing* `.gitignore` would silently overwrite it with just `.texra/\n`, destroying the user's content.
2. **`loadUserDefaults`** (`packages/cli/src/runtime/chatDefaults.ts`) swallowed every `GlobalStorageFS.readJson` failure alike via `.catch(() => undefined)`, so a corrupt user-level `config.json` silently dropped the user's chat defaults with zero signal. This was asymmetric with `loadWorkspaceCliConfig`, which already surfaces this class of failure via `configWarnings`.
3. **`getValidatedConfig`** (`src/utils/config/configUtils.ts`) used `schema.catch(defaultValue)`, so a stale or hand-edited VS Code setting that fails its schema silently reverts to the default with no trace — the "silent Zod default on persisted/user-authored data" masking pattern the repo's error-handling checklist explicitly flags as a blocker.

## Fix

Applied the repo's established `isFileNotFoundError` predicate (`@common/errors`) at each site, per the "loud read" pattern in the code-review checklist:

- **`ensureTexraGitignored`**: only treats a genuine `ENOENT` as "file absent"; any other read error rethrows before ever reaching the write, so the destructive-overwrite path is unreachable on a non-missing-file failure.
- **`loadUserDefaults`**: only silently defaults to `{}` on `ENOENT`; anything else writes a `WARN` line to stderr (mirroring the existing `configWarnings` stderr convention used for workspace config) and then continues with the fallback so the CLI still starts.
- **`getValidatedConfig`**: replaced `schema.catch(default)` with `schema.safeParse` + a channel-log warning (`@logger/logUtils`), gated on the existing `isConfigExplicitlySet` predicate so a *merely unset* setting (which legitimately fails most schemas as `undefined`) stays silent — only a value the user actually set that fails validation warns.

No new abstractions: each fix reuses an existing predicate/helper already established elsewhere in the codebase (`isFileNotFoundError`, `writeTextStderr`, `@logger/logUtils`, `isConfigExplicitlySet`).

## Regression coverage

- `ensureTexraGitignored`: added tests for create/append happy paths (previously untested) plus a repro that a non-ENOENT read failure (mocked `EACCES`) rejects instead of overwriting existing `.gitignore` content.
- `loadUserDefaults` / `resolveChatDefaults`: fixed the suite's default mock to reject with a realistic `ENOENT`-shaped error instead of a plain `Error` (which the old code and the new code both treat identically as "missing," but which would have made the fix's warning path fire on *every* existing test). Added a test that a non-ENOENT rejection warns via `writeTextStderr` and still falls through to the next default tier.
- `getValidatedConfig`: new test file (no prior suite existed) covering: valid value passes through, unset value falls back silently (no warning), and an explicitly-set invalid value warns and falls back.

## Verification

```
npx tsc --noEmit                                    # clean
npx tsc --noEmit -p tsconfig.test-kernel.json        # clean (1 pre-existing unrelated failure, confirmed via git stash)
cd packages/cli && npx tsc --noEmit -p tsconfig.json # clean
npx eslint <changed files>                           # clean (2 import/order warnings auto-fixed)
npx prettier --check <changed files>                 # clean
npx vitest run src/test-kernel/cli src/test-kernel/utils/config
  # 130 files / 1608 tests passed
npm run check:dead-code-ratchet                      # at baseline
```

Fixes #7470.

---

> [!NOTE]
> **Medium Risk**
> Touches init-time filesystem writes and shared config validation used across the extension; behavior changes are intentional but affect startup and settings fallback paths.
> 
> **Overview**
> Stops treating every config/read failure as “file missing” so user data is not clobbered and bad settings are not dropped silently.
> 
> **`texra init` / `.gitignore`:** `ensureTexraGitignored` only treats **ENOENT** as absent; **EACCES** and other read errors **rethrow** instead of writing a new `.gitignore` that would overwrite existing content.
> 
> **CLI chat defaults:** `loadUserDefaults` still ignores a missing user `config.json`, but **corrupt JSON or permission errors** emit a **stderr WARN** (aligned with workspace config warnings) before falling back.
> 
> **VS Code settings:** `getValidatedConfig` uses **`safeParse`** and logs a warning when a setting is **explicitly set** but invalid; unset values still default **without** a warning.
> 
> Regression tests cover gitignore create/append, EACCES non-overwrite, corrupt user config warning, and invalid explicit settings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3710583a26b2c3229bb1eb7690008d6165a5ee27. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/runtime/chatDefaults.ts
+++ b/packages/cli/src/runtime/chatDefaults.ts
@@ -1,11 +1,13 @@
 import { TEXRA_CONFIG_FILE_NAME } from '@platform/defaults/nodeStorage';
 import { listExecutions } from '@agent/storage';
 import { AgentCategory } from '@agent/core/definition/AgentDataclass';
+import { isFileNotFoundError } from '@common/errors';
 import {
   decideRunModel,
   type RunModelDecisionReason,
 } from '@model/runModelDecision';
 import { toNewestFirstByTimestamp } from '@utils/core';
+import { toErrorMessage } from '@utils/errors/errorMessage';
 import { GlobalStorageFS } from '@utils/files/storageFS';
 import {
   CLI_BUILTIN_DEFAULT_MODEL,
@@ -19,6 +21,7 @@ import {
   isImplicitDefaultEligible,
   pickDefaultToolUseAgent,
 } from './defaultAgents';
+import { writeTextStderr } from './logSinks';
 
 /**
  * A configured or environment agent value, trimmed and dropped if it can't be
@@ -74,11 +77,22 @@ async function loadWorkspaceDefaults(cwd: string): Promise<PartialDefaults> {
 }
 
 async function loadUserDefaults(): Promise<PartialDefaults> {
-  // A missing or corrupt user config means no user defaults:
-  // parseCliConfigValues maps the undefined fallback to {}.
-  const raw: unknown = await GlobalStorageFS.readJson(
-    TEXRA_CONFIG_FILE_NAME,
-  ).catch(() => undefined);
+  // A missing user config means no user defaults (parseCliConfigValues maps
+  // the undefined fallback to {}). Anything else — corrupt JSON, a permission
+  // error — is surfaced as a warning instead of silently dropping the user's
+  // chat defaults, mirroring loadWorkspaceCliConfig's handling of the same
+  // class of failure for the workspace config.
+  let raw: unknown;
+  try {
+    raw = await GlobalStorageFS.readJson(TEXRA_CONFIG_FILE_NAME);
+  } catch (error: unknown) {
+    if (!isFileNotFoundError(error)) {
+      writeTextStderr(
+        `WARN Could not read user config (${TEXRA_CONFIG_FILE_NAME}): ${toErrorMessage(error)}`,
+      );
+    }
+    raw = undefined;
+  }
   return defaultsFromConfigValues(parseCliConfigValues(raw));
 }
 
--- a/packages/cli/src/runtime/initConfig.ts
+++ b/packages/cli/src/runtime/initConfig.ts
@@ -8,6 +8,7 @@ import { access, mkdir, readFile, writeFile } from 'node:fs/promises';
 import path from 'node:path';
 
 import { TEXRA_STORAGE_DIR_NAME } from '@platform/defaults/nodeStorage';
+import { isFileNotFoundError } from '@common/errors';
 
 import type {
   CliApprovalPolicy,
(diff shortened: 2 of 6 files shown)


LionSR:
CI status: every check is green except \`validate\`/\`validate (linux)\`, which fail on a pre-existing, unrelated typecheck error:

\`\`\`
src/test-kernel/frontend/ExtensionAgentRuntimeHost.vitest.ts(37,7): error TS2353: Object literal may only specify known properties, and 'pending' does not exist in type 'HostInteractions'.
\`\`\`

This is not caused by this PR's diff — confirmed via \`git stash\` on this branch's base commit (80537be19) and by fetching current \`origin/main\` (173137fca): the failure reproduces on both with this PR's changes fully removed. It's in \`packages/extension\`'s \`HostInteractions\` type surface, part of the actively-migrating host-interactions-port subsystem (see #7447, #6951), and out of scope for this fix. Not attempting a fix here per this repo's convention of not rushing patches into an in-flight shared-subsystem migration from an unrelated PR.

Leaving this PR open with CI red for that one unrelated job; the #7470 fix itself is fully verified locally (typecheck/lint/prettier/tests all clean, see PR description).

PR #7491, opened 2026-07-07
fix(agent-runtime): deliver failed resume wakeups through the subagent error path

### LINKED ISSUE #7402: Surface failed resume wakeups to the orchestrator instead of only logging
## Context

Follow-up from #7319 (fix(agent-runtime): don't block the parent while resuming a suspended child).

That PR correctly stopped `delegate_agent`'s resume path from awaiting `nativeStrategy.wakeQueuedFollowUp(...)` / `wakeQueuedFollowUpStream(...)`, since blocking the parent tool call on the resumed child's full turn contradicted the async delegation contract (#7289). The wake is now dispatched fire-and-forget with only a `.catch()` that logs a warning via `logger.warn`.

Flagged by `chatgpt-codex-connector[bot]` (P2) on PR #7319 and unaddressed at merge: when a native resumed child hits an unexpected resume-storage failure (e.g. `retrieveSessionResumeData` throwing for unreadable resume storage — `src/agent/runtime/SessionResumeRetrieval.ts:110-113` — awaited in `resumeStream`, `src/tools/delegation/nativeSubagentStrategy.ts:299`), the wake promise can reject *before* `resumeQueuedToolUseSnapshot` installs the subagent's error-delivery callbacks. The orchestrator has already been told the follow-up "will process it automatically" (it was queued via `sendFollowUp`), but no async subagent error is ever delivered back — the failure only reaches a log channel, and the follow-up is left queued indefinitely with no observable delivery to the parent.

## Details

Route native wake failures through the strategy's existing error-delivery path (the same channel used to deliver a resumed subagent's terminal error back to the orchestrator), instead of only logging via `logger.warn`. Otherwise surface the failed resume in a way the orchestrator can observe and act on.

## References

- PR: #7319
- Review comment: https://github.com/LionSR/TeXRA/pull/7319#discussion_r3531608998
- `src/tools/DelegationTools.ts` (wake `.catch()` handler)
- `src/tools/delegation/nativeSubagentStrategy.ts:273-320` (`wakeQueuedFollowUp`, `resumeStream`)
- `src/agent/runtime/SessionResumeRetrieval.ts:110-113` (`retrieveSessionResumeData`)

LionSR:
## Summary

Follow-up from #7319, flagged by `chatgpt-codex-connector[bot]` (P2) on that PR and unaddressed at merge (https://github.com/LionSR/TeXRA/pull/7319#discussion_r3531608998).

`delegate_agent`'s resume path dispatches a native subagent's wake fire-and-forget (see #7289 — awaiting it would stall the parent's whole turn for as long as the child keeps running). But if the wake rejects before `NativeSubagentStrategy.resumeStream` ever reaches `resumeQueuedToolUseSnapshot` — e.g. `retrieveSessionResumeData` throwing for unreadable resume storage (`src/agent/runtime/SessionResumeRetrieval.ts:110-113`) — that turn's `onError`/`onRunError` callbacks are never installed. The failure previously only reached a `logger.warn` in `DelegationTools.ts`'s `wake.catch()`, even though the orchestrator had already been told the follow-up "will process it automatically" via `sendFollowUp`.

## Changes

- `NativeSubagentStrategy.wakeQueuedFollowUp` now catches a rejection from `wakeQueuedFollowUpStream` and routes it through `deliverSubagentError` — the same terminal error-delivery channel a resumed turn's own `onError`/`onRunError` use — then retires the strategy via `finish()`, matching how every other terminal path in this class tears down.
- `DelegationTools.ts`'s `wake.catch()` comment updated to reflect that it's now only a last-resort backstop (non-native path with no strategy to deliver through, or a failure in the delivery path itself), not the primary channel for native wake failures.

## Test plan

- [x] `npm run typecheck` passes.
- [x] Added a regression test in `NativeSubagentStrategy.vitest.ts` reproducing the exact failure scenario from the issue (`retrieveSessionResumeData` rejecting before `resumeQueuedToolUseSnapshot` is called); confirmed it fails against the pre-fix code and passes with the fix.
- [x] `npx vitest run src/test-kernel/tools/` — 83 files / 445 tests pass.
- [x] `npx eslint` on all changed files — clean.

Closes #7402

---

> [!NOTE]
> **Medium Risk**
> Touches subagent delegation/resume and parent cost accounting, but the change is a narrow error-handling path with regression tests and mirrors existing abandon()/onError behavior.
> 
> **Overview**
> Fixes a gap where **`delegate_agent` resume** tells the orchestrator a queued follow-up will run automatically, but a **fire-and-forget wake** could fail (e.g. unreadable resume storage in `retrieveSessionResumeData`) **before** resume installs `onError`/`onRunError`—leaving only a **`DelegationTools` log** and no parent-visible outcome.
> 
> **`NativeSubagentStrategy.wakeQueuedFollowUp`** now **catches** wake failures, **settles parent cost** from **`lastKnownCostUsd`** when the wake never reaches a real turn (same pattern as **`abandon()`**), **delivers** via **`deliverSubagentError`**, and **`finish()`** to clear strategy/registry state. **`DelegationTools`** documents that **`wake.catch`** is a **backstop** for non-native paths or delivery failures, not the primary native path.
> 
> Regression tests cover orchestrator error delivery, cleanup after wake failure, and cost settlement on wake failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d37962d2d357b967f796bcdf918043cc2f5bbe9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/tools/NativeSubagentStrategy.vitest.ts
+++ b/src/test-kernel/tools/NativeSubagentStrategy.vitest.ts
@@ -313,6 +313,50 @@ describe('NativeSubagentStrategy', () => {
     ).toBeUndefined();
   });
 
+  it('settles the parent usage totals with the last known cost when a wake failure has nothing else to settle from (Bugbot: wake failure locks zero cost)', async () => {
+    const childStream = 'child-stream' as StreamTabId;
+    mocks.readConfig.mockResolvedValue({ agentCategory: 'toolUse' });
+    mocks.retrieveSessionResumeData.mockRejectedValue(
+      new Error('resume storage unreadable'),
+    );
+    mocks.persistChildRunReport.mockResolvedValue({ kind: 'persisted' });
+    mocks.deliverChildRunFollowUp.mockResolvedValue({ kind: 'delivered' });
+    const settleSubagentCost = vi.fn();
+
+    const strategy = new NativeSubagentStrategy({
+      executionId,
+      agentName: 'review',
+      orchestratorStreamId: parentStreamId,
+      parentSession: ownerSession,
+      runtimeHost: { emit: vi.fn() },
+      startedAt: 0,
+      settleSubagentCost,
+    });
+    strategy.setChildStreamId(childStream);
+    // Captures a cost snapshot on the suspended turn before the wake fails —
+    // mirrors the same real-world sequence as the #7287 abandon() test above
+    // (onBeforeWaiting always runs before a turn can suspend and later be
+    // resumed/abandoned/wake-failed).
+    await strategy.onBeforeWaiting('done for now', [], [], 0.0042);
+
+    await strategy.wakeQueuedFollowUp(
+      { status: 'queued', reason: 'waiting' },
+      {},
+    );
+
+    // deliverSubagentError's own settle call (below, second call) carries no
+    // result (the wake never reached a real turn), so without the fallback
+    // the parent would only ever see a 0-cost settle. The real
+    // `subagentExecution.ts` wrapper this strategy is constructed with in
+    // production is itself idempotent (first call wins), so it's this first
+    // call — not the second — that determines what the parent's usage totals
+    // actually record.
+    expect(settleSubagentCost).toHaveBeenNthCalledWith(
+      1,
+      expect.objectContaining({ totalCostUsd: 0.0042 }),
+    );
+  });
+
   it('cleans up registry entries when the underlying execution is untracked while abandoned in WAITING', () => {
     let abandonListener: ((handle: unknown) => void) | undefined;
     const addListener = vi.fn((_id: string, cb: (handle: unknown) => void) => {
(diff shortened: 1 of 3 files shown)


LionSR:
Correct catch — fixed in d37962d2d. The wake-failure catch now settles the parent's usage totals from `this.lastKnownCostUsd` (the same fallback `abandon()` already uses for the identical reason: no real `AgentFlowResult` exists to read cost from) before calling `deliverSubagentError`. Since the real `settleSubagentCost` wrapper (constructed in `subagentExecution.ts`) is itself idempotent — first call wins — this ensures the parent sees the real captured spend instead of `deliverSubagentError`'s own no-result zero-settle. Added a regression test confirming the fix (verified it fails against the pre-fix code, restored, confirmed pass).

Issue #7467, opened 2026-07-07
Malformed provider tool_calls payload silently becomes 'no tool calls' and the run mis-completes


LionSR:
High, from the 2026-07-07 error-handling audit (CONFIRMED). `extractToolUse`'s catch skips the malformed entry, the cycle sees zero tool calls, and the run finalizes as a successful completion — the user receives a confidently wrong 'completed' result from a corrupted provider payload.

Fix (checklist §15, delete-the-guard): throw to the existing L4 classifier boundary (`classifyAgentError`) so the run fails loudly and retryably instead of mis-completing.

PR #7499, opened 2026-07-07
fix(model-handlers): centralize media/tool-attachment failure policy (#7465)

### LINKED ISSUE #7465: Media attachment failure policy is re-derived at 15 divergent call sites in model handlers
High, from the 2026-07-07 error-handling audit (verified as the M4 companion of the largest M1 cluster). 11 `createMediaMessage` drop-and-continue catch sites, 2 attachment-encode-to-null drops, 2 tool-attachment upload drops across `src/agent/modelHandlers/`. Initial-message paths let the failure fail the run; follow-up/insert paths swallow it with a warn log — the same user action (attach a file) silently vanishes or errors depending on which code path runs.

Fix (checklist §15, decide-once-carry-as-data): ONE policy owner decides fail-vs-warn-vs-surface per round type; call sites consume the decision. `MediaAttachmentProcessor` already returns per-file `{ok:false}` results, so most of the fix is deleting the per-site catches and letting the owner act on that data. Fixing this single pattern removes 15 of the model-handler scope's 19 masking sites.

LionSR:
## Summary

**#7465** — Media/tool-attachment failure policy is re-derived at 15 divergent call sites in `src/agent/modelHandlers/`. Verified the count on `main` as of this PR's base: 11 `createMediaMessage` catch-and-drop sites (`initializeMessages`/`createRoundMessages`/`addMediaToUserMessage` across Anthropic, OpenAI chat, OpenAI Responses, Google GenAI, Google Interactions, OpenRouter), 2 attachment-encode-to-null sites (`buildFunctionResponseAttachment` in Google GenAI, `buildFunctionResultImage` in Google Interactions), and 2 tool-attachment-upload sites (`uploadToolAttachments` in `anthropicTools.ts` and `openAIResponseFileUploads.ts`).

Added a single policy owner — `src/agent/modelHandlers/support/mediaAttachmentPolicy.ts` (`reportMediaAttachmentFailure`) plus a thin call-through on the base class (`ModelHandler.createMediaForRound`) — and routed all 15 sites through it instead of re-deriving fail-vs-warn locally. `initial` rounds fail the round loudly (thrown); `followUp`/`insert`/`toolAttachment` warn (one consistently formatted error-level trace entry) and continue without the attachment.

This also fixes two real divergences the audit flagged:
- `ModelHandlerOpenAIResponse.initializeMessages` (an *initial* round) used to catch-and-warn instead of failing the round, unlike every other provider's initial round.
- `ModelHandlerGoogleGenAI.createRoundMessages` (a *follow-up* round) had no catch at all and threw, unlike every other provider's follow-up round.

**#7467** — already closed independently by #7489 (merged to `main` while this branch was in flight). This branch originally carried its own duplicate fix for the same `extractToolUse` catch; rebased onto `main` and dropped the redundant hunk plus its redundant regression test in favor of `main`'s version (functionally identical — both make the malformed-payload catch throw instead of swallowing). No `#7467` changes remain in this PR.

None of the above affects the #7465 changes, which touch different methods in the same files and are independent of the #7467 fix.

## Test plan
- [x] `npm run typecheck` — clean (after rebasing onto `main`, post-#7489)
- [x] `eslint` on all changed/added files — clean
- [x] Full relevant vitest suites (`agent/modelHandlers`, `agent/followUp`, `modelHandlers`) — 440 passed, 4 pre-existing skips
- [x] New regression tests, each confirmed to fail against the pre-fix code before restoring the fix:
  - `MediaAttachmentFailurePolicy.vitest.ts` — pins the two #7465 divergence fixes (OpenAIResponse initial round now fails; Google GenAI follow-up round now warns-and-continues) plus the two unchanged behaviors
  - `AnthropicToolAttachmentUploadPolicy.vitest.ts` — `uploadToolAttachments` still degrades a failed upload to `unsupported` and continues with the rest, now reported through the shared policy owner

Closes #7465

---

> [!NOTE]
> **Medium Risk**
> Changes user-visible behavior for media failures on initial vs follow-up rounds and touches every provider’s message-building path; risk is mitigated by centralized policy and targeted regression tests.
> 
> **Overview**
> **#7465** unifies how model handlers react when user media or tool attachments fail to load, encode, or upload. A new `reportMediaAttachmentFailure` policy in `mediaAttachmentPolicy.ts` decides behavior by context: **initial** rounds **throw** (fail the run); **follow-up**, **insert**, and **toolAttachment** **log and continue** without the file, with **warn** vs **error** severity preserved per context (tool uploads stay warn-level).
> 
> All providers now build round media through **`ModelHandler.createMediaForRound`** instead of per-handler try/catch around `createMediaMessage`. Anthropic/OpenAI/OpenRouter/Google paths and tool upload helpers (`anthropicTools`, `openAIResponseFileUploads`, Google function-result encoding) route failures through the same reporter.
> 
> This fixes two inconsistencies: **OpenAI Responses** `initializeMessages` used to warn on media errors instead of failing like other providers’ first rounds; **Google GenAI** `createRoundMessages` used to throw on media errors instead of warning and continuing like other follow-up rounds.
> 
> Google GenAI/Interactions **`createMediaContent`** is now a deliberate throw (media is built in overridden `createMediaMessage`) so the base class can type media as `Part`/`Content` without unsafe casts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b3b7f462cb38eaa395bc583f732064c66797997. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
All 3 findings fixed in 0b3b7f462:

1. **Hard-coded 'process media files' operation label** — added `CONTEXT_OPERATION` keyed by context (`toolAttachment` now logs `operation: 'upload tool attachment'`, everything else keeps `'process media files'`).

2. **Dead `createMediaContent` overrides with unsafe casts** — both `ModelHandlerGoogleGenAI` and `ModelHandlerGoogleInteractions` now throw a clear error instead of casting `MediaEntry[]` to `Part[]`/`Content[]`, so a future refactor that accidentally makes these reachable fails loudly instead of sending a malformed payload to the Google SDK.

3. **JSDoc/severity mismatch for `toolAttachment`** — real finding, not just docs. Checked git history: the pre-refactor `anthropicTools.ts` upload-failure catch genuinely used `logger.warn`, so centralizing through `logSdkError` (always error-level) was a real severity escalation for that context. Added `CONTEXT_SEVERITY` so `toolAttachment` now warns (matching its prior behavior — a transient upload hiccup shouldn't render as a red error) while `followUp`/`insert` stay at error level. Updated the JSDoc to describe this explicitly instead of a blanket "warn" claim.

Verification: `npm run typecheck` clean, `eslint` clean, `vitest run` on the modelHandlers suites — 442/442 passing (2 new tests added: one pinning `toolAttachment`'s warn level, one pinning `followUp`'s unchanged error level; also fixed `AnthropicToolAttachmentUploadPolicy.vitest.ts`'s stale assertion, which had been checking the wrong severity). Confirmed both new tests fail against the pre-fix code (temporarily reverted, reran, restored).

Issue #6966, opened 2026-07-03
Stage 3c: persistence facade — atomic delete, orphan sweep, deriveResumability, repairAfterRestart


LionSR:
Part of #6951. **Stage 2/2.5 have landed. Proceed after Checkpoint A (#6978) lands.** Can run in parallel with 3a/3b. Read the proposal's "Persistence and resume" section. Honors audit A6: `StreamLogStore` and `StreamSnapshotStore` formats stay separate — this unifies **ownership**, not formats.

Status: complete. Close-out reverified 2026-07-07 on `origin/main` `fce7dcf74`; bullets landed across the Stage 3c PR train and remaining projector deletion work belongs to Stage 5 (#6968).

## Task

1. **`session.stores` facade with atomic lifecycle**: tab-delete removes `streamLogs` + `streamData` + `executions/{id}` together. Today `clearStream` still deletes only stream logs + stream data (`ProgressViewState.ts:322-333`, `StreamSnapshotStore.ts:603-608`), leaving `executions/{id}` and flow records behind. Goal entries are now forgotten by the extension and desktop delete paths (`ProgressStreamLifecycleController.ts:59-64`, `desktopAgentExecution.ts:915-916`); keep that as regression coverage rather than citing it as an open leak. **Landed in #7180 (`c932cc3`).**
2. **One ownership rule per fact**: flow record = resume SSOT; `streamData/` = display sidecars; `executions/config.json` = config. Delete the `conversation.json`/`todos.json` KV projections (`runToolUseFlow.ts:364-372` — they duplicate flow-record shared and are rewritten every node step).
3. **One `deriveResumability(executionId) → {resumable, cause}`** replacing the divergent predicates: `detectWaitingStreams.ts:59-67`, `followUpCommand.ts:48-60`, `SessionResumeRetrieval.ts:120-135`, `history.ts:292-311`, and `desktopStreamSnapshot.ts:144-150`.
4. **One `repairAfterRestart(session)`** shared by extension and desktop (today: two orchestrations in different orders, plus a duplicate desktop catch fallback, `desktopAgentExecution.ts:753-840`): status via the `restart-repair` cause + transcript group closure + outcome rewrite **together**. Repaired-to-WAITING streams get neutral group closure, not forced `'error'`. **This resolves #6938** (the catch-fallback path marking a resumable stream ERROR instead of WAITING) — verify against that issue's repro and close it with this stage.
[...]

PR #7513, opened 2026-07-07
fix(transcript): an empty source never wins conversation arbitration

LionSR:
Fast-follow to #7500 (merged before its last two cursor findings could land there). Part of #6966 bullet 2.

## The two findings (same root)

1. A present-but-**empty** `conversation.json` whose mtime is at/after the stream log won the freshness arbitration and hid a full sidecar transcript (`readConversation()` maps an empty array to `null` only AFTER arbitration had already committed to legacy).
2. With multiple log-backed sidecars sharing an `executionId`, the resolver's first pick in scan order won even when its log reconstructs empty while a sibling stream holds the real transcript.

## One protocol rule, not two patches

**An empty result never wins.** The mtime comparison (ties toward legacy, unchanged) now decides only the ORDER the two sources are tried in; a source that reads or reconstructs empty falls through to the next — symmetric in both directions (`legacy→sidecar` and `sidecar→legacy`) — until a source yields content or all are empty (`source: 'none'`).

Within the sidecar arm: the resolver's pick keeps #7433's ranking (merged; log-backed > workPlan-only > bare, scan order within a rank) as the primary criteria. Only when that pick reconstructs empty does the facade try the remaining **log-backed sibling streams** for the same execution, in the resolver's deterministic enumeration order — i.e. selection is "has-content first, then the existing criteria" as a stable sort. The has-content preference lives in `completedRunArchive`, not the generic resolver, because "reconstructs conversation content" is conversation-domain knowledge the resolver's other callers (todos reader, trace assembler) must not absorb. The sibling scan is a cold path: it runs only after an empty primary reconstruction.

§15 classification: L5 decide-once precedence owner (the facade remains the single arbitration point; no new fallback chains at call sites).

## Fixture evidence (both exact scenarios pinned)

- `never lets an empty-but-present legacy file beat a full sidecar`: empty `conversation.json` with mtime 60s AFTER the stream log, beside a full sidecar → `source: 'streamLog'`, full conversation.
- `falls through to the content-bearing sibling when the resolver picks an empty log`: two meta-matched, log-backed sidecars (`aChild@tool#id` with only `progressStatus` rows scans first; `zOrchestrator@…#id` holds the transcript), no legacy file → `source: 'streamLog'` with the sibling's streamId and messages.
- Existing cases (fresher-legacy-with-content wins, no-conversation-rows falls back to legacy, neither-source → `none`) still pass — content-bearing legacy behavior is unchanged.

## §14 accounting

`git diff --stat origin/main`: 2 files, **+205 / −25** (test +117). Exported symbols: 0 added / 0 deleted; 4 private helpers added inside the facade (`readLegacyConversationMessages`, `conversationFromStream`, `siblingConversationStreams`, `readSidecarConversation`), all ≥2-call-path or protocol-carrying. The legacy-mtime stat is now lazy (only taken when a sidecar exists to compare against).

## Verification

- root / test-kernel / cli / extension / desktop `tsc` clean.
- `vitest` green: `src/test-kernel/transcript/` (incl. the 9-case facade suite), `src/test-kernel/agent/storage/`, `src/test-kernel/agent/export/`, executions-tool suites, CLI history suites.

---

> [!NOTE]
> **Medium Risk**
> Changes precedence for completed-run conversation display/export (chat export and history) in edge cases with empty legacy projections or multi-stream sidecars; behavior is intentional but affects what users see for archived runs.
> 
> **Overview**
> **`readCompletedRunConversation`** no longer lets an empty legacy file or an empty resolver-chosen stream log “win” just because mtime or scan order favored them. **Mtime still only sets try order** (legacy first when legacy mtime ≥ sidecar, ties toward legacy); each arm must return **non-empty** content or the facade tries the other.
> 
> On the **sidecar** path, after the resolver’s primary stream, the facade can scan **other log-backed sibling streams** for the same `executionId` when the primary reconstructs no conversation-shaped messages—without changing the shared `resolvePersistedStreamIdForExecution` ranking used by todos/trace.
> 
> Two vitest cases pin the regressions: empty `conversation.json` with newer mtime vs full sidecar, and resolver picking an empty sibling log while another stream holds the transcript.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 695d232dbeb0886d2b21d8afacd34311de298999. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/transcript/completedRunArchive.vitest.ts
+++ b/src/test-kernel/transcript/completedRunArchive.vitest.ts
@@ -357,6 +357,66 @@ describe('completedRunArchive facade', () => {
     expect(sidecarWins.source).toBe('streamLog');
   });
 
+  it('never lets an empty-but-present legacy file beat a full sidecar (non-empty rule)', async () => {
+    const executionId = 'fff666fff666' as ExecutionId;
+    const streamId = 'orchestrator@deepseekproT#fff666fff666' as StreamTabId;
+    await writeSidecarFixture(executionId, streamId);
+    // Present but EMPTY legacy projection with an mtime after the stream
+    // log — under pure mtime arbitration this would win and hide the full
+    // transcript. The non-empty rule orders legacy first but falls through.
+    await getExecutionStore(executionId).writeConversation([]);
+    const legacyPath = await findFile(tempDirs.at(-1)!, 'conversation.json');
+    await setMtime(legacyPath, Date.now() + 60_000);
+
+    const result = await readCompletedRunConversation(executionId);
+    expect(result.source).toBe('streamLog');
+    expect(result.streamId).toBe(streamId);
+    expect(result.conversation).not.toBeNull();
+    expect(result.conversation!.length).toBeGreaterThan(0);
+  });
+
+  it('falls through to the content-bearing sibling when the resolver picks an empty log (non-empty rule)', async () => {
+    const executionId = '0777aa0777aa' as ExecutionId;
+    // Both streams are meta-matched and log-backed; the alphabetically
+    // first-scanned one holds only non-conversation rows, so the resolver's
+    // log-rank pick (#7433 criteria) reconstructs empty. No legacy file
+    // exists — the facade must fall through to the sibling that actually
+    // holds the transcript.
+    const emptyStream = 'aChild@tool#0777aa0777aa' as StreamTabId;
+    const fullStream = 'zOrchestrator@deepseekproT#0777aa0777aa' as StreamTabId;
+
+    const snapshots = new StreamSnapshotStore();
+    snapshots.setTaskState(emptyStream, taskState('bash'), executionId);
+    snapshots.setTaskState(fullStream, taskState('orchestrator'), executionId);
+    await snapshots.flush();
+
+    const logs = new StreamLogStore();
+    await logs.load();
+    logs.ensureStream(emptyStream);
+    logs.append(
+      emptyStream,
+      logRow(MESSAGE_TYPES.PROGRESS_STATUS, { text: 'Working...' }),
+    );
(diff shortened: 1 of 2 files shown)


LionSR:
Both findings fixed in 71f36e03c:

1. **Unnecessary meta.json reads for non-log-backed streams** (copilot) — `siblingConversationStreams` now pre-filters `persisted` to only log-backed candidates (`streamLogStore.has(streamId)`) before running the `pMap` meta scan, instead of reading every persisted stream's `meta.json` and filtering afterward. Also excludes `primary` from the candidate list up front, for the same reason.

2. **A single corrupted sibling shouldn't abort the whole scan** (claude, marked non-blocking) — wrapped `readPersistedExecutionId(streamId)` in a per-item `.catch()` that logs a warning and degrades that one candidate to a non-match, instead of letting `p-map`'s default `stopOnError: true` propagate a single sibling's read failure all the way up through `readCompletedRunConversation`. This mirrors `readMeta`'s own existing "malformed data degrades to absent" convention, just extended to genuine IO-level throws (which `readMeta`'s `tryRead` only catches for `SyntaxError`, not other error types) — logged loudly rather than silently, per the audit's "loud reads" theme.

Verification: `npm run typecheck` clean, `eslint` clean, `vitest run` on `completedRunArchive.vitest.ts` (7/7) and the full `transcript/` suite (78/78) — all passing, confirming the reordering doesn't regress the existing "falls through to content-bearing sibling" coverage. Given finding 2 was explicitly marked non-blocking and the fix is a small, directly-verifiable code change (a `.catch()` clause changing propagation), I didn't build a dedicated IO-throw reproduction test — happy to add one if requested.

Issue #7086, opened 2026-07-04
Emit the finalized assistant message into StreamLogStore with a stable id; delete the CLI synthetic-entry machinery


LionSR:
## Context

Audit B2's named upstream fix (`docs/proposals/tech-debt-audit-2026-07.md` §B2), verified feasible against `37d1211`. Not currently tracked by any sub-issue of #6950/#6951/#6952/#6953.

The CLI compensates for the transcript store lacking a guaranteed finalized assistant message:

- `appendAssistantTranscriptIfMissing` dedups the run-result fallback text against store-derived entries **by normalized text**, including a hand-coded `replaceAll('\\checkmark', '✓')` (`packages/cli/src/chat/tui/state/transcript.ts:36,46-96`) — needed because the recorder path applies replacement rules the raw `result.lastResponse` doesn't.
- Synthetics are spliced by `syntheticAfterSeq` (`subscribeStreamLog.ts:394-410`).
- Injection sites: `chatSessionController.ts:282` (WAITING turn boundary) and `:302` (tool-use final result).

## Verified insertion points

Production `StreamLogStore` writers are exactly two: `attachTranscriptRecorder` (`src/transcript/TexraTranscriptRecorder.ts`) and one legacy backfill (`ProgressViewState.ts:442`). The recorder's `stream.end` arm **already accepts an authoritative `finalText`** (`TexraTranscriptRecorder.ts:344`), and its `result` arm currently does nothing (`:385-389`). `ResultEvent` carries no `lastResponse` today (`src/agent/trace/events.ts:197`).

## Proposed

Either (a) extend `ResultEvent` with the final assistant text and have the recorder upsert a stable-id `MODEL_RESPONSE` entry when the log tail doesn't already contain it, or (b) guarantee flows always route final text through `trace.stream(...end{finalText})` where `assembly.lastResponse` is finalized (`ResponseCycleFlow.ts:376`). The mid-run WAITING boundary needs the same treatment (emit at turn boundary, not only terminal result).

Then delete `appendAssistantTranscriptIfMissing`, the `\checkmark` normalization, and the `syntheticAfterSeq` splice (~120 LoC of the most fragile CLI state code). Net-negative LoC; also removes the text-normalization mismatch at the source.

## Coordination

Touches the trace vocabulary — if extending `ResultEvent`, align with #6951's single-ownership rules (one emit rail; the fact decided once at the flow boundary and carried as data).

## Acceptance

[...]

PR #7601, opened 2026-07-08
fix(transcript): emit finalized assistant text with a stable id, delete CLI synthetic-entry machinery

LionSR:
## Summary

The CLI transcript worked around a gap in the transcript store by re-deriving the last assistant response from the run result and appending it as a synthetic entry whenever the store didn't already seem to show it — deduped by normalized text, including a hand-coded `\checkmark` <-> checkmark-glyph alias. That hack existed because the store's live `MODEL_RESPONSE` stream writes raw provider text in real time, while the run result's text has already gone through `extractResponse`'s replacement-rule cleanup, so the two copies of "the same" text could differ trivially and defeat exact-match dedup.

This fixes the root cause instead of continuing to compensate for it at the CLI edge, and deletes the compensating CLI machinery entirely (net-negative production LoC, per the issue).

## Details

- Added a new first-class trace fact, `response.finalized` (`ResponseFinalizedEvent` in `src/agent/trace/events.ts`, `AgentTrace.responseFinalized()`), decided once at the flow boundary where `assembly.lastResponse` is set.
- `ToolUseProcessNode.post()` (`src/agent/core/flows/toolUseRound/ToolUseProcessNode.ts`) now calls `logger.responseFinalized(execRes.text)` at that exact point — which fires both at a mid-run WAITING turn boundary (the tool-use loop pausing for the next user message) and at the terminal round, since both go through the same `endTurn` branch. Gated on `useStreaming` so non-streaming rounds (which already log their own formatted `MODEL_RESPONSE` line in `exec()`) don't get a duplicate.
- `TexraTranscriptRecorder` tracks the id of the round's own `MODEL_RESPONSE` stream (if one was opened) and, on `response.finalized`, upserts that entry to the authoritative text; if the round never streamed, it appends a fresh entry instead. Correlation is by id, never by comparing text, and is reset at every round boundary so an earlier round's stream id can never leak into a later round's finalize.
- Deleted the CLI's compensating machinery entirely: `appendAssistantTranscriptIfMissing`, the checkmark-alias dedup helper, the `syntheticKind: 'final'` entry kind, and the associated splice/dedup branch in `subscribeStreamLog.syncStreamLog`. `projectStreamTranscript`'s `fallbackAssistant` option and both call sites in `chatSessionController.ts` (WAITING boundary + terminal result) are gone.
- Deliberately scoped to `ToolUseProcessNode` (the tool-use round flow that the CLI, extension, and desktop chat surfaces all share) and left `ResponseCycleFlow` (reflection/workflow) untouched — that flow always streams `phaseOnly` by design (output is surfaced via export/extraction, not the live transcript), so applying the same upsert there would leak content a workflow deliberately withholds.

## Validation

- `npx tsc --noEmit` (root), `npx tsc --noEmit -p tsconfig.test-kernel.json`, and `packages/cli`'s own `tsc --noEmit -p tsconfig.json` — all clean.
- `npx eslint` on every touched file — 0 errors (3 pre-existing import-order warnings, unrelated to this diff).
- `npx prettier --check` on every touched file — clean.
- `npm test` (full suite) — 5241 passed, 4 skipped (pre-existing skips).
- New/updated coverage:
  - `src/test-kernel/agent/trace/AgentTrace.vitest.ts` — `responseFinalized` emits the event and stamps the ambient stage id.
  - `src/test-kernel/transcript/TexraTranscriptRecorder.vitest.ts` — upserts the round's streamed entry to the authoritative text, appends fresh when the round never streamed, and does not let an earlier round's stream id leak into a later round.
  - `src/test-kernel/agent/core/ToolUseProcessNodeResponseFinalized.vitest.ts` — `responseFinalized` fires only for a streamed, turn-ending round; not for non-streaming or mid-turn (tool-continuing) rounds.
  - `src/test-kernel/cli/TuiStateAndFocus.vitest.mts` — replaced the deleted synthetic-fallback tests with equivalents exercising the real mechanism (streamed-then-finalized reconciliation, no-stream append, round-boundary isolation, zero synthetic entries on projection).
- Production code net LoC: -18 (99 insertions, 117 deletions across CLI + trace/recorder files); test code grew (+108) to cover the new mechanism at three levels instead of just the deleted workaround.

Refs #7086

---

> [!NOTE]
> **Medium Risk**
> Touches shared agent trace, transcript persistence, and CLI projection paths used on every chat turn; behavior is gated and well-tested but regressions could show as missing or duplicate assistant lines.
> 
> **Overview**
> Fixes **#7086** by making the transcript store own authoritative assistant text instead of the CLI guessing from run results.
> 
> Adds a **`response.finalized`** trace event and **`AgentTrace.responseFinalized()`**. **`ToolUseProcessNode.post()`** emits it when a round **ends the turn** and used **streaming** (non-streaming rounds already log `MODEL_RESPONSE` in `exec()`).
> 
> **`TexraTranscriptRecorder`** upserts the round’s `MODEL_RESPONSE` row by stream id when streaming happened, or appends a new row when it did not; round boundaries reset correlation so ids don’t leak across turns.
> 
> The CLI **removes** `appendAssistantTranscriptIfMissing`, `syntheticKind: 'final'`, text-based dedup in stream sync, and **`projectStreamTranscript`’s `fallbackAssistant`** (including **`chatSessionController`** idle/terminal projection). Turn boundaries now rely on **sync + finalize** from the store.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9b173932d8fa3f34a2d784437bfcc68954e55c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -273,7 +273,6 @@ export function createChatSessionController(
     const { runtimeHost, approvalsUnavailable, finalize } =
       setupRunHost(sessionContext);
     const executionId = generateExecutionId();
-    let waitingTurn = 0;
     let executionRegistered = false;
     let agentSettled = false;
     session.executionId = executionId;
@@ -293,15 +292,9 @@ export function createChatSessionController(
             activeStreamId.set(resolvedStreamId);
             if (session.stopRequested) interruptActiveRun();
           },
-          onIdle: (lastResponse) => {
+          onIdle: () => {
             if (!session.streamId) return;
-            projectStreamTranscript(session.streamId, {
-              fallbackAssistant: {
-                text: lastResponse,
-                idPrefix: `waiting:${executionId}:${waitingTurn++}`,
-              },
-              finalize: true,
-            });
+            projectStreamTranscript(session.streamId, { finalize: true });
           },
         });
       })
@@ -312,17 +305,7 @@ export function createChatSessionController(
           sessionContext,
         );
         if (result.streamId) {
-          projectStreamTranscript(result.streamId, {
-            finalize: true,
-            ...(result.category === AgentCategory.ToolUse
-              ? {
-                  fallbackAssistant: {
-                    text: result.lastResponse,
-                    idPrefix: `final:${result.executionId}`,
-                  },
-                }
-              : {}),
-          });
+          projectStreamTranscript(result.streamId, { finalize: true });
         }
         notify({ kind: 'agentFinished' });
       })
--- a/packages/cli/src/chat/tui/state/cliState.ts
+++ b/packages/cli/src/chat/tui/state/cliState.ts
@@ -46,7 +46,7 @@ interface ConversationEntryBase {
   /** Entry was synthesized by the CLI and is not present in StreamLogStore. */
   readonly synthetic?: boolean;
   /** Why the CLI synthesized this entry. */
-  readonly syntheticKind?: 'final' | 'local' | 'process';
+  readonly syntheticKind?: 'local' | 'process';
   /** StreamLog head at the moment a synthetic entry was appended. */
   readonly syntheticAfterSeq?: number;
 }
--- a/packages/cli/src/chat/tui/state/subscribeStreamLog.ts
+++ b/packages/cli/src/chat/tui/state/subscribeStreamLog.ts
@@ -398,22 +398,13 @@ export function syncStreamLog(streamId: StreamTabId): void {
       if (!rendered) continue;
       logCandidates.push({ rendered, sortSeq: entry.seqNo, tieBreak: 0 });
     }
-    // The synthetic loop's duplicate check scans `logCandidates`, so synthetics
-    // push into a copy; without synthetics the log list is used as-is.
+    // Synthetic (local/process) rows are positioned by `syntheticAfterSeq`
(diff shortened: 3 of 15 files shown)


LionSR:
**Decision (2026-07-09): merge-priority; land this before #7698.**

This implements the direction both audits decided (issue #7086 / micro-audit A3): fix the missing stable id at the source and delete the CLI's synthetic-entry machinery — retiring an R1 dual-state and the most fragile CLI state code (dedup-by-normalized-text incl. the checkmark-glyph alias). Green at 13 checks.

Merge-order note: #7698 (delete TUI runtime host wrapper) touches the adjacent `subscribeRuntimeHost`/CLI-state territory — land this fix-class PR first, then rebase #7698 onto it rather than the reverse.

LionSR:
Rebased onto main at cf138f802+ (one conflict: `transcriptProjection.ts` — main's `TranscriptFallbackAssistant` interface was added for the exact fallback path this PR deletes, and this PR already rewrites both of its callers in `chatSessionController.ts`; took the deletion side). Verified post-rebase: 128 CLI suites / 1,614 tests green, transcript+agent-core suites green, typecheck clean, zero fallback residue by grep. Approved + green — ready to land.

PR #7859, opened 2026-07-09
fix(ci): isolate runtime validation credentials

### LINKED ISSUE #7858: CLI runtime validation depends on local provider credentials
## Summary

The guarded CLI runtime validator inherits provider credentials from the invoking environment. It passes on developer machines with configured keys but fails in clean CI before the internal validation model handler can take over.

## Evidence

Main CI run https://github.com/LionSR/TeXRA/actions/runs/29046706816 failed in `runtime validation (linux)` with:

> Model "deepseekproT" is not available in the active API mode (missing api key). No models are currently available.

All other main validation jobs passed.

## Expected behavior

`pnpm --filter @texra-ai/cli validate:run` should be self-contained under its guarded internal validation mode and should never depend on or expose real provider credentials.

LionSR:
## Summary

Make the guarded CLI real-runtime validator self-contained in clean CI. Every subprocess using the internal validation model now receives fake credentials for the existing provider-key list, so model availability resolution succeeds without inheriting developer secrets and the internal handler remains the only model boundary used.

## Issue acceptance gates

Closes #7858.

- The validator no longer depends on configured provider keys.
- Guarded validation subprocesses cannot inherit real provider credentials.
- The full runtime validation passes with all provider keys unset and an isolated home directory.

## Single source of truth

`validationProviderApiKeyEnv` remains the provider credential inventory. `VALIDATION_FAKE_API_KEY` owns the placeholder value, and `validationModelProviderEnv` derives the guarded subprocess environment from both.

No provider list or fake credential literal is duplicated. One derived environment object owns fake-key injection for every guarded validation subprocess.

## Net elements (R6)

- Files: 1 modified
- Exported symbols: 0
- Classes/interfaces/enums: 0
- LoC: 6 insertions, 1 deletion, net +5

## Consumer counts (R8)

n/a — no emit path or export changed.

## Host impact

Extension: None.

Desktop: None.

CLI: Validation harness only; production bundles retain the existing guarded internal-handler behavior.

## Scheduled refactoring

None.

## Validation

- Credential-free `corepack pnpm --filter @texra-ai/cli validate:run` with all ten provider keys unset and fresh `HOME`/XDG directories — passed before and after review fixes.
- GitHub Actions `runtime validation (linux)` in manually dispatched exact-SHA CI run `29051110260` — passed on the functional patch.
- `corepack pnpm exec prettier --check packages/cli/scripts/validate-run.mjs` — passed.
- `git diff --check` — passed.

### DIFF
--- a/packages/cli/scripts/validate-run.mjs
+++ b/packages/cli/scripts/validate-run.mjs
@@ -34,6 +34,7 @@ const validationEnv = 'TEXRA_INTERNAL_VALIDATE_MODEL_HANDLER';
 const validationFlagEnv = 'TEXRA_INTERNAL_VALIDATE_MODEL_HANDLER_FLAG';
 const validationFlagContent = 'texra-cli-run-validation\n';
 const validationBundleMarker = validationFlagContent.trim();
+const VALIDATION_FAKE_API_KEY = 'texra-validation-fake-key';
 const ESC = String.fromCharCode(27);
 const validationProviderApiKeyEnv = [
   'OPENAI_API_KEY',
@@ -47,6 +48,9 @@ const validationProviderApiKeyEnv = [
   'MINIMAX_API_KEY',
   'GLM_API_KEY',
 ];
+const validationModelProviderEnv = Object.fromEntries(
+  validationProviderApiKeyEnv.map((name) => [name, VALIDATION_FAKE_API_KEY]),
+);
 
 function isolatedCliHomeEnv(home, overrides = {}) {
   return {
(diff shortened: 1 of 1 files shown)


LionSR:
Independent verification (session babysitter): a diagnostic agent reproduced the runtime-validation failure key-less (isolated HOME/XDG, no provider env keys) at main 8c105bcfb, root-caused it identically — the run-path validators never injected fake keys, latent since validate-run.mjs's inception, first exercised key-less when #7743 wired the job into CI — and validated this exact mechanism (fake provider keys for the guarded subprocesses) end-to-end: full build + `validate:run` green in a completely key-less environment, typecheck clean, ModelFactoryRouting/ComputeModelOptions 48/48. `DEEPSEEK_API_KEY` is in validationProviderApiKeyEnv, covering the CLI_BUILTIN_DEFAULT_MODEL (deepseekproT) path that fails in CI. This central +7 injection in run()'s validationModel branch is cleaner than the per-call-site variant we built and discarded. Marking ready and merging to un-redden main's runtime-validation job.

PR #7887, opened 2026-07-10
fix(cli): cancel stale auto-resume after Ctrl-C

### LINKED ISSUE #7886: CLI auto-resume can restart after Ctrl-C
## What happened

A queued CLI follow-up can auto-resume a stream after the user presses Ctrl-C during resume preparation.

On current `origin/main` `c43073a64`, `tryResumeStream`:

1. checks `session.stopRequested`,
2. resets it to `false`,
3. awaits `setCliHelperModel()`, and
4. calls `resolveAndResumeStream()` without another cancellation guard.

If Ctrl-C lands during model setup, `stop()` sets `stopRequested = true` and marks the stream cancelled, but the pending wake continues afterward. The shared stream transition table permits an explicit `CANCELLED -> RUNNING` resume, so status alone does not reject the stale wake.

## Why this hurts

Ctrl-C is the user's explicit stop action. A delayed auto-resume that restarts after the stop makes the CLI feel unsafe and can continue model/tool work the user believes was cancelled.

## Reproduction

1. Put a CLI tool-use stream into `WAITING`.
2. Queue a follow-up so the wake path calls `tryResumeStream`.
3. Delay `setCliHelperModel()`.
4. Press Ctrl-C while model setup is pending.
5. Release model setup.
6. Observe that `resolveAndResumeStream` still launches the resume.

## Expected

A stop request cancels resume preparation across every async boundary. The queued message may remain durable, but the cancelled wake must not restart the stream.

## Smallest clean fix

Pass a cancellation predicate into the shared resume orchestrator and check it before and after asynchronous resume-state retrieval. The CLI can provide `() => session.stopRequested`, closing both the helper-model setup window and the persisted-state retrieval window without adding another status projection.

LionSR:
## Summary

- Let hosts provide a live cancellation predicate to the shared resume orchestrator.
- Wire CLI `session.stopRequested` into auto-resume preparation.
- Recheck cancellation immediately before launching a recovered tool-use or workflow resume.

Closes #7886.

## Issue acceptance gates

- [x] Ctrl-C during helper-model setup prevents the queued wake from resuming.
- [x] Cancellation during persisted resume retrieval prevents launch.
- [x] Cancelled preparation does not call tool-use or workflow resume ports.
- [x] Existing active-stream and duplicate-resume guards remain unchanged.

## Single source of truth

`TuiSession.stopRequested` remains the CLI owner of the user's stop decision. `ResumeStreamPorts.isCancellationRequested` exposes that live decision to `resolveAndResumeStream`, which owns the final pre-launch guard after asynchronous preparation.

No cancellation state or status projection was added. One optional predicate crosses the existing host/runtime boundary; extension and desktop callers remain unchanged because the hook is optional.

## Net elements (R6)

5 files changed; 0 files added/deleted; 58 insertions, 1 deletion, net +57 LoC. The net addition is one optional interface member, two race-focused regression tests, and the changelog entry. No exports, classes, enums, or schemas were added or removed.

## Consumer counts (R8)

n/a — no emit path, event key, or export was deleted or rerouted.

## Host impact

Extension: No behavior change; no cancellation predicate is supplied.

Desktop: No behavior change; no cancellation predicate is supplied.

CLI: Ctrl-C now cancels an automatic resume even when it arrives during asynchronous model or session-state preparation.

## Scheduled refactoring

None.

## Validation

- Focused Vitest: 2 files, 28 tests passed.
- `npm run typecheck` passed.
- `npm run lint` passed with 0 errors and 51 existing warnings.
- `git diff --check` passed.
- GitHub CI `validate (linux)` passed in 18m13s.
- The timing-sensitive window is covered deterministically with deferred helper-model setup and cancellation during resume retrieval rather than a flaky wall-clock PTY test.

---

> [!NOTE]
> **Low Risk**
> Small, optional port hook with CLI-only wiring; existing active-stream and in-flight guards stay the same.
> 
> **Overview**
> **Ctrl-C during CLI auto-resume preparation** no longer restarts a stream after helper-model setup or persisted resume lookup finishes.
> 
> The shared `resolveAndResumeStream` orchestrator accepts an optional **`isCancellationRequested`** hook on `ResumeStreamPorts`. It is checked at entry and again after async resume retrieval, so a cancelled preparation never calls tool-use or workflow resume ports. The CLI chat controller wires **`session.stopRequested`** into that hook for queued follow-up wakes.
> 
> Extension and desktop callers are unchanged because the predicate is optional. Regression tests cover cancellation during retrieval and stop while `setCliHelperModel` is in flight.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2e021d3430f03ea133a29fcbb31cf577c36b3d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,10 @@ All notable changes to this project will be documented in this file.
   progress, command results, doctor output, and structured error/instruction
   records now share one queued stdout writer.
 
+- **Ctrl-C cancels pending automatic resumes** — stopping while a queued
+  follow-up is preparing can no longer restart the cancelled stream after an
+  asynchronous model or session-state lookup finishes.
+
 ### Desktop
 
 #### Bug Fixes
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -473,6 +473,7 @@ export function createChatSessionController(
           resolveAndResumeStream(streamId, {
             runtimeHost,
             streamStatus: defaultSession().status,
+            isCancellationRequested: () => session.stopRequested,
             resolveResumeState: async () => ({
               runState: config,
               executionId,
--- a/src/agent/runtime/resolveAndResumeStream.ts
+++ b/src/agent/runtime/resolveAndResumeStream.ts
@@ -27,6 +27,8 @@ interface ResolvedResumeState {
 export interface ResumeStreamPorts {
   /** Runtime host receiving the RESUMING/WAITING status updates. */
   readonly runtimeHost: AgentRuntimeHost;
+  /** True when the host cancelled this resume while it was preparing. */
+  isCancellationRequested?(): boolean;
   /**
    * The status machine of the session that owns this stream — per-window on
    * desktop, the default session's machine in the extension/CLI. The
@@ -87,6 +89,7 @@ export async function resolveAndResumeStream(
   ports: ResumeStreamPorts,
 ): Promise<boolean> {
   if (
+    ports.isCancellationRequested?.() ||
     ports.streamStatus.isActiveOrResuming(streamId) ||
     resumeInFlight.has(streamId)
   ) {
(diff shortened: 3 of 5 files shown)


LionSR:
Superseded by #7893, which is now merged and closes #7886 with cancellation checks at each shared resume boundary plus CLI coverage.

PR #7888, opened 2026-07-10
refactor(latex): unify workspace source resolution

### LINKED ISSUE #7469: Decide-once the two remaining re-derive resolvers (execution→stream reverse-scan; dual resolveWorkspaceSourceDir)
Medium (M5, #6951 decide-once), from the 2026-07-07 audit (verified):

- `src/transcript/executionStreamResolver.ts` `resolvePersistedStreamIdForExecution` reverse-scans every `streamData/*/meta.json` on hot paths to re-derive execution→stream, a fact `registerExecution` could persist as data.
- `resolveWorkspaceSourceDir` is implemented twice with divergent heuristics (`compileCheck.ts` vs `LatexDiffManager.ts`), re-deriving a fact destroyed at location-creation time; persist the source dir on the location record.

Fix: decide-once-carry-as-data for both. Sequence after Stage 3c's persistence facade work where the write points live (#6966).

LionSR:
## Summary

Collapse the two workspace-source-directory derivations in compile checks and latexdiff into one exported resolver owned by `compileCheck.ts`. The resolver maps workspace and run-storage locations to the live workspace source folder; caller-specific result shape and external/no-workspace fallback remain at their callers.

The consolidation also removes the old path-shape ambiguity: only `runStorage` locations strip a leading `r<N>/` workflow round segment. A genuine workspace directory named `r1`, `r2`, and so on remains a real directory.

## Issue acceptance gates

Closes #7469. Item 1 was previously completed by #7551; this PR completes the remaining dual-`resolveWorkspaceSourceDir` item under the maintainer decision recorded on 2026-07-09.

- Exactly one production resolver definition remains.
- Compile check still prefers a valid workspace `OutputFileInfo.source`, then falls back to the output location.
- Latexdiff still uses the reference file's absolute parent for external/no-workspace locations.
- Run-storage round prefixes are stripped; workspace folders with the same spelling are retained.

## Single source of truth

`resolveWorkspaceSourceDir(FileLocation)` in `src/agent/output/compileCheck.ts` solely owns the workspace-root mapping decision.

The 31-line private `LatexDiffManager.resolveWorkspaceSourceDir` implementation is deleted. No module, options object, schema field, compatibility export, or persisted migration is introduced.

## Net elements (R6)

- Files: 4 modified, 0 added, 0 deleted.
- LoC: +190 / -71, net +119; production code is net negative, while characterization/regression coverage expands substantially.
- Exports: one existing resolver is made public within the internal agent-output surface.
- Classes/interfaces/enums: unchanged.

## Consumer counts (R8)

Two production consumers remain: compile check and latexdiff PDF compilation. Both call the one resolver; repo search shows exactly one definition.

## Host impact

Extension: shared workflow compile/diff behavior uses the consolidated resolver.

Desktop: same shared behavior.

CLI: same shared behavior.

## Scheduled refactoring

None. The other #7469 item was completed by #7551.

## Validation

- `corepack pnpm vitest run src/test-kernel/agent/output/LatexCompileInputDirs.vitest.ts` (8 tests)
- related compile/latexdiff focused suites during the first pass (16 tests across available files)
- targeted ESLint
- `git diff --check`
- repo search confirming one resolver definition

---

> [!NOTE]
> **Medium Risk**
> Touches workflow LaTeX compile and diff input paths across all hosts; behavior change is intentional for edge-case folder names but could affect builds that previously resolved inputs incorrectly.
> 
> **Overview**
> Workflow **compile checks** and **latexdiff PDF builds** now share a single exported `resolveWorkspaceSourceDir` in `compileCheck.ts`; the duplicate private resolver in `LatexDiffManager` is removed.
> 
> The shared logic only strips a leading `r<N>/` segment for **run-storage round outputs** (detected via `parseWorkflowOutputRoundDir` on the path under the run directory). Workspace files and `original/` snapshots keep a top-level folder literally named `r1`, `r2`, etc., so those paths map to the correct live source directory instead of being treated as workflow round prefixes.
> 
> Compile checks still prefer a workspace `outputFile.source` when it resolves, then fall back to the output location for `extraInputDirs`. Latexdiff still prepends the revised round directory and uses the shared resolver for the reference file, falling back to the reference file’s parent for external paths or when no workspace is open.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73ef541078173b8e79a7b5065d7a4eb66d6ca693. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 #### Bug Fixes
 
+- **Workflow LaTeX compilation resolves source directories consistently** —
+  compile checks and latexdiff PDF builds now share one workspace-source
+  resolver, and a real workspace folder named `r1`, `r2`, and so on is no
+  longer mistaken for a run-storage round folder.
 - **OpenRouter transient server errors no longer stall a request for up to an
   hour** — the OpenRouter SDK's built-in retry window is now capped at 30
   seconds, so a persistent 5XX surfaces through TeXRA's visible retry/failure
--- a/src/agent/output/LatexDiffManager.ts
+++ b/src/agent/output/LatexDiffManager.ts
@@ -391,6 +391,8 @@ export class LatexDiffManager {
       sourceLocation.kind === 'runStorage'
         ? path.dirname(sourceLocation.absolutePath)
         : null,
+      // Snapshot bases live under `original/`, while between-round bases live
+      // under `r<N>/`; map either back without confusing a real `r<N>` folder.
       resolveWorkspaceSourceDir(referenceLocation) ??
         path.dirname(referenceLocation.absolutePath),
     ].filter((dir): dir is string => dir !== null);
--- a/src/agent/output/LatexDiffManager.ts
+++ b/src/agent/output/LatexDiffManager.ts
@@ -16,14 +16,12 @@ import {
 } from '@shared/schemas';
 import { WorkspaceStateKey } from '@shared/state/stateKeys';
 import { LATEX_CONFIG_RANGES } from '@shared/constants/latex';
-import { parseWorkflowOutputRoundDir } from '@shared/constants/workflowOutput';
 import {
   createExternalLocation,
   createRunStorageLocation,
   createWorkspaceLocation,
   FlexibleFS,
   TaskRunFileService,
-  WorkspaceFS,
 } from '@utils/files';
 import { checkToolInstalled } from '@utils/system';
 import { getComparablePath } from '@utils/files/taskRunStorage';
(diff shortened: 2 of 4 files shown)


LionSR:
Addressed the snapshot ambiguity in 73ef54107. The resolver now examines the location beneath the execution directory: only physical r<N>/ round outputs lose that segment, while original/r<N>/... snapshots retain the genuine workspace directory. Added the snapshot regression and restored the latexdiff invariant comment. Focused tests (8), workspace typecheck, targeted ESLint, formatting, and diff checks pass.

PR #7825, opened 2026-07-09
Honest-settings trio: wire desktop auto-open-PDF, VS Code skills.enabled, desktop Goals (FS3+FS5+FS6)

LionSR:
## What / why

Fixes #7751 (host-parity audit FS3+FS5+FS6, `docs/proposals/host-parity-audit-2026-07.md`, PR #7746). Three independent advertised-but-inert settings surfaces, each wired per its section's stated "preferred" direction (wire, since the plumbing already exists on each host).

### FS3 — desktop `WORKFLOW_AUTO_OPEN_PDF` toggle no-ops
`WORKFLOW_AUTO_OPEN_PDF`'s schema advertises `hosts: ['vscode', 'desktop']`, but `DesktopProgressBridge.handleInteractionEvent` fell through `default: return` for `requestOpenFile` — the event `OutputNode` emits for both "open new output files" and the post-compile PDF/log open. Desktop already owns an `openPath` preview-with-fallback host (the same one `runExecution`'s `openWorkflowOutput` uses), so `requestOpenFile` now routes `location.absolutePath` through it, with a warn-log on failure instead of a silent drop.

### FS5 — `texra.skills.enabled` is an advertised VS Code setting that does nothing
`userVars.ts` gates `loadRuntimeSkillCatalog` on `getConfig('texra.skills.enabled', true)`, but the extension never called `setRuntimeSkillSources` — so `loadRuntimeSkillCatalog` always saw zero sources regardless of the toggle, an already-no-op. CLI and desktop both wire this at their composition root via `nodeHost.ts`'s `initializeNodeRuntimeSkills` (`defaultSkillSources` + `setRuntimeSkillSources`). `extension.ts`'s `activate()` now does the same, alongside `registerAgentFeatures()`.

Desktop's own skill-sources wiring is untouched — that was already decided and landed separately per #7692's closure (desktop gained runtime skills there); this issue's FS5 scope is the VS Code extension only.

[...]

PR #7900, opened 2026-07-10
fix(desktop): resolve revealStream category from persisted state, not session-only kind

### LINKED ISSUE #7851: DesktopProgressBridge.revealStream widens category filter using ephemeral session state
## Context

Follow-up from #7825 (Honest-settings trio: wire desktop auto-open-PDF, VS Code skills.enabled, desktop Goals (FS3+FS5+FS6)).

Flagged by `claude[bot]`'s review during #7825; not addressed before merge (noted at review time as "not feature-breaking on its own" but a worthwhile alignment).

`DesktopProgressBridge.revealStream()`'s category-filter widening reads `this.state.getStreamState(streamId)?.kind` — ephemeral, session-only state populated lazily only once a stream has emitted a live fact this session. For a goal-owned stream restored from persisted `workspaceState` that hasn't emitted any facts yet this session, `category` is `undefined`, so the filter unconditionally resets to `'all'` even when the stream's real (persisted) category would already satisfy the current filter. The extension's equivalent (`ProgressViewProvider.revealStream`) avoids this by resolving category through `buildStreamInfo`, which reads persisted config/hints instead of live-session state.

## Details

Align desktop's `revealStream` filter-visibility check with `buildStreamInfo`'s persisted-data approach (or otherwise source `category` from persisted state) instead of reading `getStreamState()`.

## References

- PR: #7825
- Review thread: https://github.com/LionSR/TeXRA/pull/7825#discussion_r3554313701

LionSR:
## What / Why

Fixes #7851.

`DesktopProgressBridge.revealStream()` read `this.state.getStreamState(streamId)?.kind`
to decide whether to widen the current category filter — that map is
ephemeral, session-only state populated lazily only once a stream has
emitted a live fact this session. For a goal-owned stream restored from
persisted `workspaceState` that hasn't emitted any facts yet this session,
`category` read as `undefined`, so the filter-visibility check
unconditionally reset the filter to `'all'` even when the stream's real
(persisted) category already satisfied the current filter.

This aligns desktop's `revealStream` with the extension's
`ProgressViewProvider.revealStream`, which resolves category through
`buildStreamInfo` — reading persisted `AgentConfig`/hints instead of
live-session state — so the filter-widening check now returns the same
answer on both hosts for a stream that hasn't emitted a fact yet this
session.

Flagged by `claude[bot]`'s review during #7825 (not addressed before merge,
per the issue).

## Net elements (R6)

- +1 import (`buildStreamInfo`, already exported and used by the extension's
  equivalent — no new abstraction).
- Net function body: -1 line (one `getStreamState()` lookup + a 2-line
  condition collapse into a single `buildStreamInfo(...) === null` check,
  the same pattern `ProgressViewProvider.revealStream` already uses).
- +1 regression test.

## Consumer counts (R8)

n/a — no helper deleted or re-routed; `buildStreamInfo` already has two
call sites in the extension's `streamInfoUtils.ts` module plus its own
internal `buildStreamInfos`, and desktop's `this.state` is the same
host-neutral `ProgressViewState` type it already expects.

## Testing

- `npx vitest run src/test-kernel/desktop/DesktopAgentExecution.vitest.mts` — 70/70 pass.
- New test `revealStream keeps a matching filter for a restored stream with no live session facts yet (issue #7851)` fails against the pre-fix code (`agentFilter: 'all'` instead of `'toolUse'`) and passes against the fix — verified by manually reverting the source change and re-running.
- `corepack pnpm --filter @texra/desktop typecheck` — clean.
- `npm run typecheck` (root: `src/` + test-kernel + CLI + extension + trace-viewer) — clean.

---

> [!NOTE]
> **Low Risk**
> Small behavioral fix in desktop progress navigation with extension parity and a targeted regression test; no auth or data-path changes.
> 
> **Overview**
> Fixes **#7851**: when jumping from Goals to a run, desktop **`revealStream`** no longer widens the progress rail filter to **`all`** just because the stream has not emitted live session facts yet this launch.
> 
> **`DesktopProgressBridge.revealStream`** now uses **`buildStreamInfo`** (persisted run config / stream hints), matching the extension’s **`revealProgressStream`**, instead of **`getStreamState(streamId)?.kind`**, which stays empty for restored “ghost” streams until facts arrive. The filter is reset only when **`buildStreamInfo(..., filter) === null`** (stream does not match the active category).
> 
> Adds a regression test for a snapshot-hydrated **toolUse** stream with **`toolUse`** filter already set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ab029c87467f37830bdeb9380c46dbc2d2a0955. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/desktopAgentExecution.ts
+++ b/packages/desktop/src/main/desktopAgentExecution.ts
@@ -14,6 +14,7 @@ import {
   type ProgressBackendInteractionPayloads,
 } from '@controllers/progressView/backend/events/ProgressInteractionHandler';
 import { ProgressBackend } from '@controllers/progressView/backend/ProgressBackend';
+import { buildStreamInfo } from '@controllers/progressView/backend/streamInfoUtils';
 import { hydrateProgressViewInquiries } from '@controllers/progressView/backend/externalInquiryHydration';
 import {
   buildApprovalRequestHandlerSet,
@@ -1087,14 +1088,19 @@ export class DesktopProgressBridge {
    * Mirrors the extension's `revealProgressStream` for the desktop Settings
    * Goals panel (issue #7751 FS6) so jumping from a goal entry to its owning
    * run works the same way on both hosts.
+   *
+   * Resolves category via `buildStreamInfo` (persisted config/hints), not
+   * `getStreamState()`'s ephemeral session-only kind, so a goal-owned stream
+   * restored from `workspaceState` that hasn't emitted a live fact yet this
+   * session still matches the current filter instead of unconditionally
+   * resetting it to 'all' (#7851).
    */
   revealStream(streamId: StreamTabId): void {
     if (!this.streamLogs.has(streamId)) {
       return;
     }
     const filter = this.state.agentCategoryFilter;
-    const category = this.state.getStreamState(streamId)?.kind;
-    if (filter !== 'all' && filter !== category) {
+    if (buildStreamInfo(this.state, streamId, filter) === null) {
       this.state.agentCategoryFilter = 'all';
     }
     this.routeToProgress();
--- a/src/test-kernel/desktop/DesktopAgentExecution.vitest.mts
+++ b/src/test-kernel/desktop/DesktopAgentExecution.vitest.mts
@@ -1821,6 +1821,48 @@ describe('DesktopProgressBridge', () => {
     }
   });
 
+  it('revealStream keeps a matching filter for a restored stream with no live session facts yet (issue #7851)', async () => {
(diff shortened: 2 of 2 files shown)


LionSR:
Merged current main after #7902 in 5ab029c87. The combined revealStream now awaits restart repair before checking streamLogs, then resolves filter visibility from persisted config/hints. Both regression tests are retained; the desktop execution suite passes (71 tests), with format and diff checks clean.

PR #7904, opened 2026-07-10
fix(setup): verify relay token auth status

### LINKED ISSUE #7891: Verify relay-token status before reporting authenticated in default setup platform
## Context

Follow-up from #7841 (feat(setup): derive default platform from shared ports).

`createDefaultSetupPlatform()`'s `defaultAuthStatus()` (`src/tools/setup/platform.ts`) now backs `probe_environment` / `verify_setup` auth reporting on CLI and desktop. It calls `SupabaseClient.isAuthenticated()`, which treats an uncached `TEXRA_RELAY_TOKEN` as authenticated without verifying it hasn't been revoked or expired. The existing CLI status path verifies the token before reporting status; this new default adapter does not, so a stale/revoked relay token can make setup report `researcherAccess.authenticated: true` ("signed in") while the credential check (`anyUsableCredentialExists`) denies usable access — surfacing a misleading "signed in" state instead of prompting the user to re-authenticate.

This was flagged during PR review (2026-07-10) and acknowledged as non-blocking-but-worth-attention; it was not addressed before merge.

## Details

- Verify relay-token validity (matching whatever the existing CLI status path already does) before returning `authenticated: true` from `defaultAuthStatus()`, or delegate to the existing host auth-status helper instead of calling `SupabaseClient.isAuthenticated()` directly.
- Add coverage for the case where a relay token is present but revoked/expired, confirming `probe_environment`/`verify_setup` report unauthenticated rather than a false "signed in".

## References

- PR: #7841

LionSR:
## Summary

- Verify configured relay tokens through the canonical status path before setup reports authentication.
- Sequence setup authentication before credential readiness so both results consume the same settled relay status.
- Preserve fallback to an authenticated Supabase session and API-key-only setup.

Closes #7891.

## Issue acceptance gates

- [x] Revoked and expired relay tokens report unauthenticated.
- [x] Credential readiness cannot race ahead and accept a rejected relay token.
- [x] A valid Supabase session remains authenticated after relay-token rejection.
- [x] API-key-only configuration remains usable and is not described as signed in.

## Single source of truth

Relay validity remains owned by fetchRelayTokenStatus. The setup tools settle auth.getStatus before credential readiness, so SupabaseClient and ServerSideKeyService read the same relay-status cache state.

No new service or public abstraction was added. Existing setup operations are sequenced at their two reporting boundaries, with one shared test assertion for the ordering contract.

## Net elements (R6)

6 files changed; 124 insertions and 11 deletions; no files, runtime types, or services added.

## Consumer counts (R8)

n/a — no emit path or public export was deleted or rerouted.

## Host impact

CLI, desktop, and extension setup reporting no longer accepts stale relay-token presence as proof of authentication or credential readiness.

## Scheduled refactoring

None.

## Validation

- Focused setup suites: 10 tests passed.
- npm run format passed.
- npm run typecheck passed.
- Targeted ESLint passed without warnings.
- git diff --check passed.

---

> [!NOTE]
> **Medium Risk**
> Changes shared setup authentication and credential reporting used by CLI, desktop, and extension onboarding; behavior is well-tested but affects how users and agents interpret sign-in state.
> 
> **Overview**
> **Setup no longer treats a configured CI relay token as signed-in until it is verified**, and **credential readiness cannot run ahead of that check**.
> 
> `defaultAuthStatus` in the setup platform now calls `fetchRelayTokenStatus` when `TEXRA_RELAY_TOKEN` (or equivalent) is set, priming the shared relay-status cache that `SupabaseClient.isAuthenticated()` reads. Expired or revoked tokens therefore report **unauthenticated** instead of appearing signed in from token presence alone.
> 
> `probe_environment` and `verify_setup` **await `auth.getStatus()` before** probing `anyUsableCredentialExists()` (they previously ran auth in parallel with credential checks). That ordering removes the race where credential reporting could settle while relay rejection was still pending.
> 
> Valid **Supabase sessions** still authenticate when relay verification fails, and **API-key-only** setups remain usable without being labeled as signed in. Tests lock in auth-before-credential ordering and relay rejection / session fallback behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 031ac32424e81f45ed76e8f0ff62758f6b5e400d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Added the import-block header and documented that fetchRelayTokenStatus primes the cache consumed by SupabaseClient.isAuthenticated() in 4e7230eb6. The focused default-platform suite still passes.

LionSR:
Addressed the readiness race in faf6e33e0. Both probe_environment and full verify_setup now settle auth.getStatus() before starting anyUsableCredentialExists(), so a rejected relay token is in the shared cache before canUseServerSideKeys() evaluates readiness. Added ordering regressions for both tools. Two setup suites pass (10 tests), full workspace typecheck passes, targeted ESLint is clean, and diff/format checks pass.

PR #7918, opened 2026-07-10
test(supabase): catch llm-zoo version skew between package.json and relay deno.json

### LINKED ISSUE #7769: llm-zoo version skew between package.json and the relay's exact deno.json pin isn't caught
## Context

Follow-up from #7760 (Fence relay shared configuration), which closed #7681.

Flagged by the TeXRA agent's automated review on #7760: https://github.com/LionSR/TeXRA/pull/7760#pullrequestreview-4664644279

## Details

#7760 pinned `supabase/functions/relay/deno.json`'s `llm-zoo` import to an exact version (`1.12.0`) and added a version-bump workflow reminder to update that pin when `llm-zoo` changes. However:

- The workspace `package.json` still pins `llm-zoo` with a caret range (`^1.12.0`).
- The version-bump reminder text only mentions refreshing the relay's Deno pin/lock — it doesn't cover the reverse case (someone bumps the relay's Deno pin without bumping `package.json`, or vice versa).
- The new `RelaySharedConfigParity.vitest.ts` test only asserts hardcoded tier/spending-limit constants are equal; it does not compare the two `llm-zoo` version pins, so version skew between the client dependency and the relay's Deno dependency is not CI-caught.

Add a check (either a small test asserting the two version strings are compatible, or an explicit note in the version-bump reminder covering both directions) so `llm-zoo` version drift between the workspace and the relay is caught the same way the tier/spending-limit drift now is.

## References

- PR: #7760
- Review: https://github.com/LionSR/TeXRA/pull/7760#pullrequestreview-4664644279

LionSR:
## What / Why

Closes #7769.

#7760 pinned `supabase/functions/relay/deno.json`'s `llm-zoo` import to an
exact version and added `RelaySharedConfigParity.vitest.ts` to catch
tier/spending-limit drift between the client and the relay. But the
workspace `package.json` still pins `llm-zoo` with a caret range
(`^1.12.0`), and nothing in CI compared it against the relay's exact
`deno.json` pin — so version skew in either direction (bump the npm dep
without the relay pin, or vice versa) would go unnoticed until runtime.

This extends the existing parity test with a third case: it reads both
`package.json`'s `dependencies["llm-zoo"]` and
`supabase/functions/relay/deno.json`'s `imports["llm-zoo"]`, strips the
range operator / `npm:llm-zoo@` prefix from each, and asserts the
resulting versions are equal. Any future skew now fails CI the same way
tier/spending-limit drift already does, symmetrically in either direction.

## Verification

- Confirmed the defect still existed at HEAD before fixing: `package.json`
  pinned `^1.12.0`, relay `deno.json` pinned the exact `1.12.0`, and the
  existing parity test only checked tiers/spending limits (no version
  comparison).
- Added the new `it()` case to `RelaySharedConfigParity.vitest.ts`.
- Ran `npx vitest run src/test-kernel/supabase/RelaySharedConfigParity.vitest.ts`
  — 3/3 pass.
- Ran `npx vitest run src/test-kernel/supabase/` — 40/40 pass (no
  regressions in sibling supabase tests).
- **Regression check (revert-style):** temporarily bumped
  `package.json`'s `llm-zoo` to `^1.13.0` (without touching the relay
  pin) and re-ran the test — it fails with a clear message identifying
  both pins and which files to update; reverted the temporary edit
  afterwards (working tree confirmed clean / matches the single intended
  diff before committing).
- `npm run typecheck` — passes.
- `npx eslint src/test-kernel/supabase/RelaySharedConfigParity.vitest.ts`
  and `npx prettier --check` on the same file — clean.

## Net elements (R6)

- 0 new files, 0 new abstractions/helpers.
- +45 lines in one existing test file (one new `it()` block using only
  `node:fs`/`node:path`/`node:url`, already used elsewhere in
  `test-kernel` for repo-root-relative file reads, e.g.
  `src/test-kernel/architecture/progressEventHandlerRetirement.vitest.ts`).
- No production code changed; no new CI workflow step (the existing
  `RelaySharedConfigParity.vitest.ts` suite already runs under `npm test`).

## Consumer counts (R8)

- N/A — no shared helper/abstraction introduced; the new assertion lives
  inline in the single test file it extends, consistent with the other
  two cases in that suite.

---

> [!NOTE]
> **Low Risk**
> Test-only change with no runtime or auth impact; failures are limited to dependency pin drift detection.
> 
> **Overview**
> Extends **`RelaySharedConfigParity.vitest.ts`** so CI fails when the workspace and Supabase relay disagree on **`llm-zoo`**, alongside the existing tier and spending-limit checks.
> 
> A **`REPO_ROOT`** helper (same pattern as other `test-kernel` file reads) loads repo files from disk. **Two new cases** assert: (1) the caret **floor** in root `package.json` matches the exact pin in `supabase/functions/relay/deno.json` after stripping `^`/`~` and `npm:llm-zoo@`; (2) the version **pnpm actually resolves** in `pnpm-lock.yaml` (regex on the top-level `llm-zoo@…` entry) matches that relay pin—so skew inside a valid caret range is caught too. Failures include messages pointing at which files to update.
> 
> No production or dependency manifest changes in this diff—tests only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47431318d3fd150499c1e1cebc731e0f4eeba8e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/supabase/RelaySharedConfigParity.vitest.ts
+++ b/src/test-kernel/supabase/RelaySharedConfigParity.vitest.ts
@@ -39,7 +39,7 @@ describe('relay shared configuration parity', () => {
     }
   });
 
-  it('keeps the workspace llm-zoo dependency in sync with the relay deno.json pin', () => {
+  it('keeps the workspace llm-zoo range floor in sync with the relay deno.json pin', () => {
     const workspacePackageJson = JSON.parse(
       readFileSync(resolve(REPO_ROOT, 'package.json'), 'utf8'),
     ) as { dependencies?: Record<string, string> };
@@ -49,7 +49,10 @@ describe('relay shared configuration parity', () => {
       'package.json is missing an llm-zoo dependency',
     ).toBeDefined();
     // Workspace pin is a caret range (e.g. "^1.12.0"); strip the range
-    // operator to compare against the relay's exact version.
+    // operator to compare against the relay's exact version. This only
+    // checks the declared floor — pnpm may resolve a higher patch/minor
+    // within the caret range, which the resolved-version check below
+    // catches.
     const workspaceVersion = workspaceRange!.replace(/^[\^~]/, '');
 
     const relayDenoJson = JSON.parse(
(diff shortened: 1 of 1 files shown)


LionSR:
Fixed the codex P2: the parity test now compares the **pnpm-resolved** llm-zoo version (parsed from `pnpm-lock.yaml`'s `packages:` entry, e.g. `llm-zoo@1.12.5:`) against the relay's exact `deno.json` pin, in a new `keeps the pnpm-resolved llm-zoo version in sync with the relay deno.json pin` test. This catches skew that hides inside the caret range (e.g. `^1.12.0` resolving to `1.12.5` while the relay stays pinned to `1.12.0`), which the original package.json-vs-deno.json floor comparison would have missed. Kept the floor check too (renamed for clarity) since it still flags intentional range drift.

Also folded in the Copilot nit: both failure messages now say to update the relay `deno.json`/`deno.lock` **and** the workspace `pnpm-lock.yaml`.

Verified: `vitest run` passes (4/4); hand-edited the lockfile copy to `llm-zoo@1.12.5:` to simulate resolved-version skew and confirmed the new assertion fails with the expected message, then reverted via `git checkout -- pnpm-lock.yaml` before committing.

PR #7967, opened 2026-07-10
refactor(cli): consolidate child-stream relationship state into one map

LionSR:
Implements #7951.

## What

Implements the accepted design in `docs/proposals/cli-child-stream-state-consolidation.md`
(#7941, closing #7864): consolidates the CLI TUI's three subagent-relationship
representations — the live `activeSubagents` roster, the retained `childStreams`
history, and the `parentStream` child→parent topology — into one CLI-local
`Map<StreamTabId, ChildStreamEntry>` owned by `packages/cli/src/chat/tui/state/childExecutions.ts`,
exposed only through pure selectors and transition functions.

- **New owner** (`childExecutions.ts`): `ChildStreamEntry`/`ChildStreamEntries`,
  the `CHILD_STREAMS` signal (private), transition functions
  (`setParentStream`, `applySubagentRoster`, `applyChildStreamRemoval`,
  `resetChildStreamEntries`), and selectors (`activeSubagentsFor`,
  `retainedChildStreamsFor`, `visibleSubagentRows`, `focusOrderDescendants`,
  `isChildStreamRemoved`, plus the derived `parentStream` and
  `childStreamEntries` computed signals). Implements the precedence rule from
  the design's "Authority and transition semantics": `removeStream` tombstone
  > explicit edge > roster-derived parent > none, with status read live from
  each child's own `StreamSlice` (no status field on `ChildStreamEntry`, so
  `updateChildStatusReferences`/`withMirroredChildStreamStatus`/the three-list
  status traversal in `setStreamStatusInCliState` are deleted entirely).
- **`cliState.ts`**: `activeSubagents`/`childStreams` removed from
  `StreamSlice`; `parentStream`/`setParentStream` are now thin re-exports of
  the `childExecutions.ts` owner (same import path, same names — zero
  call-site churn for those two symbols specifically); `removeStream` and
  `resetCliState` now delegate the child-stream half of their job to
  `applyChildStreamRemoval`/`resetChildStreamEntries`.
- **One atomic cutover**, no dual-write stage: `subscribeRuntimeHost.ts`'s
  `applyActiveSubagents`/`applyParentStream` route directly to
  `applySubagentRoster`/`setParentStream`; every CLI reader (`App.tsx`,
  `StatusBar.tsx`, `StreamTabsStrip.tsx`, `StaticConversationTranscript.tsx`,
  `SubagentList.tsx`, `ChildControlPicker.tsx`, `childControls.ts`,
  `streamViews.ts`, `focusCycle.ts`, `resumeHint.ts`,
  `focusedChildFollowUp.ts`, `runChatTui.tsx`) switched to the selectors in
  the same commit.
- **Dev harness** (`packages/cli/scripts/tui-harness.tsx`) seeds through the
  real `applySubagentRoster`/`setParentStream` transition functions instead of
  poking `StreamSlice` fields directly.
- **Bug fix in `useSignal.ts`** (found by running the existing `validate:tui`
  PTY harness against this diff, see Verification): `parentStream` is now a
  `Signal.Computed` (previously a plain `Signal.State`) — the first CLI
  `Computed` ever routed through `useSignal`. `useSyncExternalStore` can call
  our `getSnapshot` synchronously from inside `notify()` while a signal
  `.set()` call is still unwinding its watcher notifications; reading a
  `Computed` there recomputes it, which itself reads further signals — and
  `Signal.subtle.Watcher` asserts against exactly that ("signal read during
  notification phase"). `/clear` (`resetCliState()`, which fires several
  `.set()` calls back-to-back including the new `resetChildStreamEntries()`)
  crashed the harness on `main`-plus-this-diff before the fix. Fixed by
  deferring `notify()` to a microtask, mirroring `@lit-labs/signals`'s own
  `SignalWatcher`/`effectWatcher`, which already uses exactly this pattern
  internally for the same reason.

## Why

`activeSubagents`, `childStreams`, and `parentStream` were three views of one
child-stream relationship, mutated independently with duplicated
status-mirroring machinery (`updateChildStatusReferences`,
`withMirroredChildStreamStatus`, a three-list traversal in
`setStreamStatusInCliState`) that risked drift between them. The design doc's
load-bearing race census (roster-before-edge, edge-before-roster,
completion-before-terminal-status, promotion with a stale roster,
reattachment, removal with late facts) drove the transition semantics and the
test matrix below.

## Verification

```
npm run typecheck          # root + test-kernel + cli + desktop + extension + trace-viewer — all green
npm run check:cli-architecture
npm run lint                # 0 errors (pre-existing import/order warnings elsewhere, untouched files)
npm test                    # 621 files / 5459 tests passed, 1 file / 4 tests skipped (pre-existing)
npx vitest run src/test-kernel/cli/   # 128 files / 1652 tests passed
npm run validate:tui       # PTY harness: 79/80 scenarios pass; the 1 failure
                            # (model-switch-compatible-only, "GPT-5.5" missing)
                            # is a pre-existing hardcoded-model-name mismatch in
                            # validate-tui.mjs unrelated to this diff (verified:
                            # it references a model list this PR never touches)
```

`TuiStateAndFocus.vitest.mts` gained a new `describe('child-stream ordered
transition matrix', ...)` covering the design's 11-sequence ordered unit
matrix (canonical/roster-first/edge-first/status-first orderings, completion,
promotion-with-stale-roster, explicit reattachment, child removal with late
facts, fresh activation after removal, two-child retention order stability,
parent removal with late facts), asserting effective parent, active rows, and
retained rows at the load-bearing checkpoints in each sequence — driven
through the real `applySubagentRoster`/`setParentStream`/`removeStream`/
`setStreamStatusInCliState` functions, not hand-built state. One of these
(scenario 8) is a proven regression test: it fails if `setStreamStatusInCliState`'s
tombstone guard is removed (verified locally by reverting that guard).

`ChildControls.vitest.mts`, `StreamTabsStrip.vitest.mts`, `ResumeHint.vitest.mts`,
and 2 fixtures in `ConversationPane.vitest.mts` were converted from
`StreamSlice.activeSubagents`/`childStreams` fixtures to a new shared
selector-level test builder, `src/test-kernel/support/childStreamEntries.ts`
(`buildChildStreamEntries`), used by all four files — a DRY extraction, not a
new test suite (R7): it has no `it`/`describe` of its own and follows the
existing `test-kernel/support/` convention (`FakePlatform.ts`,
`asyncTestUtils.ts`, ...).

### Not completed in this PR

The design's acceptance-gate items 4-5 ask for byte-identical PTY frame
checks for order-equivalent cases via a new opt-in
`HARNESS_CHILD_EVENT_ORDER` fixture in `packages/cli/scripts/tui-harness.tsx`
plus `validate-tui.mjs` assertions. That fixture was not built in this PR —
the vitest-level ordered matrix above is real, event-sequence-driven coverage
of the same 11 scenarios (through the production transition functions), but
it is not the rendered-byte-level PTY layer the design calls for. I ran the
**existing, unmodified** `npm run validate:tui` scenarios against this diff:
79/80 pass (the 1 failure is pre-existing and unrelated — see Verification),
including a crash the harness caught in `slash-clear` that this PR's
`useSignal.ts` fix resolves (see "Bug fix in `useSignal.ts`" above) — real
evidence the rendered TUI is not regressed for every scenario the current
harness exercises, short of the new event-ordering fixture. Filing a
follow-up issue for the `HARNESS_CHILD_EVENT_ORDER` fixture and the
byte-identical frame assertions is recommended before this is considered
fully gate-complete.

## Net elements (R6)

Using the method in `fewer-elements-2026-07.md` (files from the diffstat,
`^[+-]export` over the diff, class/interface/enum declarations the same way),
measured on the actual diff (this PR replaces the design doc's estimates):

- **Production files:** 17 modified, 0 added (deepened the existing
  `childExecutions.ts` per the design's preference, no new module; includes
  the `useSignal.ts` bug fix). Design estimated "add 0... modify
  approximately 12-16": close (one more file than the high end, for the
  signal-phase fix the design didn't anticipate).
- **Production lines:** net **+320** (724 insertions / 404 deletions). Design
  estimated net **-40 to -100**; the difference is almost entirely the new
  `childExecutions.ts` transition/selector bodies plus doc comments — the
  design's own estimate flagged this range as approximate and gated a
  positive *production*-line result on maintainer re-review, which this PR
  requests.
- **Class/interface/enum declarations:** delete 1 (`ParentStreamEdgeUpdate`),
  add 1 (`ChildStreamEntry`) — **net 0**, matching the design's estimate
  exactly.
- **Authoritative mutable collections:** delete 3 (`StreamSlice.activeSubagents`,
  `StreamSlice.childStreams`, the `PARENT_STREAM` signal), add 1
  (`CHILD_STREAMS`, private) — **net -2**, matching the design's estimate
  exactly.
- **Exported-symbol delta (`^[+-]export` line count):** **+17 / -8, net +9**.
  This diverges from the design's estimate (`+2/-5, net -3` via "one narrow
  owner facade"). I chose flat top-level exports
  (`activeSubagentsFor`/`retainedChildStreamsFor`/`applySubagentRoster`/
  `applyChildStreamRemoval`/`isChildStreamRemoved`/`resetChildStreamEntries`/
  `focusOrderDescendants`/the two computed signals) over bundling them into
  one facade object, specifically so `parentStream` and `setParentStream` —
  used across 16+ consumer files and dozens of tests — keep their exact prior
  export names and import path (`./cliState`), at zero call-site churn for
  those two symbols. Semantically: 2 exported symbols are deleted
  (`registerChildStreams`, `mergeChildStreams`); 10 are genuinely new
  capability (the interface/type, the 4 new selectors, 3 of the transition
  functions, and the new `childStreamEntries` computed signal) — a net **+8**
  new *symbols*, not just diff-line noise, though inflated further to +9 by
  the mechanical line-count method double-counting unchanged functions
  (`childExecutionKey`/`childExecutionLabel`) that moved within the same file
  during a full-file rewrite. This is a deliberate, disclosed tradeoff between
  R5 (minimal consumer-facing diff) and R6 (exported-symbol delta), flagged
  here per the design's own instruction that a positive delta needs a stated
  reason and is subject to explicit maintainer re-review, not a silent
  pass/fail.
- **Callable declarations** (`function`/arrow-function declarations):
  **+15 / -18, net -3** in production code.
- **Test files:** 5 modified, 1 added (`src/test-kernel/support/childStreamEntries.ts`,
  a shared test-fixture builder, not a test suite — see "Verification" above).
- **Test lines:** net **+493** (1159 insertions / 666 deletions), close to the
  design's estimated **+200 to +320** but higher, mostly the new 11-scenario
  ordered matrix (~230 lines) plus the mechanical fixture-shape rewrite across
  4 existing files.
- **Whole-diff net lines:** **+813** (23 files, 1883 insertions / 1070
  deletions), above the design's stated **+100 to +280** range. Per the
  design's own R6 gate language ("A non-negative production-line result or
  whole-diff result above the stated range requires explicit maintainer
  re-review even if tests pass"), this PR is explicitly flagged for that
  re-review rather than asserting compliance.
- **New shared planes, vocabularies, schemas, reducers, or persistence
  stores:** **0** — nothing added under `src/shared/` or the webview; the
  event hub and `StreamSnapshotStore` are unchanged (acceptance-gate item 7).
- `activeProcesses` and its output lifecycle are untouched (acceptance-gate
  item 6).

## Consumer counts (R8)

Grep-verified caller counts (excluding `childExecutions.ts` itself) before
treating any of these as a justified shared selector:

- `activeSubagentsFor`: 6 files
- `retainedChildStreamsFor`: 2 files
- `visibleSubagentRows`: 8 files
- `applySubagentRoster`: 4 files
- `applyChildStreamRemoval`: 2 files (`cliState.removeStream`, the new ordered
  test matrix)
- `isChildStreamRemoved`: 3 files
- `childStreamEntries` (the computed signal): 17 files
- `focusOrderDescendants`: 1 file (`focusCycle.ts`'s
  `orderedDescendantsFromTree` wrapper, itself called from 2 sites) — kept as
  its own selector, not inlined, because it is exactly the composed selector
  the design specifies (retained-first-then-topology, `streams.has()`-gated)
  and is the unit under direct test in the ordered matrix.
- `resetChildStreamEntries`: 1 file (`cliState.resetCliState`) — matches the
  existing one-call-site reset-hook pattern every other slice in `cliState.ts`
  already uses.

---

> [!NOTE]
> **Medium Risk**
> Broad refactor of CLI focus routing, subagent UI, and stream removal semantics with many touchpoints; behavior is heavily tested but event-ordering edge cases are complex.
> 
> **Overview**
> **Child-stream state** moves from three independently mutated views (`StreamSlice.activeSubagents`, `childStreams`, and a writable `parentStream` map) into a single signal-backed `CHILD_STREAMS` map in `childExecutions.ts`, with transitions (`applySubagentRoster`, `setParentStream`, `applyChildStreamRemoval`) and selectors (`activeSubagentsFor`, `retainedChildStreamsFor`, `visibleSubagentRows`, derived `parentStream`). Child **status** is no longer mirrored on parent slices; selectors read each child's `StreamSlice` instead, so status-copy machinery in `cliState` is removed.
> 
> **Consumers** (App, status bar, stream tabs, subagent panel, child controls, resume hints, runtime subscription, TUI harness) now pass `childStreamEntries` and call the selectors rather than reading slice fields. `removeStream` / `resetCliState` tombstone or reset relationship state through the new module.
> 
> **`useSignal`** defers watcher `notify()` to a microtask so subscribing to `Signal.Computed` (e.g. `parentStream`) does not recompute during the notification phase—fixing harness crashes on rapid resets like `/clear`.
> 
> Tests add an ordered transition matrix and shared `buildChildStreamEntries` fixtures; production net is more lines than the design estimated due to the new owner implementation and comments.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1dbf6c4a017356cef5bac75efaebcb31288ada4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/scripts/tui-harness.tsx
+++ b/packages/cli/scripts/tui-harness.tsx
@@ -69,6 +69,12 @@ import {
   type ConversationEntry,
 } from '../src/chat/tui/state/cliState';
 import {
+  activeSubagentsFor,
+  applySubagentRoster,
+  childStreamEntries,
+  visibleSubagentRows,
+} from '../src/chat/tui/state/childExecutions';
+import {
   focusedChildFollowUpRoute,
   stoppedFocusedChildFollowUpMessage,
 } from '../src/chat/tui/state/focusedChildFollowUp';
@@ -704,14 +710,6 @@ function makeChildEntries(agent: string, action: string): ConversationEntry[] {
   ];
 }
 
-function stoppedChild(child: ActiveChildInfo): ActiveChildInfo {
-  return {
-    ...child,
-    status: STREAM_STATUS.STOPPED,
-    startedAt: undefined,
-  };
-}
-
 function isDifferentExecution(
   child: ActiveChildInfo,
   executionId: string,
@@ -1043,12 +1041,18 @@ if (SHOW_CHILDREN) {
   const activeSubagents = childStreams.filter(
     (child) => child.status !== STREAM_STATUS.ERROR,
   );
+  // Seed through the real transition functions rather than poking StreamSlice
+  // fields directly: register the complete roster first (so the errored
+  // child gets a retained row), then re-apply the roster with it omitted —
+  // mirroring the production sequence where a roster stops including a child
+  // once it leaves the runtime's active registry, while its retained history
+  // survives.
+  applySubagentRoster(STREAM_ID, childStreams);
+  applySubagentRoster(STREAM_ID, activeSubagents);
   patchStream(STREAM_ID, (slice) => ({
     ...slice,
     status: STREAM_STATUS.RUNNING,
     runStartedAt: startedAt,
-    activeSubagents,
-    childStreams,
     activeProcesses: [
       {
         kind: 'process' as const,
@@ -1078,16 +1082,11 @@ if (SHOW_CHILDREN) {
     const addNestedChildren =
       SHOW_NESTED_CHILDREN && child.agentName === 'strategy';
     setParentStream(streamId, STREAM_ID);
+    if (addNestedChildren) applySubagentRoster(streamId, [nestedStrategyChild]);
     patchStream(streamId, (slice) => ({
       ...slice,
       status: child.status,
       description: `${child.agentName} sub-workflow`,
-      activeSubagents: addNestedChildren
-        ? [nestedStrategyChild]
-        : slice.activeSubagents,
-      childStreams: addNestedChildren
-        ? [nestedStrategyChild]
-        : slice.childStreams,
       activeProcesses: addNestedChildren
         ? [nestedStrategyProcess]
         : slice.activeProcesses,
@@ -1283,22 +1282,20 @@ if (SHOW_AGENT_PROPOSAL) {
 
 function markHarnessInterrupted(): void {
   canInterrupt = false;
-  const parentSlice = streams.get().get(STREAM_ID);
   const childStreamIds = new Set(
-    [
-      ...(parentSlice?.activeSubagents ?? []),
-      ...(parentSlice?.childStreams ?? []),
-    ]
+    visibleSubagentRows(STREAM_ID, childStreamEntries.get(), streams.get())
       .map((child) => child.childStreamId)
       .filter((streamId): streamId is string => streamId !== undefined),
   );
+  // Clear active roster membership through the real transition function;
+  // each retained row's status comes from its own StreamSlice, set by the
+  // per-child patchStream loop below.
+  applySubagentRoster(STREAM_ID, []);
   patchStream(STREAM_ID, (slice) => ({
     ...slice,
     status: STREAM_STATUS.STOPPED,
     runStartedAt: undefined,
-    activeSubagents: [],
     activeProcesses: [],
-    childStreams: slice.childStreams.map(stoppedChild),
     entries: [
       ...slice.entries,
       {
@@ -1420,10 +1417,15 @@ function markHarnessExecutionStopped(executionId: string): void {
   const parentSlice = streams.get().get(STREAM_ID);
   if (!parentSlice) return;
 
+  const activeSubagentRows = activeSubagentsFor(
+    STREAM_ID,
+    childStreamEntries.get(),
+    streams.get(),
+  );
   const executionRows = [
-    ...parentSlice.activeSubagents,
+    ...activeSubagentRows,
     ...parentSlice.activeProcesses,
-    ...parentSlice.childStreams,
+    ...visibleSubagentRows(STREAM_ID, childStreamEntries.get(), streams.get()),
   ];
   const executionRow = executionRows.find(
     (child) => child.executionId === executionId,
@@ -1431,16 +1433,16 @@ function markHarnessExecutionStopped(executionId: string): void {
   if (!executionRow) return;
 
   const messageId = `harness-killed-${executionId}-${Date.now()}`;
+  applySubagentRoster(
+    STREAM_ID,
+    activeSubagentRows.filter((child) =>
+      isDifferentExecution(child, executionId),
+    ),
+  );
   patchStream(STREAM_ID, (slice) => ({
     ...slice,
-    activeSubagents: slice.activeSubagents.filter((child) =>
-      isDifferentExecution(child, executionId),
-    ),
     activeProcesses: slice.activeProcesses.filter((child) =>
       isDifferentExecution(child, executionId),
-    ),
-    childStreams: slice.childStreams.map((child) =>
-      child.executionId === executionId ? stoppedChild(child) : child,
     ),
     entries: [
       ...slice.entries,
@@ -1476,6 +1478,7 @@ function handleHarnessSubmit(line: string): void {
   if (focusedChildRoute.kind === 'reject') {
     appendHarnessAssistantTranscript(
       stoppedFocusedChildFollowUpMessage({
+        childStreamEntries: childStreamEntries.get(),
         parentStream: parentStream.get(),
         streamId: focusedChildRoute.streamId,
         streams: streams.get(),
--- a/packages/cli/scripts/tui-harness.tsx
+++ b/packages/cli/scripts/tui-harness.tsx
@@ -59,8 +59,6 @@ import {
   activeStreamId as activeStreamIdSignal,
   rootRunStartAvailable,
   rootStreamId,
-  parentStream,
-  setParentStream,
   resetCliState,
   sessionMeta,
   setCliSessionModelOverride,
@@ -72,6 +70,8 @@ import {
   activeSubagentsFor,
   applySubagentRoster,
   childStreamEntries,
+  parentStream,
+  setParentStream,
   visibleSubagentRows,
 } from '../src/chat/tui/state/childExecutions';
 import {
--- a/packages/cli/src/chat/tui/App.tsx
+++ b/packages/cli/src/chat/tui/App.tsx
@@ -12,6 +12,7 @@ import {
 } from '@common/constants/streamStatus';
 import {
   TODO_STATUS,
+  type ActiveChildInfo,
   type StreamLifecycleStatus,
   type StreamTabId,
   type TodoItem,
@@ -66,6 +67,10 @@ import {
   parentStream as parentStreamSignal,
   streams as streamsSignal,
 } from './state/cliState';
+import {
+  childStreamEntries as childStreamEntriesSignal,
+  visibleSubagentRows,
+} from './state/childExecutions';
 import { focusedChildInputDisabledMessage } from './state/focusedChildFollowUp';
 import { nextFocusBack, nextFocusForward } from './state/focusCycle';
 import { streamDisplayLabel } from './state/streamViews';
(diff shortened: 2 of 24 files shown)


LionSR:
Applied the three review findings on this branch (pushed to `claude/issue-7951-child-stream-impl`):

**1. Deleted the `parentStream`/`setParentStream` re-export shim in `cliState.ts`** (claude bot, R1 ruling — a code-to-code shim can't merge without a dated ledger row, so the preferred arm is deletion in the same PR). Retargeted **10 consumers** to import directly from `childExecutions.ts`, the actual owner:
- `packages/cli/src/chat/tui/state/focusCycle.ts`
- `packages/cli/src/chat/tui/state/transcript.ts`
- `packages/cli/src/chat/tui/state/subscribeRuntimeHost.ts`
- `packages/cli/src/chat/tui/App.tsx`
- `packages/cli/src/chat/tui/runChatTui.tsx`
- `packages/cli/src/chat/tui/panes/StreamTabsStrip.tsx`
- `packages/cli/src/chat/tui/panes/StaticConversationTranscript.tsx`
- `packages/cli/src/chat/tui/panes/StatusBar.tsx`
- `packages/cli/scripts/tui-harness.tsx`
- `src/test-kernel/cli/TuiStateAndFocus.vitest.mts`

(Several already imported other symbols from `childExecutions.ts`, so this only merges the import lines; a couple needed a new import line.)

**2. Fixed `applySubagentRoster` writing/re-rendering on identical snapshots** (Copilot). It previously wrote every accepted roster child unconditionally and set `changed = true` regardless of whether anything actually differed, so `CHILD_STREAMS.set()` fired — and every `useSignal` subscriber re-rendered — even for a byte-identical roster resent by the runtime on every poll. Added `subagentEntryUnchanged`/`summaryUnchanged` helpers that compare only the fields this function persists (`summary`, `activeParentStreamId`, `retainedParentStreamId`, `retainedOrder`) and skip the write when nothing changed. Left the tombstone > explicit-edge > roster-derived > none precedence logic untouched. Added test 12 in the `child-stream ordered transition matrix` describe block in `TuiStateAndFocus.vitest.mts`: applying an identical roster snapshot (fresh row objects, same field values) twice leaves `childStreamEntries.get()` reference-equal — i.e., no store write on the second call.

**3. Fixed a post-unmount microtask leak in `useSignal.ts`** (codex P2). The queued-notify microtask could fire after the subscriber unmounted, since cleanup only called `watcher.unwatch(signal)` — the pending microtask would still call `notify()` and re-arm via `watcher.watch()`, leaking the watcher. Added a `disposed` flag set in the cleanup and checked at the top of the microtask; it's scoped per `subscribe()` invocation, so React StrictMode's dev-mode mount/cleanup/remount gets its own fresh watcher/flag pair rather than sharing state with the torn-down one. No dedicated `useSignal` test harness exists in this repo today, so per scope I did not build one — flagging this here rather than adding new test infrastructure.

**Verification:**
- `npx vitest run src/test-kernel/cli/` — 128 test files, 1653 tests, all passing (includes the new no-op-roster-write test).
- `corepack pnpm --filter @texra-ai/cli typecheck` — clean.
- `npm run typecheck` (full workspace: root, test-kernel, cli, trace-viewer, desktop) — clean.
- `npx eslint` on all 13 touched files — clean.

PR #5776, opened 2026-06-10
Converge workspace config on `.texra/config.json` across hosts

LionSR:
## Summary

Unifies workspace configuration storage across the CLI, desktop, and extension by adopting a shared `.texra/config.json` file at the project root. Previously, each host stored workspace config in different locations (CLI: `.texra/config.json`, desktop: internal per-workspace store, extension: VS Code settings). This change makes the desktop and CLI converge on the same file, enabling checked-in project configuration to behave identically in both hosts.

The extension continues using VS Code settings (`.vscode/settings.json`) as its workspace config source, maintaining the VS Code idiom.

## Issue acceptance gates

Implements the workspace config convergence design:
- ✅ CLI reads/writes `/.texra/config.json`
- ✅ Desktop reads `/.texra/config.json` with fallback to legacy internal store
- ✅ Desktop performs one-time migration from legacy store to project file
- ✅ CLI wires user-level config (`~/.texra/global-storage/config.json`, the same file the chat-defaults loader already read) into the `global` config target, matching desktop's existing global store (`<userData>/config/global.json`)
- ✅ Shared constants and paths eliminate duplication

## Single source of truth

**Module:** `src/platform/defaults/nodeStorage.ts`

Exports `workspaceTexraConfigPath()`, `TEXRA_CONFIG_FILE_NAME`, and `TEXRA_STORAGE_DIR_NAME` as the canonical definitions of the shared project config location. All hosts (CLI, desktop) and config initialization code import from this module rather than re-declaring paths.

## Duplication and abstraction budget

**Removed duplication:**
- Eliminated `workspaceCliConfigPath()` from `packages/cli/src/runtime/cliConfig.ts`; now uses `workspaceTexraConfigPath()` from platform defaults
- Consolidated `CLI_CONFIG_DIR` and `CLI_CONFIG_FILE` constants into `TEXRA_STORAGE_DIR_NAME` and `TEXRA_CONFIG_FILE_NAME` in the platform layer
- Removed per-host magic strings for output truncation (`OUTPUT_MAX_CHARS`); now uses `TERMINAL_OUTPUT_MAX_CHARS` from `src/hosts/terminalHost.ts`
- Consolidated tool timeout defaults into `src/shared/constants/toolDefaults.ts` (used by both tool implementations and progress UI)

**New abstractions:**
- `workspaceTexraConfigPath(workspaceRoot)` — owns the invariant that workspace config lives at `/.texra/config.json`
[...]

PR #8002, opened 2026-07-10
fix(desktop): unify desktop data root onto CLI's ~/.texra scheme

### LINKED ISSUE #7987: Unify Desktop data root onto ~/.texra so Desktop and CLI share workspace memory/history (narrowed from #3865)
## Problem (narrowed from #3865 — the one user-facing remainder)

Desktop roots memory/history/executions under Electron `userData` while the CLI roots the same data under `~/.texra` — so a user working in the same workspace from Desktop and CLI sees **two different histories and memory views**. Storage-identity computation is already unified (`workspaceStorageId()`, #3993) and workspace *config* already converged (#5776); the data ROOT is the last divergence.

## Decision (maintainer-delegated, 2026-07-11)

Unify the desktop data root onto the CLI's `~/.texra` scheme (the host-neutral choice; `userData` is the Electron-idiom outlier), with a one-time migration.

## Implementation sketch (cleanest path)

1. Point desktop's storage provider (`packages/desktop/src/main/platform/` — the piece that currently derives roots from `app.getPath('userData')`) at the same `~/.texra/global-storage`/workspace-keyed roots the CLI's `nodeStorage.ts` uses, through the existing `workspaceStorageId()`.
2. One-time migration on desktop startup: if the legacy `userData` store exists and the target does not, MOVE it (per-workspace-key directory move; log per directory; never overwrite an existing target — skip and warn, matching the #5776 config-migration pattern).
3. Keep a read-only legacy fallback for one release window with a dated #6981 ledger row (R1), then delete.
4. Tests: extend the desktop storage suite with migration cases (fresh install, legacy-present, both-present-skip).

## Explicitly rejected (recorded so the umbrella never regrows)

- **Cross-process locking as a prerequisite** — deferred until a real corruption incident. Concurrent runs already write distinct execution directories; config is last-write-wins and low-stakes. Building a lock layer now is speculative infrastructure (R4/R5).
- **Extension convergence** — the extension keeps VS Code `workspaceState`/`settings.json` idioms per its explicit opt-out in #5776 and the host-parity audit's fenced-deserved register (§5). Not drift.
- **Un-freezing `platform()` for multi-workspace processes** — rejected trap: deep plumbing for a niche need; workspace switch via relaunch stays acceptable.

Supersedes the #3865 umbrella (closed with the Checkpoint-B evidence package).

LionSR:
Fixes #7987.

## Summary

Points desktop's storage provider at the same `~/.texra/global-storage` /
`~/.texra/workspace-storage` roots the CLI's `nodeStorage.ts` already
computes via `workspaceStorageId()`, instead of rooting under Electron's
`app.getPath('userData')`. This directly delivers the issue's outcome:
**Desktop and CLI now share the same workspace memory/history/executions
view** for a workspace worked on from both hosts.

The issue's sketch predates the maintainer's binding fact that desktop has
never shipped a public release, so this is simplified from the sketch:

- **No dated read-only fallback, no #6981 ledger row.** After the one-time
  migration attempt below, only the new `~/.texra` root is ever read. There
  is no ongoing legacy read path to delete later.
- **Migration is best-effort and one-time**, purely to protect the
  maintainer's own dev machines (which already have real `userData`-rooted
  stores): on startup, `migrateLegacyDesktopDataRoot()` moves each top-level
  storage directory (`global-storage`) and each per-workspace-key directory
  under `workspace-storage` independently from the legacy `userData` root
  into the new `~/.texra` root. It never overwrites an existing target
  directory — it skips that one directory and logs a warning instead
  (mirrors the #5776 config-migration pattern: log per directory, never
  clobber). Idempotent: once a legacy directory has moved, it no longer
  exists, so later launches are cheap no-ops.

## What changed

- `packages/desktop/src/main/platform/paths.ts` — new
  `resolveDesktopDataRoot()`: returns the CLI's `DEFAULT_NODE_STORAGE_ROOT`
  (`~/.texra`) by default; when the e2e/dev harness's
  `TEXRA_DESKTOP_E2E_USER_DATA_PATH` override is set (isolating Electron's
  own `userData` profile — see `packages/desktop/src/main/index.ts`), the
  data root stays colocated with that same isolated profile instead of
  diverging onto a developer's real `~/.texra`, so relaunch/e2e tests keep
  working unchanged.
- `packages/desktop/src/main/platform/dataRootMigration.ts` (new) —
  `migrateLegacyDesktopDataRoot()`, the best-effort one-time move described
  above.
- `packages/desktop/src/main/platform/index.ts` — wires the new data root
  and migration call in ahead of constructing `WorkspaceStorageProvider`;
  everything else in this file (global state/config stores, secrets) still
  reads directly from `userDataPath`, unchanged.

## Consumer counts (R8)

Grepped every `userData` reference under `packages/desktop/src` before and
after:

- `packages/desktop/src/main/desktopAppLog.ts` (logs dir) — unaffected,
  still under `app.getPath('userData')`.
- `packages/desktop/src/main/desktopStreamSnapshot.ts` /
  `packages/desktop/src/main/index.ts` (`streams.json` progress snapshot) —
  unaffected.
- `packages/desktop/src/main/platform/index.ts` — `state/global.json`,
  `config/global.json`, `secrets.json` all still open directly under
  `userDataPath`; only the `WorkspaceStorageProvider` root argument changed.
- `packages/desktop/src/main/desktopUpdateChecker.ts` — no `userData` or
  storage-root coupling at all; confirmed unaffected.

Only the one storage-provider root construction changes; every other
`userData` consumer keeps its existing path.

## Net elements (R6)

- **+1 file**: `dataRootMigration.ts` — a genuinely new one-time behavior
  (directory move with skip-and-warn semantics), not an extraction of
  existing logic, so it isn't a single-caller-helper anti-pattern; it has
  exactly one caller (`platform/index.ts`) by design, matching the #5776
  migration precedent.
- **+1 function** (`resolveDesktopDataRoot`) in the existing
  `paths.ts`, alongside `resolveWorkspacePath`/`resolveResourcesPath` — same
  options/env-override shape already used there.
- **No new abstraction layers, no ports, no ledger.** No wrapper functions
  added — `initializeElectronPlatform` calls `resolveDesktopDataRoot()` and
  `migrateLegacyDesktopDataRoot()` directly.

## Tests

- Extended `src/test-kernel/desktop/ElectronCompositionRoot.vitest.mts`
  with a case proving `resolveDesktopDataRoot()` defaults to the CLI's
  `~/.texra` and only diverges under the e2e override.
- Extended `src/test-kernel/desktop/ElectronPlatformAdapters.vitest.mts`
  (the existing desktop storage suite) with the requested migration trio,
  plus two extra cases for idempotency and the same-root no-op:
  - **fresh** — no legacy store: no-op, nothing logged.
  - **legacy-present** — legacy `global-storage` and a workspace-key
    directory move cleanly into the empty target root.
  - **both-present-skip** — `global-storage` collides at the target (left
    untouched on both sides, warned) *and*, in the same run, one
    `workspace-storage` key collides (skipped+warned) while a second,
    non-colliding key still migrates — proving the per-directory (not
    whole-tree) granularity the issue asked for.
  - idempotency (second call after a successful move is a silent no-op) and
    same-root no-op (guards the e2e-override branch) round out the suite.

```
npx vitest run src/test-kernel/desktop src/test-kernel/platform/WorkspaceStorage.vitest.ts --no-file-parallelism
# 43 files, 447 tests passed
```

(`--no-file-parallelism` avoids cross-test-file timeout flakes under this
worktree's concurrent CI load; the same suites are green either way, just
occasionally slower under contention.)

`npx tsc --noEmit -p packages/desktop/tsconfig.main.json` — clean (covers
every file touched in this PR). The full `npm run typecheck` chain (root →
test-kernel → cli → trace-viewer → desktop, across all workspaces) was run
too; root and test-kernel stages passed before this PR body was written
(the chain uses `&&`, so reaching the CLI/desktop stages implies the earlier
stages were clean).

---

> [!NOTE]
> **Medium Risk**
> Changes where desktop reads/writes execution and workspace history on disk at startup; migration is best-effort and non-throwing but wrong paths could split or hide data until manually fixed.
> 
> **Overview**
> **Desktop and CLI now share the same persisted memory/history/executions tree** by pointing `WorkspaceStorageProvider` at `~/.texra` (via `resolveDesktopDataRoot()` and `DEFAULT_NODE_STORAGE_ROOT`) instead of Electron `userData`. Global state, secrets, and config under `userData` are unchanged.
> 
> On startup, **`migrateLegacyDesktopDataRoot()`** best-effort moves `global-storage` and each `workspace-storage/<key>` from the old `userData` root into `~/.texra`, **per directory**, skipping targets that already exist (warn only). E2E/dev keeps the data root colocated with the isolated profile when `TEXRA_DESKTOP_E2E_USER_DATA_PATH` is set.
> 
> **Crash reporting** adds `platformInit.dataRoot` to `sensitivePaths` so payloads are scrubbed like `userData`. Vitest covers `resolveDesktopDataRoot` and migration (fresh, move, skip-on-collision, unreadable legacy, idempotent, same-root).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1595aba1ca9c94c755c80d9b7650f949be7ab3cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/index.ts
+++ b/packages/desktop/src/main/index.ts
@@ -962,6 +962,7 @@ if (protocolLifecycle.shouldContinue) {
               sensitivePaths: [
                 platformInit.workspacePath,
                 app.getPath('userData'),
+                platformInit.dataRoot,
               ],
               log: console,
             });
--- a/packages/desktop/src/main/platform/dataRootMigration.ts
+++ b/packages/desktop/src/main/platform/dataRootMigration.ts
@@ -6,6 +6,9 @@ import { dirname, join } from 'node:path';
 // Local imports - utils
 import { toErrorMessage } from '@utils/errors/errorMessage';
 
+// Type imports - node
+import type { Dirent } from 'node:fs';
+
 export interface DesktopDataRootMigrationLogger {
   info(message: string): void;
   warn(message: string): void;
@@ -13,6 +16,15 @@ export interface DesktopDataRootMigrationLogger {
 
 const GLOBAL_STORAGE_DIR = 'global-storage';
 const WORKSPACE_STORAGE_DIR = 'workspace-storage';
+
+function errorCodeOf(error: unknown): string | undefined {
+  return typeof error === 'object' &&
+    error != null &&
+    'code' in error &&
+    typeof (error as { code?: unknown }).code === 'string'
+    ? (error as { code: string }).code
+    : undefined;
+}
 
 async function moveDirectoryIfAbsent(
   legacyPath: string,
@@ -23,7 +35,7 @@ async function moveDirectoryIfAbsent(
   if (!existsSync(legacyPath)) return;
   if (existsSync(targetPath)) {
     logger.warn(
-      `[desktop] Skipping legacy data migration for "${label}": a directory already exists at ${targetPath}.`,
+      `[desktop] Skipping legacy data migration for "${label}": a directory already exists at ${targetPath}. Legacy data is still at ${legacyPath}; move it manually if needed.`,
     );
     return;
   }
@@ -32,8 +44,13 @@ async function moveDirectoryIfAbsent(
     await rename(legacyPath, targetPath);
     logger.info(`[desktop] Migrated legacy "${label}" to ${targetPath}.`);
   } catch (error) {
+    // Surface the errno code (e.g. EXDEV on some Windows redirected-profile
+    // setups, EACCES on permission-restricted trees, ENOSPC when the target
+    // volume is full) so a failure is diagnosable from the log line alone —
+    // this stays non-throwing/best-effort either way.
+    const code = errorCodeOf(error);
     logger.warn(
-      `[desktop] Failed to migrate legacy "${label}" to ${targetPath}. Cause: ${toErrorMessage(error)}`,
+      `[desktop] Failed to migrate legacy "${label}" to ${targetPath}${code ? ` (${code})` : ''}. Legacy data is still at ${legacyPath}. Cause: ${toErrorMessage(error)}`,
     );
   }
 }
@@ -73,7 +90,19 @@ export async function migrateLegacyDesktopDataRoot(
   const legacyWorkspaceRoot = join(legacyRoot, WORKSPACE_STORAGE_DIR);
   if (!existsSync(legacyWorkspaceRoot)) return;
 
-  const entries = await readdir(legacyWorkspaceRoot, { withFileTypes: true });
+  // Best-effort here too: an unreadable legacy directory (EACCES, or
+  // ENOTDIR if something unexpected occupies that path) must not abort
+  // startup — it only means this run can't migrate workspace-storage.
+  let entries: Dirent[];
+  try {
+    entries = await readdir(legacyWorkspaceRoot, { withFileTypes: true });
+  } catch (error) {
+    const code = errorCodeOf(error);
+    logger.warn(
+      `[desktop] Failed to read legacy workspace-storage directory ${legacyWorkspaceRoot}${code ? ` (${code})` : ''}. Cause: ${toErrorMessage(error)}`,
+    );
+    return;
+  }
   for (const entry of entries) {
     if (!entry.isDirectory()) continue;
     await moveDirectoryIfAbsent(
--- a/packages/desktop/src/main/platform/dataRootMigration.ts
+++ b/packages/desktop/src/main/platform/dataRootMigration.ts
@@ -1,0 +1,86 @@
+// Node imports
+import { existsSync } from 'node:fs';
+import { mkdir, readdir, rename } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
+
+// Local imports - utils
+import { toErrorMessage } from '@utils/errors/errorMessage';
+
+export interface DesktopDataRootMigrationLogger {
+  info(message: string): void;
+  warn(message: string): void;
+}
+
+const GLOBAL_STORAGE_DIR = 'global-storage';
+const WORKSPACE_STORAGE_DIR = 'workspace-storage';
+
+async function moveDirectoryIfAbsent(
+  legacyPath: string,
+  targetPath: string,
+  label: string,
+  logger: DesktopDataRootMigrationLogger,
+): Promise<void> {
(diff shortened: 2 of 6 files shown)


LionSR:
Addressed the four review findings on `packages/desktop/src/main/platform/dataRootMigration.ts` (codex P2, Copilot, and Claude's EXDEV note):

1. **Unreadable legacy dir can't abort startup** — wrapped the `readdir(legacyWorkspaceRoot)` call in try/catch. An EACCES/ENOTDIR (or any other read failure) now logs a warning and returns instead of throwing out of `migrateLegacyDesktopDataRoot`, keeping the whole migration best-effort, not just the per-directory moves.
2. **Collision warnings now include the legacy path** — both the "already exists at target" skip-warning and the failed-move warning now print `legacyPath` alongside `targetPath`, so a maintainer can find and manually recover the un-migrated data.
3. **Differentiated failure diagnostics** — failure warnings now surface the errno `code` (e.g. `EXDEV`, `EACCES`, `ENOSPC`) inline, so EXDEV on a Windows redirected-profile cross-volume rename reads differently from a permission or disk-space failure, while staying non-throwing.
4. **Crash-event scrubbing parity** — grepped `scrubDesktopCrashEvent`'s call site (`packages/desktop/src/main/index.ts`) and confirmed `sensitivePaths` already scrubbed `app.getPath('userData')` before this PR. Since workspace/global storage now lives under the new `~/.texra` data root instead of `userData`, that root would otherwise leak unredacted into crash payloads. Threaded `dataRoot` out of `initializeElectronPlatform`'s return value and added it to `sensitivePaths` for parity with the existing `userData` scrubbing.

Also extended `src/test-kernel/desktop/ElectronPlatformAdapters.vitest.mts`: a new case where the legacy `workspace-storage` path is a file (readdir → ENOTDIR) proving startup doesn't abort, and added legacy-path assertions to the existing collision-warning test.

**Verification:**
- `npx vitest run src/test-kernel/desktop src/test-kernel/platform/WorkspaceStorage.vitest.ts --no-file-parallelism` → 43 files, 448 tests passed
- `npx tsc --noEmit -p packages/desktop/tsconfig.main.json` → clean

PR #8007, opened 2026-07-10
fix(export): own the attachment-marker block vocabulary in normalizeConversation

LionSR:
## The leak

`completedRunArchive.ts`'s `userMessageEntryToMessages` reconstructs a
transcript row's attachment markers by mapping each persisted
`MediaAttachmentKind` straight into a content-block literal:

```ts
...attachments.map((kind) => ({ type: kind })),
```

This silently pins the literal strings `'image'` / `'document'` that
`normalizeConversation.ts`'s `blocksToUserParts` switch recognizes as
attachment markers. Nothing enforced the two staying in sync — if
`MediaAttachmentKind` (`src/shared/schemas/progressView/data.ts`) ever grew
or renamed a member, `completedRunArchive.ts` would go on producing
`{ type: <newKind> }` blocks that `normalizeConversationForExport` silently
drops (no attachment marker, no error, no test failure at either site).

## The owner it deepens

`normalizeConversation.ts` already owns the Anthropic-shaped content-block
vocabulary — it's the file with the `blocksToUserParts` recognition switch
for `'image'`/`'document'` (plus the `'input_image'`/`'image_url'`/
`'input_file'`/`'file'` provider aliases). It now also owns *constructing*
the marker block for a given `MediaAttachmentKind`, via a new exported
`mediaAttachmentKindToContentBlock(kind: MediaAttachmentKind): ContentBlock`.
The function's body is an exhaustive switch over `MediaAttachmentKind`
(`default: assertNever(...)`), so adding or renaming a kind fails to
compile right there — at the owner — until the constructor (and the
neighboring recognition logic it sits beside) is updated. The vocabulary
can no longer drift out of sync silently.

`completedRunArchive.ts` now calls this constructor instead of building the
block itself:

```ts
...attachments.map(mediaAttachmentKindToContentBlock),
```

It imports only the `MediaAttachmentKind` type (already required to type
`attachments`) and the constructor — it holds no independent knowledge of
what an attachment-marker block looks like, so a vocabulary change that
breaks the mapping now breaks the build at the call site too (the call no
longer type-checks against a changed/removed constructor signature),
alongside the exhaustiveness failure at the owner.

## Net elements (R6)

- **+1 exported function** (`mediaAttachmentKindToContentBlock`) in the
  owner module — the inverse of the recognition logic that already lived
  there, not a new abstraction layer.
- **0 new files, 0 new exported types** (`ContentBlock` stays
  module-private; the constructor's return type is inferred from it).
- Net LOC: `normalizeConversation.ts` +31 (mostly the doc comment
  explaining the compile-time link), `completedRunArchive.ts` −8/+7
  (literal construction replaced by a call, comment updated to match),
  test suite +43 (one new round-trip case covering both
  `MediaAttachmentKind` members).

## Consumer counts (R8)

- `mediaAttachmentKindToContentBlock`: 1 production caller
  (`completedRunArchive.ts`'s `userMessageEntryToMessages`), 1 test caller
  (the new round-trip case).
- Repo-wide grep for hand-built attachment-marker blocks
  (`{ type: kind }`, `{ type: 'image' }`, `{ type: 'document' }`) across
  `src/agent/export` and `src/transcript` turns up no other duplicate
  construction site.

## Tests

- Extended `src/test-kernel/commands/history/normalizeConversation.vitest.ts`
  with a `mediaAttachmentKindToContentBlock` describe block that round-trips
  every `MediaAttachmentKind` (`'image'`, `'document'`) through
  `normalizeConversationForExport` and asserts the resulting
  `user-message`/`attachment` `ExportNode` shape — proves the constructor's
  output is exactly what the recognition switch expects.
- `src/test-kernel/transcript/completedRunArchive.vitest.ts`'s existing
  sidecar-fixture test already exercises the consumer call site end-to-end
  (`attachments: ['image']` → `{ type: 'image' }` in the reconstructed
  conversation) and continues to pass unchanged.
- `npx vitest run src/test-kernel/commands/history/normalizeConversation.vitest.ts src/test-kernel/transcript/completedRunArchive.vitest.ts` — 43 passed.
- `npm run typecheck` (all workspaces).

---

> [!NOTE]
> **Low Risk**
> Narrow refactor of how attachment markers are constructed for completed-run export; behavior for current `image`/`document` kinds should be unchanged, with stronger compile-time coupling to the schema.
> 
> **Overview**
> **Attachment-marker blocks** (`{ type: 'image' }` / `{ type: 'document' }`, no bytes) are now built only through a new exported **`mediaAttachmentKindToContentBlock`** in `normalizeConversation.ts`, with an exhaustive switch over `MediaAttachmentKind` and `assertNever` so new or renamed kinds fail at compile time instead of being dropped silently during export.
> 
> **`completedRunArchive`**’s `userMessageEntryToMessages` no longer maps kinds to `{ type: kind }` inline; it calls that helper so transcript reconstruction stays aligned with what `blocksToUserParts` recognizes.
> 
> Tests add a **round-trip** case that exercises every `MediaAttachmentKind` through `normalizeConversationForExport` and asserts the expected `user-message` / `attachment` export nodes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9a578d1abfaddcc06e4311352a0ea7155374e68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/export/normalizeConversation.ts
+++ b/src/agent/export/normalizeConversation.ts
@@ -23,6 +23,8 @@ import {
   isFunctionCallOutputItem,
 } from '@agent/modelHandlers/openai/openAIResponseContent';
 import { isResponseFunctionToolCallItem } from '@agent/modelHandlers/openai/responseStreamEvents';
+import type { MediaAttachmentKind } from '@shared/schemas';
+import { assertNever } from '@utils/core';
 import type { Part } from '@google/genai';
 import type {
   ChatCompletionMessageParam,
@@ -152,6 +154,35 @@ function googlePartToBlocks(part: Part): ContentBlock[] {
     return [{ type: mimeType?.startsWith('image/') ? 'image' : 'document' }];
   }
   return [];
+}
+
+/**
+ * Anthropic-shaped content block for a media-attachment marker (no bytes) —
+ * the two-member vocabulary `blocksToUserParts` below recognizes as
+ * `'image'` / `'document'`. This is the ONE place that vocabulary is
+ * spelled out; callers that need to synthesize an attachment marker block
+ * (e.g. reconstructing a conversation from transcript rows that only
+ * recorded the attachment kind, never the bytes) call this constructor
+ * instead of writing `{ type: kind }` themselves, so they carry no
+ * independent knowledge of the literal strings.
+ *
+ * The switch is exhaustive over {@link MediaAttachmentKind}
+ * (`src/shared/schemas/progressView/data.ts`): adding or renaming a kind
+ * there fails to compile here via `assertNever` until this function (and
+ * `blocksToUserParts`'s recognition of the same two literals) is updated —
+ * the vocabulary can't drift silently out of sync again.
+ */
+export function mediaAttachmentKindToContentBlock(
+  kind: MediaAttachmentKind,
+): ContentBlock {
+  switch (kind) {
+    case 'image':
+      return { type: 'image' };
+    case 'document':
+      return { type: 'document' };
+    default:
+      return assertNever(kind, `Unmapped media attachment kind: ${String(kind)}`);
+  }
 }
 
 function blocksToUserParts(blocks: ContentBlock[]): UserPart[] {
--- a/src/agent/export/normalizeConversation.ts
+++ b/src/agent/export/normalizeConversation.ts
@@ -181,7 +181,10 @@ export function mediaAttachmentKindToContentBlock(
     case 'document':
       return { type: 'document' };
     default:
-      return assertNever(kind, `Unmapped media attachment kind: ${String(kind)}`);
+      return assertNever(
+        kind,
+        `Unmapped media attachment kind: ${String(kind)}`,
+      );
   }
 }
 
--- a/src/test-kernel/commands/history/normalizeConversation.vitest.ts
+++ b/src/test-kernel/commands/history/normalizeConversation.vitest.ts
@@ -653,7 +653,16 @@ describe('mediaAttachmentKindToContentBlock', () => {
     // themselves. Prove the round trip for every kind the schema defines, so
     // the two stay in sync: this module's own switch (not the caller's) is
     // what breaks the build if a kind is added or renamed.
-    const kinds: MediaAttachmentKind[] = ['image', 'document'];
+    //
+    // The `Record<MediaAttachmentKind, true>` (not a plain array literal)
+    // is what makes this exhaustive: adding a member to `MediaAttachmentKind`
+    // without adding it here fails to compile, so this test can't silently
+    // stop covering a kind the way a hand-maintained array could.
+    const KIND_COVERAGE: Record<MediaAttachmentKind, true> = {
+      image: true,
+      document: true,
+    };
+    const kinds = Object.keys(KIND_COVERAGE) as MediaAttachmentKind[];
 
     for (const kind of kinds) {
       const nodes = normalize([
--- a/src/test-kernel/commands/history/normalizeConversation.vitest.ts
+++ b/src/test-kernel/commands/history/normalizeConversation.vitest.ts
@@ -11,8 +11,12 @@
 
 import { describe, expect, it } from 'vitest';
 
-import { normalizeConversationForExport } from '@agent/export/normalizeConversation';
+import {
+  mediaAttachmentKindToContentBlock,
+  normalizeConversationForExport,
+} from '@agent/export/normalizeConversation';
 import type { ExportNode } from '@agent/export/schemas';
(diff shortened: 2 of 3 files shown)


LionSR:
## Review triage

Reviewed the two Cursor/TeXRA-review passes (no issues), Copilot's review (2 line comments), and the `claude` PR comment (endorses the single-caller extraction). All review comments/reviews landed on commit `f0aa3c2` and only the format-bot commit followed, so all findings are current against HEAD.

### Fixed (pushed, `e9a578d1a`)

**Copilot — test exhaustiveness isn't enforced** (`normalizeConversation.vitest.ts:657`, ~line 649 in the review). Real: `const kinds: MediaAttachmentKind[] = ['image', 'document'];` is a plain array literal typed as `MediaAttachmentKind[]` — adding a third `MediaAttachmentKind` member wouldn't fail this test to compile, it would just silently stop covering the new kind, undermining the exact "vocabulary can't drift silently" claim the PR makes about the production code. Fixed by deriving `kinds` from `Object.keys(KIND_COVERAGE)` where `KIND_COVERAGE: Record<MediaAttachmentKind, true>` — a missing member is now a type error at the test itself. Cheap, mechanical, directly reinforces the PR's own thesis, so fixed now rather than deferred.

### Noted, not fixed (follow-up grade)

**Copilot — `completedRunArchive.ts` picks up a transitive `openai/*` runtime dependency** (`completedRunArchive.ts:15`). Verified as real, not stale: `normalizeConversation.ts` does non-type-only import `isAssistantMessage`/`isToolMessage` from `openai/lib/chatCompletionUtils` and `assertToolCallsAreChatCompletionFunctionToolCalls` from `openai/lib/parser` (plus value imports from `@agent/modelHandlers/openai/*`), and those are CJS modules esbuild can't tree-shake at the property level — so anything pulling in `mediaAttachmentKindToContentBlock` now eagerly loads that code. Before this PR, `completedRunArchive.ts` had zero provider-SDK imports; `src/transcript/index.ts` already re-exports `completedRunArchive.ts`, and CLI TUI entry points (`chatSessionController.ts`, `runChatTui.tsx`) already import from `@transcript`, so this is now on the CLI's always-loaded path regardless of which provider the user is running — the same class of thing `ModelFactory.ts`'s `ProviderHandlerLoader` lazy-loading exists specifically to avoid.

That said: I checked the actual openai/lib files in question (`chatCompletionUtils.js` is 15 lines, `parser.js` is 175 lines, importing only `../error.js`) — this isn't the full OpenAI SDK client, it's two tiny leaf utility modules, so the practical runtime/bundle cost here is negligible (not the "increase load time / bundle surface" magnitude Copilot's wording implies). The "proper" fix — splitting the marker-vocabulary (`ContentBlockSchema`/`ContentBlock`/`mediaAttachmentKindToContentBlock`) into a dependency-free module — cuts directly against this PR's own stated design goal of co-locating the constructor with the `blocksToUserParts` recognition switch it inverts, so a kind rename fails to compile at one site. That's a real trade-off (co-location vs. dependency isolation) that deserves its own deliberate call, not a rushed split in this triage pass. Filing as a note rather than a blocking fix given the low measured impact; happy to open a follow-up issue if you'd rather track it.

**Copilot — single-caller extraction** (`normalizeConversation.ts:175`, flagged in the `claude` PR comment, not a separate finding to act on): already addressed inline by the `claude` review comment with the correct reasoning — this isn't a DRY-motivated pass-through wrapper, it's a genuine exhaustive `switch` + `assertNever` deliberately co-located with the recognition logic it inverts. Matches the repo's accepted carve-out for single-caller test/vocabulary helpers with real compile-time-enforced logic. No action needed.

### Verified
- `gh pr view 8007` / `gh api .../pulls/8007/comments` for every review + line comment; confirmed comment `commit_id` (`f0aa3c2`) vs. current HEAD (`0a50190` → now `e9a578d1`) — only a formatting-only commit sits between them, diffed to confirm no logic changed.
- Read `src/agent/export/normalizeConversation.ts` (full file) and `src/transcript/completedRunArchive.ts` (imports + call site) at HEAD.
- Grepped `src/agent/runtime/ModelFactory.ts` to confirm the repo's existing lazy-provider-SDK-loading precedent (`ProviderHandlerLoader`).
- Read `node_modules/openai/lib/chatCompletionUtils.js` and `parser.js` to size the actual transitive dependency.
- Ran `npx vitest run src/test-kernel/commands/history/normalizeConversation.vitest.ts src/test-kernel/transcript/completedRunArchive.vitest.ts` (43 passed) and the full `npm run typecheck` (all workspaces, exit 0) after the fix.

PR #8005, opened 2026-07-10
refactor(agent): confine workspace-snapshot legacy migration to hydration boundary

LionSR:
## Problem

`AgentWorkspaceStateSnapshotSchema` (`src/agent/core/state/AgentWorkspaceState.ts:337-387` at the audited HEAD) is a union whose legacy arm migrates a pre-refactor top-level `{todos, plan}` shape into the current `workPlan` shape. That union was being re-evaluated far past hydration:

- `ToolUseCycleNode.prep()` — every tool-use round
- `ResponseCycleNode.prep()` / `MediaExtractionNode.prep()` — every reflection round
- `RoundPersistedFlow`'s per-step defense-in-depth revalidation (`sharedSchema.safeParse`) for the reflection flow — every single node step, not just every round

In every one of those call sites the value being parsed was never raw persisted/external data — it was this same run's own `AgentWorkspaceState.toSnapshot()` output from the previous round, already canonical. Re-running the legacy-migration union there is dead work, and worse, it means a corrupted mid-flow record that happened to look like the legacy shape would be silently "fixed" instead of failing loudly.

## Fix (APoSD: boundary gets deeper, consumers get simpler)

Split hydration from re-derivation in `AgentWorkspaceState`:

- `AgentWorkspaceState.fromSnapshot(snapshot: unknown)` — unchanged; still the legacy-migrating union parse. Kept to exactly the true hydration boundaries:
  - `ToolUsePrepareNode.exec()`'s session-init resume branch (runs once per tool-use session)
  - `runReflectionFlow()`'s resume read of `flowRecord.shared` (runs once per reflection-flow resume)
- `AgentWorkspaceState.fromCanonicalSnapshot(snapshot: AgentWorkspaceSnapshot)` — new. Takes the canonical *type*, validates only `AgentWorkspaceCurrentSnapshotSchema` (the union's existing "current format" arm, now exported instead of newly invented), never the legacy arm. Used everywhere a node re-derives state from its own prior `toSnapshot()` output: `ToolUseCycleNode.prep()`, `ResponseCycleNode.prep()`, `MediaExtractionNode.prep()`.
[...]

PR #8009, opened 2026-07-10
fix(agent): single-owner FlowRecord.shared legacy migration for tool-use resume

LionSR:
## Problem

Tool-use `FlowRecord.shared` legacy-format migration was parsed/validated
independently at two sites:

- `src/agent/runtime/SessionResumeRetrieval.ts:174-190`
  (`retrieveToolUseResumeData`)
- `src/agent/implementations/flows/tooluse/runToolUseFlow.ts:318-369`

Both called `migrateSharedState` on the same persisted `FlowRecord.shared`
bytes and independently derived `messages` / `stateSlices` /
`modelHandlerCompatibilityKey`.

## Tracing the two call paths

Every real resume enters through `retrieveSessionResumeData` **before**
`runToolUseFlow` ever runs:

- `resolveAndResumeStream` → `retrieveSessionResumeData` → builds
  `resume.snapshot` → `resumeToolUseSnapshot` → `resumeQueuedToolUseSnapshot`
  → `resumeToolUseFromSnapshot` → `runToolUseFlow({ resumeSnapshot: snapshot, ... })`.
- `nativeToolUseStrategy.runTurn` (every subsequent turn of a native
  subagent) also calls `retrieveSessionResumeData` directly, then
  `resumeToolUseFromSnapshot` with the same `resume.snapshot`.

So whenever `runToolUseFlow` receives a non-null `input.resumeSnapshot`, the
resume boundary has *already* migrated the same `flowRecord.shared` bytes and
strictly validated the result via `CurrentToolUseFlowRecordStateSchema`
(`SessionResumeRetrieval.ts`). `runToolUseFlow`'s own independent migrate +
ad hoc `modelHandlerCompatibilityKey` backfill on lines 318-369 was
re-deriving values the boundary had already computed and validated for the
exact same bytes.

The only path where `runToolUseFlow` genuinely sees a persisted record the
boundary never touched is a fresh launch (`executeAgent`/`runToolUseAgent`,
no `resumeSnapshot`) that happens to find a leftover flow record for its
execution id — that fallback is kept unchanged.

## The true read boundary

**`SessionResumeRetrieval.retrieveToolUseResumeData`** is the boundary:

- It is invoked first, chronologically, on every resume path.
- It is the only site performing full Zod validation
  (`CurrentToolUseFlowRecordStateSchema.safeParse`) of the migrated shape —
  `runToolUseFlow`'s old logic trusted `migrateSharedState`'s loose output
  with no schema check.
- Its own doc comment states its purpose: "provides functions to retrieve
  resume data from persisted state."

## Fix

`runToolUseFlow.ts` now branches on `input.resumeSnapshot`:

- **Present** (the boundary already ran): consume its validated
  `messages` / `stateSlices` (`run`/`workspace`/`user`) /
  `modelHandlerCompatibilityKey` fields directly via a new
  `buildResumedSharedFromSnapshot` helper, instead of re-deriving them.
  `migrateSharedState` is still called once here, but only for its
  *structural* unwrap (nested `{ state: {...} }` container) so pass-through
  fields outside the snapshot's narrower contract (`systemPrompt`,
  `lastError`, `userCancelledRetry`, ...) survive the self-heal write —
  these matter because a resume that continues mid-cycle (persisted cursor
  at `ToolUseWaitNode`, not `ToolUsePrepareNode`) reads `shared.systemPrompt`
  etc. straight from the persisted blob, not from the snapshot.
- **Absent** (defensive fallback; the boundary was never consulted for this
  call): unchanged — migrate + backfill the compatibility key locally so
  `PersistedFlow.ensureRecord`'s unchecked raw read never sees a stale
  legacy shape.

`SessionResumeRetrieval.ts` is untouched — it remains the sole owner.

## Boundary REGISTER entry

| Format | Owner | Consumers |
| --- | --- | --- |
| Tool-use `FlowRecord.shared` legacy migration (nested `{state}` unwrap, `conversation`→`messages` rename, `modelHandlerCompatibilityKey` backfill) | `SessionResumeRetrieval.retrieveToolUseResumeData` (`src/agent/runtime/SessionResumeRetrieval.ts`) — first read of the persisted record into memory on every genuine resume path; sole site doing strict Zod validation via `CurrentToolUseFlowRecordStateSchema` | 1: `runToolUseFlow` (`src/agent/implementations/flows/tooluse/runToolUseFlow.ts`) — consumes `input.resumeSnapshot`'s validated fields via `buildResumedSharedFromSnapshot`; never independently re-derives them when the boundary already ran |

## Net elements (R6)

+1 exported function (`buildResumedSharedFromSnapshot`, single-caller, kept
inline in `runToolUseFlow.ts` — no new module/file). No new port, facade, or
template method. Net removal of duplicate value-derivation logic (compat-key
inference/backfill) from the resume-snapshot branch of `runToolUseFlow.ts`;
the defensive no-snapshot fallback is unchanged.

## Consumer counts (R8)

- `SessionResumeRetrieval.retrieveToolUseResumeData` (the boundary): 1
  definition, invoked from 2 call paths (`resolveAndResumeStream`,
  `nativeToolUseStrategy.runTurn`) — unchanged by this PR.
- `buildResumedSharedFromSnapshot` (the new consumer-side helper): 1 caller
  (`runToolUseFlow`'s resume-snapshot branch) — justified as an extraction
  for unit-testability of the self-heal payload construction, not a
  general-purpose abstraction; verified via `grep -rn buildResumedSharedFromSnapshot src`.

## Regression proof

Added a new suite in the boundary's own test file,
`src/test-kernel/agent/SessionResumeRetrieval.vitest.ts`:

1. `normalizes a flat legacy conversation-keyed flow record for tool-use
   resume` — closes a coverage gap for the flat (non-nested) legacy
   `conversation` shape (previously only the nested `{state:{conversation}}`
   variant was covered).
2. `runToolUseFlow consumes the resume boundary instead of re-parsing >
   hydrates a legacy-shaped flow record through the single boundary, then
   self-heals via its canonical fields` — seeds a legacy nested-`state`
   record (with a `systemPrompt` pass-through field the snapshot's contract
   doesn't carry), reads it through the single boundary
   (`retrieveSessionResumeData`), then asserts `buildResumedSharedFromSnapshot`
   reproduces the boundary's canonical `messages` / `stateSlices` /
   `modelHandlerCompatibilityKey` values while preserving `systemPrompt`.

Proved the regression case without `git stash` (`git diff` → patch,
`git checkout --` to revert just `runToolUseFlow.ts`, ran the suite, `git
apply` to restore):

```
# before the fix (runToolUseFlow.ts reverted to origin/main):
✗ hydrates a legacy-shaped flow record through the single boundary...
  TypeError: buildResumedSharedFromSnapshot is not a function
  (9/10 other tests in the file still pass)

# after `git apply` restores the fix:
✓ 10/10 tests pass
```

## Tests

- `npx vitest run src/test-kernel/agent/SessionResumeRetrieval.vitest.ts` — 10/10 pass
- `npx vitest run src/test-kernel/agent/runtime/resumeToolUseSnapshot.vitest.ts src/test-kernel/agent/followUp/ToolUseFollowUp.vitest.ts src/test-kernel/agent/runtime/AgentRunLifecycle.vitest.ts src/test-kernel/agent/runtime/ExecutionRegistry.vitest.ts src/test-kernel/tools/NativeToolUseStrategy.vitest.ts` — 83/83 pass
- `npx vitest run src/test-kernel/agent` (full tree) — 1233 passed, 4 skipped
- `npx tsc --noEmit -p tsconfig.json` — clean
- `npx tsc --noEmit -p tsconfig.test-kernel.json` — clean
- `npx eslint` on both changed files — clean

---

> [!NOTE]
> **Medium Risk**
> Touches persisted session resume and self-heal writes for tool-use flows; logic is narrowed to avoid drift between two parsers but still affects mid-cycle resume state.
> 
> **Overview**
> **Tool-use resume** no longer migrates or re-derives `FlowRecord.shared` twice. When `runToolUseFlow` receives `resumeSnapshot` from `SessionResumeRetrieval`, it self-heals the KV blob via new `buildResumedSharedFromSnapshot`, taking canonical `messages`, state slices, and `modelHandlerCompatibilityKey` from that snapshot while `migrateSharedState` only unwraps structure so pass-through fields like `systemPrompt` survive.
> 
> If there is no `resumeSnapshot` (e.g. leftover record on a fresh launch), the existing local migrate + compatibility-key backfill path is unchanged.
> 
> Tests add coverage for flat legacy `conversation` records and assert the boundary + self-heal helper stay aligned without duplicate parsing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c237e7589f45842e2229f0cc1f25a31b8b21d5f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/implementations/flows/tooluse/runToolUseFlow.ts
+++ b/src/agent/implementations/flows/tooluse/runToolUseFlow.ts
@@ -113,6 +113,33 @@ export function getToolUseFlowErrorResult(
   return error instanceof ToolUseFlowError ? error.result : undefined;
 }
 
+/**
+ * Build the canonical self-heal payload for a resumed flow record's `shared`
+ * blob from the resume boundary's already-validated snapshot
+ * (`SessionResumeRetrieval.retrieveToolUseResumeData` -- the single owner of
+ * FlowRecord.shared's legacy-format migration and modelHandlerCompatibilityKey
+ * backfill). `structuralBase` is the record's own `migrateSharedState` output
+ * (structural unwrap only), which supplies any pass-through fields the
+ * snapshot's narrower resume contract doesn't carry (e.g. `systemPrompt`,
+ * `lastError`) so they survive the write-back untouched.
+ */
+export function buildResumedSharedFromSnapshot(
+  structuralBase: ToolUseRunShared,
+  snapshot: ToolUseSessionSnapshot,
+): ToolUseRunShared {
+  return {
+    ...structuralBase,
+    messages: snapshot.messages,
+    modelHandlerCompatibilityKey:
+      snapshot.modelHandlerCompatibilityKey ?? undefined,
+    stateSlices: {
+      runStateSnapshot: snapshot.run,
+      workspaceSnapshot: snapshot.workspace,
+      userChannels: snapshot.user,
+    },
+  };
+}
+
 interface ToolUseFlowContext {
   readonly ownerSession: SessionHandle;
   readonly session: ToolUseSessionLifecycle;
@@ -331,7 +358,32 @@ export async function runToolUseFlow<C = unknown>(
         logger.warn('Failed to parse flow record shared state, starting fresh');
         await kv.delete(flowKey(executionId));
         flowRecord = null;
+      } else if (input.resumeSnapshot) {
+        // The resume boundary (SessionResumeRetrieval.retrieveToolUseResumeData)
+        // already migrated this record's legacy shapes and strictly validated
+        // the result into `input.resumeSnapshot` before this flow was ever
+        // launched -- it is the single owner of FlowRecord.shared's
+        // legacy-format parsing and modelHandlerCompatibilityKey backfill (see
+        // its CurrentToolUseFlowRecordStateSchema). Consume its canonical
+        // fields directly here instead of re-deriving them; `migrateSharedState`
+        // above is only used for its structural unwrap so that pass-through
+        // fields the snapshot's narrower contract doesn't carry (systemPrompt,
+        // lastError, ...) survive this self-heal write of the KV blob that
+        // PersistedFlow.ensureRecord's unchecked raw read (below) depends on.
+        flowRecord.shared = buildResumedSharedFromSnapshot(
+          migrationResult.data,
+          input.resumeSnapshot,
+        );
+        await kv.write(
+          flowKey(executionId),
+          stampFlowRecordSchemaVersion(flowRecord),
+        );
       } else {
+        // Defensive fallback only: no resumeSnapshot means the resume
+        // boundary above was never consulted for this call (e.g. a fresh
+        // launch that happens to find a leftover record for its execution
+        // id). Migrate/backfill here so PersistedFlow.ensureRecord's
+        // unchecked raw read never sees a stale legacy shape.
         let migratedData = migrationResult.data;
         let shouldWriteShared = migrationResult.migrated;
         const sharedModel = migratedData.stateSlices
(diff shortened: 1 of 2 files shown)


LionSR:
Merge disposition: cursor's two Low notes acknowledged — buildResumedSharedFromSnapshot is a local named function exported for direct test access (the alternative, testing only through the full flow, would need a heavier harness; accepting the export as test surface, not a shared abstraction), and the rationale comments will thin naturally when #8005's P1 fix lands on this boundary (see cross-PR note there).

PR #8011, opened 2026-07-10
fix(tools): stop re-deriving KV filename vocabulary in executionKvFiles

LionSR:
## The leak

`src/tools/executions/executionKvFiles.ts` hand-copied the KV filename
vocabulary its two owners already hold privately:

- `ExecutionKVStore.ts:44-53`'s private `KEYS` map (explicitly commented
  "implementation detail — not exported"), including the `child-` prefix
  used for per-child execution records.
- `persistedFlow.ts:13-16`'s `flowKey()`, whose doc comment already claims
  to be the "single source of truth" for the `flow_` prefix — a claim that
  `executionKvFiles.ts` quietly violated by re-deriving `'flow_'` itself.

Both owners could drift a key/prefix and `isKVFile` would silently stop
filtering it out of `/executions/{id}/files` and the CLI's history file
listing (`packages/cli/src/runtime/history/generatedFiles.ts`), leaking
internal KV blobs into user-facing file listings.

## The fix — owners deepen, leak is deleted

- `ExecutionKVStore.ts` now exports `isReservedKvKeyName(key)`, derived
  directly from its `KEYS` map (a new `SINGLE_VALUE_KEYS` object feeds both
  `KEYS` and the reserved-name `Set`), covering the seven single-value keys
  plus the `child-` prefix.
- `persistedFlow.ts` now exports `FLOW_KEY_PREFIX`; `flowKey()` is rewritten
  to build off it instead of the inline `'flow_'` literal.
- `executionKvFiles.isKVFile` is rewritten to strip the on-disk `.json`
  suffix and defer to `isReservedKvKeyName` + `FLOW_KEY_PREFIX` — the
  hand-copied `KV_FILES` literal `Set` is **deleted**.
- Same-owner cleanup while in `ExecutionKVStore.ts`: its own two internal
  `'child-'` re-derivations (`readChildren`'s `listKeys('child-')` and
  `key.replace('child-', '')`) now route through the same
  `CHILD_KEY_PREFIX` constant instead of repeating the literal.

`src/tools/` already depends on `@agent/storage` elsewhere, so this adds no
new import-boundary exception (`src/tools` and `src/agent/*` are both
VS Code-free zones).

## Net elements (R6)

- **+2 exports**: `isReservedKvKeyName` (ExecutionKVStore), `FLOW_KEY_PREFIX`
  (persistedFlow) — both are the owning modules deepening their existing
  public surface, not new abstractions.
- **-1 hand-copied literal set** (`KV_FILES`) and **-2 inline `'child-'`
  literals** in `ExecutionKVStore.ts`.
- **0 new files, 0 new layers/facades.**
- Net: `+119 / -17` lines across 7 files, but the production-code delta is
  `+41 / -17` (the rest is regression test coverage); the only production
  file that shrinks in independent knowledge is `executionKvFiles.ts`
  (22 lines touched, net -0, but it no longer owns any vocabulary of its
  own — every reserved name/prefix it recognizes now traces to exactly one
  owning module).

## Consumer counts (R8)

- `isKVFile` (the leaking function) has exactly 2 call sites, both already
  existing: `src/tools/executions/runDirectoryFiles.ts` and
  `packages/cli/src/runtime/history/generatedFiles.ts` (aliased as
  `isHistoryKvFile`). Both now transitively depend on the two real owners
  instead of a third, independently-drifting copy — this PR does not add a
  new consumer, it collapses the vocabulary the existing 2 consumers see
  down to 1 owner each (`ExecutionKVStore`, `persistedFlow`).
- `isReservedKvKeyName` has 1 caller today (`executionKvFiles.ts`) — no new
  single-caller helper was invented for its own sake; it is an export of an
  already-private map that a second module needed to stop re-deriving.

## Tests

- `src/test-kernel/agent/storage/ExecutionKVStore.vitest.ts`: new
  `describe('isReservedKvKeyName', …)` block pinning all 7 single-value
  keys, the `child-` prefix, and rejected non-reserved keys.
- `src/test-kernel/agent/PersistedFlow.vitest.ts`: new test pinning
  `flowKey()` is built from the exported `FLOW_KEY_PREFIX`.
- `src/test-kernel/tools/ExecutionsToolWorkspaceFiles.vitest.ts`: new
  end-to-end regression test that writes every reserved KV filename
  (including `child-*.json` and `flow_*.json`) plus a generated file into a
  real run directory and asserts `/executions/{id}/files` still filters all
  of them out through the real `ExecutionsTool` → `listRunDirectoryFiles` →
  `isKVFile` chain.

Ran: `npx tsc --noEmit` (root), full `npm run typecheck` (all workspaces),
targeted vitest suites above plus the broader `src/test-kernel/tools`,
`src/test-kernel/agent/storage`, and `PersistedFlow`/`RoundPersistedFlow`
suites (82 files / 532 tests, all green), and `eslint` on every changed
file (clean after one auto-fixed import-order warning).

---

> [!NOTE]
> **Low Risk**
> Behavioral change is limited to file-list filtering logic with explicit regression tests; no auth, persistence format, or runtime execution path changes.
> 
> **Overview**
> **Stops `executionKvFiles` from maintaining a duplicate list of internal KV filenames**, so `/executions/{id}/files` and CLI history listings stay aligned when storage keys change.
> 
> `ExecutionKVStore` now exports **`isReservedKvKeyName`** (derived from its key map plus `child-` prefix) and **`persistedFlow`** exports **`FLOW_KEY_PREFIX`**; `flowKey()` builds on that constant. **`isKVFile`** drops the hand-copied `KV_FILES` set, requires a **`.json`** suffix before classifying a name, then delegates to those owners. **`readChildren`** uses the same `CHILD_KEY_PREFIX` instead of inline `'child-'` literals.
> 
> Regression tests cover reserved-key recognition, end-to-end filtering via `ExecutionsTool`, and **extensionless** files named like reserved keys (must remain visible).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f1a9fbb659f6d1204dba3271de494c75093105e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/node/persistedFlow.ts
+++ b/src/agent/node/persistedFlow.ts
@@ -10,9 +10,12 @@ import { toErrorMessage } from '@utils/errors/errorMessage';
 import { BaseNode, Flow, type Action } from '.';
 import type { z } from 'zod';
 
+/** Prefix for a flow record's KV key. Single source of truth for callers deriving it. */
+export const FLOW_KEY_PREFIX = 'flow_';
+
 /** KV key for a flow record. Single source of truth for the prefix. */
 export function flowKey(runId: string): string {
-  return `flow_${runId}`;
+  return `${FLOW_KEY_PREFIX}${runId}`;
 }
 
 const CHANNEL = 'PersistedFlow';
--- a/src/agent/storage/ExecutionKVStore.ts
+++ b/src/agent/storage/ExecutionKVStore.ts
@@ -59,7 +59,7 @@ const KEYS = {
   child: (id: string) => `${CHILD_KEY_PREFIX}${id}`,
 } as const;
 
-/** Single-value keys, derived from KEYS so the reserved-name check below never drifts. */
+/** Single-value keys, derived from SINGLE_VALUE_KEYS so the reserved-name check below never drifts. */
 const RESERVED_KEY_NAMES = new Set<string>(Object.values(SINGLE_VALUE_KEYS));
 
 /**
--- a/src/agent/storage/ExecutionKVStore.ts
+++ b/src/agent/storage/ExecutionKVStore.ts
@@ -41,7 +41,10 @@ import {
 // Key constants (implementation detail — not exported)
 // ============================================================================
 
-const KEYS = {
+/** Prefix for a per-child execution record's KV key. */
+const CHILD_KEY_PREFIX = 'child-';
+
+const SINGLE_VALUE_KEYS = {
   META: 'meta',
   CONFIG: 'config',
   REPORT: 'report',
@@ -49,8 +52,27 @@ const KEYS = {
   CONVERSATION: 'conversation',
   WORKSPACE_FILES: 'workspace-files',
   RESULT_META: 'result-meta',
-  child: (id: string) => `child-${id}`,
 } as const;
+
+const KEYS = {
+  ...SINGLE_VALUE_KEYS,
+  child: (id: string) => `${CHILD_KEY_PREFIX}${id}`,
+} as const;
+
+/** Single-value keys, derived from KEYS so the reserved-name check below never drifts. */
+const RESERVED_KEY_NAMES = new Set<string>(Object.values(SINGLE_VALUE_KEYS));
+
+/**
+ * True when `key` is one of ExecutionKVStore's reserved keys — a single-value
+ * key (meta, config, report, todos, conversation, workspace-files,
+ * result-meta) or a per-child record key (`child-{id}`). Exported so callers
+ * that walk an execution's storage directory (e.g.
+ * `src/tools/executions/executionKvFiles.ts`) can recognize internal KV
+ * entries without re-deriving this vocabulary themselves.
+ */
+export function isReservedKvKeyName(key: string): boolean {
+  return RESERVED_KEY_NAMES.has(key) || key.startsWith(CHILD_KEY_PREFIX);
+}
 
 const CHANNEL = 'ExecutionKVStore';
 export const EXECUTION_META_SCHEMA_VERSION = 1;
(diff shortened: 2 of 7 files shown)


LionSR:
## Review disposition

All review comments were on commit `1876b96` (the only commit at review time) and remain applicable to HEAD as of this reply.

### Fixed — bare reserved-name false-positive filtering (real bug, confirmed)

Cursor Bugbot (medium), Codex (P2), and Copilot independently flagged the same defect in `isKVFile`:

```ts
const key = name.replace(/\.json$/, '');
return isReservedKvKeyName(key) || key.startsWith(FLOW_KEY_PREFIX);
```

`String.replace(/\.json$/, '')` is a no-op when the name has no `.json` suffix, so `key` falls through unchanged to `isReservedKvKeyName`. A generated file whose *basename* happens to equal `meta`, `config`, `report`, `todos`, `conversation`, `workspace-files`, or `result-meta` — with no extension — was therefore wrongly hidden from `/executions/{id}/files` and the CLI's history listing, where the pre-PR `KV_FILES` set only ever matched the full `name.json` string. Verified by tracing `KVStore.keyToPath` (`src/common/storage/KVStore.ts`): every real KV entry is written as `` `${key}.json` `` on disk, so a name without that suffix can never legitimately be one.

Fix (`src/tools/executions/executionKvFiles.ts`): require the `.json` suffix before applying the reserved-name/prefix check at all — `isKVFile` now returns `false` immediately for any name that isn't `*.json`, then defers to the same owner-exported vocabulary as before for names that are. This also formally decides the `child-`/`flow_` prefix question the pre-PR code left inconsistent (those prefix checks ran on the raw name regardless of extension); since real prefixed entries are always written as `.json` too, gating uniformly on the suffix changes no real-world behavior while eliminating the reported false positive. Added a regression test (`ExecutionsToolWorkspaceFiles.vitest.ts`) writing extensionless files named `meta`, `config`, `report`, `child-def456` and asserting they now survive the `/executions/{id}/files` filter.

### Fixed — stale comment (Copilot, wording-accuracy nit)

`RESERVED_KEY_NAMES`'s doc comment said "derived from KEYS" but the code derives it from `SINGLE_VALUE_KEYS` (`Object.values(SINGLE_VALUE_KEYS)`). Corrected the comment to name the actual source.

### Verification

- `npx tsc --noEmit` (root) — clean.
- `npm run typecheck` (all workspaces: root, test-kernel, cli, trace-viewer, desktop) — clean.
- Targeted: `ExecutionsToolWorkspaceFiles.vitest.ts`, `ExecutionKVStore.vitest.ts`, `PersistedFlow.vitest.ts` — 61/61 passed, including the new regression test.
- Broader: `src/test-kernel/tools`, `src/test-kernel/agent/storage` — 80 files / 524 tests passed.
- `PersistedFlow.vitest.ts` + `RoundPersistedFlowCompileRepair.vitest.ts` — 9/9 passed.
- `eslint` on all three touched files — clean.

Pushed as `8f1a9fbb6` on `claude/leak-kv-vocabulary`.

PR #8013, opened 2026-07-10
refactor(agent): delete resume-time system-message refresh — preserve provider prompt-cache prefix

LionSR:
## Reworked per maintainer ruling

@LionSR's comment on the original relocation — **"no this breaks prompt
caching!!!!!!! Should not update"** — is a ruling on the *behavior*, not on
where it lived. The prior version of this PR moved the ~75-line resume-time
system-message refresh (`refreshPersistedSystemMessage` /
`withFirstBlockReplaced`) from `ToolUsePrepareNode` into a `ModelHandler`
port method (`refreshSystemMessage` / `refreshFirstContentBlockText`) —
correct relocation, but it kept the refresh happening on every resume,
which the maintainer is right to reject: rewriting `messages[0]` with
freshly rebuilt system text on every tool-use resume changes the leading
bytes of the request and invalidates the provider's prefix-based prompt
cache (OpenAI Chat/Responses, OpenRouter).

**This version deletes the refresh outright instead of relocating it.**
Resumed tool-use runs now keep their persisted system message
byte-identical to what was snapshotted, preserving the provider's cached
prefix.

## What changed

- `src/agent/modelHandlers/ModelHandler.ts`: `refreshSystemMessage` and its
  `refreshFirstContentBlockText` helper (plus `TEXT_BLOCK_TYPES`) are
  deleted entirely. The file is now byte-identical to `main`.
- `src/agent/types/IModelHandler.ts`: the `refreshSystemMessage` `Pick`
  entry is removed. Byte-identical to `main`.
- `src/agent/implementations/flows/tooluse/nodes/ToolUsePrepareNode.ts`:
  the resume branch passes `snapshot.messages` straight through — no call
  into the model handler, no local re-derivation. (The `refreshPersistedSystemMessage`
  / `withFirstBlockReplaced` local functions this PR originally relocated
  stay deleted, as they were on the prior version of this branch.)
- `systemPrompt` (used only to resupply `system` per-call to Anthropic/Google,
  which never store it in `messages` at all) is still rebuilt fresh on
  resume. That mechanism was introduced by a separate, already-on-`main`
  bug fix (`92159f74a`, fixing tool-use rounds silently running with *no*
  system prompt for Anthropic/Google) and is orthogonal to the `messages[0]`
  prefix-caching concern this PR addresses — it's a per-call parameter, not
  part of the cached message array, and it already gets resupplied on every
  round regardless of resume.

## Net elements (R6)

Production code only:

- `ToolUsePrepareNode.ts`: net **-8** (the refresh call site collapses to a
  direct field assignment; the comment explains the new cache-preservation
  contract instead of the provider-shape mechanics).
- `ModelHandler.ts` / `IModelHandler.ts`: net **-59** combined — fully
  reverted to `main`.

Overall vs. `main`: **2 files touched** (down from 5), **+58/-93** (down
from the prior version's +294/-89). Strongly net-negative, as expected for
a straight deletion.

## Consumer counts (R8)

`grep -rn "refreshSystemMessage\|refreshFirstContentBlockText\|refreshPersistedSystemMessage\|withFirstBlockReplaced"` across `src/` and
`packages/`: **0 matches**. No handler ever overrode the base method, and
nothing else called it — confirmed clean deletion, not a shim.

## Tests

**Deleted** (asserted the refresh behavior that no longer exists):
- `src/test-kernel/agent/modelHandlers/ModelHandlerContinuationContract.vitest.ts`:
  the entire `describe('model handler system message refresh contract
  (tool-use resume)')` block — 7 cases (OpenAI chat, OpenAI Responses, o1-style,
  OpenRouter-native-with-Anthropic-config, and the 3-case Anthropic/Google
  GenAI/Google Interactions no-op loop).
- `src/test-kernel/agent/followUp/ToolUsePrepareNodeTranscriptLog.vitest.ts`:
  the `describe('ToolUsePrepareNode resume (system message refresh
  delegation)')` test that asserted delegation to the port method.

**Added** — the new cache-preservation regression test, in the same file:
`describe('ToolUsePrepareNode resume (prompt-cache preservation)')` asserts
`result.result.messages` is `toBe` (reference-identical to)
`snapshot.messages`, i.e. resume never rewrites `messages[0]`.

Net test count: 606 → 599 (7 removed, 1 removed + 1 added elsewhere), all
passing.

## Ran

- `npx vitest run` on the PR's original targeted suites (modelHandlers,
  followUp, `runtime/resumeToolUseSnapshot`, `SessionResumeRetrieval`,
  `SessionResumeResolution`, `SessionResumeExitPolicy`,
  `ResumeExtensionToolUseSnapshot`, `ToolUseDispatchParallel`) — **81 files
  / 599 tests passed**, 4 pre-existing skips, unrelated.
- `npm run typecheck` (full workspace: root, test-kernel, `texra`,
  `@texra-ai/cli`, `@texra/trace-viewer`, `@texra/desktop`) — clean.
- `npx eslint` on all changed files — clean.

## Tradeoff, made explicit

A system-prompt or tool-config edit made **between suspend and resume no
longer propagates into an already-suspended run's prefix** — the resumed
run keeps whatever system text was frozen at snapshot time. This is the
maintainer's explicit call: staleness is preferable to invalidating the
provider's prefix-based prompt cache on every resume.

**Alternative reading considered and not taken:** the comment could also
be read more narrowly as "don't update *unconditionally* — only rewrite
when the rebuilt text actually differs from what's persisted" (cache-neutral
in the common case). I implemented the plain reading (never touch
`messages[0]` on resume) rather than a conditional-diff variant, since (a)
it's the simplest correct interpretation of "should not update", (b) it
needs zero new mechanism, and (c) a diff-based variant still has to inspect
provider message shape to compare against, which reintroduces exactly the
kind of `messages[0]`-shape knowledge this PR is trying to keep out of both
`ToolUsePrepareNode` and `ModelHandler`. Happy to add the diff-gated
variant instead if that's what's wanted.


LionSR:
no this breaks prompt caching!!!!!!!


Shout not update

LionSR:
## Review disposition

Read every review (2× bot review, `claude[bot]` finding comment, Copilot inline comments, Copilot's summary review) plus the owner comment. Verified each against HEAD (`2173c9425`) in a fresh worktree; ran targeted vitest, full `npx tsc --noEmit`, and eslint on the touched file after the fix below.

### 1. @LionSR: "no this breaks prompt caching!!!!!!! Should not update" — **BLOCKED, needs your call**

Verified this is **not new behavior introduced by this PR**. `git log -p` on `origin/main` shows `refreshPersistedSystemMessage`/`withFirstBlockReplaced` — the exact logic this PR relocates — was introduced 3 days before this PR opened, in `91fcffacb6` ("refactor(agent): avoid recomputing systemText on tool-use resume", Jul 7 10:24) and `92159f74a` ("fix(agent): supply systemPrompt to tool-use round flow", Jul 7 10:15), both already on `main`. This PR (`2db88b0f3`) is a byte-for-byte behavior-preserving move of that logic from `ToolUsePrepareNode` into `ModelHandler.refreshSystemMessage` — confirmed independently by both the `texra-review` bot ("body... structurally identical... byte-for-byte identical") and my own diff of the deleted functions vs. the new ones.

So the prompt-caching interaction you're flagging exists on `main` today, independent of this PR: on tool-use resume, `messages[0]` (system role for OpenAI Chat/OpenRouter, or the first content block for o1-style `supportsSystemPrompt:false` handlers) gets its text replaced with freshly rebuilt system text reflecting current workspace/tool state — which, when it differs from what was last sent, invalidates the provider's prefix-based prompt cache for that conversation on every resume. Anthropic/Google are unaffected (`requiresPerCallSystemPrompt` no-ops them; they never store `system` in `messages`).

Because this PR is a pure relocation, I did **not** unilaterally change the refresh behavior — reverting it, gating it, or making it cache-aware is a product tradeoff (fresh workspace state on resume vs. cache cost) that isn't mine to make. What I can tell you: merging this PR as-is doesn't make the caching situation any worse than `main` already is, but it does make the design more "official" (a base-class port method every handler inherits, vs. one call site) — which raises the stakes of getting the tradeoff right before it spreads further. Options for you to pick from: (a) revert the resume-refresh feature entirely (back to using the snapshot's frozen system text verbatim), (b) skip the `messages[0]` rewrite specifically for the chat-completions-style providers that benefit from prefix caching (OpenAI Chat, OpenRouter) while keeping it for o1-style/Responses, or (c) only rewrite when the new system text actually differs from what's persisted (cache-neutral when nothing changed, which is the common case). Happy to implement whichever direction you want in a follow-up.

### 2. Copilot inline: `refreshFirstContentBlockText` doc/code mismatch (line 193, `ModelHandler.ts`) — **fixed**

Confirmed real: the guard only checked `typeof type !== 'string'`, so a first content block of, e.g., `type: 'image_url'` would pass and get a stray `text` field spliced onto it (structurally invalid for the OpenAI API), contradicting the function's own doc comment ("non-text first block: leave the message alone"). This is inherited unchanged from the pre-existing `withFirstBlockReplaced` (same bug on `main` today, not introduced by this PR) — but since this PR is the one moving/renaming the function, and the fix is a minimal one-line guard, I fixed it here rather than deferring: restricted the rewrite to the two known text-carrying block types (`'text'`, `'input_text'`) via `TEXT_BLOCK_TYPES`. Pushed as `2173c9425`. Ran the full `ModelHandlerContinuationContract`/`ToolUsePrepareNodeTranscriptLog` suites (20 tests) plus the PR's full targeted suite (592 passed/4 skipped, pre-existing skips unrelated), `npx tsc --noEmit` clean, `eslint` clean on the changed file.

### 3. `texra-review` bot: `supportsSystemPrompt` truthiness change (line 1076) — **dismissed, bot itself marked it safe**

The bot's own comment concludes "No fix needed" — `ModelCapabilities.supportsSystemPrompt` is a required `boolean` (verified `openAIMessageUtils.ts:251`), always populated via `structuredClone(config.capabilities)` in the constructor, so `undefined` is unreachable. Not an actionable finding.

### 4. Copilot inline: `buildSnapshot` test helper casts an incomplete object to `ToolUseSessionSnapshot` (line 108, `ToolUsePrepareNodeTranscriptLog.vitest.ts`) — **dismissed, consistent with existing pattern**

Real observation, but matches the established test-kernel convention already used two functions above it in the same file (`buildServices` does the identical `as ToolUseServices<unknown>` cast) and throughout the test-kernel generally for test-double builders. Not a new pattern this PR introduces; not worth a special-cased fully-typed construction here given the accepted convention.

### Verified
- `git log -p origin/main -- src/agent/implementations/flows/tooluse/nodes/ToolUsePrepareNode.ts` — confirmed `refreshPersistedSystemMessage`/`withFirstBlockReplaced` predate this PR by 3 days (commits `91fcffacb6`, `92159f74a`, both on `main` before this PR opened).
- `git show 2db88b0f3^:.../ToolUsePrepareNode.ts` diffed against the new `ModelHandler.ts` code — confirmed the move is behavior-preserving.
- `src/agent/modelHandlers/ModelHandler.ts` (full diff + surrounding `refreshSystemMessage`/`refreshFirstContentBlockText`) — applied and verified the block-type guard fix.
- `src/agent/modelHandlers/openai/openAIMessageUtils.ts:230-330` (`initializeChatMessages`) — confirmed how `messages[0]` content is built for o1-style handlers, and that `insertMediaIntoChatUserMessage`/`addMediaToUserMessage` (`modelHandlerOpenAI.ts:1310-1323`, `modelHandlerOpenRouterNative.ts:755`) can unshift non-text blocks onto the *last* user message, which is reachable at index 0 in a single-round o1-style conversation — the scenario the guard fix protects against.
- `src/test-kernel/agent/followUp/ToolUsePrepareNodeTranscriptLog.vitest.ts` (full file) — confirmed `buildSnapshot`'s cast pattern matches `buildServices` above it.
- Ran: targeted vitest (`ModelHandlerContinuationContract`, `followUp`, `runtime/resumeToolUseSnapshot`, `SessionResumeRetrieval`, `ToolUseDispatchParallel`) — 592 passed / 4 skipped (pre-existing); `npx tsc --noEmit` — clean; `eslint src/agent/modelHandlers/ModelHandler.ts` — clean.

**Verdict: blocked** on item 1 pending your design call on the resume-refresh-vs-prompt-caching tradeoff (pre-existing on `main`, not introduced here). Item 2 fixed and pushed; items 3-4 dismissed with evidence above.

LionSR:
@LionSR — reworked per your ruling: this is no longer a relocation, it's a
deletion (`2173c9425` → `fb0e8d6dc`).

**What changed:** `ModelHandler.refreshSystemMessage` /
`refreshFirstContentBlockText` and the `IModelHandler` port entry are gone
— both files are now byte-identical to `main`. `ToolUsePrepareNode`'s
resume branch passes `snapshot.messages` straight through with no
rewrite. Net production diff flipped from **+205** (5 files) to **-35**
(2 files) vs. `main`.

**Tradeoff, stated explicitly:** a system-prompt or tool-config edit made
between suspend and resume no longer propagates into an
already-suspended run's prefix — the resumed run keeps whatever system
text was frozen at snapshot time, in exchange for never invalidating the
provider's prefix-based prompt cache on resume.

**One thing I left untouched on purpose:** `systemPrompt` is still
rebuilt fresh on resume for Anthropic/Google, which pass `system`
per-call rather than storing it in `messages` at all — that mechanism is
a separate, already-on-`main` fix (`92159f74a`) for tool-use rounds
silently running with *no* system prompt for those providers, and it's
orthogonal to the `messages[0]` prefix-caching concern here (it's a
per-call parameter, already resupplied every round regardless of
resume, never part of the cached array). Flag if you want that touched
too.

**Alternative reading I considered and didn't take:** "should not update"
could mean "don't rewrite unconditionally, only when the text actually
changed" (cache-neutral in the common case) rather than "never rewrite."
I went with the plain reading (never touch `messages[0]` on resume) since
a diff-gated variant would still need to inspect provider message shape
to compare against — reintroducing the exact knowledge this PR removes.
Say the word if you meant the narrower version instead.

Tests: deleted the 7 refresh-behavior cases + the delegation test;
added one regression test asserting the resumed messages array is
reference-identical to the persisted snapshot. 599/599 targeted tests
pass, full `npm run typecheck` clean, eslint clean. Full detail in the
PR body.

PR #8014, opened 2026-07-10
fix(transcript): eager-apply overlay for missing outputs and compile failures

LionSR:
## The race

`StreamSnapshotStore` is the single writer of `streamData/{id}/*`. When a
stream hasn't been seeded from disk yet, every mutator has to choose between
applying to memory now (so a caller can read its own write back) and letting
the in-flight seed's `applyStreamData()` — which does
`this.<accumulator>.set(stream, data.<accumulator>)`, an unconditional
overwrite — land after it and clobber the update.

`addOutputFiles`/`addUsage` already carry the fix (added in 609bf0046, "fix:
keep snapshot output usage updates readable"): apply to memory eagerly, and
if the stream isn't seeded yet, additionally record the patch in a
per-accumulator overlay map. `applyStreamData()` replays that overlay on top
of the freshly-read disk state once the seed lands, so the eager write can
never be clobbered.

`updateMissingOutputs`/`updateCompileFailures` never got this treatment —
they still went through the plain `mutate()` path, which *defers the entire
apply* until after the seed resolves. Concretely:

1. Stream `X` isn't seeded (e.g. it was outside a partial `preload()`).
2. `updateMissingOutputs(X, { 1: ['next.tex'] })` calls `mutate()`, which sees
   the stream isn't synchronously mutable and queues the whole apply behind
   `ensureSeeded()` (the in-flight disk read for `X`) via `queueAfterSeed`.
3. A caller that reads `getMissingOutputs(X)` in the same tick — before the
   seed's disk read resolves — sees **the pre-mutation value**, not what it
   just wrote. Anything that branches on that read (surfacing a missing-file
   marker in the UI, or persisting a decision made off `getMissingOutputs`)
   silently loses the update until the seed's microtask eventually catches up
   — the same lost-update class `addOutputFiles`/`addUsage` were fixed for,
   just half-applied across the four mutators.

Two new regression tests
(`src/test-kernel/transcript/StreamSnapshotStore.vitest.ts`) reproduce this
exact interleaving — `preload([STREAM])` leaves `OTHER_STREAM` unseeded, then
`updateMissingOutputs`/`updateCompileFailures` on `OTHER_STREAM` is
immediately followed by a synchronous `getMissingOutputs`/
`getCompileFailures` read, mirroring the existing "returns output files
immediately for streams outside a partial preload" test for
`addOutputFiles`. Verified failing pre-fix (checked out against
`origin/main`'s `StreamSnapshotStore.ts` with the new tests applied):

```
AssertionError: expected undefined to deeply equal [ 'next.tex' ]
AssertionError: expected undefined to deeply equal [ { round: 1, …(4) } ]
```

## The ownership fix

Doctrine: the cure for an internal race is single-owner/sequenced design
that makes the race impossible, not caller discipline. Rather than
hand-rolling a third bespoke eager-apply+overlay block for
`updateMissingOutputs`/`updateCompileFailures`, this generalizes the shape
`addOutputFiles`/`addUsage` already proved into one shared internal helper,
`mutateWithOverlay`, and deletes the two bespoke overlay blocks
(`addOutputFilesOverlay`, `addUsageOverlay`) in favor of it:

- `mutateWithOverlay<T, P>(stream, overlayPatch, overlay, mergePatch,
  applyToMemory, persist)` always applies to memory immediately. A seeded
  stream also persists immediately. An unseeded stream instead merges
  `overlayPatch` into `overlay` (via `mergePatch`) and queues a no-op behind
  the seed so `applyStreamData()`'s reconciliation picks it up.
- `applyStreamData()`'s post-seed reconciliation now handles all three
  round-keyed overlays (output files, missing outputs, compile failures)
  the same way, plus the pre-existing usage overlay.
- `parseRoundPatch`/`applyRoundPatch` replace three near-duplicate
  functions (`applyRoundKeyedPatch`, `parseOutputFilesPatch`,
  `applyOutputFilesPatch`) with one generic round→value-patch parse/apply
  pair shared by all three round-keyed accumulators.
- `addOutputFiles`, `updateMissingOutputs`, `updateCompileFailures`, and
  `addUsage` — all four mutators — now call `mutateWithOverlay`, so the
  guarantee is structural: a fifth round/usage mutator added later gets it
  by construction instead of needing to remember to hand-roll it.

`setTodos`/`setPlan`/`clearMissingOutputs`/`queueMetaPatch` still use the
existing plain `mutate()` (and `queueMetaPatch`'s own pre-existing meta
overlay) — out of scope; this PR only touches the four round/usage patch
mutators named in the finding.

## Net elements (R6)

- **Deleted:** `applyRoundKeyedPatch`, `parseOutputFilesPatch`,
  `applyOutputFilesPatch`, `addOutputFilesOverlay`, `addUsageOverlay` (5
  private methods).
- **Added:** `parseRoundPatch`, `applyRoundPatch` (2 generic private
  methods, replacing the first 3 above), `mutateWithOverlay` (1 shared
  private method, replacing the bespoke branching `addOutputFiles`/
  `addUsage` used to each hand-roll), `mergeRoundPatch`/`mergeUsagePatch`
  (2 pure module-level merge functions), 2 new overlay `Map` fields
  (`missingOutputsOverlays`, `compileFailuresOverlays` — required: these are
  genuinely new per-accumulator state, not a wrapper).
- Net production diff is +187/-103 lines in `StreamSnapshotStore.ts`; the
  increase is the fix itself (two more accumulators gaining eager-apply +
  overlay coverage — two new overlay maps, two new reconciliation blocks in
  `applyStreamData`, plus JSDoc on the new shared helper) rather than
  incidental abstraction — the mechanism itself is smaller than four
  independent hand-rolled copies would have been (1 shared helper instead of
  what would otherwise be 4).
- No new coordinator/layer: `mutateWithOverlay` lives on `StreamSnapshotStore`
  itself, next to the `mutate()` it complements — the sole owner of this
  state — not a new class or facade.

## Consumer counts (R8)

- `mutateWithOverlay`: 4 call sites (`addOutputFiles`, `updateMissingOutputs`,
  `updateCompileFailures`, `addUsage`) — all four call sites the finding
  named.
- `parseRoundPatch`/`applyRoundPatch`: 3 call sites each (output files,
  missing outputs, compile failures) plus 1 additional `applyRoundPatch`
  call from `applyStreamData`'s overlay reconciliation.
- `mergeRoundPatch`: 3 call sites (the three round-keyed
  `mutateWithOverlay` calls). `mergeUsagePatch`: 1 call site (`addUsage`).

## Tests

- Extended `src/test-kernel/transcript/StreamSnapshotStore.vitest.ts` with
  two regression tests encoding the lost-update interleaving for
  `updateMissingOutputs`/`updateCompileFailures` (verified failing pre-fix,
  passing post-fix).
- `npx vitest run src/test-kernel/transcript/StreamSnapshotStore.vitest.ts` —
  31/31 pass.
- `npx vitest run src/test-kernel/transcript/` — 93/93 pass.
- Broader transcript-adjacent suites (desktop stream snapshot/agent
  execution, CLI chat session controller/run execution, progress backend,
  execution stream resolver, completed run archive, output discovery,
  round-map schemas) — 209/209 pass.
- `npm run typecheck` — clean across the whole workspace (root, test-kernel,
  extension, cli, trace-viewer, desktop).

---

> [!NOTE]
> **Medium Risk**
> Changes core transcript persistence and seed/overlay ordering for multiple sidecar mutators; incorrect replay could drop or resurrect missing-output/compile-failure markers, though behavior is heavily regression-tested.
> 
> **Overview**
> Fixes a **lost-update race** on unseeded streams (e.g. tabs outside a partial `preload()`): `updateMissingOutputs` and `updateCompileFailures` used deferred `mutate()` and callers could not read their own writes synchronously, unlike `addOutputFiles` / `addUsage`.
> 
> **`StreamSnapshotStore`** now routes those mutators—and refactors the existing output/usage paths—through shared **`mutateWithOverlay`**: apply to memory immediately, record per-stream overlay patches when disk seed is still in flight, and replay overlays in **`applyStreamData`** after load so seeding never clobbers eager updates. **`clearMissingOutputs`** uses the same reset-aware overlay so clear/update interleaving replays in call order.
> 
> Round-keyed patching is unified via **`parseRoundPatch`** / **`applyRoundPatch`** and **`mergeRoundPatch`**; missing outputs get **`mergeMissingOutputsOverlay`** for clear vs update ordering. Vitest adds regressions for immediate read-back, clear/update ordering, and compile failures on partially preloaded streams.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a993827f7ba74e5f4958f72fec852adc5d25f791. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/transcript/StreamSnapshotStore.vitest.ts
+++ b/src/test-kernel/transcript/StreamSnapshotStore.vitest.ts
@@ -582,6 +582,52 @@ describe('StreamSnapshotStore', () => {
     });
   });
 
+  it('replays clearMissingOutputs before a later updateMissingOutputs on an unseeded stream', async () => {
+    // clearMissingOutputs must not stay on the plain deferred mutate() path
+    // while updateMissingOutputs eagerly overlays: on an unseeded stream that
+    // ordering let the seed's overlay replay (update) land, then the clear
+    // (queued behind the same seed) run afterward and wipe it out regardless
+    // of call order. Here the clear fires first, so the later update must
+    // survive.
+    const dir = streamDataDir(OTHER_STREAM);
+    await StorageFS.ensureDir(dir);
+    await StorageFS.write(
+      path.join(dir, 'missingOutputs.json'),
+      JSON.stringify({ '0': ['stale.tex'] }),
+    );
+
+    const store = new StreamSnapshotStore();
+    await store.preload([STREAM]);
+
+    store.clearMissingOutputs(OTHER_STREAM);
+    store.updateMissingOutputs(OTHER_STREAM, { 1: ['next.tex'] });
+    expect(store.getMissingOutputs(OTHER_STREAM)).toEqual({ 1: ['next.tex'] });
+    await store.flush();
+
+    const raw = await StorageFS.readJson(path.join(dir, 'missingOutputs.json'));
+    expect(raw).toEqual({ '1': ['next.tex'] });
+  });
+
+  it('replays a later clearMissingOutputs over an earlier updateMissingOutputs on an unseeded stream', async () => {
+    const dir = streamDataDir(OTHER_STREAM);
+    await StorageFS.ensureDir(dir);
+    await StorageFS.write(
+      path.join(dir, 'missingOutputs.json'),
+      JSON.stringify({ '0': ['stale.tex'] }),
+    );
+
+    const store = new StreamSnapshotStore();
+    await store.preload([STREAM]);
+
+    store.updateMissingOutputs(OTHER_STREAM, { 1: ['next.tex'] });
+    store.clearMissingOutputs(OTHER_STREAM);
+    expect(store.getMissingOutputs(OTHER_STREAM)).toEqual({});
+    await store.flush();
+
+    const raw = await StorageFS.readJson(path.join(dir, 'missingOutputs.json'));
+    expect(raw).toEqual({});
+  });
+
   it('returns compile failures immediately for streams outside a partial preload without erasing disk markers', async () => {
     const dir = streamDataDir(OTHER_STREAM);
     await StorageFS.ensureDir(dir);
--- a/src/test-kernel/transcript/StreamSnapshotStore.vitest.ts
+++ b/src/test-kernel/transcript/StreamSnapshotStore.vitest.ts
@@ -552,6 +552,65 @@ describe('StreamSnapshotStore', () => {
     });
   });
 
+  it('returns missing outputs immediately for streams outside a partial preload without erasing disk markers', async () => {
+    // Same race as the output-files case above, replayed for
+    // updateMissingOutputs: a stream outside the preloaded set is still
+    // unseeded when the mutation lands, so the seed's disk read is racing
+    // the caller's read of its own write. Regression for the "half-fixed"
+    // eager-apply overlay gap (missingOutputs/compileFailures lacked the
+    // guarantee addOutputFiles/addUsage already had).
+    const dir = streamDataDir(OTHER_STREAM);
+    await StorageFS.ensureDir(dir);
+    await StorageFS.write(
+      path.join(dir, 'missingOutputs.json'),
+      JSON.stringify({ '0': ['prior.tex'] }),
+    );
+
+    const store = new StreamSnapshotStore();
+    await store.preload([STREAM]);
+
+    store.updateMissingOutputs(OTHER_STREAM, { 1: ['next.tex'] });
+    // Fails pre-fix: the plain mutate() path defers the whole apply behind
+    // the in-flight seed, so this synchronous read-back still sees nothing.
+    expect(store.getMissingOutputs(OTHER_STREAM)[1]).toEqual(['next.tex']);
+    await store.flush();
+
+    const raw = await StorageFS.readJson(
+      path.join(dir, 'missingOutputs.json'),
+    );
+    expect(raw).toMatchObject({
+      '0': ['prior.tex'],
+      '1': ['next.tex'],
+    });
+  });
+
+  it('returns compile failures immediately for streams outside a partial preload without erasing disk markers', async () => {
+    const dir = streamDataDir(OTHER_STREAM);
+    await StorageFS.ensureDir(dir);
+    const prior = compileFailure('prior.tex', 0);
+    const next = compileFailure('next.tex', 1);
+    await StorageFS.write(
+      path.join(dir, 'compileFailures.json'),
+      JSON.stringify({ '0': [prior] }),
+    );
+
+    const store = new StreamSnapshotStore();
+    await store.preload([STREAM]);
+
+    store.updateCompileFailures(OTHER_STREAM, { 1: [next] });
+    // Fails pre-fix for the same reason as the missing-outputs case above.
+    expect(store.getCompileFailures(OTHER_STREAM)[1]).toEqual([next]);
+    await store.flush();
+
+    const raw = await StorageFS.readJson(
+      path.join(dir, 'compileFailures.json'),
+    );
+    expect(raw).toMatchObject({
+      '0': [prior],
+      '1': [next],
+    });
+  });
+
   it('makes task state readable immediately while preserving later seeded sidecars', async () => {
     const dir = streamDataDir(STREAM);
     await StorageFS.ensureDir(dir);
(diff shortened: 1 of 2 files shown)


LionSR:
## Review disposition

Verified every review comment against HEAD (`a993827f7`, on top of the reviewed commit `7237e5e1`). New commit pushed to fix a real, high-severity finding two bots caught independently.

### Fixed

**cursor[bot] "Clear loses later missing updates" (High) + chatgpt-codex-connector "Preserve ordering with clearMissingOutputs" (P2)** — both correctly identified the same bug, from different angles. `clearMissingOutputs` still went through the plain deferred `mutate()` path while this PR moved `updateMissingOutputs`/`updateCompileFailures` to eager-apply-plus-overlay. On an unseeded stream, `mutate()`'s queued apply chains onto the seed promise *after* `applyStreamData`'s overlay reconciliation already ran, so `clearMissingOutputs` always replayed last **regardless of call order** — either wiping a later `updateMissingOutputs`, or resurrecting stale disk rounds a clear should have erased. I traced the exact interleaving through `queueAfterSeed`/`ensureSeeded`/`applyStreamData`, reproduced both directions as new regression tests (`replays clearMissingOutputs before a later updateMissingOutputs...` / `replays a later clearMissingOutputs over an earlier updateMissingOutputs...`), confirmed both **fail** against the pre-fix branch (one even threw `ENOTEMPTY` from the resulting corrupted disk state), then fixed by moving `clearMissingOutputs` onto `mutateWithOverlay` too, via a `reset`-aware overlay (`RoundOverlay<T>` + `mergeMissingOutputsOverlay`) so clear and update replay in call order. Pushed as `a993827f7`. Root typecheck, test-kernel typecheck, and `src/test-kernel/transcript/` (95 tests) all pass; `npx eslint`/`prettier --check` clean.

### Deferred (real, but out of scope for this PR)

**Copilot: `updateMissingOutputs` casts `raw as string[]` without validation** — real gap (a malformed progress event could store non-`string[]` data that later crashes `cloneRoundIndexed`), but this is pre-existing behavior, unchanged by this PR (verified identical on `origin/main`, `applyRoundKeyedPatch`'s `(raw) => raw as string[]` normalize function, line 556). Fixing it here would be unrelated defensive-programming scope creep on a focused race fix; better as its own PR alongside the analogous cast pattern check, if any, elsewhere.

**chatgpt-codex-connector: "Re-read late overlays before seeding" (P2)** — real: `applyStreamData` captures `outputFileOverlay`/`missingOutputsOverlay`/`compileFailuresOverlay`/`usageOverlay` as locals *before* the `await this.hydrateRunStateFromMeta(...)` call, so an overlay-recording mutation that races in during that await window (a real gap — `hydrateRunStateFromMeta` does a genuine disk read) gets its eager memory write correctly applied, but its disk persistence silently skipped and its overlay-map entry orphaned. However, this exact capture-before-await pattern **already existed pre-PR** for `outputFileOverlay`/`usageOverlay` (verified on `origin/main`) — this PR only extends the same existing (also-affected) pattern to two more accumulators, it doesn't introduce the shape. Fixing it properly means re-reading all four overlay maps after the await, uniformly, which is a distinct, larger change than this PR's scope (a focused parity fix for `updateMissingOutputs`/`updateCompileFailures`). Filing as a follow-up rather than blocking here.

### Verified

- `gh pr view 8014 --json reviews,comments` + `gh api .../pulls/8014/comments` — read every bot/human comment and review.
- Checked comment `commit_id` against `git log` on `origin/claude/race-snapshot-overlay`: all four inline findings (cursor, Copilot, 2×codex) target the reviewed HEAD (`7237e5e1`/`bbff671b`), not stale commits.
- The prior `claude` review and `github-actions`/TeXRA-agent review (both pre-dating or concurrent with the bot comments) found no issues; I independently re-verified their trace of `mutateWithOverlay`/`applyStreamData` and it holds.
- Read `src/transcript/StreamSnapshotStore.ts` in full at HEAD, diffed against `origin/main` for both flagged-but-pre-existing patterns (Copilot's cast, codex's capture-before-await) to confirm they weren't introduced by this PR.
- Reproduced the clear/update race both directions against the pre-fix branch (`git stash` back to `7237e5e1`) before writing the fix — both new tests failed as expected, confirming they're real regressions and not vacuous.
- Ran `npx tsc --noEmit` (root) and `npx tsc --noEmit -p tsconfig.test-kernel.json` clean; `npx vitest run src/test-kernel/transcript/` (95/95), plus `progressView/ProgressBackend`, `schemas/StreamDataRoundMaps`, `desktop/DesktopStreamSnapshot`, `desktop/DesktopAgentExecution` (231/231 combined) — all green post-fix.
- `npx eslint`/`npx prettier --check` clean on both touched files.

PR #8015, opened 2026-07-10
fix(cli): fuse resume()'s root-run slot check+claim into one atomic step

LionSR:
## The race

`resume()` in `packages/cli/src/chat/chatSessionController.ts` checked
`chatTuiCanStartRootRun(session)` as its first statement, but only *claimed*
the root-run slot (`session.runPromise = …`) after four `await`s and several
session-field mutations later:

```
resume(A):
  if (!chatTuiCanStartRootRun(session)) return;   // check
  await resolveCliResumeSnapshot(id);              // ← suspension point
  ... session.streamId = ...
  await getDefaultStreamLogStore().ensureLoaded(...);
  await snapshotStore.load(...);
  await snapshotStore.read(...);
  ...
  session.runPromise = setCliHelperModel(...).then(...);  // claim (finally)
```

`tryResumeStream()` already claims synchronously (fused check+claim, no
`await` in between — landed in #7925). So the interleaving from the finding
plays out as:

1. `resume(A)` passes its check (slot free) and suspends at
   `resolveCliResumeSnapshot(id)`.
2. A follow-up wake fires `tryResumeStream(B)` for a different stream. It
   sees the slot still free (A hasn't claimed yet), claims it, and starts
   real work (`snapshotStore.preload`, `resolveAndResumeStream`, …).
3. `resume(A)` wakes back up, runs to completion of its own steps, and
   unconditionally overwrites `session.runPromise` at the end — clobbering
   B's claim while B's async work is still in flight.
4. Two concurrent runs now share one `session`; further mutations from
   either can slip in, including a third caller once B "finalizes" against
   state A already stomped on.

## The ownership fix

Per the maintainer's doctrine, the cure for an internal race is
single-owner/sequenced design that makes the race impossible, not a new
coordination layer. `sessionRunState.ts` already owns the run-state fields
(`markChatTuiRunPending`, `markChatTuiRunCompleted`,
`chatTuiCanStartRootRun`); it was missing the *atomic* check-and-claim
primitive, so callers had to hand-roll the two-step "check, then (much
later) claim" convention themselves — and `resume()` did it wrong.

Added `tryClaimRootRunSlot(session, runPromise, runtimeHost?)` to
`sessionRunState.ts`: it fuses `chatTuiCanStartRootRun` +
`markChatTuiRunPending` into one synchronous call. Both `resume()` and
`tryResumeStream()` now call it as their literal first statement, before any
`await` — there is no longer a window in either function where a check can
pass and a claim can still be pending. `resume()` is restructured to build a
deferred placeholder promise up front, claim with it immediately, then
forward that deferred's settlement to the real run chain once it's built
further down (`runChain.then(resolveRunPromise, rejectRunPromise)`) — so
`resume()`'s own returned promise keeps its documented fire-and-forget
contract (settles after rehydration), while `session.runPromise` (used by
exit-drain's `await session.runPromise` and by `chatTuiCanStartRootRun`)
still only settles once the continued run truly finishes or is interrupted.

The two-step convention is gone: neither `resume()` nor `tryResumeStream()`
calls `chatTuiCanStartRootRun`/`markChatTuiRunPending` separately anymore —
both route through the one atomic primitive.

## Regression test

Extended `src/test-kernel/cli/chatSessionController.vitest.mts` (mirrors the
existing "reserves the root-run slot before tryResumeStream awaits persisted
state" test in the same describe block):

- **`reserves the root-run slot before resume() awaits the resolved
  snapshot`** — asserts `session.runPromise` is claimed synchronously,
  before `resume()`'s first `await` settles.
- **`resume() suspended on the resolved snapshot keeps a concurrent
  follow-up wake from also claiming the root-run slot`** — encodes the exact
  interleaving from the finding: `resume(A)` is suspended at
  `resolveCliResumeSnapshot` (mocked as a controllable deferred), then
  `tryResumeStream('stream-b')` fires while A is still suspended. Asserts B's
  `snapshotStore.preload` was never called (B bailed out synchronously) and
  `tryResumeStream` resolved `false` — exactly one caller ever holds the
  slot.

**Proved the regression test fails pre-fix** (git stash forbidden, per
instructions): saved the fix as a patch, `git checkout --` reverted
`chatSessionController.ts`/`sessionRunState.ts` to origin/main, reran the
suite — both new tests failed (`expected undefined to be defined` on
`session.runPromise`, since pre-fix `resume()` hadn't claimed yet at that
point) — then `git apply`'d the patch back and confirmed all 16 (now
18-including-new) tests pass.

## Net elements (R6)

- **+1 exported function**: `tryClaimRootRunSlot` in `sessionRunState.ts` —
  the single atomic primitive both hazard sites now share.
- **0 new layers/coordinators**: no new class, no new module, no new
  subscribe surface — the fix deepens the existing state-owner module.
- **-1 duplicated convention**: the "check, then claim much later" pattern
  that only `tryResumeStream()` got right (and `resume()` didn't) is gone;
  both call sites now share one call.
- Test file: +2 regression tests (71 lines), reusing the existing
  `deferred()` helper and mock scaffolding already in the suite — no new
  test infrastructure.

## Consumer counts (R8)

- `tryClaimRootRunSlot`: **2** production callers (`resume()`,
  `tryResumeStream()` in `chatSessionController.ts`) — exactly the two
  hazard sites from the finding, plus 1 doc-comment self-reference.
- `chatTuiCanStartRootRun`: **5** other production consumers remain
  (`runChatTui.tsx`, `agentModelCommands.ts`, `handleSlashCommand.ts`,
  `apiModeCommands.ts`, plus its own use inside `tryClaimRootRunSlot`) — kept
  exported; these are legitimate read-only availability checks, not claims.
- `markChatTuiRunPending`: **2** other production callers remain
  (`startRootRun()` in `chatSessionController.ts`, and `startSession()`'s
  outer reservation in `runChatTui.tsx`, which already claims before its own
  first `await`) — kept exported, not folded into `tryClaimRootRunSlot`
  since those call sites don't have the check-then-claim gap this fix
  closes.

## Tests run

- `npx vitest run src/test-kernel/cli/chatSessionController.vitest.mts` — 16/16 pass
- `npx vitest run src/test-kernel/cli/TuiStateAndFocus.vitest.mts src/test-kernel/cli/SessionResumeExitPolicy.vitest.mts src/test-kernel/cli/SlashCommandDispatch.vitest.mts src/test-kernel/cli/SessionResumeResolution.vitest.mts` — 143/143 pass
- `npx vitest run src/test-kernel/cli` (full CLI suite) — 128 files / 1657 tests pass
- `npm run typecheck` — clean across all workspace packages

---

> [!NOTE]
> **Medium Risk**
> Changes CLI session run orchestration and concurrency boundaries; behavior is well-tested but mistakes could still affect resume/stop or concurrent stream wakes.
> 
> **Overview**
> Fixes a race where **`resume()`** could pass a root-run availability check, suspend on rehydration **`await`s**, and let **`tryResumeStream()`** (or another **`resume()`**) claim the same **`session`** slot—then overwrite **`session.runPromise`** when the first call finished setup.
> 
> Adds **`tryClaimRootRunSlot`** in **`sessionRunState.ts`**, fusing **`chatTuiCanStartRootRun`** and **`markChatTuiRunPending`** into one synchronous step. **`resume()`** and **`tryResumeStream()`** now call it as their first statement, before any **`await`**. **`resume()`** claims with a deferred placeholder promise and forwards settlement to the real run chain so exit-drain still waits for the agent, while honoring **Ctrl-C** during rehydration (no **`resumeToolUseFromSnapshot`** after **`stop()`**).
> 
> Regression coverage in **`chatSessionController.vitest.mts`** plus a host mock baseline entry for **`@agent/runtime/executeAgent`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d023dd48ee11c74d46440ba373a3ce81934d1993. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/host-agent-mock-baseline.json
+++ b/host-agent-mock-baseline.json
@@ -34,6 +34,11 @@
     {
       "file": "src/test-kernel/cli/chatSessionController.vitest.mts",
       "form": "mock",
+      "specifier": "@agent/runtime/executeAgent"
+    },
+    {
+      "file": "src/test-kernel/cli/chatSessionController.vitest.mts",
+      "form": "mock",
       "specifier": "@agent/runtime/executionRegistry"
     },
     {
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -69,6 +69,7 @@ import {
   markChatTuiRunCompleted,
   markChatTuiRunPending,
   publishChatTuiRootRunStartAvailability,
+  tryClaimRootRunSlot,
   type TuiSession,
 } from './tui/state/sessionRunState';
 import { createTuiHostInteractions } from './tui/state/subscribeApprovals';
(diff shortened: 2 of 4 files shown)


LionSR:
## Disposition

All review comments are on commit `b8102d29` (the PR's only commit), so nothing here is stale.

### Fixed (pushed in `0856887`)

**1. Resume ignores cancellation after claim** — flagged independently by Cursor Bugbot (medium), Copilot, and Codex (P2). Verified against HEAD: the early `tryClaimRootRunSlot` call in `resume()` makes `chatTuiCanStopActiveRun()` report the run as stoppable (`session.runPromise` is set, `session.streamId` is either cleared or freshly set) well before rehydration (`resolveCliResumeSnapshot`, `getDefaultStreamLogStore().ensureLoaded`, `snapshotStore.load`/`read`) finishes. `resume()` never re-read `session.stopRequested` after those awaits — unlike `tryResumeStream()`, which does exactly this check (`if (session.stopRequested) return false;`) right after its own preparatory awaits. So a Ctrl-C during resume rehydration set the flag but had no effect: the resumed agent started anyway, and per `reportRunFailure`'s convention any later failure in that unwanted run would be silently swallowed and reported as a clean exit.

   Fix: added the same stop-check to `resume()`, placed right after `setupRunHost(sessionContext)` and before building `runChain` — this is the point after all of resume()'s preparatory awaits, and placing it after `setupRunHost` means a cancel-during-rehydration bail can still call the real `finalize()` to tear down the runtime host/listeners that were just allocated, then settle the claimed promise. This mirrors the existing established pattern: `startRootRun` similarly checks `stopRequested` right after its stream resolves (`if (session.stopRequested) interruptActiveRun();`), and `tryResumeStream` checks it before doing any real mutation.

   Regression test added: `honors a Ctrl-C issued while resume() is still rehydrating and never starts the resumed run` in `chatSessionController.vitest.mts` — suspends `resume()` inside `getDefaultStreamLogStore().ensureLoaded()` (via `setDefaultStreamLogStore`), fires `ctrl.stop()` while suspended, resolves the suspension, and asserts `resumeToolUseFromSnapshot` is never called and the slot is released (`runCompleted === true`). Confirmed this test fails pre-fix (`expected "vi.fn()" to not be called at all, but actually been called 1 times`) by temporarily removing just the new stop-check block and rerunning, then restored it.

**2. Doc comment overstated ("every entry point MUST call this")** — flagged independently by claude[bot] and Copilot. Verified: `startRootRun` claims the slot via `markChatTuiRunPending` directly, not via `tryClaimRootRunSlot`, and has no `chatTuiCanStartRootRun` check at all — safe only because it never suspends (`await`s) before claiming, so there's no check-then-await race window for the new primitive to close. Narrowed the doc comment on `tryClaimRootRunSlot` to scope the MUST to "entry points that await before they would otherwise claim" and explicitly noted `startRootRun`'s different-but-safe pattern, so a future reader doesn't read `startRootRun` as silently violating the contract.

### Not separately actioned

- Nothing else survived triage — the remaining bot comments (github-actions review, copilot-pull-request-reviewer overview, claude[bot]'s full review) were all "no issues found" / summary comments, not findings.

### Verification performed

- `npx vitest run src/test-kernel/cli/chatSessionController.vitest.mts` — 17/17 pass (16 original + 1 new)
- `npx vitest run src/test-kernel/cli/{TuiStateAndFocus,SessionResumeExitPolicy,SlashCommandDispatch,SessionResumeResolution}.vitest.mts` — 143/143 pass
- `npx vitest run src/test-kernel/cli` (full CLI suite) — 128 files / 1658 tests pass
- `npm run typecheck` — clean across all workspace packages (extension, cli, trace-viewer, desktop)
- `npx eslint` on all three touched files — clean (0 errors)
- Proved the new regression test fails pre-fix by reverting just the `resume()` stop-check, rerunning, then restoring.

LionSR:
Fixed the QA-2 host-mock ratchet failure (27→28 `vi.mock('@agent/...')` sites).

Evaluated avoidance first: the new race-interleaving test needs to control the timing of `resumeToolUseFromSnapshot` from `@agent/runtime/executeAgent` (an await-suspension point the test must hold open to reproduce the interleaving). The source file already imported from that specifier pre-PR, but no prior test in this file mocked it, so this is a genuinely new mock site rather than an avoidable one.

Deliberate growth: race-interleaving regression test needs to control resolveCliResumeSnapshot timing (and its downstream resumeToolUseFromSnapshot await). Added the one site to `host-agent-mock-baseline.json`, kept sorted per the ratchet's own comparator. Reran the ratchet, `chatSessionController.vitest.mts`, `ResumeCommand.vitest.mts`, and CLI typecheck — all green.

PR #8016, opened 2026-07-10
fix(cli): give the interactive chat TUI exclusive SIGINT/SIGTERM ownership

LionSR:
## The race

Every interactive entry point that eventually mounts the chat TUI —
`texra chat`, the default-command launcher (`texra`/`texra orchestrate`),
`texra setup`, and `texra resume` — called plain `initCliPlatform()`. By
default that installs its own SIGINT/SIGTERM handlers
(`installCliShutdownSignalHandlers`, via `process.once`). Later in the same
call chain, once Ink actually mounts, `runChatTui.tsx`'s `runChat()` installs
a *second*, independent set of SIGINT/SIGTERM handlers (via `process.on`) to
own interrupt/exit UX (double-Ctrl-C to force-exit, resumable-idle exits,
terminal-mode restore, persistence drain).

Both handler sets react to the *same* signal and each ends in its own
`process.exit()` call. The platform handler's chain is
`await lifecycle.runShutdown(); await flushNdjsonStdout(); process.exit(...)`;
the TUI's chain does its own teardown (unmount Ink, restore terminal modes,
print the resume hint, drain persistence, run the same shutdown sequence)
before its own `process.exit(...)`. Node invokes both listeners synchronously
back-to-back, but each is async past its first `await`, so which one's
`process.exit()` actually fires first — and therefore whether the terminal
gets restored, whether the resume hint gets printed, whether usage logs get
flushed — was an unspecified race between two independently-scheduled
async completions.

The await-suspension point that made this a *real*, not just theoretical,
race: `runChat()` awaits `initCliPlatform()` (installs the platform handler),
then awaits onboarding / model resolution (real I/O — network, disk), *then*
mounts Ink and installs its own handler. A signal arriving in that window hit
only the platform handler; a signal arriving after Ink mounts hit both.

## The ownership fix

Per the verified redesign: no new coordinator layer, no signal-arrival
tie-breaking — the cure is a single owner. The option to suppress the
platform handler already existed (`installSignalHandlers: false`, already
used correctly by `packages/cli/scripts/tui-harness.tsx`); it just wasn't
threaded through the real interactive entry points, so the "caller
discipline" of remembering to pass it was the failure mode.

- **New `initInteractiveCliPlatform`** (`initPlatform.ts`) wraps
  `initCliPlatform` and forces `installSignalHandlers: false`. It's the one
  place that owns the "this path is about to hand control to the
  TUI" decision — mirroring the existing `initLocalCliPlatform` wrapper
  pattern in the same file, not a new abstraction layer.
- The four real interactive entry points now call it instead of plain
  `initCliPlatform`: `runChatTui.tsx`'s `runChat()` (the `chat` command and
  every launcher branch that reaches it), `orchestrate.ts`'s
  `runOrchestration()` (default command), `setup.ts`'s `runSetup()`, and
  `resumeExecution.ts`'s `runResumeExecution()`. Headless commands
  (agents/agentsRun/auth/doctor/history/init/latex/memory/models/multiAgent/
  relayTokens/tools/workflow/runModel) are untouched and keep calling plain
  `initCliPlatform`, so they keep the platform handler as before.
- **Shared shutdown sequence**: extracted `runCliPlatformShutdownSequence`
  (lifecycle shutdown, best-effort, then the NDJSON flush) out of
  `installCliShutdownSignalHandlers`'s handler body. `runChatTui.tsx`'s own
  `runPlatformShutdown()` now delegates to the same function instead of only
  calling `lifecycle.runShutdown()` — so the TUI's exclusive signal path
  performs *exactly* the sequence the suppressed platform handler would have,
  with one definition instead of two that could drift.

Net effect: for the four interactive entry points, the TUI's own
`process.on('SIGINT'/'SIGTERM', ...)` handlers (installed once Ink mounts)
are now the *only* registered listeners for those signals — the race is
eliminated by construction, not by scheduling luck.

## Net elements (R6)

- `initInteractiveCliPlatform` — **1 new function**, 4 callers (chat via
  runChatTui.tsx, orchestrate.ts, setup.ts, resumeExecution.ts); not a
  single-caller extraction.
- `runCliPlatformShutdownSequence` — **1 new function**, 2 callers
  (`installCliShutdownSignalHandlers`'s handler, `runChatTui.tsx`'s
  `runPlatformShutdown`); genuine dedup (both previously hand-rolled the same
  try/catch shutdown+flush sequence, now one definition).
- 0 new files, 0 new ports/interfaces, 0 new coordinator/orchestration layers.
- Production diff: 4 import swaps + 2 small function additions in
  `initPlatform.ts`, plus doc comments at each call site. The bulk of the
  +319/-32 total is new regression-test code (~230 lines across 4 test
  files).

## Consumer counts (R8)

- `initCliPlatform` (unchanged, still exported): kept by every headless
  command file (`agents.ts`, `agentsRun.ts`, `auth.ts`, `chatgptAuth.ts`,
  `doctor.ts`, `history.ts`, `init.ts`, `latex.ts`, `memory.ts`, `models.ts`,
  `multiAgent.ts`, `relayTokens.ts`, `tools.ts`, `workflow.ts`,
  `runtime/runModel.ts`) and by `initLocalCliPlatform`/
  `initInteractiveCliPlatform` internally — 16 consumers, confirmed via
  `grep -rn initCliPlatform packages/cli/src`.
- `initInteractiveCliPlatform`: 4 consumers (`runChatTui.tsx`,
  `orchestrate.ts`, `setup.ts`, `resumeExecution.ts`) — every path that
  actually reaches `runChatTui`'s `runChat()`.
- `installCliShutdownSignalHandlers`: unchanged, 1 caller
  (`initCliPlatform`'s first-init branch), still exported for its existing
  direct test coverage.
- `runCliPlatformShutdownSequence`: 2 consumers as above.

## Tests

- `src/test-kernel/cli/CliInitPlatform.vitest.mts` — new
  `describe('CLI platform interactive signal ownership', ...)`: asserts
  `initInteractiveCliPlatform` installs **no** platform SIGINT/SIGTERM
  handler and the TUI's simulated `process.on` mount is the sole owner;
  asserts a plain (headless-style) `initCliPlatform` call **does** install
  the platform handler; and a third test reproduces the two-owner race
  against real production wiring when the interactive wrapper is skipped.
- `src/test-kernel/cli/InitPlatformSignalHandlers.vitest.mts` — new test for
  `runCliPlatformShutdownSequence` (shutdown-then-flush ordering,
  best-effort on a failing `runShutdown`, safe on an `undefined` lifecycle).
- `src/test-kernel/cli/SetupCommand.vitest.mts` and
  `ResumeCommand.vitest.mts` — updated their `@cli/runtime/initPlatform`
  mocks from `initCliPlatform` to `initInteractiveCliPlatform` (matching the
  real import change) and added a wiring assertion that `runSetup`/
  `runResumeExecution` call `initInteractiveCliPlatform` with the expected
  context.

**Proved fail-before/pass-after** (checkout, not stash — `git show
origin/main:<path> > <path>` to revert just the source file under test, then
restored the fix afterward):
- `initPlatform.ts` reverted → `CliInitPlatform.vitest.mts`'s
  `initInteractiveCliPlatform` test and
  `InitPlatformSignalHandlers.vitest.mts`'s `runCliPlatformShutdownSequence`
  test both fail with `TypeError: ... is not a function` (the exports don't
  exist pre-fix).
- `setup.ts` + `resumeExecution.ts` reverted → the new
  `SetupCommand.vitest.mts` / `ResumeCommand.vitest.mts` wiring tests fail
  (`No "initCliPlatform" export is defined on the "@cli/runtime/initPlatform"
  mock`) and cascade into 8 pre-existing tests in those files also failing,
  since the real `runSetup`/`runResumeExecution` call an unmocked
  `initCliPlatform`.

All reverts were restored from a scratch backup before committing; the
committed diff is the fixed version throughout.

## Verified

- `packages/cli/src/runtime/initPlatform.ts`,
  `packages/cli/src/chat/tui/runChatTui.tsx`,
  `packages/cli/src/commands/orchestrate.ts`,
  `packages/cli/src/commands/setup.ts`,
  `packages/cli/src/runtime/resumeExecution.ts`,
  `packages/cli/scripts/tui-harness.tsx` (unaffected — already used
  `installSignalHandlers: false` directly).
- Traced every `initCliPlatform`/`runChat` caller in `packages/cli/src`
  (`agents.ts`, `agentsRun.ts`, `auth.ts`, `chatgptAuth.ts`, `doctor.ts`,
  `history.ts`, `init.ts`, `latex.ts`, `memory.ts`, `models.ts`,
  `multiAgent.ts`, `orchestrate.ts`, `relayTokens.ts`, `setup.ts`,
  `tools.ts`, `workflow.ts`, `runtime/resumeExecution.ts`,
  `runtime/runModel.ts`) to confirm exactly 4 reach `runChatTui`.
- `src/test-kernel/cli/CliInitPlatform.vitest.mts`,
  `InitPlatformSignalHandlers.vitest.mts`, `SetupCommand.vitest.mts`,
  `ResumeCommand.vitest.mts` — read in full before extending.

## Tests run

- `npx vitest run src/test-kernel/cli/` — 128 files, 1661 tests passed.
- `npm run typecheck` — clean (root, test-kernel, extension, CLI,
  trace-viewer, desktop).
- `npx eslint` / `npx prettier --check` on every changed file — clean.

---

> [!NOTE]
> **Medium Risk**
> Changes process signal ownership and shutdown ordering on interactive CLI exits; mistakes could skip usage-log flush or leave duplicate handlers, but scope is limited to CLI interactive entry points with substantial regression tests.
> 
> **Overview**
> Fixes a **race where the platform and the Ink chat TUI both registered SIGINT/SIGTERM handlers**, so two async shutdown paths could compete on `process.exit()` and leave teardown order undefined (terminal restore, resume hint, usage flush).
> 
> Interactive entry points (`texra chat`, `orchestrate`, `setup`, `resume`) now call **`initInteractiveCliPlatform`** instead of plain `initCliPlatform`. The platform handler **stays live** through onboarding, model resolution, and the orchestration launcher; **`handOffCliShutdownSignalHandlers()`** in `runChatTui` removes only the listeners this module installed **immediately before** the TUI registers its own `process.on` handlers once Ink mounts.
> 
> **`runCliPlatformShutdownSequence`** centralizes lifecycle shutdown plus NDJSON flush for both the platform signal path and the TUI’s `exitNow` / resumable-idle exits. Regression tests cover handoff ordering, headless vs interactive wiring, and the old double-owner case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6fd9594de99e71eb671e44277a7afd75f184b9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/tui/runChatTui.tsx
+++ b/packages/cli/src/chat/tui/runChatTui.tsx
@@ -43,6 +43,7 @@ import { resolveChatDefaults } from '@cli/runtime/chatDefaults';
 import { seedCliRosterFromDefaultTeam } from '@cli/runtime/defaultTeamRoster';
 import { CliExitCode } from '@cli/runtime/exitCodes';
 import {
+  handOffCliShutdownSignalHandlers,
   initInteractiveCliPlatform,
   runCliPlatformShutdownSequence,
   setCliHelperModel,
@@ -271,11 +272,12 @@ export async function runChat(
     return { exitCode: CliExitCode.Usage };
   }
 
-  // installSignalHandlers: false (baked into initInteractiveCliPlatform) —
-  // this function is about to install its own SIGINT/SIGTERM/SIGHUP handlers
-  // (below, once Ink mounts) and owns teardown exclusively from here on; see
-  // initInteractiveCliPlatform's doc comment for why a second, platform-level
-  // handler set would race it.
+  // The platform's own SIGINT/SIGTERM handler stays live through onboarding
+  // and model resolution below — this function does not suppress it. Once
+  // Ink actually mounts (below), handOffCliShutdownSignalHandlers() removes
+  // it immediately before this function installs its own process.on pair, so
+  // exactly one owner is ever registered for a given signal; see
+  // initInteractiveCliPlatform's doc comment for the full handoff design.
   await initInteractiveCliPlatform({ ...context, quietLogs: true });
   // First-run gate (interactive only; headless already rejected above). A
   // credential-less user signs in or saves a key here; the apiMode + model
@@ -987,6 +989,12 @@ export async function runChat(
     clearPendingExit();
     ink.unmount();
   }
+  // Ownership transfers right here, not any earlier: everything above this
+  // line (initInteractiveCliPlatform, onboarding, model resolution) ran with
+  // the platform's own handler still live, so a signal during that window
+  // still got a graceful shutdown. This removes it and makes the handlers
+  // installed below the sole owner.
+  handOffCliShutdownSignalHandlers();
   process.on('SIGINT', handleSigint);
   process.on('SIGTERM', handleSigterm);
   process.on('SIGHUP', handleSighup);
--- a/packages/cli/src/chat/tui/runChatTui.tsx
+++ b/packages/cli/src/chat/tui/runChatTui.tsx
@@ -42,7 +42,11 @@ import { firstRunSetupAgentOverride } from '@cli/onboarding/setupContinuation';
 import { resolveChatDefaults } from '@cli/runtime/chatDefaults';
 import { seedCliRosterFromDefaultTeam } from '@cli/runtime/defaultTeamRoster';
 import { CliExitCode } from '@cli/runtime/exitCodes';
-import { initCliPlatform, setCliHelperModel } from '@cli/runtime/initPlatform';
+import {
+  initInteractiveCliPlatform,
+  runCliPlatformShutdownSequence,
+  setCliHelperModel,
+} from '@cli/runtime/initPlatform';
 import {
   formatCliNoAvailableModelsRecovery,
   selectCliRunnableModel,
@@ -267,7 +271,12 @@ export async function runChat(
     return { exitCode: CliExitCode.Usage };
   }
 
-  await initCliPlatform({ ...context, quietLogs: true });
+  // installSignalHandlers: false (baked into initInteractiveCliPlatform) —
+  // this function is about to install its own SIGINT/SIGTERM/SIGHUP handlers
+  // (below, once Ink mounts) and owns teardown exclusively from here on; see
+  // initInteractiveCliPlatform's doc comment for why a second, platform-level
+  // handler set would race it.
+  await initInteractiveCliPlatform({ ...context, quietLogs: true });
   // First-run gate (interactive only; headless already rejected above). A
   // credential-less user signs in or saves a key here; the apiMode + model
   // resolution below then see the freshly-set credentials in the same process.
@@ -870,12 +879,15 @@ export async function runChat(
     ]);
   };
   // These TUI exit paths call process.exit() directly, so bin/texra.ts's
-  // `finally` (which runs platform shutdown) never fires. Run it here too so
-  // shutdown handlers — notably UsageLogService.dispose(), which flushes any
-  // queued usage entries — execute before the process dies. runShutdown is
-  // idempotent, so the normal return path can still rely on bin/texra.ts.
+  // `finally` (which runs platform shutdown) never fires. Run the same
+  // shutdown sequence the (suppressed) platform SIGINT/SIGTERM handlers
+  // would have run — lifecycle shutdown (notably UsageLogService.dispose(),
+  // which flushes any queued usage entries) then the NDJSON flush — so it
+  // still happens once before the process dies. runCliPlatformShutdownSequence
+  // is idempotent-safe to call again, so the normal return path can still
+  // rely on bin/texra.ts's own `finally`.
   const runPlatformShutdown = (): Promise<void> =>
-    tryPlatform()?.lifecycle.runShutdown() ?? Promise.resolve();
+    runCliPlatformShutdownSequence(tryPlatform()?.lifecycle);
   const exitNow = (exitCode: number): void => {
     exiting = true;
     removeProcessHandlers();
--- a/packages/cli/src/commands/orchestrate.ts
+++ b/packages/cli/src/commands/orchestrate.ts
@@ -90,11 +90,14 @@ async function runOrchestration(context: CliContext): Promise<number> {
 
   await notifyCliUpdate(context);
 
-  // Every branch below either mounts the chat TUI directly (chat/preset/
-  // setupAgentOverride) or hands off to runResumeExecution (which also
-  // mounts it) — this is one of the real interactive entry points, so signal
-  // ownership must stay with the TUI once it mounts (see
-  // initInteractiveCliPlatform).
+  // This is one of the real interactive entry points (see
+  // initInteractiveCliPlatform): most branches below mount the chat TUI
+  // directly (chat/preset/setupAgentOverride) or hand off to
+  // runResumeExecution (which also mounts it), at which point the TUI takes
+  // over signal ownership. The `help` and `exit` launcher actions below mount
+  // nothing and return instead — the platform's own handler, still installed
+  // by initInteractiveCliPlatform, covers those the same way it would a
+  // headless command.
   await initInteractiveCliPlatform({
     ...context,
     quietLogs: true,
--- a/packages/cli/src/commands/orchestrate.ts
+++ b/packages/cli/src/commands/orchestrate.ts
@@ -8,7 +8,7 @@ import { AgentCategory } from '@agent/core/definition/AgentDataclass';
 import { firstRunSetupAgentOverride } from '../onboarding/setupContinuation';
 import { CliExitCode } from '../runtime/exitCodes';
 import { listCliHistoryEntries } from '../runtime/history';
-import { initCliPlatform } from '../runtime/initPlatform';
+import { initInteractiveCliPlatform } from '../runtime/initPlatform';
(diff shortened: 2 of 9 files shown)


LionSR:
## Review disposition (commit 3345574d → a6fd9594d)

Five reviewers (cursor bugbot, `claude` review, codex, and two `github-actions` texra-review passes) independently converged on the same finding, so I verified it first and led with it.

### Fixed

**Pre-mount SIGINT/SIGTERM gap** (cursor, claude[bot], codex P2, github-actions ×2 — `packages/cli/src/runtime/initPlatform.ts:192-197` et al.)

Verified against HEAD: `initInteractiveCliPlatform` forced `installSignalHandlers: false` at *init* time, but the TUI doesn't install its own `process.on('SIGINT'/'SIGTERM', ...)` until `runChatTui.tsx:990`, well after `await initInteractiveCliPlatform(...)` (line 279), `maybeRunCliOnboarding` (network I/O), and model resolution. `orchestrate.ts`'s `runOrchestration()` has an even longer pre-mount window (also awaits `getCliModelAccessList` and mounts the separate `runOrchestrationTui()` launcher, which has no signal handling of its own — confirmed via grep). Before this PR, that whole window had exactly one graceful handler (the platform's); after the original fix, it had zero — a signal there now hits Node's default action and skips `UsageLogService.dispose()` / `flushNdjsonStdout()`, a real regression vs. `main`, not just an unfixed edge case.

Fix (pushed to `claude/race-signal-handlers`, commit `a6fd9594d`): stopped suppressing the platform handler at init time. It now stays live through onboarding/model-resolution/the orchestration launcher. Added `handOffCliShutdownSignalHandlers()` (removes exactly the `process.once` listeners `installCliShutdownSignalHandlers` installed, tracked by reference — not every listener on the process) and call it in `runChatTui.tsx` immediately before the TUI installs its own `process.on` pair. Ownership now transfers at the actual moment the TUI takes over, not pre-emptively, so both the post-mount race (the PR's original target) and the pre-mount gap (this finding) are closed. `orchestrate.ts`'s `help`/`exit` branches and `resumeExecution.ts`'s non-resumable-snapshot early return — which never reach `runChat()` at all — now correctly keep the platform handler for their whole (short) lifetime instead of running with zero handlers.

As a side effect this also resolves the **Copilot type-guard nit** ("the `installSignalHandlers` field is silently overridden") for free: `initInteractiveCliPlatform`'s parameter type now `Omit`s `installSignalHandlers` entirely, since there's no override left to silently defeat.

Also fixed the **codex/copilot-adjacent doc-comment inaccuracy** in `orchestrate.ts:93-97` (comment claimed every branch mounts the TUI or hands to resume; `help`/`exit` do neither) while touching that comment for the handoff-design rewrite.

Verification: `npx vitest run src/test-kernel/cli/` — 128 files / 1662 tests pass (rewrote the `CliInitPlatform.vitest.mts` ownership test to assert the platform handler stays `once`-registered through `initInteractiveCliPlatform` and is only removed on explicit handoff; added a `handOffCliShutdownSignalHandlers` unit test in `InitPlatformSignalHandlers.vitest.mts`; updated stale "suppresses the platform's own handler" comments in `SetupCommand.vitest.mts`/`ResumeCommand.vitest.mts` to describe the handoff instead). `npm run typecheck` clean across all packages. `eslint`/`prettier --check` clean on every touched file.

### Noted, not fixed (follow-up grade)

**Missing direct `runChatTui.tsx` wiring test** (github-actions, low severity) — no test asserts `runChat()` itself calls `initInteractiveCliPlatform`/now `handOffCliShutdownSignalHandlers` (only `setup`/`resume`'s wiring is directly asserted). Agreed this is a real coverage gap, but closing it means driving `runChat()`'s full Ink-mount path under test (terminal-capability mocks, model resolution, etc.) — a heavier lift than this fix warrants. The existing `CliInitPlatform.vitest.mts` ownership tests plus the new `handOffCliShutdownSignalHandlers` unit test cover the underlying mechanism; a `runChatTui.tsx`-specific integration test is reasonable future work, not a blocker.

### Verified

- `packages/cli/src/runtime/initPlatform.ts` (full diff + surrounding context), `packages/cli/src/chat/tui/runChatTui.tsx` (mount-time signal-handling section, lines ~270-330 and ~950-995), `packages/cli/src/commands/orchestrate.ts` (full file, including the `help`/`exit` branches), `packages/cli/src/commands/setup.ts`, `packages/cli/src/runtime/resumeExecution.ts` (including its non-resumable early-return branch) — read in full against HEAD (`3345574d`), not a stale commit; all review comments were posted against this same commit, confirmed via `git log`.
- `src/test-kernel/cli/CliInitPlatform.vitest.mts`, `InitPlatformSignalHandlers.vitest.mts`, `SetupCommand.vitest.mts`, `ResumeCommand.vitest.mts` — read in full before extending/rewriting.
- Ran the PR's own suite (`npx vitest run src/test-kernel/cli/`, 128 files / 1662 tests) and `npm run typecheck` (all packages) against the fixed worktree.

**Verdict: recycling-after-fix.** Pushed to `claude/race-signal-handlers` (`a6fd9594d`).

PR #8068, opened 2026-07-11
refactor(export): isolate attachment marker vocabulary

### LINKED ISSUE #8029: Avoid transitive openai/* runtime import in completedRunArchive.ts via normalizeConversation
## Context

Follow-up from #8007 (fix(export): own the attachment-marker block vocabulary in normalizeConversation).

`Copilot`'s review comment flagged this on the PR; the author's own triage comment verified it as real (not stale) and explicitly filed it as "follow-up grade" rather than a blocking fix, offering to open a tracking issue.

## Details

`completedRunArchive.ts` now imports `mediaAttachmentKindToContentBlock` from `@agent/export/normalizeConversation`. That module has non-type-only runtime imports from `openai/lib/chatCompletionUtils` (`isAssistantMessage`/`isToolMessage`) and `openai/lib/parser` (`assertToolCallsAreChatCompletionFunctionToolCalls`), plus value imports from `@agent/modelHandlers/openai/*`. These are CJS modules esbuild can't tree-shake at the property level, so anything importing `mediaAttachmentKindToContentBlock` now eagerly loads that code.

Before this PR, `completedRunArchive.ts` had zero provider-SDK imports. `src/transcript/index.ts` already re-exports `completedRunArchive.ts`, and CLI TUI entry points (`chatSessionController.ts`, `runChatTui.tsx`) already import from `@transcript` — so this is now on the CLI's always-loaded path regardless of which provider the user is running. This is the same class of thing `ModelFactory.ts`'s `ProviderHandlerLoader` lazy-loading exists specifically to avoid.

The author's triage measured the actual transitive cost as small (`chatCompletionUtils.js` is 15 lines, `parser.js` is 175 lines, importing only `../error.js`), so this isn't urgent — but it's a real, deliberate trade-off (co-locating `mediaAttachmentKindToContentBlock` with the `blocksToUserParts` recognition switch it inverts vs. dependency isolation) that was accepted rather than fixed. Consider extracting the attachment-marker vocabulary (`ContentBlockSchema`/`ContentBlock`/`mediaAttachmentKindToContentBlock`) into a small dependency-free module (re-exported from `normalizeConversation.ts` to keep the current API), so transcript code doesn't pull in provider-SDK leaf modules at runtime.

## References

- PR: #8007
- Review comment: https://github.com/LionSR/TeXRA/pull/8007#discussion_r3562470664
- Author triage: https://github.com/LionSR/TeXRA/pull/8007#issuecomment-4940692149
- `src/agent/export/normalizeConversation.ts`
- `src/transcript/completedRunArchive.ts`
- Lazy-loading precedent: `src/agent/runtime/ModelFactory.ts` (`ProviderHandlerLoader`)

LionSR:
## Summary

Move the media-attachment marker constructor into a small provider-free module. `completedRunArchive.ts` now reconstructs image and document markers without loading the full conversation normalizer and its OpenAI runtime dependencies.

The former `normalizeConversation.ts` export had no remaining consumer after the archive import moved, so it is deleted rather than retained as an unowned compatibility re-export.

Closes #8029.

## Issue acceptance gates

- `completedRunArchive.ts` no longer imports `normalizeConversation.ts`.
- The attachment vocabulary has no provider SDK or model-handler imports.
- Every `MediaAttachmentKind` has an explicit marker mapping checked by TypeScript.
- Attachment markers still round-trip to the expected export nodes.
- The CLI eager archive graph reaches zero OpenAI SDK inputs.
- No compatibility re-export or second marker implementation remains.

## Single source of truth

`attachmentMarkerVocabulary.ts` is the sole owner of the mapping from persisted `MediaAttachmentKind` values to byte-free content blocks. Transcript reconstruction consumes it directly, and conversation normalization recognizes those marker values without owning their construction.

The new module contains one mapping, one private structural type, and one constructor. It replaces the larger provider-coupled switch and removes more lines than it adds; no factory, class, compatibility wrapper, or second marker implementation is introduced.

## Net elements (R6)

- Files changed: 4.
- Diffstat: 41 insertions, 56 deletions (net -15).
- New files: 1, containing 17 lines.
- Semantic exported symbols: net 0; the constructor moves from the provider-coupled module to its canonical vocabulary module.
- Classes/enums: 0.
- Private interfaces: +1.

## Consumer counts (R8)

- Direct production consumer: `completedRunArchive.ts`.
- Focused test consumer: one test module covering direct construction and normalization round trips.
- Compatibility consumers: 0.

## Host impact

Extension: no behavior change.

Desktop: removes one eager provider-SDK path from the startup graph.

CLI: removes OpenAI SDK modules from the completed-run archive eager graph.

## Validation

- focused Vitest: 2 files, 44 tests passed
- root source typecheck
- test-kernel typecheck
- CLI typecheck
- targeted ESLint (0 errors, 0 warnings)
- CLI esbuild metafile: the archive eager graph contains the vocabulary, excludes `normalizeConversation`, and reaches zero OpenAI SDK inputs
- integration validation with #8066: full desktop build, unpacked macOS arm64 package, and unchanged `check:desktop-package` all pass; the desktop startup graph excludes OpenAI, Anthropic, and Google SDKs

## Residual risk

Provider-specific conversation normalization remains intentionally provider-coupled and is still loaded when an export actually needs it. This change only removes that graph from transcript archive reconstruction.

### DIFF
--- a/src/agent/export/attachmentMarkerVocabulary.ts
+++ b/src/agent/export/attachmentMarkerVocabulary.ts
@@ -1,0 +1,17 @@
+import type { MediaAttachmentKind } from '@shared/schemas';
+
+const ATTACHMENT_MARKER_TYPE_BY_KIND = {
+  image: 'image',
+  document: 'document',
+} as const satisfies Record<MediaAttachmentKind, string>;
+
+interface AttachmentMarkerContentBlock {
+  type: (typeof ATTACHMENT_MARKER_TYPE_BY_KIND)[MediaAttachmentKind];
+}
+
+/** Build the byte-free content block used to mark an archived attachment. */
+export function mediaAttachmentKindToContentBlock(
+  kind: MediaAttachmentKind,
+): AttachmentMarkerContentBlock {
+  return { type: ATTACHMENT_MARKER_TYPE_BY_KIND[kind] };
+}
--- a/src/agent/export/normalizeConversation.ts
+++ b/src/agent/export/normalizeConversation.ts
@@ -23,8 +23,6 @@ import {
   isFunctionCallOutputItem,
 } from '@agent/modelHandlers/openai/openAIResponseContent';
 import { isResponseFunctionToolCallItem } from '@agent/modelHandlers/openai/responseStreamEvents';
-import type { MediaAttachmentKind } from '@shared/schemas';
-import { assertNever } from '@utils/core';
 import type { Part } from '@google/genai';
 import type {
   ChatCompletionMessageParam,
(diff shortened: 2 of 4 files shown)


LionSR:
Addressed the compatibility-surface review in `fe93bef7a`: the unused `normalizeConversation` re-export and identity-only test are removed. `attachmentMarkerVocabulary` is now the sole constructor path; the two focused suites still pass (44 tests), along with source, test-kernel, and CLI type checks and targeted lint.

PR #8075, opened 2026-07-11
fix(auth): make relay-token invalid status sticky against stale probes

LionSR:
## Finding

**MED race in the relay CI-token status cache** (`src/auth/relayToken.ts`,
`fetchRelayTokenStatus`): an async tier-config probe that started before a
live 401 rejection landed could resolve *after* `markRelayTokenRejected()`
had already marked the token `invalid`, and unconditionally overwrite the
cache back to `valid` — clobbering fresher, authoritative evidence with a
stale one.

Interleaving:

1. `fetchRelayTokenStatus(token)` misses the cache and kicks off a
   tier-config probe.
2. Before the probe resolves, a live relay call gets a real 401 and calls
   `markRelayTokenRejected(token)` — the cache is now `invalid`.
3. The in-flight probe (fetched from data that predates the rejection)
   resolves `valid` and `fetchRelayTokenStatus` writes it into the cache
   unconditionally, clobbering the fresher `invalid` entry.
4. Synchronous credential checks (`getCachedRelayTokenState`,
   `isAuthenticated`, relay-call fallback) now disagree with reality until
   the TTL happens to expire.

## Fix

Per the verified redesign, `fetchRelayTokenStatus`'s cache-write path is now
the single owner of the downgrade rule: it re-reads `statusCache.get(token)`
immediately before writing, and never overwrites an already-`invalid` entry
with a probe result. `invalid` is sticky until an explicit refresh (cache
reset) — a stale `valid`/`unknown` probe outcome can no longer undo a fresher
rejection recorded via `markRelayTokenRejected`.

```ts
const result = await probeRelayTokenStatus(token, fetchImpl);
if (result.state !== 'unknown') {
  const current = statusCache.get(token);
  if (current?.state === 'invalid') {
    return current;
  }
  statusCache.set(token, result);
}
return result;
```

Returning the sticky `current` (not just skipping the write) keeps the
function's return value consistent with what `getCachedRelayTokenState`
reports — callers of `fetchRelayTokenStatus` no longer see `valid` for a
token the cache already treats as rejected.

## Test

Extended `src/test-kernel/cli/RelayTokens.vitest.mts` with the exact
interleaving: kick off a probe with a controllable/deferred `fetch`, call
`markRelayTokenRejected` while it's still in flight, then resolve the probe
with a `valid` tier-config response. Asserts the cache stays `invalid`.

Proved the regression test fails pre-fix (reverted the `relayToken.ts` change
locally and reran):

```
AssertionError: expected 'valid' to be 'invalid' // Object.is equality
Expected: "invalid"
Received: "valid"
```

## Net elements (R6)

- 0 new files, 0 new exports, 0 new types/interfaces.
- +1 re-read + sticky-guard branch inside the existing single write site in
  `fetchRelayTokenStatus` — no new layer, no new cache, no new function.
- +1 regression test in the existing relay-token suite.

## Consumer counts (R8)

`fetchRelayTokenStatus` has 3 call sites, all unaffected in shape (still
`await fetchRelayTokenStatus(token)`), only benefiting from the corrected
sticky-invalid semantics:

- `src/tools/setup/platform.ts:143`
- `src/auth/SupabaseClient.ts:311` (same module also owns
  `markRelayTokenRejected` at `:238`, the source of the race)
- `packages/cli/src/runtime/supabaseAuth.ts:286`

## Verified

- `src/auth/relayToken.ts` — the cache-write path and its callers
  (`markRelayTokenRejected`, `getCachedRelayTokenState`,
  `resetRelayTokenTierCacheForTests`).
- `src/auth/SupabaseClient.ts` — confirmed `markRelayTokenRejected` is the
  live-401 caller that races the probe.
- `src/tools/setup/platform.ts`, `packages/cli/src/runtime/supabaseAuth.ts` —
  confirmed both other consumers only `await` the result, unaffected by the
  return-value change (which now can never return a stale `valid` where the
  cache holds `invalid`).
- `src/test-kernel/cli/RelayTokens.vitest.mts` — full suite (14 tests)
  passing post-fix; new test proven to fail pre-fix by temporarily reverting
  the `relayToken.ts` change.

---

> [!NOTE]
> **Medium Risk**
> Touches relay CI token auth caching used on model/relay paths; behavior change is narrowly scoped to not clobbering fresher rejections, but incorrect logic could still mis-report token validity.
> 
> **Overview**
> Fixes a race where an in-flight tier-config probe could **overwrite** or **return** a fresher `invalid` from `markRelayTokenRejected` (live relay 401) with a stale `valid` or `unknown`.
> 
> **`fetchRelayTokenStatus`** now re-reads `statusCache` after `probeRelayTokenStatus` completes. If the entry is already `invalid`, it returns that and skips writing the probe result—so return value and cache stay aligned for callers like auth profile checks.
> 
> Adds two regression tests in **`RelayTokens.vitest.mts`**: probe resolves `valid` or `unknown` after rejection while still in flight; both must keep `invalid`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97bccc508fda193402145799f98815c2a605da53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/auth/relayToken.ts
+++ b/src/auth/relayToken.ts
@@ -114,6 +114,15 @@ export async function fetchRelayTokenStatus(
   if (cached) return cached;
   const result = await probeRelayTokenStatus(token, fetchImpl);
   if (result.state !== 'unknown') {
+    // Re-read immediately before writing: a concurrent live 401
+    // (markRelayTokenRejected) may have landed while this probe was in
+    // flight, and is authoritative evidence fresher than whatever the probe
+    // observed. `invalid` is sticky — never let a stale probe result
+    // clobber it back to `valid`; only an explicit cache reset clears it.
+    const current = statusCache.get(token);
+    if (current?.state === 'invalid') {
+      return current;
+    }
     statusCache.set(token, result);
   }
   return result;
--- a/src/auth/relayToken.ts
+++ b/src/auth/relayToken.ts
@@ -113,16 +113,19 @@ export async function fetchRelayTokenStatus(
   const cached = statusCache.get(token);
   if (cached) return cached;
   const result = await probeRelayTokenStatus(token, fetchImpl);
+  // Re-read immediately before returning/writing: a concurrent live 401
+  // (markRelayTokenRejected) may have landed while this probe was in
+  // flight, and is authoritative evidence fresher than whatever the probe
+  // observed. `invalid` is sticky against a stale probe result — even an
+  // `unknown` one — until an explicit cache reset or TTL expiry clears it.
+  // Checking this ahead of the `result.state !== 'unknown'` cache-write
+  // guard (rather than inside it) keeps the *return value* consistent with
+  // the cache for every probe outcome, not just settled ones.
+  const current = statusCache.get(token);
+  if (current?.state === 'invalid') {
+    return current;
+  }
   if (result.state !== 'unknown') {
-    // Re-read immediately before writing: a concurrent live 401
-    // (markRelayTokenRejected) may have landed while this probe was in
-    // flight, and is authoritative evidence fresher than whatever the probe
-    // observed. `invalid` is sticky — never let a stale probe result
-    // clobber it back to `valid`; only an explicit cache reset clears it.
-    const current = statusCache.get(token);
-    if (current?.state === 'invalid') {
-      return current;
-    }
     statusCache.set(token, result);
   }
   return result;
--- a/src/test-kernel/cli/RelayTokens.vitest.mts
+++ b/src/test-kernel/cli/RelayTokens.vitest.mts
@@ -8,6 +8,7 @@ import {
   fetchRelayTokenStatus,
   getCachedRelayTokenState,
   getConfiguredRelayToken,
+  markRelayTokenRejected,
   resetRelayTokenTierCacheForTests,
 } from '@auth/relayToken';
 import {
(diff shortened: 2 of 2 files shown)


LionSR:
## Review triage

Read all 4 reviews (Cursor Bugbot, Copilot, `claude[bot]`, `github-actions`/TeXRA review-agent) plus their 4 inline comments. All submitted at HEAD (commit `57d2a6aa0`), no staleness — no intervening commits on this branch.

**Real, fixed:**

- **Cursor Bugbot (MED) + Copilot (independently, same line) — "sticky invalid skips unknown probes"**: `fetchRelayTokenStatus`'s re-read-before-write guard only ran inside the `result.state !== 'unknown'` branch. If a live 401 (`markRelayTokenRejected`) landed while a probe was in flight *and* that probe resolved `unknown` (timeout/5xx/parse failure — never cached), the function returned `unknown` straight from the probe result, bypassing the cache entirely — reintroducing exactly the return-value/cache divergence this PR exists to close. Both `getUserAuthContext` (`status.state !== 'invalid'`) and `getCliAuthProfile` (`status.state === 'invalid'`) trust the return value, so a just-rejected token could still read as usable/unverified-but-active. Confirmed by reverting and reproducing: pre-fix the new regression test failed (`expected 'invalid', received 'unknown'`).

  Fix: moved the cache re-read/sticky-return ahead of the `result.state !== 'unknown'` write-guard, so the sticky-`invalid` check now applies to every probe outcome — the write-guard (never cache `unknown`) is unchanged and still gates only the `statusCache.set`. Added a regression test mirroring the existing one but with the probe resolving via a `500` (→ `unknown`), asserting both the resolved value and the cache stay `invalid`. Proved it fails pre-fix (reverted `relayToken.ts` locally, reran — same "unknown" leak). Strengthened the original race test too, per Copilot's second finding below.

- **Copilot — original regression test only asserted the cache, not the resolved return value**: a fix that skipped the cache write but still resolved `{state:'valid'}` would have passed the old test. Added `await expect(pending).resolves.toEqual({state:'invalid'})` to the existing test alongside the cache assertion.

**Real, fixed as a drive-by (trivial, same comment block):**

- **`claude[bot]` — doc nit**: "only an explicit cache reset clears it" overstated the guarantee; the entry also self-clears on the 5-min TTL (`SERVER_SIDE_CACHE_TTL_MS`, pre-existing). Reworded to "until an explicit cache reset or TTL expiry clears it" in both the source comment and the mirrored test comment.

**Verified**

- `src/auth/relayToken.ts` — full diff re-read; confirmed the re-read/sticky-return now sits above the `result.state !== 'unknown'` write-guard so it covers `valid`/`invalid`/`unknown` probe outcomes uniformly, while `unknown` results still never get cached (unchanged behavior).
- `src/test-kernel/cli/RelayTokens.vitest.mts` — full suite: 15/15 passing post-fix (was 14, +1 test: the new `unknown`-probe race). Reverted `relayToken.ts` only (kept both new/updated tests) and reran: both the original race test (now asserting resolved value) and the new `unknown`-probe test fail pre-fix with the exact stale-value symptom described by Cursor/Copilot — confirms both fixes are load-bearing, not just cosmetic.
- Callers unaffected in shape: `src/tools/setup/platform.ts:143` (discards return value), `src/auth/SupabaseClient.ts:311`, `packages/cli/src/runtime/supabaseAuth.ts:286` — both real consumers now correctly see `invalid` instead of a leaked `unknown` on this interleaving.
- No `vscode` import in either touched file (`src/auth/` is a VS Code-allowed zone but doesn't use it here regardless); no new files/exports/types; net diff stays a same-function reordering + comment fix + 1 new test.

**Verdict: recycling-after-fix.** Both Cursor Bugbot and Copilot independently caught a real gap in the fix's own safety property (return value could still leak `unknown` on the exact race this PR targets); fixed with a same-shape reorder (no new branch, no new function) plus a proving regression test. `claude[bot]`'s doc nit folded in as a zero-cost drive-by. `github-actions`/TeXRA agent review found no issues (pre-existing gap wasn't in its scope of what it checked, since it evaluated the settled-result path only).

PR #8076, opened 2026-07-11
fix(webview): reset pasted-image state when follow-up-input's stream changes

LionSR:
## Finding

**HIGH race (webview):** the progress view reuses a **single** `<follow-up-input>` Lit instance across the active-stream context switch (`ToolUseStreamContent` consumes `streamStateContext`/`streamInfo` via `@consume`, per `BaseStreamContent`). `FollowUpInput` keeps its pasted-image attachment state (`pendingImages`, `imagePasteRevision`, `sendAfterImagePastes`) as private instance fields with no reset keyed on which stream it's currently bound to.

Failure story: paste an image into the follow-up box while viewing stream A, switch to stream B before hitting send, then hit send — the image (and the `[fileName]` token already inserted into the textarea, via `.value`) rides along and is attached to stream B's follow-up instead of stream A's.

## Fix

Deepen the owner (`FollowUpInput`) using the SAME stream-identity source `TodoList`/`PlanView` already key their `collapseKey` off of (`streamInfo.name`, wired in `ToolUseStreamContent.render()`):

- Added a `streamId` property to `FollowUpInput` (`packages/extension/src/progressView/frontend/components/FollowUpInput.ts`).
- In `willUpdate`, when `streamId` changes, reset `pendingImages`, `sendAfterImagePastes`, and bump `imagePasteRevision` (the same revision guard `attachPastedImages` already checks — `pasteRevision !== this.imagePasteRevision` — so any paste still in flight from the old stream becomes a no-op instead of leaking a late-resolving image into the new stream).
- Wired `.streamId=${streamInfo.name}` in `ToolUseStreamContent.render()` (`packages/extension/src/progressView/frontend/components/ToolUseStreamContent.ts`), mirroring the existing `.collapseKey=${streamInfo.name}` bindings on `<todo-list>`/`<plan-view>` two lines above.

View-level, stateless-renderer-compatible fix: no new bus emit, no new subscribe surface, no new port — one property + one `willUpdate` branch, following the exact pattern `TodoList.collapseKey` already established.

## Tests

No existing suite covered `FollowUpInput` specifically — `packages/extension` has zero `*.test.*`/`*.vitest.*` files; `npm test` only runs `src/test-kernel/**/*.vitest.{ts,mts}`. There IS existing shared Lit-component test infra (`src/test-kernel/settings/litComponentTestUtils.ts`'s `useLitComponentTestDom`, already reused by other progress-view component suites such as `TaskGroupListIndex.vitest.ts` and `TerminalOutput.vitest.ts`), so I added a new suite file in that same area rather than inventing new test plumbing:

- `src/test-kernel/progressView/FollowUpInputStreamSwitch.vitest.ts` — mounts a real `<follow-up-input>` custom element, seeds `pendingImages` (standing in for a completed paste — this harness's jsdom has no `FileReader`/`ClipboardEvent` wiring), then asserts:
  1. switching `streamId` drops the pending image before send (the regression case);
  2. a same-stream re-render (e.g. new `value` while streaming) does NOT drop it (guards against an over-eager reset).

Rendering `<follow-up-input>` for the first time in this harness surfaced a real gap in the shared jsdom setup: Web Awesome's `wa-details` (the "Followup" collapsible wrapper) constructs a `MutationObserver` in `firstUpdated()`, which wasn't in `useLitComponentTestDom`'s global allowlist (only `ResizeObserver` was, for `wa-textarea`). Extended that SAME allowlist/replacement list with `MutationObserver` (jsdom's own `window.MutationObserver`) rather than adding a parallel polyfill — same pattern as the existing `ResizeObserverStub`/dialog-polyfill entries in that file.

**Proven to fail pre-fix**: reverted the two source files to `origin/main` (`git checkout HEAD -- FollowUpInput.ts ToolUseStreamContent.ts`) with the new test file left in place — `drops a pending pasted image left over from the previously bound stream` failed with the stale image still attached (`AssertionError: expected [ { fileName: 'pasted-image-1.png', ... } ] to deeply equal []`); `git apply`'d the fix patch back and both tests passed. No `git stash` used.

## Net elements (R6)

- **+1 property** (`FollowUpInput.streamId`) — no new class, port, or abstraction layer.
- **+1 `willUpdate` branch** — same shape as the existing `polishRevision`/`recording` branches already in that method.
- **+1 template binding** (`ToolUseStreamContent.ts`, one line) — same shape as the two adjacent `collapseKey` bindings.
- **+1 test suite file** + **+1 allowlist entry** in the existing shared jsdom harness (no new harness, no new file for that harness).
- **0 deletions** — nothing pre-existing duplicated this reset logic to remove.

## Consumer counts (R8)

- `FollowUpInput` (`<follow-up-input>`) has exactly **one** call site: `ToolUseStreamContent.render()`. `streamId` is a single-purpose property read only in that component's own `willUpdate` — not a shared helper, so no multi-caller bar applies.
- `useLitComponentTestDom` (the harness touched for `MutationObserver`) already has 5 consumers before this PR (`SettingsReliabilityPlacement`, `ProviderKeyModal`, `ModelSelectionProviderKeys`, `TerminalOutput`, `TaskGroupListIndex`), now 6 with `FollowUpInputStreamSwitch` — extending its existing global allowlist is squarely in "called from multiple locations" territory, not a new single-caller abstraction.

## Verified

- `packages/extension/src/progressView/frontend/components/FollowUpInput.ts`
- `packages/extension/src/progressView/frontend/components/ToolUseStreamContent.ts`
- `packages/extension/src/progressView/frontend/components/BaseStreamContent.ts` (confirms the single shared `streamContext`-consuming instance, i.e. the race's root cause)
- `packages/extension/src/progressView/frontend/components/TodoList.ts` (the `collapseKey` precedent this fix mirrors)
- `src/test-kernel/settings/litComponentTestUtils.ts` (shared Lit jsdom harness)
- `src/test-kernel/progressView/TaskGroupListIndex.vitest.ts`, `TerminalOutput.vitest.ts` (existing usage precedent for the harness)
- Confirmed no pre-existing `FollowUpInput` test coverage exists anywhere in the repo (`packages/extension` has 0 `*.test.*`/`*.vitest.*` files).

---

> [!NOTE]
> **Medium Risk**
> Changes which stream receives pasted follow-up attachments; fix is narrow and tested but touches the send path for multimodal follow-ups.
> 
> **Overview**
> Fixes a race where the progress view **reuses one** `<follow-up-input>` across active streams: pasted images (and in-flight paste work) from stream A could be sent on a follow-up after switching to stream B.
> 
> **`FollowUpInput`** now takes a **`streamId`** (same identity as `streamInfo.name` / existing `collapseKey` wiring). On **`streamId`** change in **`willUpdate`**, it clears **`pendingImages`**, **`sendAfterImagePastes`**, and **`pendingImagePastes`**, and bumps **`imagePasteRevision`** so late async pastes from the old stream are ignored. **`ToolUseStreamContent`** binds **`.streamId=${streamInfo.name}`**.
> 
> Adds **`FollowUpInputStreamSwitch.vitest.ts`** for stream-switch vs same-stream re-render behavior, and exposes **`MutationObserver`** on the shared Lit jsdom harness so **`follow-up-input`** ( **`wa-details`** ) can mount in tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d97dc0485fd5446e6b143dc0f80bf7ba61184634. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/src/progressView/frontend/components/FollowUpInput.ts
+++ b/packages/extension/src/progressView/frontend/components/FollowUpInput.ts
@@ -174,7 +174,14 @@ export class FollowUpInput extends LitElement {
     // streams, so any image pasted while bound to the previous stream must
     // not ride along to whichever stream is active when send eventually
     // fires. In-flight pastes are invalidated via imagePasteRevision so
-    // their async completion is a no-op (see attachPastedImages).
+    // their async completion is a no-op (see attachPastedImages). Also drop
+    // any still-in-flight paste promises from the old stream: emitSend()
+    // gates on pendingImagePastes.size, so a stale entry left behind here
+    // would block (or, via flushPendingImagePasteSend, mis-time) a send
+    // issued in the newly-bound stream until the old-stream paste settles.
+    // Clearing the set is safe — attachPastedImages already no-ops on a
+    // revision mismatch, and the promise's own .finally() tolerates
+    // deleting an already-absent entry.
     if (
       changedProperties.has('streamId') &&
       changedProperties.get('streamId') !== undefined
@@ -182,6 +189,7 @@ export class FollowUpInput extends LitElement {
       this.imagePasteRevision += 1;
       this.pendingImages = [];
       this.sendAfterImagePastes = false;
+      this.pendingImagePastes.clear();
     }
 
     // React to shouldFocus property change
--- a/packages/extension/src/progressView/frontend/components/FollowUpInput.ts
+++ b/packages/extension/src/progressView/frontend/components/FollowUpInput.ts
@@ -124,6 +124,16 @@ export class FollowUpInput extends LitElement {
   @property({ type: Boolean, reflect: true }) visible = false;
   @property({ attribute: false }) value = '';
   @property({ attribute: false }) queuedMessages: string[] = [];
+  /**
+   * Identity of the stream this instance is currently bound to (the same
+   * `streamInfo.name` TodoList/PlanView key off of as `collapseKey`). The
+   * progress view reuses a single `<follow-up-input>` instance across the
+   * active-stream context switch, so any pasted-image state pending here
+   * must be reset when the bound stream changes — otherwise an image
+   * attached while viewing one stream can be delivered to whichever stream
+   * is active when the user later hits send.
+   */
+  @property({ type: String }) streamId = '';
 
   @property({ attribute: false }) shouldFocus = false;
   @property({ attribute: false }) polishedText: string | null = null;
@@ -160,6 +170,20 @@ export class FollowUpInput extends LitElement {
   });
 
   protected override willUpdate(changedProperties: PropertyValues): void {
+    // React to streamId property change: this instance is reused across
+    // streams, so any image pasted while bound to the previous stream must
+    // not ride along to whichever stream is active when send eventually
+    // fires. In-flight pastes are invalidated via imagePasteRevision so
+    // their async completion is a no-op (see attachPastedImages).
+    if (
+      changedProperties.has('streamId') &&
+      changedProperties.get('streamId') !== undefined
+    ) {
+      this.imagePasteRevision += 1;
+      this.pendingImages = [];
+      this.sendAfterImagePastes = false;
+    }
+
     // React to shouldFocus property change
     if (changedProperties.has('shouldFocus') && this.shouldFocus) {
       this.focusInput({ scrollIntoView: true }).then(() => {
--- a/packages/extension/src/progressView/frontend/components/ToolUseStreamContent.ts
+++ b/packages/extension/src/progressView/frontend/components/ToolUseStreamContent.ts
@@ -72,6 +72,7 @@ export class ToolUseStreamContent extends BaseStreamContent {
 
       <follow-up-input
         .visible=${true}
+        .streamId=${streamInfo.name}
         .value=${currentState.ui.followUpText}
         .queuedMessages=${currentState.queuedFollowUps}
         .shouldFocus=${currentState.ui.shouldFocusFollowUp}
--- a/src/test-kernel/progressView/FollowUpInputStreamSwitch.vitest.ts
+++ b/src/test-kernel/progressView/FollowUpInputStreamSwitch.vitest.ts
@@ -27,6 +27,7 @@ type FollowUpInputInternals = HTMLElement & {
   streamId: string;
   updateComplete: Promise<boolean>;
   pendingImages: ExtractedClipboardImage[];
+  pendingImagePastes: Set<Promise<void>>;
   emitSend: () => void;
 };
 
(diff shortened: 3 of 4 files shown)


LionSR:
## Review triage

Three automated review threads landed on this PR, all anchored on the same `willUpdate` `streamId`-reset block (`FollowUpInput.ts:172-190`):

**1. codex (P2) + Copilot — CONFIRMED, fixed.** Both flagged that the `streamId` reset clears `pendingImages`/`sendAfterImagePastes`/bumps `imagePasteRevision`, but leaves any still-in-flight entries in `pendingImagePastes` (the `Set<Promise<void>>` tracking in-progress paste reads) untouched. Traced the actual failure mode: `emitSend()` gates on `pendingImagePastes.size > 0`; a stale promise from the *previous* stream would silently delay a send issued in the *newly-bound* stream until that unrelated paste resolves — and if the user switches streams again before it resolves, `flushPendingImagePasteSend()` re-fires `emitSend()` against whichever stream is active *then* (routed via `getActiveFollowUpText()`'s `activeStreamId` read in `eventHandlers.ts`), not necessarily the stream the user actually sent from. Fixed by adding `this.pendingImagePastes.clear()` to the same reset branch — safe because `attachPastedImages` already no-ops on `pasteRevision !== this.imagePasteRevision`, and the promise's own `.finally()` tolerates deleting an already-absent Set entry. Added a regression test (`FollowUpInputStreamSwitch.vitest.ts`) that reproduces the pre-fix block (verified it fails without the `.clear()` call, passes with it).

**2. cursor bugbot (Medium) — real, but follow-up, not fixed here.** Distinct bug: round-tripping the *same* stream (A → B → back to A) without sending now drops the pending image (since leaving A clears `pendingImages`), while the `[fileName]` chip token survives because per-stream text (`followUpText`) is persisted in store state independently of the component-local image bytes. Confirmed this is a regression introduced by *this* PR's fix specifically — pre-fix, `pendingImages` was never reset at all, so a same-stream round trip preserved it (at the cost of the original cross-stream leak this PR closes). Note: `handleFollowUpSend`'s existing "orphan gate" (`eventHandlers.ts` — only attaches images whose `[fileName]` token survives in the submitted text) means this manifests as a message going out with an orphaned chip marker and no image data, not a wrong-stream image delivery — lower severity than the two race-condition findings above. A correct fix requires promoting `pendingImages` from component-local state to per-stream store state (mirroring how `followUpText` is already persisted), which is materially larger surface than the two single-line fixes here and outside this PR's R6 footprint. Filing as a follow-up rather than blocking/expanding scope on this PR.

## Verified
- `packages/extension/src/progressView/frontend/components/FollowUpInput.ts` (fix + existing `attachPastedImages`/`emitSend`/`flushPendingImagePasteSend`)
- `packages/extension/src/progressView/frontend/components/ToolUseStreamContent.ts` (`.streamId` binding)
- `packages/extension/src/progressView/frontend/eventHandlers.ts` (`handleFollowUpSend`, `getActiveFollowUpText` — confirmed the "whichever stream is active at dispatch time" routing that makes the codex/Copilot race concrete, and the orphan-gate that limits the cursor finding's severity)
- `src/test-kernel/progressView/FollowUpInputStreamSwitch.vitest.ts` (added regression case)
- No human/maintainer comment on this PR — all three findings are bot reviews, no ruling to defer to.

## Verification run
- `npx tsc --noEmit` (root, covers `packages/extension/src`) — clean
- `npx tsc --noEmit -p tsconfig.test-kernel.json` — clean
- `npm run typecheck` (full: root + test-kernel + extension + cli + trace-viewer + desktop) — clean, exit 0
- `npx vitest run src/test-kernel/progressView` — 36 files / 206 tests passed
- `npm test` (full suite) — passed (see CI/commit for full run)

PR #8080, opened 2026-07-11
fix(agent): recognize live Anthropic server-tool blocks in conversationFormat

LionSR:
## Summary

Round-2 APoSD audit finding (HIGH): `conversationFormat.ts`
(`src/agent/storage/`) and `normalizeConversation.ts` (`src/agent/export/`)
independently classify Anthropic content-block types, and the two
vocabularies had drifted apart.

- `conversationFormat.ts`'s `formatConversationBlock` switch had no cases
  for `server_tool_use`, `web_search_tool_result`, or `web_fetch_tool_result`
  — the three server-tool block types `AnthropicStreamHandler.ts:267,280,284`
  actually emits — so they fell through to the `default:` branch and got
  JSON-dumped verbatim (including `web_search_result`'s
  `encrypted_content` field and the full fetched page text).
- `normalizeConversation.ts`'s `assistantBlockToNode` switch carried three
  `code_execution_tool_result` / `bash_code_execution_tool_result` /
  `text_editor_code_execution_tool_result` cases that no model handler in
  this codebase emits — dead speculative support.

## Fix

1. Added the three live cases to `conversationFormat.ts`'s switch, mapping
   onto its existing `[tool_use: ...]` / `[tool_result: ...]` marker
   shapes:
   - `server_tool_use` → reuses `formatToolUseMarker` (same shape as a
     regular `tool_use` block).
   - `web_search_tool_result` → new `formatWebSearchResultMarker`: a
     compact `title (url), title (url), ...` list instead of JSON-dumping
     each result's `encrypted_content`.
   - `web_fetch_tool_result` → new `formatWebFetchResultMarker`: a
     `title (url)` marker instead of JSON-dumping the fetched page's full
     text.
2. Deleted the three dead code-execution cases from
   `normalizeConversation.ts` (deletion beats speculative support — they
   now fall through to the existing `default: return null`, same as any
   other unrecognized block type).
3. Cross-referenced each switch's doc comment to the other file/switch by
   name, so the two vocabularies can't drift blind again. Went with doc
   comments over a shared type→category const (the
   `mediaAttachmentKindToContentBlock` pattern from #8007) because the two
   switches map the same three block types to genuinely different output
   shapes (a truncated marker string vs. a structured `ExportNode` variant
   per type) — a shared const would be a new abstraction layer for
   classification logic that's already this thin (3-way switch in each
   file).

Extended `ExecutionsConversationFormat.vitest.ts` with the three live
block types (verified failing pre-fix — see Tests below) and repurposed
the now-stale `normalizeConversation.vitest.ts` "handles
code_execution_tool_result" test into a "does not map unused
code-execution block types" test covering all three deleted cases.

## Net elements (R6)

- **Added**: 2 private helper functions (`formatWebSearchResultMarker`,
  `formatWebFetchResultMarker`) in `conversationFormat.ts` — both are
  single-purpose formatters for the two Anthropic block content shapes
  that need bespoke compaction (not a shared abstraction; each is called
  from exactly one switch case).
- **Deleted**: 3 dead switch cases (`code_execution_tool_result`,
  `bash_code_execution_tool_result`, `text_editor_code_execution_tool_result`)
  in `normalizeConversation.ts`.
- **No new files, no new exports, no new types.** Net diff is +145/−13
  across 2 source files + 2 test files; the two added helpers are pure
  functions colocated with (and reusing `formatToolResultMarker` as a
  fallback from) the switch they serve.

## Consumer counts (R8)

- `formatConversationBlock`'s switch (edited) is exercised by
  `formatConversationContent`, consumed by:
  - `src/tools/executions/conversationFormat.ts` → `ExecutionsTool`'s
    `/executions/{id}/conversation` endpoint
  - `packages/cli/src/runtime/history/conversationFormat.ts` → the CLI's
    `texra history` / resume previews
  Both are pure call-through consumers of the shared formatter — no
  changes needed on their side; they inherit the fix automatically.
- `assistantBlockToNode`'s switch (edited) is exercised by
  `normalizeConversationForExport`, the sole normalization entry point for
  every chat-export renderer (markdown/LaTeX/HTML) per the module's own
  docstring — grepped, no other caller.

## Tests

- `npx vitest run src/test-kernel/tools/ExecutionsConversationFormat.vitest.ts` —
  3 new cases added (`server_tool_use`, `web_search_tool_result`,
  `web_fetch_tool_result`); confirmed failing pre-fix (implemented the
  tests first, ran them against the unmodified `conversationFormat.ts`,
  all 3 failed with the raw-JSON `default:` output), then passing after
  the fix. 9/9 pass post-fix.
- `npx vitest run src/test-kernel/commands/history/normalizeConversation.vitest.ts` —
  21/21 pass; repurposed test now asserts the three dead code-execution
  types produce no `tool-result` node.
- `npm run typecheck` (root tsc, test-kernel, `texra` extension, CLI,
  trace-viewer, desktop) — exit 0.

## Verified

- `src/agent/storage/conversationFormat.ts` (edited)
- `src/agent/export/normalizeConversation.ts` (edited)
- `src/agent/modelHandlers/support/AnthropicStreamHandler.ts` (confirmed
  the three live emitted block types at the cited line numbers)
- `src/agent/types/ServerToolTypes.ts` (`mapAnthropicWebSearchEntries`,
  `WebFetchResult` shapes — informed the marker formats without importing
  SDK types into the duck-typed `conversationFormat.ts`)
- `src/tools/executions/conversationFormat.ts`,
  `packages/cli/src/runtime/history/conversationFormat.ts` (both
  pass-through consumers, no changes needed)
- `src/test-kernel/tools/ExecutionsConversationFormat.vitest.ts`,
  `src/test-kernel/commands/history/normalizeConversation.vitest.ts`
  (edited)
- Grepped the whole tree for `code_execution_tool_result` /
  `bash_code_execution_tool_result` / `text_editor_code_execution_tool_result`
  to confirm the only other reference was the now-repurposed test.

---

> [!NOTE]
> **Low Risk**
> Display-only formatting and export normalization with no auth or persistence changes; behavior is tightened and covered by new tests.
> 
> **Overview**
> **Conversation text views** (`conversationFormat`) now handle the three Anthropic server-tool block types the stream handler actually emits (`server_tool_use`, `web_search_tool_result`, `web_fetch_tool_result`) instead of JSON-dumping them. `server_tool_use` reuses the existing `[tool_use: …]` marker; search and fetch results get compact `title (url)` markers so `encrypted_content` and full page bodies never leak into Executions `/conversation` or CLI `texra history` output. Archived completed-run `web_fetch_tool_result` shapes (top-level `url`/`title` without nested `content`) are handled alongside the live SDK shape.
> 
> **Export normalization** (`normalizeConversation`) drops speculative `code_execution_*_tool_result` cases that no handler emits; those blocks now follow the same `default: return null` path as other unknown types. Doc comments on both switches cross-reference each other so the shared three-type vocabulary stays aligned across marker strings vs structured `ExportNode`s.
> 
> Tests cover the new formatters and assert the removed code-execution types no longer produce `tool-result` export nodes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f07486c939e48f72503d8e4739325e3e5f5a0e63. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/export/normalizeConversation.ts
+++ b/src/agent/export/normalizeConversation.ts
@@ -244,11 +244,7 @@ function assistantBlockToNode(block: ContentBlock): ExportNode | null {
         input: prettyJson(block.input ?? {}),
       };
 
-    // Tool result blocks: Anthropic tool_result + server-side code execution results
     case 'tool_result':
-    case 'code_execution_tool_result':
-    case 'bash_code_execution_tool_result':
-    case 'text_editor_code_execution_tool_result':
       return {
         kind: 'tool-result',
         text:
@@ -257,6 +253,13 @@ function assistantBlockToNode(block: ContentBlock): ExportNode | null {
             : prettyJson(block.content ?? ''),
       };
 
+    // Anthropic server-side tool blocks (the provider executes these, not a
+    // local tool handler). This vocabulary must stay in sync with the
+    // `formatConversationBlock` switch in `@agent/storage/conversationFormat`
+    // — both classify the same three live Anthropic block types emitted by
+    // `AnthropicStreamHandler` (`server_tool_use`, `web_search_tool_result`,
+    // `web_fetch_tool_result`), just into different output shapes (a
+    // structured `ExportNode` here vs. a truncated marker string there).
     case 'server_tool_use':
       if (block.name === 'web_search') {
         const query =
--- a/src/agent/storage/conversationFormat.ts
+++ b/src/agent/storage/conversationFormat.ts
@@ -153,25 +153,41 @@ function formatWebSearchResultMarker(
 }
 
 /**
- * Anthropic `web_fetch_tool_result` content is a
+ * A live Anthropic `web_fetch_tool_result` block's `content` is a
  * `{type: 'web_fetch_result', url, content: {title, ...}}` object (or an
- * error object) — rendering it through the generic block formatter would
- * JSON-dump the fetched page's full text, so this collapses it to a
- * `title (url)` marker instead. Falls back to `formatToolResultMarker` for
- * the error shape.
+ * error object). A completed-run's transcript-sidecar archive
+ * (`webFetchEntryToMessages` in `@transcript/completedRunArchive`)
+ * reconstructs the same block type with no nested `content` at all —
+ * `url`/`title`/`page_content` sit directly on the block instead. Both
+ * `ExecutionsTool`'s `/conversation` endpoint and the CLI's `texra history`
+ * read completed-run conversations through that archive, so this must
+ * recognize both shapes; rendering either through the generic block
+ * formatter would JSON-dump the fetched page's full text, so this collapses
+ * whichever shape is present to a `title (url)` marker instead. Falls back
+ * to `formatToolResultMarker` for the live error shape.
  */
 function formatWebFetchResultMarker(
-  content: unknown,
+  block: Record<string, unknown>,
   options: ConversationFormatOptions,
 ): string {
-  if (!isObject(content) || content.type !== 'web_fetch_result') {
-    return formatToolResultMarker(content, options);
-  }
   const marker = options.truncationMarker ?? DEFAULT_TRUNCATION_MARKER;
-  const url = asText(content.url);
-  const title = objectStringField(content.content, 'title');
-  const label = title && url ? `${title} (${url})` : title || url;
-  return `[tool_result: ${truncate(label, options.toolBlockLimit, marker)}]`;
+  const content = block.content;
+  if (isObject(content) && content.type === 'web_fetch_result') {
+    const url = asText(content.url);
+    const title = objectStringField(content.content, 'title');
+    const label = title && url ? `${title} (${url})` : title || url;
+    return `[tool_result: ${truncate(label, options.toolBlockLimit, marker)}]`;
+  }
+  const archivedUrl = asText(block.url);
+  const archivedTitle = asText(block.title);
+  if (archivedUrl || archivedTitle) {
+    const label =
+      archivedTitle && archivedUrl
+        ? `${archivedTitle} (${archivedUrl})`
+        : archivedTitle || archivedUrl;
+    return `[tool_result: ${truncate(label, options.toolBlockLimit, marker)}]`;
+  }
+  return formatToolResultMarker(content, options);
 }
 
 /**
@@ -277,7 +293,7 @@ function formatConversationBlock(
     case 'web_search_tool_result':
       return formatWebSearchResultMarker(block.content, options);
     case 'web_fetch_tool_result':
-      return formatWebFetchResultMarker(block.content, options);
+      return formatWebFetchResultMarker(block, options);
     default:
       return truncate(
         stringifyConversationValue(block),
--- a/src/agent/storage/conversationFormat.ts
+++ b/src/agent/storage/conversationFormat.ts
@@ -128,6 +128,53 @@ function formatToolResultMarker(
 }
 
 /**
+ * Anthropic `web_search_tool_result` content is an array of
+ * `{type: 'web_search_result', title, url, ...}` entries (or an error
+ * object) — rendering each entry through the generic block formatter would
+ * JSON-dump its `encrypted_content` field, so this collapses the array to a
+ * compact `title (url)` list instead. Falls back to `formatToolResultMarker`
+ * for the error shape.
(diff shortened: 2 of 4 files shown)


LionSR:
## Review disposition (PR #8080)

All six review-thread comments (Codex, `texra-review`/github-actions bot, Copilot ×4) plus the `claude[bot]` PR-comment convergent review independently identified **the same single real defect**, all posted against commit `f27e907e` after it landed (not stale):

### Real, fixed now — `formatWebFetchResultMarker` assumed only the live Anthropic SDK shape
`formatWebFetchResultMarker(block.content, options)` assumed `block.content = {type:'web_fetch_result', url, content:{title}}` (confirmed against `AnthropicStreamHandler.ts`). But completed-run conversations reconstructed from the transcript-sidecar archive (`completedRunArchive.ts`'s `webFetchEntryToMessages`) put `url`/`title`/`page_content` **directly on the block**, with no nested `content` at all. `ExecutionsTool`'s `/conversation` endpoint and the CLI's `texra history` both read completed-run conversations through that archive path, so `block.content` was `undefined` there and the new switch case fell back to `formatToolResultMarker(undefined, …)`, producing an empty `[tool_result: ]` for every archived web-fetch — a regression versus the pre-PR `default:` branch, which at least JSON-dumped the block (showing url/title, if also leaking full page text).

Fixed in `f07486c9` by branching `formatWebFetchResultMarker` on whichever shape is present (nested live-SDK `content` vs. flat archived fields), passing the whole block through instead of just `block.content`. Added a regression test (`ExecutionsConversationFormat.vitest.ts`) covering the archived shape and asserting `page_content` is never leaked. `web_search_tool_result` and `server_tool_use` do **not** have this bug — both the archive reconstruction and the live SDK put `name`/`input`/`content` in the same place for those two types, verified by every reviewer including `claude[bot]`'s independent re-derivation.

Verified: `npx vitest run src/test-kernel/tools/ExecutionsConversationFormat.vitest.ts src/test-kernel/commands/history/normalizeConversation.vitest.ts` (46/46 pass), `npm run typecheck` (root tsc, test-kernel, extension, CLI, trace-viewer, desktop — exit 0), `npx prettier --check` on both touched files.

### Real, deferred as follow-up — `normalizeConversation.ts`'s `web_fetch_tool_result` case has the mirror-image pre-existing bug
The `texra-review` bot and `claude[bot]` both flagged that `assistantBlockToNode`'s `web_fetch_tool_result` case (`normalizeConversation.ts:281-287`) reads `block.url`/`block.title`/`block.page_content` at the top level unconditionally — correct for the archived shape, but wrong for genuine live-Anthropic-shaped blocks reaching it through the legacy `conversation.json` KV fallback, which nests those fields under `block.content`. This case is **not touched by this PR's diff** (only the doc comment directly above it was added, cross-referencing the `conversationFormat.ts` switch) — confirmed via `git diff` against the pre-PR base. Deferring rather than fixing in this PR: it's a distinct bug in a different consumer (chat export markdown/HTML/LaTeX, not the executions/history path this PR targets), fixing it correctly needs the same live-vs-archived branch this PR just added on the other side plus export-specific test coverage, and conflating it here would blow past this PR's stated scope (R6/R8 net-element accounting). Will file a tracking issue for a narrow follow-up PR mirroring today's fix onto `normalizeConversation.ts`.

## Verified
- `src/agent/storage/conversationFormat.ts` (the fix)
- `src/transcript/completedRunArchive.ts` (`webFetchEntryToMessages`, `webSearchEntryToMessages` — confirmed the shape mismatch is web-fetch-only)
- `src/agent/modelHandlers/support/AnthropicStreamHandler.ts` (live SDK shapes)
- `src/tools/ExecutionsTool.ts`, `packages/cli/src/runtime/history.ts` (confirmed both consume `readCompletedRunConversation` → the archive path with the flat shape)
- `src/agent/export/normalizeConversation.ts`, `src/agent/export/loadChatExportInput.ts` (confirmed the mirror-image pre-existing bug and that it's untouched by this PR's diff)
- `src/test-kernel/tools/ExecutionsConversationFormat.vitest.ts` (new regression test)

**Verdict: recycling-after-fix.** One real regression fixed and pushed (`f07486c9`); one real but genuinely pre-existing, out-of-diff bug in a different file/consumer deferred to a follow-up issue rather than scope-creeping this PR.

Issue #7213, opened 2026-07-05
GPT-5.5/ChatGPT-subscription (Codex) backend has no context compaction of any kind


LionSR:
## Context

Follow-up from investigating a user-visible "Manual context compaction is not yet available for this model. Stay tuned!" message for GPT-5.5 accessed via the ChatGPT subscription (Codex) backend.

`ModelHandlerOpenAIResponse.shouldCompact()` (`src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts`) gates both the user-triggered "compact now" action AND the threshold-based automatic compaction behind a single flag:

```ts
private shouldCompact(): boolean {
  if (!this.supportsManualCompaction) {
    this.compactionRequested = false;
    return false;
  }
  ...
}
```

`src/model/providerCapabilities.ts`'s `chatgpt-subscription` profile hardcodes `supportsManualCompaction: false`. So for GPT-5.5 (and any other Codex-eligible model) routed through the ChatGPT subscription, compaction never happens — not just the manual button, the automatic threshold-based relief valve too. Long conversations on this backend just keep growing input tokens until they hit the context window ceiling, with nothing to relieve the pressure.

This isn't a TeXRA implementation gap — it's a hard property of the unofficial Codex backend. Per `modelHandlerCodex.ts`'s own doc comment, the backend "requires `store: false`" on every request (confirmed by direct probing, same as the `background:true` → 400 result already documented there). No server-side storage means:
- No `previous_response_id` to chain off of → `supportsResponseChaining: false`
- No stored response for OpenAI's `/responses/compact` endpoint to act on → `supportsManualCompaction: false`

So the existing compaction mechanism (which calls `client.responses.compact(...)` against a stored server-side response) fundamentally cannot work for this backend. This compounds the timeout risk addressed in the reasoning-effort clamp (PR #7208): this backend already has no background/polling mode, and now also no way to shrink an over-long conversation.

## Details

Add a **client-side** compaction fallback for provider profiles where `supportsManualCompaction` is false but context pressure still needs relief — i.e., summarize the conversation locally and resend a shorter history, instead of calling OpenAI's stateful `/responses/compact` endpoint.

[...]

PR #8081, opened 2026-07-11
fix(model-handlers): commit compaction payload and chain-anchor clear together

LionSR:
## Problem

`ModelHandlerOpenAIResponse.compactConversation()` / `compactConversationClientSide()`
commit two things when compaction succeeds:

1. **The chain anchor** — `chainState.clearChainForCompaction()` nulls
   `previousResponseId` on `ResponseChainState` (`src/agent/modelHandlers/openai/ResponseChainState.ts`).
   This is a handler-lifetime field, so once cleared it stays cleared across
   any later retry within the same turn.
2. **The computed payload** — `this.compactionResult` (compacted messages +
   token estimate), a handler field the caller reads to build the outer
   `create()` request.

`createResponseImpl()` (`src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts:1444` pre-fix)
unconditionally wiped `this.compactionResult` at the top of every call:

```ts
// Clear any stale compaction result from previous attempts. ...
this.compactionResult = undefined;
```

PocketFlow's `Node._exec` (`src/agent/node/index.ts:172`) retries a failed
`exec()` with the *same* `prepRes`, so `ModelInvocationNode.exec()`
(`src/agent/core/flows/ModelInvocationNode.ts:128`) resends the identical
`messages` array reference on a same-turn retry. On that retry the anchor
clear (#1) had already committed permanently on `chainState`, but the payload
(#2) had just been thrown away — and `chainState.cumulativeInputTokens` is
only updated from a *successful* response (`finalizeResponse`, line ~555), so
`shouldCompact()` still reads the pre-compaction count and returns `true`
again. The handler silently redid the whole `/responses/compact` round trip
for input it had already compacted, purely because the payload's retry
lifetime didn't match the anchor mutation's.

## Fix

Key `compactionResult` to the exact `messages` reference it was computed
from (`sourceMessages`). `createResponseImpl()` now only discards a cached
result when the call's `messages` is a genuinely different reference (a new
turn), and reuses it directly — skipping the network round trip — when it's
the same reference (a same-turn retry). This makes the anchor clear and the
payload commit/survive together instead of diverging.

No new abstraction: `ResponseChainState` and `compactionResult` are the
existing owners: the payload is now trivially “this call's input” which is
already how PocketFlow's own retry model treats `prepRes.messages`.

## Regression test

`src/test-kernel/agent/modelHandlers/OpenAIResponseCompaction.vitest.ts` —
new case *"reuses a successful compaction across a same-turn retry instead of
re-compacting (chain-anchor/payload commit race)"*: turn 2 crosses the
compaction threshold, compacts successfully, then the outer `create()` call
fails once (simulating a transient error) forcing exactly the same
`messages` array to be retried. Asserts `/responses/compact` is invoked
exactly once across both attempts.

Verified failing pre-fix (`git checkout` on the production file only, test
kept): `compactRequests` had length 2 instead of 1. Re-applied the fix and
reran — passes.

## Net elements (R6)

- 0 new files, 0 new classes/functions.
- +1 field (`sourceMessages`) on the existing `compactionResult` object
  literal type in both `compactConversation()` and
  `compactConversationClientSide()`.
- 1 unconditional wipe → 1 conditional wipe (same call site).
- 1 new `if` branch in the existing `shouldCompact()` dispatch (reuse path),
  no new method.
- Net diff: +35/-5 in production code, +77 lines of test.

## Consumer counts (R8)

- `compactionResult` field: 2 producers (`compactConversation`,
  `compactConversationClientSide`), 5 pre-existing readers within the same
  class (`getBestInputTokenEstimate`, the background-resume context builder,
  `createResponseImpl`'s reuse/wipe logic, `applyCompactionState`) — all
  private to `ModelHandlerOpenAIResponse`; no other file reads this field, so
  widening its shape is confined to one class.
- `ResponseChainState.clearChainForCompaction()`: 2 callers, both in this
  same file — unchanged by this PR.

## Verified

- `src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts` — the fix.
- `src/agent/modelHandlers/openai/ResponseChainState.ts` — confirmed
  `clearChainForCompaction()` is the permanent, cross-retry mutation this fix
  reconciles the payload's lifetime with.
- `src/agent/node/index.ts` — confirmed `Node._exec` retries `exec(prepRes)`
  with the same `prepRes` object across attempts.
- `src/agent/core/flows/ModelInvocationNode.ts` — confirmed `exec()` passes
  `prepRes.messages` straight through to `createResponse()`, so the retry
  reuses the identical array reference.
- `src/test-kernel/agent/modelHandlers/OpenAIResponseCompaction.vitest.ts` —
  existing coverage + new regression test.

---

> [!NOTE]
> **Medium Risk**
> Touches conversation compaction and request input selection in the OpenAI Responses handler; incorrect cache logic could drop messages or skip compaction, but behavior is heavily regression-tested.
> 
> **Overview**
> Fixes a **retry race** in OpenAI Responses compaction: after a successful compact, `clearChainForCompaction()` stayed committed across PocketFlow same-turn retries, but `createResponseImpl` **always cleared** `compactionResult`, so failed outer `create()` calls re-ran `/responses/compact` unnecessarily.
> 
> **`compactionResult`** now records **`sourceMessages`** (the `messages` array reference compaction used). At the start of `createResponseImpl`, the cache is dropped only when `sourceMessages !== messages`; when they match, the handler **reuses** the compacted payload instead of compacting again. **`applyCompactionState`** clears `compactionResult` on **every successful** response so a cache keyed only by reference cannot leak into the next turn when `shared.messages` is mutated in place.
> 
> Regression tests cover same-turn retry (one compact call), cross-turn reuse with a stable array reference (incremental send, no stale compaction), and client-side compaction bookkeeping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3407d5873501d8bd49e00f125610f275be305c2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts
+++ b/src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts
@@ -675,10 +675,21 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
    * Result from compactConversation including messages and state updates.
    * State updates are returned but not applied - caller is responsible for
    * applying them only after successful API call to prevent stale state on retry.
+   *
+   * `sourceMessages` is the exact `messages` array reference compaction ran
+   * against — it's how {@link createResponseImpl} recognizes a same-turn retry
+   * (PocketFlow's `Node._exec` reuses the same `prepRes`, hence the same
+   * `messages` reference, across retry attempts) and reuses this result
+   * instead of re-running compaction. That reuse is what keeps this payload's
+   * retry lifetime matched to {@link ResponseChainState.clearChainForCompaction}'s
+   * anchor clear, which already survives retries permanently — without it, the
+   * anchor clear alone would survive while this payload got wiped every
+   * attempt, forcing a redundant re-compaction on each retry.
    */
   private compactionResult?: {
     compactedMessages: ResponseInputItem[];
     tokensAfter: number;
+    sourceMessages: ResponseInputItem[];
   };
 
   /**
@@ -855,7 +866,7 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
       // Store compacted messages for use in this request.
       // Mark as pending compaction - state will be finalized after successful API call.
       // This prevents stale state if API call fails and needs retry.
-      this.compactionResult = { compactedMessages, tokensAfter };
+      this.compactionResult = { compactedMessages, tokensAfter, sourceMessages: messages };
 
       return compactedMessages;
     } catch (err) {
@@ -1005,6 +1016,7 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
       // so an output-token underestimate could let through a request the backend
       // then rejects.
       tokensAfter: this.estimateResentInputTokens(compactedMessages),
+      sourceMessages: messages,
     };
     return compactedMessages;
   }
@@ -1429,9 +1441,17 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
       }
     }
 
-    // Clear any stale compaction result from previous attempts. A retained
-    // pending response is handled above before this state can be discarded.
-    this.compactionResult = undefined;
+    // Clear any stale compaction result from a genuinely different input. A
+    // same-turn retry (PocketFlow's Node._exec reuses the same prepRes, hence
+    // the same `messages` reference, across retry attempts) keeps its cached
+    // result below instead — otherwise the chain anchor that compaction
+    // already cleared on chainState (which survives retries permanently)
+    // would outlive this payload, forcing a redundant re-compaction on every
+    // retry. A retained pending response is handled above before this state
+    // can be discarded.
+    if (this.compactionResult?.sourceMessages !== messages) {
+      this.compactionResult = undefined;
+    }
 
     // Route through isBackgroundModeActive() so provider-profile policy actually
     // gates the request, not just the predicate.
@@ -1471,7 +1491,17 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
     let compactedThisCall = false;
     // Store compacted messages for return value (captured when compaction succeeds)
     let compactedMessages: ResponseInputItem[] | undefined;
-    if (this.shouldCompact()) {
+    if (this.compactionResult?.sourceMessages === messages) {
+      // Same-turn retry of an input that already compacted successfully on a
+      // prior attempt (see the cache check above): reuse it instead of
+      // hitting the compact endpoint again. Re-running compaction here would
+      // be a silent no-op from the caller's perspective but a real, wasted
+      // API round trip, since chainState's anchor clear already committed
+      // permanently and doesn't need redoing.
+      effectiveMessages = this.compactionResult.compactedMessages;
+      compactedThisCall = true;
+      compactedMessages = this.compactionResult.compactedMessages;
+    } else if (this.shouldCompact()) {
       // Capture whether this was a manual request before clearing the flag.
       const wasManualRequest = this.compactionRequested;
       // Clear manual compaction flag now that compaction is being attempted.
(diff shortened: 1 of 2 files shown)


LionSR:
## Review disposition (commit 3407d5873)

**real-fix-now — applied.** cursor[bot] (HIGH), chatgpt-codex-connector[bot] (P1), Copilot (2 comments), and a Claude review all converged on the same root cause, independently and unanimously: keying `compactionResult` reuse on `sourceMessages === messages` also matches the *next turn*, not just a same-turn retry.

Verified against HEAD (`82e3bce`, the only commit after all these reviews, formatting-only): `ModelInvocationNode.post()` mutates `shared.messages` in place via `replaceMessagesInPlace` (`target.length = 0; target.push(...)`, `CommonCycleTypes.ts:79-82`), and other sites (`ToolUseDispatchNode.ts:715,727`) `.push()` onto the same array too. So the array reference routinely survives across turns, not just across `Node._exec` retries of one turn — the premise the original fix relied on. `compactionResult` was never cleared after a *successful* call (`applyCompactionState()`'s own doc comment even said so, stale from the pre-PR unconditional-wipe behavior), so a later turn reusing the same reference would silently resend the prior turn's stale `compactedMessages` and drop everything appended since (tool outputs, new user turns) — a worse regression than the wasted-API-call bug this PR set out to fix.

**Fix applied:** `applyCompactionState()` (called only from `finalizeResponse()`'s success paths — non-streaming, streaming, WebSocket, and background-resume — never on a failed attempt still awaiting retry) now clears `compactionResult` unconditionally after consuming it. This keeps the original same-turn-retry reuse intact (a failed attempt's `compactionResult` survives because `finalizeResponse` never runs for it) while making it impossible for the cache to leak into a later turn. Updated the stale doc comments on `applyCompactionState()` and the `compactionResult` field that claimed the field survived until "the start of the next `createResponse()` call."

**Regression test added:** drives the handler the way the real PocketFlow cycle does — one shared `messages` array, mutated in place across three turns (turn 1: no compaction; turn 2: crosses threshold, compacts; turn 3: new message appended onto the *same* array reference, must NOT reuse turn 2's stale payload). Verified it fails on the pre-fix code (turn 3 wrongly resent turn 2's stale compacted content and dropped the newly appended message, exactly matching the bots' predicted failure mode) and passes with the fix.

Also updated the pre-existing `#7213` client-side-compaction test, which read `handler.compactionResult` *after* `createResponse()` resolved — that read is now `undefined` by design (cleared on success), so the test now snapshots `tokensAfter` from inside the outer `create()` mock (the correct in-flight window) and additionally asserts `compactionResult` is cleared post-success.

**Verified (this pass):**
- `src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts` — `applyCompactionState()`, `finalizeResponse()`'s 4 success call sites, the `sourceMessages` wipe/reuse branch, `getBestInputTokenEstimate()`'s only other reader.
- `src/agent/core/flows/ModelInvocationNode.ts`, `CommonCycleTypes.ts` (`replaceMessagesInPlace`) — confirmed in-place mutation preserves reference identity across turns.
- `src/test-kernel/agent/modelHandlers/OpenAIResponseCompaction.vitest.ts` — added the cross-turn regression test; reproduced pre-fix failure by temporarily reverting only the new clear line, confirmed both the new test and the updated `#7213` test fail without the fix and pass with it.
- Full model-handler suite (`src/test-kernel/agent/modelHandlers/`): 433 passed, 4 skipped, 0 failed.
- `npm run typecheck` (root, test-kernel, extension, CLI, trace-viewer, desktop): all clean.
- `eslint` on both changed files: clean.

No other findings — Copilot's two comments and the `claude[bot]` review comment describe the same issue as the two bot findings above; nothing else was raised.

**Verdict: recycling-after-fix.**

PR #8082, opened 2026-07-11
fix(tools): add disk-based resume fallback for claude_agent sessions

LionSR:
## Finding

**MED race** (round-2 APoSD audit): `claude_agent` tool resume has no disk-based
fallback when the in-memory session registry is torn down mid-approval —
asymmetric with `codex.ts`, which handles this case.

Pinned repro (pre-fix, `HEAD` = `b235d14e5`):

`src/tools/claudeAgent.ts` `ClaudeAgentTool.execute()` called
`resumeAgentCliSession(ClaudeAgentSessions, { id: input.session_id, ... })`
unconditionally whenever `input.session_id` was set — with no
`ClaudeAgentSessions.isActive(...)` gate. `resumeAgentCliSession` (in
`src/tools/agentCliShared.ts`) throws a `ToolError` the instant
`store.lookup(id)` is `undefined`:

```ts
const stored = store.lookup(id);
if (!stored) {
  throw new ToolError(
    `${labels.notActiveLabel} '${id}' is not active. It may have completed or been stopped; start a new session without ${labels.idParamName}.`,
  );
}
```

The in-memory `ClaudeAgentSessions` map is torn down on extension
reload/crash (it's a process-local `Map`), so any follow-up on a still-valid
Claude Code CLI session (the SDK can resume it from disk via its `resume`
option) became a hard tool error instead of continuing. `codex.ts` already
guards this exact case:

```ts
if (input.thread_id && CodexThreads.isActive(input.thread_id)) {
  return resumeAgentCliSession(CodexThreads, { ... });
}
// Fall through when the thread's in-memory loop is gone (extension
// reload, crash): createCodexThread resumes via the SDK from disk.
```

`claude_agent` had no equivalent — it's the asymmetry flagged by the audit.

## Fix

Mirror `codex.ts`'s pattern in `src/tools/claudeAgent.ts`:

- Gate the resume branch on `ClaudeAgentSessions.isActive(input.session_id)`
  before calling `resumeAgentCliSession`.
- When inactive, fall through to `launchClaudeAgentSession`, now seeded with
  the stale `session_id` as the first turn's `resumeSessionId` — the SDK's
  `resume` option — so the conversation continues from disk instead of
  starting a blind new session or erroring.
- `startClaudeAgentLoop` registers the known `session_id` up front via
  `onSessionStart` (reusing the same `registerSession` helper `onTurnSuccess`
  already used), mirroring codex's `registerThread` reuse — this closes the
  race where a concurrent call with the same `session_id` during the first
  fallback turn could bypass the in-memory guard and start a second parallel
  fallback loop.

No new abstraction: `registerSession` factors out the two call sites
(`onSessionStart`, `onTurnSuccess`) that already existed inline in
`onTurnSuccess`, exactly as codex's `registerThread` does for its two sites.

## Net elements (R6)

- **+0 new files** of production code (mirrors an existing pattern in the
  same file's sibling, `codex.ts` — no new module/port/facade).
- **+1 test file** (`ClaudeAgentResumeFallback.vitest.ts`), encoding the
  failure story directly against the tool's `execute()`.
- Net production diff: `src/tools/claudeAgent.ts` +41/-17 lines — one added
  parameter (`resumeSessionId`), one `isActive` gate, one small
  `registerSession` closure that *replaces* duplicate inline registration
  logic previously only in `onTurnSuccess` (no growth in call sites, just
  DRY'd across the two sites the same way `codex.ts` already had DRY'd).

## Consumer counts (R8)

- `registerSession` (new local closure, file-scoped, not exported): **2**
  call sites (`onSessionStart`, `onTurnSuccess`) — both pre-existing
  call sites, not a speculative single-caller extraction.
- `resumeSessionId` param on `startClaudeAgentLoop`: **1** call site
  (`launchClaudeAgentSession`), matching codex's `Thread`/`thread.id`
  resume-context threading (codex doesn't need an explicit param since
  `thread.id` is intrinsic to the `Thread` object it already threads
  through).

## Verified

- `src/tools/codex.ts` — read the exact fallback shape (`isActive` gate +
  fall-through launch + `registerThread` reused for `onSessionStart` /
  `onTurnSuccess`) mirrored here.
- `src/tools/claudeAgent.ts` — the fixed file.
- `src/tools/agentCliShared.ts` — confirmed `resumeAgentCliSession` throws
  `ToolError` on a registry miss (the exact defect).
- `src/tools/agentCliSessionStores.ts`, `src/tools/agentCliSessionRegistry.ts`
  — `ClaudeAgentSessions`/`AgentCliSessionRegistry` API (`isActive`,
  `register`, `release`).
- `src/agent/runtime/childRunLoop.ts` — `ChildRunStrategy.onSessionStart`
  contract ("registered up front... so a concurrent call... can't bypass
  the in-memory guard").
- `src/tools/core/base.ts` — confirmed `BaseTool.call()` catches `ToolError`
  and returns `{status:'error', ...}` rather than rejecting (shaped the
  test assertions).

## Tests

New: `src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts`

1. **Torn-down registry falls through instead of throwing** — `session_id`
   set, `ClaudeAgentSessions.isActive(...)` false (empty registry, as after
   an extension reload): asserts `result.status === 'executed'` (not
   `'error'`) and that a fresh child-run loop was launched.
2. **Fallback seeds the SDK resume option** — same setup, additionally
   drives the captured `strategy.launch(...)` and asserts the mocked SDK
   `query()` was called with `options.resume === 'stale-session'` — proving
   the fix truly resumes from disk rather than silently dropping context
   and starting a blind new conversation (a fix that just swallows the
   error would still pass test 1 but fail this one).
3. **Active session still takes the resume path** — `session_id` present
   AND registered as active: asserts the follow-up is enqueued
   (`resumeAgentCliSession`'s path) and no fresh loop is launched — guards
   against an overcorrection that always falls through.

**Proven to fail pre-fix**: reverted `src/tools/claudeAgent.ts` via
`git apply -R` on the isolated diff, re-ran the suite — tests 1 and 2 failed
(`expected 'error' to be 'executed'`, `startChildRunLoop` called 0 times)
exactly as the bug predicts; test 3 (active-session guard) passed unchanged
both before and after, confirming it's correctly scoped to the new gate only.
Then reapplied the fix and confirmed all 3 pass.

Commands run:

```
npx vitest run src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts \
  src/test-kernel/agent/followUp/ClaudeAgentProgressEvents.vitest.ts \
  src/test-kernel/agent/followUp/CodexProgressEvents.vitest.ts \
  src/test-kernel/tools/AgentCliSessionRegistry.vitest.ts \
  src/test-kernel/tools/ClaudeAgentConfig.vitest.ts
# 5 files, 18 tests passed

npm run typecheck
# tsc --noEmit (all workspaces) — clean

npx eslint src/tools/claudeAgent.ts src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts
# clean
```

`src/test-kernel/tools/` full run had 4 pre-existing unrelated failures
(`PRPollingSourceAnnotationPages`, `PRPollingSourceCiStarted`,
`SubagentExecutionChildStreamId`, `ToolAvailabilityAppSignals` — timeouts
under parallel load unrelated to agent-CLI resume). Verified
`SubagentExecutionChildStreamId` times out identically with
`src/tools/claudeAgent.ts` reverted to pre-fix content — confirmed
pre-existing, unrelated to this change.

---

> [!NOTE]
> **Medium Risk**
> Changes agent-CLI session routing and follow-up behavior after registry loss; regression tests cover the main paths but a narrow concurrent-fallback race remains (documented, same as codex).
> 
> **Overview**
> **Fixes `claude_agent` follow-ups after extension reload or crash** when the in-memory `ClaudeAgentSessions` registry no longer has the id but the SDK can still resume from disk.
> 
> `execute()` now only calls `resumeAgentCliSession` when `ClaudeAgentSessions.isActive(session_id)`; otherwise it falls through to a new child-run launch, passing the stale id as `resumeSessionId` so the first SDK turn uses `options.resume`. Session registration is centralized in a `registerSession` helper and, for fallback resumes, wired to `onSessionStart` so the id is active before the first turn finishes—matching `codex.ts`.
> 
> Adds **`ClaudeAgentResumeFallback.vitest.ts`** covering no-throw fallback, SDK `resume` seeding, eager `onSessionStart` registration, and unchanged behavior when the session is still active.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f75e70d64d29395227ef9d28ac1d1aafe6ba771e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts
+++ b/src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts
@@ -193,6 +193,40 @@ describe('claude_agent tool — resume fallback for a torn-down registry', () =>
     expect(callArgs.options.resume).toBe('stale-session');
   });
 
+  it('registers the fallback session_id up front via onSessionStart, ahead of the first turn completing', async () => {
+    setupCommonMocks();
+    mocks.query.mockReturnValue(
+      streamMessages([
+        {
+          type: 'result',
+          subtype: 'success',
+          session_id: 'stale-session',
+          result: 'Resumed and continued.',
+        },
+      ]),
+    );
+
+    const tool = new ClaudeAgentTool();
+    await tool.call({
+      prompt: 'continue the refactor',
+      session_id: 'stale-session',
+    });
+
+    expect(mocks.startChildRunLoop).toHaveBeenCalledTimes(1);
+    const [loopParams] = mocks.startChildRunLoop.mock.calls[0] as [
+      { strategy: ChildRunStrategy<unknown> },
+    ];
+
+    // Pins the eager-registration mechanism itself: onSessionStart must make
+    // the stale session_id observably active synchronously, independent of
+    // whether the first turn has resolved yet.
+    expect(ClaudeAgentSessions.isActive('stale-session')).toBe(false);
+    loopParams.strategy.onSessionStart?.({
+      executions: { getAgentHandleByStream: () => undefined } as any,
+    } as any);
+    expect(ClaudeAgentSessions.isActive('stale-session')).toBe(true);
+  });
+
   it('still enqueues a follow-up (no fresh launch) when session_id IS active in the registry', async () => {
     setupCommonMocks();
     ClaudeAgentSessions.register('sess-resumed', {
--- a/src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts
+++ b/src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts
@@ -1,0 +1,217 @@
+// Regression coverage for the claude_agent tool's resume fallback: when a
+// caller passes a session_id whose in-memory ClaudeAgentSessions registry
+// entry is gone (extension reload, host crash, or a stale id from an older
+// run), the tool must NOT throw a "not active" ToolError like a raw
+// resumeAgentCliSession() call would. It must mirror codex.ts's pattern —
+// fall through to launching a fresh session that resumes the SDK session
+// from disk via the `resume` option, rather than silently starting a blind
+// new conversation or surfacing an error to the caller.
+
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+import { createRunTrace, StreamLogStore } from '@transcript';
+import type { AgentTrace } from '@agent/trace';
+import type {
+  ChildRunPorts,
+  ChildRunStrategy,
+} from '@agent/runtime/childRunLoop';
+import type { ExecutionId, StreamTabId } from '@shared/schemas';
+import { ClaudeAgentSessions } from '@tools/agentCliSessionStores';
+import type { ChildStream } from '@tools/childStream';
+
+const mocks = vi.hoisted(() => ({
+  requestBashApproval: vi.fn(),
+  getCurrentToolContexts: vi.fn(),
+  registerExecution: vi.fn(),
+  getExecutionStore: vi.fn(),
+  ensureRunDir: vi.fn(),
+  createChildStream: vi.fn(),
+  startChildRunLoop: vi.fn(),
+  currentSession: vi.fn(),
+  query: vi.fn(),
+}));
+
+vi.mock('@tools/approval/bashApproval', () => ({
+  requestBashApproval: mocks.requestBashApproval,
+  buildBashApprovalRejectedResult: vi.fn(),
+}));
+
+vi.mock('@agent/followUp/ToolFileInteractionContext', () => ({
+  getCurrentToolContexts: mocks.getCurrentToolContexts,
+}));
+
+vi.mock('@agent/runtime/RunContext', () => ({
+  getRunContextExecutionId: (ctx: any) => ctx?.executionId,
+  getRunContextStreamId: (ctx: any) => ctx?.streamId,
+  getRunContextWorkingDirectory: (ctx: any) => ctx?.workingDirectory,
+  getRunContextRuntimeHost: (ctx: any) => ctx?.runtimeHost,
+}));
+
+vi.mock('@agent/runtime/SessionHandle', () => ({
+  currentSession: mocks.currentSession,
+}));
+
+vi.mock('@agent/storage', () => ({
+  registerExecution: mocks.registerExecution,
+  getExecutionStore: mocks.getExecutionStore,
+}));
+
+vi.mock('@utils/files/taskRunStorage', () => ({
+  ensureRunDir: mocks.ensureRunDir,
+}));
+
+vi.mock('@tools/childStream', () => ({
+  createChildStream: mocks.createChildStream,
+}));
+
+vi.mock('@agent/runtime/childRunLoop', () => ({
+  startChildRunLoop: mocks.startChildRunLoop,
+}));
+
+vi.mock('@tools/claudeAgentConfig', () => ({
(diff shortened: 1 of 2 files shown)


LionSR:
## Triage disposition

Verified all 4 findings against HEAD at the time they were posted (`03990eb22`, the PR's only commit — every review comment postdates it, so nothing was stale). Fixed in `f75e70d64`.

### Fixed now

1. **Race-closure overclaim** (claude[bot] + codex P2, `claudeAgent.ts:403-476`/`:429`) — CONFIRMED. Read `codex.ts`'s `registerThread`: it has the identical await-before-registration shape (`isActive` gate in `execute()` → several awaits → `onSessionStart` fires inside `startCodexLoop`), and its own doc comment makes the same "can't bypass the in-memory guard" overclaim. So this is a pre-existing gap this PR mirrors faithfully, not a new defect it introduces — but the PR's comment/description language claimed the fallback "closes" the race, which overstates it. **Fixed**: softened the doc comment on `resumeSessionId` to say it *narrows* the window and explicitly names the residual gap, matching codex's inherited shape, instead of claiming closure.
   - **Not fixed (follow-up)**: the actual synchronization fix (reserve the session id synchronously before the async launch chain, as codex's comment suggests) needs a real primitive — `AgentCliSessionRegistry.register()` requires a full entry with `executions: ExecutionRegistry`, which doesn't exist until the SDK session starts, so a true "reserve" would need a placeholder-state concept the registry doesn't have today. That's legitimate scope creep for a symmetric fix across both `claudeAgent.ts` and `codex.ts` (the gap predates this PR in codex), not a minimal diff on this PR. Recommend a follow-up issue covering both files together rather than fixing one asymmetrically here.

2. **Non-null assertion nit** (claude[bot], `claudeAgent.ts:475`) — CONFIRMED, trivial. Fixed: captured `params.resumeSessionId` into a local `const fallbackSessionId` before the closure, dropping the `!`.

3. **Missing test for the eager-registration mechanism** (claude[bot], test file `:217`) — CONFIRMED gap. The three existing tests cover fallback-vs-resume routing and SDK `resume` seeding, but none exercised `onSessionStart` itself. Fixed: added a 4th test that drives `strategy.onSessionStart` directly and asserts `ClaudeAgentSessions.isActive('stale-session')` flips from `false` to `true` synchronously — pinning the actual registration mechanism, not just its downstream effects.

### Verification

- `npx vitest run src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts` — 4/4 pass (new test included).
- `npx vitest run src/test-kernel/tools/` — 79 files / 480 tests pass (full tools suite, no regressions).
- `npx eslint src/tools/claudeAgent.ts src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts` — clean.
- `npm run typecheck` (root tsc, test-kernel tsc, cli, trace-viewer, desktop) — all clean.

### Verdict: recycling-after-fix

The three fixed items were low-risk, minimal-diff corrections (a doc comment, a non-null assertion, a test case). The one deferred item (actual race closure) is a real but pre-existing, cross-file architectural gap shared with `codex.ts` — appropriately scoped as a follow-up rather than folded into this PR.

PR #8184, opened 2026-07-11
fix(cli-tui): align spinner, subagent-panel, and overflow-marker conventions

LionSR:
## Summary

Fixes the `tui-conventions` group of adversarially-verified UI-consistency findings (4 findings, 1 high).

### 1. `@inkjs/ui` Spinner → shared-Clock text indicator (HIGH) — fixed
`FormFrame.tsx`'s `renderAsyncListFormTransient` (used by AgentListForm, ModelListForm, ResumeListForm, MemoryListForm, SkillsListForm, ToolsListForm) and `ApiModeForm.tsx` rendered `@inkjs/ui`'s `<Spinner>`, which runs its own independent `cli-spinners`-driven `setInterval` (~80ms) outside the TUI's shared clock — the exact per-component-interval anti-pattern `useLiveNowMs` exists to avoid (see `ConversationPane.tsx`'s `ThinkingRow`, which already uses the shared clock for the same "work in progress" job).

Added `packages/cli/src/chat/tui/ui/LoadingIndicator.tsx`: a `<Text>`-based row that ticks a plain ASCII spin cycle (`|/-\`) off the existing `useLiveNowMs` shared 1 Hz clock, mirroring `ThinkingRow`. Swapped both `@inkjs/ui` `<Spinner>` call sites for it.

**`@inkjs/ui` dependency**: left in `package.json`. It's still used by `StatusBar.tsx`'s `<Badge>` and by `packages/cli/src/onboarding/runOnboarding.tsx`'s `<Spinner>` — neither is in this finding's scope (`consolidationSafe` names only the 2 call sites fixed here), so removing the dependency now would be an unscoped, unverified change to those two other surfaces.

### 2. Key-hint format divergence (StatusBar vs KeyHints) — investigated, left as-is with evidence
Attempted the recommended fix (StatusBar's bracket-style `[Key]action`/double-space bindings row re-skinned toward `KeyHints`' unbracketed/middot format). Reverted after verification surfaced a real behavior regression, not just a cosmetic diff:

[...]

PR #8177, opened 2026-07-11
fix(ui): consolidate empty-state markup onto renderEmptyState()

LionSR:
## Summary

Consolidates the 6 `empty-states`-group findings onto the shared `renderEmptyState()` helper (`src/shared/wa/emptyState.ts`) and one shared absolute-timestamp formatter (`@shared/utils/string`).

## Per-finding disposition

1. **Hand-rolled `.empty-state` markup in settingsView lists (GoalTab, MemoryList, HistoryList x2)** — Fixed. All four call sites now call `renderEmptyState({ icon, title, body?, headingTag: 'h3', className: 'empty-state' })`, matching `ProgressApp.ts`'s existing usage. `className: 'empty-state'` keeps the outer `<section>` on the shared `commonViewStyles` `.empty-state` rule (flex-column, centered `wa-icon` at 2.5x, italic), so no new CSS was needed for these three components.
2. **"Empty-state hero pattern hand-rolled in three settingsView lists" (duplicate of #1)** — Same fix; this is the duplicate the task description flagged, deduped against #1 rather than done twice.
3. **GoalTab's shadowing local `.empty-state` CSS rule** — Fixed by deletion. Removed GoalTab's local `.empty-state { padding; text-align; color }` block entirely (it silently overrode the shared `commonViewStyles` rule via cascade order); GoalTab now falls through to the same shared rule as History/Memory via `renderEmptyState()`'s `className` option, so no CSS is left to shadow anything.
4. **TaskGroupList's hand-rolled getting-started empty state with action buttons** — Fixed. Now calls `renderEmptyState()` with its existing `actions` array (label/icon/onClick), which already supports this exact shape (used identically by `ProgressApp.ts`'s "No runs yet" panel) — no extension to the helper was needed, so per the task's guidance ("extend ... ONLY if ... 2+ callers benefit — else compose") this is pure reuse, not a new API surface. Added scoped `.log-placeholder .empty-state-*` CSS in `logStyles.ts` (icon size, title/body typography, action row) to replace the now-dead `.log-placeholder-actions` rule it superseded. The finding's `consolidationSafe` was "report-only/low-confidence" because TaskGroupList's trigger (`hasStreams`, per-filter) differs from ProgressApp's (`hasAnyStreams$`, global) — that's about *when* each empty state shows, not whether the same rendering helper applies, so the fix stands.
[...]

PR #8176, opened 2026-07-11
fix(ui): standardize forms/toggles on wa-* components

LionSR:
## Summary

Fixes the 4 adversarially-verified `forms` findings from the UI-consistency audit.

### 1. ExternalInquiryPanel native `<textarea>` → wa-textarea (med)
`ExternalInquiryPanel` rendered its answer and session-links inputs as raw
native `<textarea>` elements with ~90 lines of hand-rolled focus/placeholder/
border CSS, while every sibling multiline input in progressView — including
this same component's inherited rejection-feedback box (`BaseFeedbackPanel`),
`FollowUpInput`, and `WorkflowToolUseFollowupSection` — uses `wa-textarea`.
Swapped both textareas to `wa-textarea` (`rows`/`resize="vertical"`, same
`.value=${live(...)}` + `@input` pattern already used by `FollowUpInput`),
deleted the now-dead focus/placeholder/border/background CSS, and re-pointed
the remaining min/max-height rules at `::part(textarea)` (same technique
`FollowUpInput` already uses for its own sizing). Event handlers now read
`.value` off the host element (same shape `UserQuestionPanel.updateFreeText`
already uses) instead of casting to `HTMLTextAreaElement`.

### 2. UserQuestionPanel native checkbox/radio → wa-checkbox/wa-radio (low)
Swapped the hand-rolled `<input type="checkbox"/"radio">` option list for
`wa-checkbox` (multiSelect questions) and `wa-radio-group` + `wa-radio`
(single-select questions), matching `AgentSelectionPanel` (wa-checkbox) and
`InstructionPanel` (wa-radio-group/wa-radio) elsewhere in the same webview
bundle. `commonViewStyles`' existing `wa-checkbox`/`wa-radio` compaction rules
now apply automatically. Behavior is unchanged: single-select semantics were
already "exactly one radio checked" even with native inputs, and
`wa-radio-group`'s built-in arrow-key navigation is a strict a11y upgrade over
the old plain `<label>` list, not a regression.

### 3. InstructionPanel vestigial hidden input → deleted (low)
Deleted `<input type="hidden" id="sessionType" .value=${session.sessionType}>`
sitting next to the `<wa-radio-group>` that already owns the same state.
Re-grepped the whole repo (`src/`, `packages/`) for
`getElementById('sessionType')` / `querySelector('#sessionType')` before
deleting — zero readers, confirming the finding.

### 4. wa-checkbox vs wa-switch for catalog-row "enabled" toggles (med)
[...]

PR #8173, opened 2026-07-11
fix(a11y): add keyboard activation to preset-card, parent-link, and proposal file-name spans

LionSR:
## Summary

Fixes the three `a11y-clickables` findings from the adversarial UI-consistency audit: three click-only clickable divs/spans that had no `role`/`tabindex`/`@keydown`, so keyboard and screen-reader users could never reach or activate them. Two of the three (`preset-card`, `.parent-link`) already shipped dead `:focus-visible` CSS assuming this was wired up — that CSS is now live.

## Per-finding disposition

1. **`MultiAgentTab.ts` `.preset-card` (team-preset selection)** — Fixed. Added `role="button"`, `tabindex="0"`, and a `handlePresetKey` Enter/Space handler on the card `<div>`, mirroring `GoalTab.ts`'s `.goal-row` and `AgentSelectionPanel.ts`'s `.agent-list-item`. The nested `wa-button` delete control already calls `event.stopPropagation()` on click, so the card's new keydown handler doesn't interfere with it (native button semantics own its own Enter/Space activation when it has DOM focus).
2. **`StreamHeader.ts` `.parent-link` span (go-to-parent-stream nav)** — Fixed. Added `role="button"`, `tabindex="0"`, and a `handleParentLinkKey` Enter/Space handler, reusing the existing `.parent-link:focus-visible` CSS that was previously unreachable. Kept the bespoke `.parent-link` class (icon + chip styling) rather than swapping to `.clickable-link` — the finding's recommendation is the accessibility *behavior*, and the visual design here is deliberately different from a plain text link (icon, padding, muted color), so this is a like-for-like behavior fix, not a restyle.
3. **`ProposalRequestPanel.ts` `.workflow-proposal__file-name` spans (click a file name to open it)** — Fixed. Added conditional `role="button"`/`tabindex="0"` (only on `clickable` spans — the read-only `Memories` group still renders as inert text, matching prior behavior), a `handleFileKey` delegate on the container using the shared `getComposedPathElement` helper (same delegation shape as `FileList.ts`'s `handleFileKey`), and a new `.workflow-proposal__file-name:focus-visible` outline (this file had no pre-existing focus-visible CSS to reactivate, so one was added to match the outline used by `.clickable-link`/`.parent-link`).

All three: same visual appearance, same click behavior — purely additive keyboard-activation parity (Enter and Space only; other keys are no-ops, verified in tests).

## In-flight PR interaction

- **#8163** (model-option provider icons) — different file, no interaction.
[...]

PR #8185, opened 2026-07-11
fix(ui): consolidate hand-rolled buttons onto wa-button + shared helper

LionSR:
Fixes 5 buttons-lens UI-consistency findings.

## Findings addressed

1. **TaskGroupList.ts** — "Show N older …" reveal control was a raw `<button>` with ~20 lines of hand-rolled border/hover/focus-visible CSS duplicating `wa-button`. Swapped to `wa-button` (`outlined`, `small`) and deleted the dead CSS in `logEntryStyles.ts`.
2. **StreamTabs.ts** — the expand/collapse chevron hand-rolled a `<button>` reset instead of the `wa-button action-icon-button` pattern used one element over (the delete button). Sizing now lives locally in `StreamTab.styles.ts` via `::part(base)` (this shadow root doesn't load `commonViewStyles`, so `.action-icon-button`'s cross-component rules don't actually apply there — kept behavior-identical rather than relying on a class that's inert in this scope).
3. **GettingStartedBanner.ts** — redefined its own `26px` button `min-height` instead of the shared `24px` `.actions` rule in `bannerStyles.ts` used by the other four inline banners. Kept the local flex-wrap layout class, added `actions` to pick up the shared sizing, deleted the bespoke override.
4. **packages/desktop/src/renderer/main.ts** — hand-wrote `wa-button` action markup for chrome buttons that the same directory's `actionButtons.ts` helper (already proven in `logsDrawer.ts`) covers. Routed Open Folder / Logs (×2) / Back to launcher / New run / Settings through `renderIconActionButton`/`renderLabeledActionButton`. The **Commands** button is intentionally left hand-rolled: it has no icon (the helper requires one) and needs `aria-haspopup`, which isn't part of the helper's option surface — documented inline.
5. **wa-button-group + external-tooltip workaround**, independently reimplemented in `LatexDiffsSection.ts`, `MemoryItem.ts`, and `StreamHeader.ts` (because `wa-button-group`'s corner-fusion CSS breaks if a `<wa-tooltip>` sits between grouped buttons). Extended `actionButtons.ts` with `renderIconActionButtonParts`/`renderLabeledActionButtonParts`, returning the button and tooltip as separate templates instead of concatenating them, and converged all three call sites onto it.
   - StreamHeader's toolbar additionally moved from a delegated `[data-command]` click handler to per-button `onClick` closures (the helper doesn't support delegated dispatch), which let `handleToolbarClick` be deleted outright.
   - LatexDiffsSection's two file-utility buttons (set current / clear) — cited in the same finding's line range but not actually inside a `wa-button-group` — also moved off their delegated `[data-file-action][data-file-type]` handler for the same reason, since it was a clean fit once switched to direct `onClick`.

`actionButtons.ts` gains two small optional fields needed by these call sites: `hidden` (reflects the `hidden` attribute) and `ariaHidden` (explicit `aria-hidden="true"/"false"`, independent of `hidden`).

## Net elements (R6)

**Added**: `renderIconActionButtonParts`, `renderLabeledActionButtonParts`, and `hidden`/`ariaHidden` options on the existing `IconActionButtonOptions` — all in the already-shared `src/shared/wa/actionButtons.ts`, not a new module.
**Deleted**: TaskGroupList's `.log-reveal-button`/`:hover`/`:focus-visible` CSS; `StreamTab.styles.ts`'s full `.tab-expand` reset (replaced by a 3-line override); LatexDiffsSection's `renderFileUtilityButton` method and the `[data-diff-action]` branch of its click handler (the whole handler is now file-action-only... actually fully removed along with the file-action branch too — see diff); StreamHeader's `handleToolbarClick` method and its now-unused `getComposedPathElement` import.
Net production diff across the 9 touched components/styles + `actionButtons.ts`: **-40 lines** (244 insertions, 284 deletions per `git diff --stat`), before the 4 new regression-test files.

## Consumer counts (R8)

- `renderIconActionButtonParts`/`renderLabeledActionButtonParts`: **3 callers** (LatexDiffsSection.ts, MemoryItem.ts, StreamHeader.ts) — the finding's central ask, not a single-caller extraction.
- `hidden` option: **1 caller so far** (main.ts's "Back to launcher" button) — an additive field (3 lines) on an already-multi-consumer shared helper (`actionButtons.ts` has 9 importers across desktop/webview/progressView/settingsView), not a new abstraction.
- `ariaHidden` option: **1 caller so far** (StreamHeader.ts's toolbar) — same additive-field reasoning; genuinely reusable (any icon-only toolbar that keeps an inert button in the layout but wants it out of the a11y tree).

## In-flight PR interaction

#8164 (StreamHeader `wa-tooltip` for goal-chip/progress-badge) touches `StreamHeader.ts` lines ~470-520; this PR touches the toolbar button-group block (~340-450) and deletes `handleToolbarClick` (~529-537, untouched by #8164). No line overlap; should merge cleanly in either order.

## Tests

- Extended `TaskGroupListIndex.vitest.ts` with a `tagName` assertion locking in the `wa-button` conversion.
- Added 4 new suites (no prior suite existed for these components), per the `useLitComponentTestDom` + new-suite-when-none-exists precedent (#8164): `StreamTabsExpandChevron.vitest.mts`, `StreamHeaderToolbar.vitest.mts`, `LatexDiffsSectionActions.vitest.mts`, `MemoryItemActionGroup.vitest.mts`.
- `npm run typecheck` (all workspaces) and the full `progressView`/`webview`/`settings`/`desktop` vitest suites (100 files, 715 tests) pass.

---

> [!NOTE]
> **Low Risk**
> UI-only refactors with preserved event contracts and new component tests; no auth, data, or backend logic changes.
> 
> **Overview**
> Standardizes **Web Awesome** action controls across desktop, progress, settings, and webview by routing repeated `wa-button` markup through **`@shared/wa/actionButtons`**, and extends that helper so **`wa-button-group`** layouts can keep tooltips outside the group.
> 
> **Shared helper (`actionButtons.ts`):** Adds **`renderIconActionButtonParts`** / **`renderLabeledActionButtonParts`** (button + sibling tooltip templates), plus optional **`hidden`** and **`ariaHidden`** on icon buttons.
> 
> **Call-site migrations:** Desktop chrome and empty-workspace actions use **`renderIconActionButton`** / **`renderLabeledActionButton`** (Commands palette trigger stays hand-rolled). **Stream header** toolbar builds from **`*Parts`**, drops delegated **`[data-command]`** handling in favor of per-button **`onClick`**. **LaTeXDiffs** and **memory item** grouped actions use the same **`*Parts`** pattern; file-utility buttons dispatch events directly. **Task group list** “show older” controls become **`wa-button`** (custom reveal CSS removed). **Stream tab** child expand chevron becomes **`wa-button`** with local **`::part(base)`** styling. **Getting started banner** picks up shared **`.actions`** sizing instead of a local min-height override.
> 
> **Tests:** New/updated Vitest coverage for toolbar tooltips/commands, expand chevron contract, memory/LaTeXDiffs button groups, and reveal **`WA-BUTTON`** tag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 330d3ddeaacf08926cde83948954c0f9ae2252e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
## Disposition: recycling-after-fix

Reviewed every inline comment (2) and review body (5, incl. Copilot's quota no-op) — all post-date the PR's single commit `a7f787501`, so none were pre-addressed.

**Fixed:**
1. `src/test-kernel/progressView/StreamHeaderToolbar.vitest.mts` duplicated an existing suite for the same component. Merged origin/main into this branch (which brought in `StreamHeader.vitest.mts` from #8164, landed after this PR was opened) and folded the toolbar `describe` block into it as a sibling `describe('stream-header toolbar', ...)`, deleting the standalone file. Also fixed a latent bug this surfaced: two `describe` blocks in the same file must share a single top-level `useLitComponentTestDom(...)` call (see `TaskGroupListIndex.vitest.ts`'s pattern) — calling it once per `describe` rebinds `customElements` to a second jsdom instance without `stream-header` registered on it, silently leaving `element.shadowRoot` undefined for the second block. Verified both blocks pass together now.
2. Added a one-line comment on `renderLabeledActionButtonParts` per the suggestion from two independent review passes, noting it's exported for API symmetry with `renderIconActionButtonParts`/`renderLabeledActionButton` rather than current caller count (not a §13 blocker per either reviewer, one of whom explicitly declined to flag it as a violation).

**Dismissed:** none — no stale/wrong findings on this PR; both real comments were genuine and unaddressed (single-commit PR, no prior fix round unlike #8165).

Verified: full diff via `gh pr diff`, both review bodies + inline comments, `git log` timestamps vs review timestamps, merged origin/main (StreamHeader.vitest.mts add/add-adjacent from #8164 — no conflict, StreamHeader.ts merged cleanly), full `npx vitest run` suite, `npm run typecheck`.

LionSR:
Resolved the merge conflict with origin/main (6 sibling UI PRs: #8165, #8172, #8173, #8176, #8177, #8184).

**Conflicts / merge risk found and fixed:**

1. **`src/test-kernel/progressView/StreamHeader.vitest.mts`** — real 3-way conflict. Main (#8173/#8164) had folded the `parent-link keyboard activation` and `tooltips` suites into one top-level `describe('stream-header', ...)` wrapping a single `useLitComponentTestDom` call; this branch had folded its own `toolbar` suite the same way but as a *separate* top-level describe. Combined all three suites (`parent-link keyboard activation`, `tooltips`, `toolbar`) under the one `describe('stream-header', ...)` block with a single shared `useLitComponentTestDom` call, per the trap the file's own comment documents (a second top-level call leaves the second suite's custom elements unregistered). Also merged `mount()`/`baseStream()` to keep both main's `overrides` param (needed by the parent-link tests) and this branch's `STREAM_PHASE.RUNNING`/`unsupportedCommands` defaults (needed for the toolbar buttons to render).

2. **`StreamHeader.ts`, `StreamTabs.ts`, `TaskGroupList.ts`** — auto-merged with no conflict markers, and on inspection correctly combine main's parent-link a11y attributes (`role`/`tabindex`/keydown), goal-chip/progress-badge `wa-tooltip[for]` anchoring, and the getting-started empty-state helper with this branch's `renderIconActionButtonParts` toolbar consolidation, deleted `handleToolbarClick`, and the `StreamTabs` chevron / `TaskGroupList` reveal-button `wa-button` migrations.

3. **Silent semantic break caught by typecheck/tests, not git**: `TaskGroupList.ts`'s auto-merge was textually clean but broken — main's #8177 (empty-states) independently dropped `waIcon` from an unrelated import line (its own render path stopped calling it directly, using `renderEmptyState`'s icon-name strings instead), and git's line-level 3-way merge silently kept that deletion even though this branch's already-migrated `renderRevealButton` (`wa-button` + `waIcon(...)`) still needs it. Caught by `npm run typecheck` (`TS2304: Cannot find name 'waIcon'`) and by the progressView vitest suite (3 failing tests, `ReferenceError: waIcon is not defined`). Restored the import.

**Verified:**
- Repo-wide `npm run typecheck` clean (root, test-kernel, extension, CLI, trace-viewer, desktop).
- `progressView` + `webview` + `settings` + `desktop` vitest suites: 104 files / 742 tests passing.
- No leftover conflict markers anywhere in the tree.

Pushed as `330d3ddea` (merge commit, two parents: prior branch tip `2236afb46` + `origin/main` `280a233e3`).

PR #8202, opened 2026-07-11
fix(exec): delegate array-form abort/force-kill to execa natives

### LINKED ISSUE #8183: execUtils re-implements execa v9 natives: cancelSignal and forceKillAfterDelay
## What

\`src/utils/system/execUtils.ts\` hand-rolls two things modern execa provides natively:

- a manual \`addEventListener('abort')\` → SIGTERM listener (\`installAbortListener\`, ~lines 247-258) ≙ execa's \`cancelSignal\` option;
- a manual SIGKILL-after-5s timer (\`forceKillTimeoutId\`, ~lines 238-244) ≙ execa's \`forceKillAfterDelay\` option.

## What must NOT change

The bespoke \`setTimeout\` + process-group kill for the **shell/string form** (\`signalProcessGroup\`: negative-PID on POSIX, tree-kill on Windows) is the wrapper's genuine value-add — execa's own timeout reaps only the shell and leaves piped children holding stdout. Keep it; only the abort listener and force-kill timer are delegable.

Behavior-sensitive (process teardown paths), so this wants focused kill-semantics testing rather than a drive-by swap. Found in the convergence audit (2026-07-11).

LionSR:
Fixes #8183.

## What

`executeCommand` in `src/utils/system/execUtils.ts` spawns via two shapes:

- **array form** (`execa(cmd, args, ...)`) — a single, non-detached child process.
- **shell/string form** (`execa(command, { shell: true, ... })`) — optionally `detached`, since it can front piped children (`find | head`) that need process-group teardown.

Previously *both* forms went through the same hand-rolled `installAbortListener` (a manual `AbortSignal` → `SIGTERM` listener) and `forceKillTimeoutId` (a manual SIGKILL-after-5s escalation), both driving termination through `signalProcessGroup` (negative-PID on POSIX / `tree-kill` on Windows).

The array form doesn't need any of that — there's no process group to reach, so `signalProcessGroup`'s negative-PID attempt there always failed and silently fell back to a plain `process.kill(pid, signal)`. It already delegated its `timeout` straight to execa's native option; this PR extends that same delegation to abort and force-kill by passing `cancelSignal`/`forceKillAfterDelay` into the array-form `execa()` call, and scopes `installAbortListener()`/`forceKillTimeoutId` to the shell/string form only, where they remain required exactly as documented in the issue ("What must NOT change").

Result interpretation (`exitCode`/`stderr` on abort) now merges the shell path's `shellAborted` flag with execa's own `result.isCanceled`, so `'Command aborted by user'` / exit code `130` are produced identically regardless of which path aborted the process.

## Net elements (R6)

- Deleted: the unconditional `installAbortListener()` call on the array-form path (now conditional on `!Array.isArray(command)`), and this path's implicit dependency on the hand-rolled `terminateSubprocess`/`forceKillTimeoutId` machinery — it now goes through execa's own `cancelSignal`/`forceKillAfterDelay` natives instead.
- Added: `cancelSignal`/`forceKillAfterDelay` passed into the array-form `execa()` call (2 lines); an `aborted` merge (`shellAborted || result.isCanceled`) replacing the old `shellAborted`-only check (1 line); explanatory comments.
- Kept unchanged, per the issue's own "What must NOT change": the shell/string form's hand-rolled `signalProcessGroup`-based abort listener, timeout, and force-kill timer — execa's native mechanisms only ever call `subprocess.kill()` on the single tracked child, not the process group, so they can't safely replace this path (piped children would be orphaned, reproducing the exact bug `signalProcessGroup` exists to avoid).
- Net diff: `src/utils/system/execUtils.ts` +19/-4 lines; no new dependency (execa is already a direct dependency, already imported in this file).

## Consumer counts (R8)

`executeCommand` has ~20 call sites across 15 files:

- **Array-form** (17 call sites, unaffected in practice except where noted): `applyPath.ts`, `githubSubscriptionTool.ts`, `isGitRepository.ts`, `workspaceInfo.ts` (×3), `img.ts` (×2), `worktreeInfo.ts` (×2), `toolUtils.ts`, `latexdiff.ts`, `diffCommandExecutor.ts` (×2), `updateChecker.ts` (CLI), `desktopSetupTerminal.ts`. Only `src/tools/grep.ts`'s `rg` call passes a live `signal` (agent-run cancellation) today — this is the one array-form site whose abort semantics actually changed underlying mechanism (from the degenerate hand-rolled fallback to execa's native `cancelSignal`), verified behavior-identical by the new regression test.
- **Shell/string-form** (3 call sites, byte-for-bit unchanged): `bash.ts` (×2, foreground + background — the primary abort/timeout consumer, covered by the existing `aborts shell command process groups including children` test) and `desktopTerminalRunner.ts` (desktop setup terminal).

## Tests

- Extended `src/test-kernel/utils/system/SystemUtils.vitest.ts` (existing `executeCommand` describe block, R7) with `aborts array-form commands via execa native cancelSignal`, which spawns a long-lived array-form Node subprocess, aborts it mid-flight, and asserts the same `exitCode`/`stderr`/process-teardown contract as the pre-existing shell-form abort test.
- Regression-proven: temporarily reverted just the `cancelSignal`/`forceKillAfterDelay` array-form delegation (keeping the new `!Array.isArray(command)` guard that skips the hand-rolled listener) — the new test then hung and failed with `Test timed out in 10000ms`, confirming it catches an incomplete delegation. Restored before committing.
- `npx vitest run src/test-kernel/utils/system/SystemUtils.vitest.ts` — 15/15 passed.
- `npx vitest run src/test-kernel/tools/BashTool.vitest.ts src/test-kernel/tools/BashToolErrorFeedback.vitest.ts src/test-kernel/tools/Wolfram.vitest.ts` — 16/16 passed (shell-form + array-form consumer suites).
- `npm run typecheck` — clean across the whole workspace.

Closes #8183

---

> [!NOTE]
> **Medium Risk**
> Touches shared command execution and cancellation used by many tools; shell-form behavior is unchanged but array-form abort now relies on execa natives.
> 
> **Overview**
> **Array-form** `executeCommand` now passes `cancelSignal` and `forceKillAfterDelay` into execa instead of using the hand-rolled `AbortSignal` listener and process-group kill path. **Shell/string** commands still use `installAbortListener` and `signalProcessGroup` so piped children are not orphaned.
> 
> Aborted runs are normalized the same way for both shapes: `shellAborted` is combined with execa's `result.isCanceled` so callers still get exit code **130** and `'Command aborted by user'` when stderr is empty.
> 
> A regression test asserts array-form abort via execa's native cancel path (non-Windows).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cabb9a702de7399860597d2ecfe87d1d9e0f0fbd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/utils/system/SystemUtils.vitest.ts
+++ b/src/test-kernel/utils/system/SystemUtils.vitest.ts
@@ -43,6 +43,14 @@ async function waitForProcessExit(pid: number): Promise<void> {
   throw new Error(`Process ${pid} was still running after abort`);
 }
 
+async function waitForCondition(condition: () => boolean): Promise<void> {
+  for (let attempt = 0; attempt < 50; attempt += 1) {
+    if (condition()) return;
+    await sleep(20);
+  }
+  throw new Error('Timed out waiting for condition');
+}
+
 describe('executeCommand', () => {
   const tempDirs: string[] = [];
 
@@ -101,6 +109,34 @@ describe('executeCommand', () => {
     assert.equal(result.timedOut, false);
     assert.equal(result.stderr, 'Command aborted by user');
     await waitForProcessExit(childPid);
+  });
+
+  it('aborts array-form commands via execa native cancelSignal', async () => {
+    if (process.platform === 'win32') return;
+
+    const controller = new AbortController();
+    let childPid: number | undefined;
+    const promise = executeCommand(
+      [process.execPath, '-e', 'setTimeout(() => {}, 60000)'],
+      {
+        signal: controller.signal,
+        timeout: 60_000,
+        onPid: (pid) => {
+          childPid = pid;
+        },
+      },
+    );
+
+    await waitForCondition(() => childPid !== undefined);
+    assert.ok(childPid && childPid > 0);
+
+    controller.abort();
+    const result = await promise;
+
+    assert.equal(result.success, false);
+    assert.equal(result.timedOut, false);
+    assert.equal(result.stderr, 'Command aborted by user');
+    await waitForProcessExit(childPid!);
   });
 });
 
--- a/src/test-kernel/utils/system/SystemUtils.vitest.ts
+++ b/src/test-kernel/utils/system/SystemUtils.vitest.ts
@@ -119,7 +119,11 @@ describe('executeCommand', () => {
       },
     );
 
-    for (let attempt = 0; attempt < 50 && childPid === undefined; attempt += 1) {
+    for (
+      let attempt = 0;
+      attempt < 50 && childPid === undefined;
+      attempt += 1
+    ) {
       await sleep(20);
     }
     assert.ok(childPid && childPid > 0);
--- a/src/test-kernel/utils/system/SystemUtils.vitest.ts
+++ b/src/test-kernel/utils/system/SystemUtils.vitest.ts
@@ -43,14 +43,6 @@ async function waitForProcessExit(pid: number): Promise<void> {
   throw new Error(`Process ${pid} was still running after abort`);
 }
 
-async function waitForCondition(condition: () => boolean): Promise<void> {
-  for (let attempt = 0; attempt < 50; attempt += 1) {
-    if (condition()) return;
-    await sleep(20);
-  }
-  throw new Error('Timed out waiting for condition');
-}
-
 describe('executeCommand', () => {
   const tempDirs: string[] = [];
 
@@ -127,7 +119,9 @@ describe('executeCommand', () => {
       },
     );
 
-    await waitForCondition(() => childPid !== undefined);
+    for (let attempt = 0; attempt < 50 && childPid === undefined; attempt += 1) {
+      await sleep(20);
+    }
     assert.ok(childPid && childPid > 0);
 
     controller.abort();
@@ -135,6 +129,7 @@ describe('executeCommand', () => {
 
     assert.equal(result.success, false);
     assert.equal(result.timedOut, false);
+    assert.equal(result.exitCode, 130);
     assert.equal(result.stderr, 'Command aborted by user');
     await waitForProcessExit(childPid!);
   });
--- a/src/utils/system/execUtils.ts
+++ b/src/utils/system/execUtils.ts
@@ -229,6 +229,11 @@ export async function executeCommand(
     let shellTimedOut = false;
     let shellAborted = false;
 
+    // Only the shell/string form needs this hand-rolled abort + force-kill
+    // machinery: it terminates via `signalProcessGroup` (negative-PID /
+    // tree-kill) so piped children don't outlive the shell. The array form
+    // spawns a single non-detached process, so it delegates straight to
+    // execa's own `cancelSignal` / `forceKillAfterDelay` natives below.
     const terminateSubprocess = (signal: NodeJS.Signals): void => {
       const pid = subprocess.pid;
       if (!pid) return;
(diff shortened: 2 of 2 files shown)


LionSR:
## Review triage (verified at HEAD)

**Verified:**
- Cloned `execa@9.6.1` (the version pinned/installed in this repo) and read `lib/terminate/timeout.js` / `lib/terminate/cancel.js` / `lib/methods/main-async.js`; also ran an empirical repro (`execa(..., { timeout: 100 })` with no `cancelSignal`).
- `src/utils/system/execUtils.ts` (both forms of `executeCommand`), `src/test-kernel/utils/system/SystemUtils.vitest.ts` (both abort tests), `src/tools/grep.ts` (the one array-form call site that passes `signal` today).

**Fixed (pushed in 95f9126):**
1. **`waitForCondition` single-caller extraction** (claude[bot], nit) — real violation of this repo's abstraction-cost guardrail (single-caller helpers aren't DRY). Inlined the 50×20ms poll loop directly into the new array-form abort test, per the reviewer's own suggested option.
2. **Missing `exitCode` assertion** (github-actions/texra-review, Low) — the PR description claimed the new test asserts the exitCode/stderr/process-teardown contract, but it only checked `success`/`timedOut`/`stderr`. Added `assert.equal(result.exitCode, 130)`.

Both suites re-run green after the fix: `SystemUtils.vitest.ts` 15/15, `BashTool.vitest.ts`+`BashToolErrorFeedback.vitest.ts`+`Wolfram.vitest.ts` 17/17, full workspace `npm run typecheck` clean.

**Dismissed as stale/incorrect:**
- **github-actions/texra-review Finding 1 (Medium)** — claims execa v9 sets `result.isCanceled = true` for both `cancelSignal` **and** `timeout`, so a timed-out array-form command would wrongly report `exitCode: 130` / `stderr: 'Command aborted by user'`. This is false for the execa version actually pinned here: `isCanceled` and `timedOut` are set from a single `context.terminationReason`, guarded by `??=` so only the *first* termination cause ever wins — `lib/terminate/timeout.js:18` sets `'timeout'`, `lib/terminate/cancel.js:17` sets `'cancel'`, and they're read back as mutually exclusive booleans in `lib/methods/main-async.js:172-173`. Confirmed empirically: `execa(cmd, { timeout: 100 })` with no `cancelSignal` → `{ timedOut: true, isCanceled: false }`. No fix needed; the suggested `&& !timedOut` patch would be a no-op change to already-correct behavior.

**Follow-up-grade (not blocking, not applied here):**
- **codex P2** — "relying only on `cancelSignal` can leave `await subprocess` hanging for array-form commands that spawn descendants inheriting stdio (e.g. `['bash', '-c', 'sleep 60 & wait']`), since execa's native path doesn't destroy hung streams the way the old 5s hand-rolled fallback did." This is architecturally plausible — I read `lib/terminate/kill.js`'s `forceKillAfterDelay` path and it does SIGKILL escalation only, no stream destruction, unlike the old `terminateSubprocess` fallback this PR now skips for array-form. But it's not a regression in any *currently exercised* call site: the PR body's own consumer audit shows only `grep.ts`'s `rg` invocation passes a live `signal` on the array-form path today, and `rg` is a plain single process (no inherited-stdio descendants), so its pipe closes cleanly on SIGKILL. Worth a tracking issue if a future array-form caller ever needs `signal` + a shell-spawning command, but out of scope for this fix (which is specifically about not needing the process-group machinery for the *current* array-form shape).

**Verdict: merge-now** (after the two nits above, now pushed).

PR #7964, opened 2026-07-10
feat(desktop): add lightweight update-check notification (#7682)

LionSR:
## What

Adds a lightweight update-check notification to the desktop app, closing
#7682.

## Why

Desktop was the only TeXRA distribution channel with no update surface at
all (no `updater`/`autoUpdater`/`checkForUpdate` anywhere in
`packages/desktop/src`, confirmed still true at this PR's base commit).
The issue's maintainer-approved decision (2026-07-09) picked **arm (b)**:
a lightweight version-check toast, explicitly *not* a full
`electron-updater` (feed generation / signing / rollback are out of
scope), sequenced after Stage-5 close-out (#6968, closed).

## What changed

- `packages/desktop/src/main/desktopUpdateChecker.ts` (new): polls the
  public `texra-ai/texra-desktop-releases` repo's latest GitHub release
  (unauthenticated REST API), compares against the running build's
  version with `semver`, and hands a newer release off to a caller-supplied
  `notify` callback. Throttled to once/day via two new desktop-only
  `GlobalStateKey` entries (`DESKTOP_UPDATE_CHECK_LAST_CHECKED_AT`,
  `DESKTOP_UPDATE_CHECK_LAST_NOTIFIED_VERSION` in
  `src/shared/state/stateKeys.ts`, following the existing
  `DESKTOP_CRASH_REPORTING_ENABLED` desktop-only-key precedent), and
  notifies at most once per release version. Skips entirely for
  unpackaged/dev runs and when `TEXRA_NO_UPDATE_CHECK` is set (mirrors the
  CLI's `updateChecker.ts` opt-out).
- `packages/desktop/src/main/index.ts`: wires the check into
  `createWindow`, firing once on startup (non-blocking,
  `.catch(reportAsyncError)`). The `notify` callback reuses the existing
  `dialog.showMessageBox` pattern already used throughout desktop
  (`showInfoMessage`/`showErrorMessage`) with a "Download" button that
  opens the GitHub release page via `shell.openExternal` — no download, no
  install.
- `packages/desktop/package.json`: adds `semver` (dependency) /
  `@types/semver` (devDependency), reusing the exact version pins already
  used by the CLI's `updateChecker.ts`.

Explicitly out of scope (per the decision): `electron-updater`,
`latest.yml` feeds, code-signing interplay, auto-install.

## Verification commands

```
cd packages/desktop && npx tsc --noEmit -p tsconfig.main.json   # clean
npm run typecheck                                                 # clean (covers desktop)
[...]

PR #8203, opened 2026-07-11
fix(cli): throttle the CLI update check to once per day

### LINKED ISSUE #8168: CLI update check hits the npm registry on every interactive launch (no cache, blocks up to 2.5s)
## What

\`packages/cli/src/runtime/updateChecker.ts\` fetches \`<registry>/<pkg>/latest\` inline on interactive launch with a 2.5s timeout and no result caching, so every TUI start pays a network round-trip (or the full timeout when offline/slow).

## Suggestion

Adopt the caching layer of the \`update-notifier\` approach: persist the last check result + timestamp (e.g. in the existing JsonStore) and only re-check after a TTL (update-notifier defaults to 1/day), ideally off the critical startup path. The rest of the file (self-install via detected package manager, Homebrew tap refresh + \`brew info\` parsing) does more than update-notifier and should stay.

Found in a reinvented-wheel audit (2026-07-11). Related prior art: update-notifier does registry check + cache + detached process; only its check-cache concept is worth adopting here, not the whole dep necessarily.

LionSR:
## What

`packages/cli/src/runtime/updateChecker.ts`'s `notifyCliUpdate` hit the npm
registry (or refreshed the Homebrew tap + ran `brew info`) on every
interactive `texra chat`/`texra orchestrate` launch, with no caching — every
launch paid a network round-trip, or the full 2.5s timeout when offline.

The desktop update checker already solved this (#7964,
`checkForDesktopUpdate` in `desktopUpdateChecker.ts`): a daily throttle
persisted in global state, where the stamp is only written after a
*successful* fetch so a flaky/offline check retries on the very next launch
instead of being suppressed for a full day.

## Fix

Add `checkCliUpdateAvailable`, a small orchestration function that mirrors
`checkForDesktopUpdate`'s throttle/persist-on-success shape (same 24h
interval, same "only persist the timestamp after a successful fetch" rule),
gated on a new `GlobalStateKey.CLI_UPDATE_CHECK_LAST_CHECKED_AT` key.

`notifyCliUpdate` runs before `initInteractiveCliPlatform` (so `platform()`
isn't available yet), so it opens the same global `state.json` that
`createCliStateStores` opens later, directly via `JsonStore.open` +
`createNodeStorageProvider().getGlobalStoragePath()` — same file, same shape,
no new storage layer.

Deliberately NOT ported: the desktop's concurrent-check coalescing (Electron
can re-invoke the checker on window recreation; the CLI only ever calls
`notifyCliUpdate` once per process, already guarded by the existing
`notified` flag) and the "don't re-notify the same version" dedup (the CLI's
prompt is already interactive/opt-in each time it fires, unlike the desktop's
passive native dialog — the issue's problem is the *network hit*, not
prompt-nagging).

## Net elements (R6)

- **Added**: 1 exported function (`checkCliUpdateAvailable`) + its options
  interface, 1 new `GlobalStateKey` enum member, 1 new `const
  CHECK_INTERVAL_MS`.
- **Removed**: the inline unconditional `fetchLatestHomebrewFormulaVersion()`
  / `fetchLatestCliVersion()` call in `notifyCliUpdate` is now routed through
  the throttled helper — no separate cleanup needed, it's a straight swap at
  the same call site.
- **No new dependency**: reuses the existing `JsonStore` /
  `createNodeStorageProvider` platform primitives already used by
  `cliStateStores.ts` for the same global `state.json` file.

## Consumer counts (R8)

- `checkCliUpdateAvailable` — 1 production caller (`notifyCliUpdate`) + a
  dedicated `describe` block in the existing
  `src/test-kernel/cli/UpdateChecker.vitest.mts` suite (4 new tests:
  first-launch check, not-newer no-op, same-day throttle, failed-fetch stamp
  is not persisted).
- `GlobalStateKey.CLI_UPDATE_CHECK_LAST_CHECKED_AT` — 1 write site, 1 read
  site (both inside `checkCliUpdateAvailable`), asserted directly in the new
  tests (mirrors how the desktop suite asserts
  `DESKTOP_UPDATE_CHECK_LAST_CHECKED_AT`).

## Testing

- `npx vitest run --config vitest.config.mjs src/test-kernel/cli` — 130 test
  files / 1677 tests pass, including the 4 new regression tests (which fail
  against the pre-fix code: before this change there is no throttle at all,
  so a test asserting "second call within the same day makes no additional
  fetch" fails).
- `npx vitest run --config vitest.config.mjs src/test-kernel/desktop/DesktopUpdateChecker.vitest.mts` — unaffected, still 12 tests passing (confirms the desktop checker's own shape/tests are untouched).
- `npm run typecheck` — clean across all workspace packages.
- `npx eslint` / `npx prettier --check` on the touched files — clean.

Closes #8168

---

> [!NOTE]
> **Low Risk**
> Best-effort startup path only; storage or network errors stay silent and do not change auth or session behavior.
> 
> **Overview**
> **Stops the CLI from hitting npm or Homebrew on every interactive launch** by throttling version lookups to at most once per 24 hours, aligned with the desktop update checker.
> 
> `notifyCliUpdate` no longer calls `fetchLatestCliVersion` / `fetchLatestHomebrewFormulaVersion` unconditionally. It opens global `state.json` early (before platform init) and routes through new **`checkCliUpdateAvailable`**, which reads/writes `GlobalStateKey.CLI_UPDATE_CHECK_LAST_CHECKED_AT`. The last-check timestamp is **only saved after a successful fetch**, so offline or registry failures retry on the next launch instead of waiting a full day.
> 
> Four unit tests cover first launch, not-newer, same-day throttle, and failed-fetch without persisting the stamp.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 465d2f32b0167d584a62fc9ef4d5f9171af0d5e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/runtime/updateChecker.ts
+++ b/packages/cli/src/runtime/updateChecker.ts
@@ -1,10 +1,14 @@
 import { spawn } from 'node:child_process';
+import * as path from 'node:path';
 import { fileURLToPath } from 'node:url';
 
 import { gt as semverGt, valid as semverValid } from 'semver';
 import { z } from 'zod';
 
+import { JsonStore } from '@platform/defaults/jsonStore';
+import { createNodeStorageProvider } from '@platform/defaults/nodeStorage';
 import { parseJsonWith } from '@common/parsing/safeParseJson';
+import { GlobalStateKey } from '@shared/state/stateKeys';
 import { executeCommand } from '@utils/system/execUtils';
 
 import {
@@ -16,6 +20,7 @@ import {
 import { CliExitCode } from './exitCodes';
 import { askCliQuestion, writeTextStderr } from './logSinks';
 import { createCliStyle, type CliStyle } from './style';
+import type { StateStore } from '@platform/interfaces';
 
 /** Published package name on npm; the `texra` bin lives here. */
 const CLI_PACKAGE_NAME = '@texra-ai/cli';
@@ -30,6 +35,11 @@ const DEFAULT_REGISTRY = 'https://registry.npmjs.org';
 const DEFAULT_TIMEOUT_MS = 2500;
 const HOMEBREW_COMMAND_TIMEOUT_MS = 10000;
 const UPDATE_CHECK_SKIP_ENV = 'TEXRA_NO_UPDATE_CHECK';
+/**
+ * Check for a new release at most once per day, matching the desktop update
+ * checker's cadence (see `desktopUpdateChecker.ts`'s `CHECK_INTERVAL_MS`).
+ */
+const CHECK_INTERVAL_MS = 24 * 60 * 60 * 1000;
 
 /** Package manager the running binary was installed with. */
 export type InstallMethod = 'npm' | 'pnpm' | 'yarn' | 'bun' | 'brew';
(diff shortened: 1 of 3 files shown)


LionSR:
## Review triage

Codex posted 3 P2 findings on `packages/cli/src/runtime/updateChecker.ts`. Verified all three at branch HEAD.

**1. Guard update-check state failures — CONFIRMED, fixed (465d2f32b).**
`notifyCliUpdate` opened the global `state.json` via `JsonStore.open` and awaited `checkCliUpdateAvailable` with no `try`/`catch`. `JsonStore.open`'s `mkdir`, `readJsonRecord` (non-ENOENT errors rethrow), and `flush`'s `mkdir`/`write-file-atomic` can all reject (EACCES, ENOSPC under the global-storage dir). Both call sites (`chat.ts`, `orchestrate.ts`) `await notifyCliUpdate(context)` unguarded, so a filesystem error there would propagate straight out and abort interactive CLI startup — contradicting the function's own doc comment ("Failures are silent so a flaky network never gets between the user and their session"). Fixed by wrapping the state-store open + throttle check in `try { … } catch { return; }`, matching the existing fail-silent contract for the network-fetch path.

**2. Concurrent global-state clobbering — CONFIRMED but not fixed here; follow-up-grade.**
`JsonStore` snapshots `this.data` at `open()` time and flushes that snapshot on `update()`, with no re-read/merge before writing — so a second concurrent `texra` process writing an unrelated key during this check's ~2.5-10s network wait could get clobbered. This is a pre-existing design property of `JsonStore` shared by every other consumer of `state.json` (not introduced by this PR), and a correct fix (merge-on-write or field-level update) belongs in `JsonStore` itself, affecting all callers — out of proportion for this PR's scope. Noting for a dedicated follow-up; the race window is narrow (two concurrent CLI processes, one touching global settings inside the other's brief network wait).

**3. Don't stamp failed Homebrew refreshes — CONFIRMED but not fixed here; follow-up-grade.**
`fetchLatestHomebrewFormulaVersion` discards the result of `brew update --quiet` and always attempts `brew info` on the (possibly stale) local cache — this is pre-existing, intentional, and has its own regression test ("still reads formula info when the Homebrew tap refresh fails"). The gap is that `checkCliUpdateAvailable` treats any non-empty version as "successful fetch" for throttle-persist purposes, so an offline/failed brew-tap refresh that still yields a cached version gets the 24h throttle stamped, contrary to the PR's own stated retry-on-failure intent. A correct fix requires broadening the `fetchLatest` contract to separate "refresh succeeded" from "got a usable version," touching `fetchLatestHomebrewFormulaVersion`'s return shape and its 4 existing tests plus `checkCliUpdateAvailable`'s options — a larger, riskier change than this PR's scope. Flagging as a follow-up; impact is narrow (Homebrew-only, requires an offline/failed tap refresh plus a stale local cache) and non-crashing (worst case: a stale check silenced for 24h).

## Verified
- `packages/cli/src/runtime/updateChecker.ts` (full file, both HEAD and post-merge)
- `packages/cli/src/commands/chat.ts`, `packages/cli/src/commands/orchestrate.ts` (both unguarded `await notifyCliUpdate` call sites)
- `src/platform/defaults/jsonStore.ts` (confirmed `open`/`update` can throw)
- `src/test-kernel/cli/UpdateChecker.vitest.mts` (existing coverage, incl. the Homebrew "still reads on failed refresh" test)

## Testing after merge + fix
- Merged `origin/main` (1 conflict in `updateChecker.ts`, resolved by combining this PR's throttle with main's `resolveCliCwd`/cwd-aware brew runner)
- `npx vitest run --config vitest.config.mjs src/test-kernel/cli` — 130 files / 1679 tests pass
- `npm run typecheck` — clean across all workspace packages
- `npx eslint` / `npx prettier --check` on touched file — clean

Verdict: fixed finding #1 (real crash-risk on filesystem errors); #2 and #3 confirmed but deferred as follow-ups (pre-existing `JsonStore` design limitation and a narrow Homebrew-only throttle-accuracy gap, both requiring broader contract changes than this PR's scope).

PR #8205, opened 2026-07-11
refactor(cli): route CliSecrets through the shared JsonStore

### LINKED ISSUE #8182: CliSecrets duplicates JsonStore internals (write chain, tolerant read, atomic write) — converge like electronSecrets
## What

\`packages/cli/src/runtime/cliSecrets.ts\` hand-rolls three things \`JsonStore\` (\`src/platform/defaults/jsonStore.ts\`) already owns: an ordered mutation queue (\`mutationQueue\` ≙ JsonStore's \`writeChain\`), a tolerant read with hand-guard (\`isSecretFileData\` ≙ a zod \`z.record(z.string())\` at the store boundary), and the atomic write (now write-file-atomic after #8167, but still local plumbing). ~55 of its ~65 store LOC duplicate JsonStore internals.

## Why not done in the audit PRs

The desktop sibling \`electronSecrets.ts\` already wraps JsonStore for the same job and is the model — but converging the CLI needs a decision: JsonStore does not currently expose file-mode control, and the secrets file needs 0o600 (dir 0o700). Either JsonStore grows a \`mode\` option (small, benefits all stores) or the CLI keeps a thin permission layer. Env-var precedence (\`cliEnvValue\` overlay) stays either way.

Follow-up to #8167. Found in the convergence audit (2026-07-11).

LionSR:
## What

`packages/cli/src/runtime/cliSecrets.ts` hand-rolled a mutation queue, a tolerant read, and an atomic write that `JsonStore` (`src/platform/defaults/jsonStore.ts`) already owns and that `ElectronSecrets` already wraps for the desktop host. `CliSecrets` now goes through `JsonStore` the same way, and the ~55 duplicated store LOC (mutation queue, `isSecretFileData` guard, hand-rolled `mkdir`/`chmod`/`write-file-atomic`) are deleted.

## Decision: JsonStore grows two small options

The issue calls out the file-mode blocker explicitly; while wiring the convergence I found the read-tolerance semantics also needed a decision (see "Correctness note" below), so `JsonStore` grows **one options bag** (`JsonStoreOptions`) with two fields, both opt-in and defaulting to prior behavior for the other 5 existing `JsonStore.open()` call sites (unchanged, still compile/typecheck clean):

- `mode?: number` — POSIX mode for the store file (`0o600` for secrets); the containing directory is created/chmod'd to the same owner bits plus execute (`0o600` -> `0o700`), matching what `CliSecrets.writeSecrets()` did by hand.
- `strict?: boolean` — rethrow malformed JSON instead of `JsonStore`'s default self-healing "treat as empty" (see below).

## Correctness note: preserved, not regressed, #7478/#8167

`JsonStore`'s default read behavior (see `JsonStore.vitest.mts`, "recovers from malformed JSON stores and overwrites on the next write") silently discards corrupt JSON and starts the store from `{}`. That's the right default for state/config caches, but it is **exactly** the bug #7478 ("fix(cli): don't wipe CLI secrets on a transient read failure") fixed for secrets: a read-mutate-write store that silently defaults a parse failure to `{}` will happily overwrite the file with `{}` + one new key on the very next write, permanently destroying every other stored credential. `CliSecrets` opens its `JsonStore` with `strict: true` to keep rethrowing on corrupt JSON, so a subsequent `set()`/`delete()` aborts instead of wiping — same guarantee as before, now expressed as an owner option instead of duplicated logic. The existing regression test for that bug (`aborts a write instead of wiping the file when the read fails for a reason other than a missing file`) is untouched and still passes.

`CliSecrets` also keeps its per-call freshness: each `get`/`set`/`delete`/`listStoredKeys` opens its own `JsonStore` (rather than caching one for the instance's lifetime), so a fresh read happens on every operation — preserving safety across concurrently running `texra` invocations (separate processes), which is why `set()`/`delete()` originally did a full read before every write (see #8167's write-path hardening for the same file).

## Net elements (R6)

- `packages/cli/src/runtime/cliSecrets.ts`: 126 -> 82 lines (-44). Deleted: `mutationQueue`, `isSecretFileData`, `readSecrets()`, `writeSecrets()`.
- `src/platform/defaults/jsonStore.ts`: 107 -> 160 lines (+53). Added: `JsonStoreOptions` (2 fields, both documented), `dirModeFor()`, `ensureDir()` (shared by `open()`/`flush()`, 2 call sites).
- Net across the two production files: **+9 lines**, trading duplicated CLI-only logic for two small, documented, opt-in options on the shared owner that the desktop `ElectronSecrets` sibling and any future `JsonStore` consumer can also use.
- Tests: +49 lines (2 new tests: file/dir-mode regression in `CliSecrets.vitest.mts`; `strict: true` + `mode` regression in `JsonStore.vitest.mts`).

## Consumer counts (R8)

`JsonStore.open()` has 6 call sites across `electronSecrets.ts`, `cliSecrets.ts`, `desktopStreamSnapshot.ts`, `platform/index.ts` (x6 opens), `cliStateStores.ts` (x2), `initPlatform.ts` (x2). The new options are optional with defaults matching prior behavior — the other 5 call sites are untouched and unaffected (confirmed by a clean `npm run typecheck` across the whole workspace).

## Dependency decision

No new dependency. `write-file-atomic` was already a `JsonStore`/repo dependency (added via #8167); `CliSecrets` no longer imports it directly.

## Test plan

- `npx vitest run src/test-kernel/cli/CliSecrets.vitest.mts src/test-kernel/desktop/JsonStore.vitest.mts` — 7/7 pass (2 new: file/dir mode regression, `strict` regression).
- `npx vitest run src/test-kernel/cli/CliInitPlatform.vitest.mts src/test-kernel/cli/CliSecrets.vitest.mts src/test-kernel/cli/CliSupabaseAuth.vitest.mts src/test-kernel/desktop/ElectronAgentDirectories.vitest.mts src/test-kernel/desktop/ElectronConfig.vitest.mts src/test-kernel/desktop/ElectronPlatformAdapters.vitest.mts src/test-kernel/desktop/JsonStore.vitest.mts src/test-kernel/desktop/ElectronSecretsBootstrap.vitest.mts src/test-kernel/desktop/ElectronSecretStorageWarningDialog.vitest.mts` — 63/63 pass (every suite touching `JsonStore`/`CliSecrets`/`ElectronSecrets`).
- `npm run typecheck` — clean across the whole workspace (root, test-kernel, extension, CLI, trace-viewer, desktop).
- `npx eslint` / `npx prettier --write` on all changed files — clean.

Closes #8182

---

> [!NOTE]
> **Medium Risk**
> Touches credential persistence and read-failure semantics; behavior is intentionally preserved via strict mode and serialized mutations, with regression tests.
> 
> **Overview**
> **`CliSecrets`** no longer implements its own read-mutate-write, atomic flush, and directory `chmod` logic; it persists `secrets.json` via **`JsonStore`** with **`mode: 0o600`** and **`strict: true`**, aligned with desktop **`ElectronSecrets`**. Each operation still opens a fresh store for cross-process freshness; **`set()`/`delete()`** remain serialized through an instance **`mutationQueue`** so concurrent mutations in one process do not clobber each other. **`getStored`** only returns string values.
> 
> **`JsonStore`** gains optional **`JsonStoreOptions`**: **`mode`** applies owner-only file permissions and **`0o700`** on the parent directory (via **`ensureDir`** / **`dirModeFor`**); **`strict`** rethrows malformed JSON instead of self-healing to `{}`. Defaults preserve existing call sites.
> 
> Tests cover overlapping **`set()`** merges, non-string secret values, POSIX permissions, and **`strict: true`** parse failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0eff2bc948325f00183174c00413b01ff15d1c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/runtime/cliSecrets.ts
+++ b/packages/cli/src/runtime/cliSecrets.ts
@@ -1,15 +1,9 @@
 // Standard library imports
-import { chmod, mkdir, readFile } from 'node:fs/promises';
 import path from 'node:path';
 
-// Third-party imports
-import writeFileAtomic from 'write-file-atomic';
-
 // Local imports - platform
+import { JsonStore } from '@platform/defaults/jsonStore';
 import { DEFAULT_NODE_STORAGE_ROOT } from '@platform/defaults/nodeStorage';
-
-// Local imports - common
-import { isFileNotFoundError } from '@common/errors';
 
 // Local imports - CLI runtime
 import { cliEnvValue } from './cliContext';
@@ -17,27 +11,25 @@ import { cliEnvValue } from './cliContext';
 // Type imports - platform
 import type { PlatformSecrets } from '@platform/secrets';
 
-type SecretFileData = Record<string, string>;
-
-function isSecretFileData(value: unknown): value is SecretFileData {
-  return (
-    typeof value === 'object' &&
-    value !== null &&
-    !Array.isArray(value) &&
-    Object.values(value).every((entry) => typeof entry === 'string')
-  );
-}
+/** Secrets file is owner-only: `0o600` (containing dir gets `0o700`). */
+const SECRETS_FILE_MODE = 0o600;
 
 /**
  * CLI secret storage.
  *
  * Environment variables remain the highest-priority source so automation can
  * keep using ephemeral keys. Values written by CLI login are persisted under
- * the user's TeXRA state directory.
+ * the user's TeXRA state directory through the shared `JsonStore` (the same
+ * owner `ElectronSecrets` wraps for the desktop host), with `strict: true`
+ * so a corrupt secrets file aborts a write instead of silently wiping every
+ * other stored credential (see `JsonStoreOptions.strict`).
+ *
+ * Each operation opens its own `JsonStore` rather than caching one for the
+ * lifetime of this instance: separate `texra` invocations are separate
+ * processes, and a fresh open re-reads the file so a concurrently running
+ * CLI invocation's write isn't clobbered by a stale in-memory snapshot.
  */
 export class CliSecrets implements PlatformSecrets {
-  private mutationQueue: Promise<void> = Promise.resolve();
-
   constructor(private readonly filePath = cliSecretsPath()) {}
 
   async get(key: string): Promise<string | undefined> {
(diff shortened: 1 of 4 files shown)


LionSR:
## Review triage disposition

Verified every review comment against branch HEAD (`64124ebc`, now `a0eff2bc`). Two consensus findings were real and are fixed; one narrower finding is left as a follow-up.

### Fixed (pushed in `a0eff2bc9`)

1. **Overlapping writes lose secrets** (Cursor Bugbot — High; Codex P2; TeXRA review bot — Finding 1 medium; also flagged in the `claude[bot]` summary comment as "needs a fix... before merge"). Confirmed real: deleting the instance-level `mutationQueue` meant each `set()`/`delete()` opened an independent `JsonStore` off the same on-disk snapshot with its own `writeChain`; two overlapping mutations on one `CliSecrets` instance (e.g. `Promise.all([secrets.set(...), secrets.set(...)])`) would race and the later flush would silently drop the earlier key. Fix reintroduces a per-instance mutation queue around `openStore()+mutate+flush`, restoring the exact intra-process serialization the old hand-rolled code had, while reads (`get`/`getStored`/`listStoredKeys`) still open a fresh store per call (preserving the cross-process freshness the PR was built around). Added a regression test (`merges overlapping set() calls instead of one silently dropping the other`) that fails on the pre-fix code and passes now.

2. **`isSecretFileData` guard's per-value type check** (`claude[bot]` suggestion + Codex P2 "Validate stored CLI secret values"). Confirmed real, narrow: `store.get<string | undefined>(key, undefined)` was a bare cast with no runtime check, so a wrong-shaped value for one key (only reachable via manual `secrets.json` edits) would flow through to `getApiKey`'s truthiness check as if it were a valid credential. Restored the `typeof value === 'string'` check in `getStored()`. Added a regression test.

### Left as follow-up, not fixed here

- **Codex P2 "Preserve env-only CLI runs with unwritable storage"**: `JsonStore.open()` chmods the secrets directory (`0o700`) on every read, not just writes, which is new for `CliSecrets` specifically (the pre-existing `mkdir` step already ran on `open()` for all 6 `JsonStore` consumers before this PR). On a genuinely read-only-mounted storage root, this could make a pure `getApiKey()` read throw instead of falling through to the env-var, which the class's own docstring promises stays available for automation. Real but narrow (requires a first-ever invocation, no `secrets.json` yet, on a read-only-mounted home) and fixing it cleanly means deciding whether reads should tolerate `ensureDir` failure — that's a scope decision for `JsonStore`'s general contract (used by 5 other consumers), not a one-line fix in this PR. Flagging as a follow-up rather than blocking here.
- **PR body's R8 consumer count** (`claude[bot]`, "6 vs 7 `JsonStore.open(` call sites in `platform/index.ts`"): body-accuracy nit only, no code impact, not amending the body for it.

### Test/verification evidence
- `npx vitest run src/test-kernel/cli/CliSecrets.vitest.mts src/test-kernel/desktop/JsonStore.vitest.mts` — 9/9 pass (2 new).
- Full PR-listed suite (`CliInitPlatform`, `CliSecrets`, `CliSupabaseAuth`, `ElectronAgentDirectories`, `ElectronConfig`, `ElectronPlatformAdapters`, `JsonStore`, `ElectronSecretsBootstrap`, `ElectronSecretStorageWarningDialog`) — 65/65 pass.
- `npm run typecheck` — clean across the whole workspace (root, test-kernel, extension, CLI, trace-viewer, desktop).
- `npx eslint` / `npx prettier --check` on both changed files — clean.
- Merged `origin/main` into the branch first (clean merge, no conflicts) before applying the fix.

Verdict: **recycling-after-fix**, ready to merge once CI is green on `a0eff2bc9`.

Issue #7694, opened 2026-07-09
Decision: declare defaultSession() permanent + invert the alias (delete Shared*/StreamStatusService module exports, 16 sites)


LionSR:
**TD-3 (notable, decision-needed).** `defaultSession()` is permanent architecture wearing a "transitional shim" costume — 40 occurrences / 24 production files, and three audit corpora have now re-litigated it. The actually-deletable debt is the **module-level singleton exports** (`SharedExecutionRegistry`, `StreamStatusService`, `getDefaultStreamLogStore`, ...), not the function: hosts import the singletons directly at 16 sites, which is what made the A4/#7640 class of bug possible (a guard reading the global while runs write the session machine).

**Decision + mechanical inversion:** declare the posture (defaultSession() is the sanctioned single-session entry; the invariant is "no session-scoped mutable module export"), then invert the alias — construct the singletons INSIDE `defaultSession()` and delete the `Shared*`/`StreamStatusService` module exports; 16 call-site edits go through `defaultSession().x`/`currentSession().x`. Atomic per singleton, no dual state; R4 classifies rename-with-alias-retirement as deletion.

Record the ruling in session-scoped-runtime-architecture.md; execution can ride the D1 sweep (#6982).

From the 2026-07 macroscopic architecture review — full measurement + rejected traps in docs/proposals/state-of-the-architecture-2026-07.md and agent-sdk-north-star.md (PR #7676).

https://claude.ai/code/session_01ACrdwMVd2zrKSYDbh28Jmn

## Decision (2026-07-09, maintainer-approved)

**Declare + invert.** `defaultSession()` is declared **permanent** (the sanctioned single-session entry); the invariant of record is **"no session-scoped mutable module export."** Execute the inversion: construct the singletons inside `defaultSession()`, delete the `Shared*`/`StreamStatusService` module exports, retarget the 16 call sites through `defaultSession().x`/`currentSession().x`. This mechanically eliminates the #7640 bug class (no direct singleton import can bypass the owning session again). **Sequence: ride D1 Sweep 2 (#6982)** so the files are touched once; atomic per singleton, no dual state.

## Implementation sketch (cleanest path, per the recorded decision — atomic per singleton, rides #6982)
Per singleton (repeat ×3: `SharedExecutionRegistry`, `StreamStatusService`, `SharedExecutionSubscriptionBinder`; then `getDefaultStreamLogStore`/queue accessors the same way):
[...]

Issue #6329, opened 2026-06-20
Split desktop agent execution into focused host controllers


LionSR:
## Finding

`packages/desktop/src/main/desktopAgentExecution.ts` is a 1,300+ line host bridge that combines progress storage, event-bus subscriptions, permission UI routing, message handling, session snapshot restore, task resume, and execution startup. The top-level class has to know about progress webview updates, approvals, external inquiries, task state, snapshots, and `runExecution` orchestration in one place.

## Why this creates complexity

Desktop host infrastructure and agent-runtime orchestration change independently, but this module makes them a single editing surface. It amplifies changes across permissions, progress UI, resume, and execution plumbing, and it makes it hard to test one host behavior without loading the rest of the desktop execution stack.

## Refactoring approach

Split this into focused host controllers: a progress-event bridge, an approval/external-inquiry bridge, a task/session restore service, and a small execution coordinator that composes them. Keep `createDesktopAgentExecution` as the composition root so existing desktop callers do not need to know the internal split.

LionSR:
@claude **Scouting pass — investigate and propose, do not attempt a big-bang refactor.** This is `risk:high` (1,337-line stateful host bridge with intertwined progress/approval/resume/snapshot state). Please:

1. Map `packages/desktop/src/main/desktopAgentExecution.ts` into its distinct concerns (progress-event bridge, approval routing, snapshot/resume hydration, execution startup) and post a short comment with a proposed slicing plan (ordered, each slice independently shippable and low-risk).
2. Then implement **only the first, safest slice** — extracting the progress-event bridge (the `ProgressBackend`/`buildStreamInfo`/restored-stream hydration wiring, ~lines 178-512) into a new module under `packages/desktop/src/main/` (e.g. `desktopProgressEventBridge.ts`), keeping `createDesktopAgentExecution` as the composition root so callers are unchanged.

Do NOT touch approval or resume/snapshot paths in this PR. Add unit tests for the extracted bridge. Run `npm run typecheck` and `npm test`.

---
_Generated by [Claude Code](https://claude.ai/code)_

LionSR:
[...]

PR #8207, opened 2026-07-11
fix(desktop): rebind still-active executions across window recreation

### LINKED ISSUE #8148: Desktop window recreation hides active runs and drops terminal projection
## Summary

Closing a desktop window deliberately allows its executions to continue, but it also disposes every window-owned progress consumer. When a window is recreated, active execution ids are removed from restored ghost state without being rebound to the new window. The run can therefore continue without a rail entry, live progress, cancellation controls, or a terminal snapshot update.

## Current behavior

1. `DesktopProgressBridge.dispose()` tears down the host interactions, subscriptions, backend, and session maps, then calls `session.dispose({ keepActiveExecutions: true })`.
2. The disposed session-progress bridge rejects later presentation events so a closed renderer is not repopulated.
3. A newly created bridge hydrates persisted ghosts, then removes ghosts whose execution ids are still active in-process.
4. No process-level owner reattaches those active executions to the new bridge.

The individual guards are reasonable, but together they make an active run invisible after window recreation.

## Layering violation

Execution lifetime is process/session-scoped, while the renderer bridge is window-scoped. The policy to keep an execution alive currently lives at the window teardown boundary, but the projection and persistence needed to observe that execution remain owned by the disposed window. This couples a longer-lived runtime object to a shorter-lived UI adapter.

## Refactoring direction

Introduce one process-level desktop execution/projection owner, or an explicit rebind operation that attaches every active execution in the retained session to a newly created window bridge. The runtime and terminal snapshot owner must outlive individual renderer instances; a renderer should only subscribe and unsubscribe.

## Acceptance criteria

- Closing and reopening the desktop window preserves a visible rail entry for every still-active execution.
- The recreated window receives subsequent live and terminal events exactly once.
- Terminal snapshot persistence continues while no renderer exists.
- The user can cancel or inspect a run after reopening the window.
- Restored ghosts and rebound active runs cannot create duplicate entries.
- A lifecycle test covers close, event emission while closed, reopen, and terminal completion.

## Evidence

- Window teardown retains executions after disposing UI consumers: https://github.com/LionSR/TeXRA/blob/5726afa3a36498cdaa92ae779d5088dc43af5048/packages/desktop/src/main/desktopAgentExecution.ts#L685-L701
- Restored active executions are removed from ghost ownership: https://github.com/LionSR/TeXRA/blob/5726afa3a36498cdaa92ae779d5088dc43af5048/packages/desktop/src/main/desktopSessionProgressBridge.ts#L202-L217
- Presentation events become no-ops after bridge disposal: https://github.com/LionSR/TeXRA/blob/5726afa3a36498cdaa92ae779d5088dc43af5048/packages/desktop/src/main/desktopSessionProgressBridge.ts#L221-L232

LionSR:
## Summary

Closes #8148.

Closing a desktop window deliberately keeps its executions running
headless (`SessionHandle.dispose({ keepActiveExecutions: true })`,
#6329), but it also disposes every window-owned progress consumer for
that window (`this.unsubscribe()`, backend listeners, `sessionProgress`).
The `keepActiveExecutions` retained session stays in `liveSessions` (so
`getAllActiveExecutionIds()` still sees the run), but nothing in the
newly created window's session ever subscribes to it.

Concretely: a newly created `DesktopProgressBridge`'s startup repair
(`repairOrphanedStreamsAfterRestart` → `refreshActiveExecutionIds` →
`forgetActiveRestoredStreams`) saw the restored ghost's `executionId`
was still active and simply *forgot the ghost bookkeeping* — it never
attached anything new. The run kept executing with no rail owner: no
live progress, no cancellation/inspection route (`stopAgentStream`
looks up the handle via the *new* window's own `session.executions`,
which never had it), and a stale on-disk snapshot (nothing called
`streamSnapshotStore.upsert` again).

## Fix

- `src/agent/runtime/SessionHandle.ts`: add
  `findActiveAgentExecutionHandle(executionId)`, a cross-session lookup
  sibling to the existing `getAllActiveExecutionIds()` — both are the
  only sanctioned way to reach the private `liveSessions` set (the
  "no session-scoped mutable module export" invariant, #7694).
- `packages/desktop/src/main/desktopAgentExecution.ts`: add
  `rebindActiveExecutions`, called at the start of
  `refreshActiveExecutionIds` (before the ghost is forgotten). For each
  restored stream whose `executionId` is active but not yet tracked in
  *this* window's `session.executions`, it looks up the handle
  cross-session, `track()`s it in the new session's registry (so
  `stopAgentStream`/`getAgentHandleByStream` can find it — cancel and
  inspect), and reattaches its live trace via the existing
  `session.attachRunTrace(handle.trace, streamId)` — the exact hookup a
  normal launch already uses (`AgentLaunchContext.ts`), just pointed at
  the new window's session instead of the old one. `handle.result`
  settling untracks it again, so a finished run isn't stranded as
  "still active" in the new window's registry.

No new dependency, no new layer — the desktop bridge already had a
`forgetActiveRestoredStreams` path for exactly this "restored ghost
whose execution is active" case; this fix inserts the missing rebind
step ahead of it and reuses `SessionHandle.attachRunTrace`, an existing
public method.

## Net elements (R6)

- **Net production LOC**: +58 (`desktopAgentExecution.ts` +38,
  `SessionHandle.ts` +20). No files added or deleted.
- **New exports**: 1 — `findActiveAgentExecutionHandle`. Justified by
  module encapsulation, not caller count: `liveSessions` is a private
  module-scoped `Set` in `SessionHandle.ts`; the only way code outside
  that module can query it is through an exported accessor, exactly
  like the existing `getAllActiveExecutionIds` /
  `killAllSessionBackgroundProcesses` siblings in the same file.
- **New private methods**: 1 — `DesktopProgressBridge.rebindActiveExecutions`,
  a decomposition of the existing `refreshActiveExecutionIds` (not a
  new abstraction layer).
- **No new session/registry sharing**: verified against the Stage-5
  multi-window isolation gate (#6968) — the fix forwards a *specific*
  handle's trace into the new session's existing `events` hub (the same
  mechanism a normal launch uses); it never shares an `events`,
  `executions`, or `status` instance across sessions. The full
  `multi-window session isolation` suite in
  `DesktopAgentExecution.vitest.mts` still passes unchanged.

## Consumer counts (R8)

- `findActiveAgentExecutionHandle`: 1 production caller
  (`desktopAgentExecution.ts`'s `rebindActiveExecutions`), mirroring the
  2-caller `getAllActiveExecutionIds` it sits next to — both exist
  solely to cross the `liveSessions` encapsulation boundary, not as a
  general-purpose "shared" utility.
- `rebindActiveExecutions`: 1 caller (`refreshActiveExecutionIds`),
  extracted for readability from an already-multi-step repair method,
  not introduced as a standalone shared helper.

## Tests

New regression test in `src/test-kernel/desktop/DesktopAgentExecution.vitest.mts`
(`window recreation rebind (#8148) > reattaches a still-active
execution to the newly created window instead of stranding it`):
constructs a real `AgentExecutionHandle` + trace, tracks it in window
A's session, disposes window A (`keepActiveExecutions` path), emits a
trace event while no window exists (must not throw), then constructs
window B with a matching on-disk ghost snapshot and asserts:

- the handle is now tracked in window B's `session.executions` and
  reachable via `getAgentHandleByStream` (cancel/inspect target),
- the ghost bookkeeping (`sessionProgress.hasRestoredStream`) is
  cleared — no duplicate ghost + live entry,
- a subsequent live `status` event reaches window B and refreshes the
  on-disk snapshot store (`streamSnapshotStore.upsert`),
- window A's messages are unaffected (no cross-window leak — same
  invariant the existing multi-window isolation suite locks),
- terminal completion (`handle.settleResult` + trace `result`) untracks
  the handle from window B's registry.

Proven-to-fail: reverted the two production files to `origin/main` and
reran this test — it fails at the very first rebind assertion
(`expected undefined to be AgentExecutionHandle{…}`), confirming the
regression and that the test actually exercises the fix.

Ran:
- `npx vitest run src/test-kernel/desktop/DesktopAgentExecution.vitest.mts src/test-kernel/desktop/DesktopSessionProgressBridge.vitest.mts src/test-kernel/desktop/DesktopProgressIpc.vitest.mts src/test-kernel/agent/runtime/SessionHandle.vitest.ts src/test-kernel/agent/runtime/ExecutionRegistry.vitest.ts` — all passing (including the full `src/test-kernel/desktop/` suite, 458/458).
- `npx tsc --noEmit`, `tsc --noEmit -p tsconfig.test-kernel.json`, `pnpm --filter @texra/desktop typecheck` — clean.
- `npx eslint` + `npx prettier --check` on the three touched files — clean.

---

> [!NOTE]
> **Medium Risk**
> Touches desktop multi-window execution ownership and cross-session registry tracking; behavior is narrow and well-tested but mistakes could affect cancel/stop paths or session lifecycle leaks.
> 
> **Overview**
> Fixes **#8148**: when a desktop window closes with `keepActiveExecutions`, runs can keep going headless while the new window only dropped the restored ghost and never subscribed to the live handle—no rail, progress, cancel, or fresh snapshots.
> 
> **Desktop startup repair** now calls **`rebindActiveExecutions`** before **`forgetActiveRestoredStreams`**. For each restored stream whose `executionId` is still process-active but not tracked on *this* window, it looks up the handle on another live session, **`track()`s** it locally, overwrites status from the **live** owning session (not stale disk), **`attachRunTrace`** for UI updates, and on completion **`untrackActiveAgentExecution`** so the retained pre-close session does not leak in `liveSessions`.
> 
> **`SessionHandle`** adds cross-session **`findActiveAgentExecutionHandle`** and **`untrackActiveAgentExecution`**, mirroring **`getAllActiveExecutionIds`** as the sanctioned way past private `liveSessions`.
> 
> Regression tests cover rebind after window A dispose → window B open, live progress/snapshot refresh, dual-session untrack on settle, and stale **RUNNING** ghost vs live **WAITING**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ea8fddbca7a63f1b81a4dfd20593b876497cbe2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/desktopAgentExecution.ts
+++ b/packages/desktop/src/main/desktopAgentExecution.ts
@@ -48,6 +48,7 @@ import type {
 import { resumeToolUseSnapshot } from '@agent/runtime/resumeToolUseSnapshot';
 import { selectAutoOpenFinalOutput } from '@agent/runtime/selectAutoOpenFinalOutput';
 import {
+  findActiveAgentExecutionHandle,
   getAllActiveExecutionIds,
   SessionHandle,
 } from '@agent/runtime/SessionHandle';
@@ -744,11 +745,48 @@ export class DesktopProgressBridge {
   } {
     const activeExecutionIds = new Set(getAllActiveExecutionIds());
     const allExecutionIds = this.getRestartRepairExecutionIdMap();
+    // Rebind BEFORE forgetting the ghost: a restored stream's executionId
+    // being "active" only means some live session still owns it -- possibly
+    // a previous window's session, retained post-dispose so headless runs
+    // stay visible to process-wide guards (`keepActiveExecutions`, #6329).
+    // Without this, forgetting the ghost here (below) leaves the run with no
+    // rail entry owner at all: the old window's bridge already stopped
+    // forwarding events (disposed), and nothing else ever subscribed this
+    // window to them. See #8148.
+    this.rebindActiveExecutions(activeExecutionIds, allExecutionIds);
     this.sessionProgress.forgetActiveRestoredStreams(
       activeExecutionIds,
       allExecutionIds,
     );
     return { activeExecutionIds, allExecutionIds };
+  }
+
+  /**
+   * Reattach every still-active execution a restored (ghost) stream points
+   * at to THIS window's session, so window recreation doesn't strand a
+   * headless-continuing run on its previous window's disposed bridge (#8148).
+   * Idempotent (guarded by `this.session.executions.getHandle`), since
+   * startup repair calls `refreshActiveExecutionIds` more than once.
+   */
+  private rebindActiveExecutions(
+    activeExecutionIds: ReadonlySet<string>,
+    allExecutionIds: ReadonlyMap<StreamTabId, ExecutionId>,
+  ): void {
+    for (const [streamId, snapshot] of this.sessionProgress.restoredStreams) {
+      const executionId = snapshot.executionId ?? allExecutionIds.get(streamId);
+      if (!executionId || !activeExecutionIds.has(executionId)) continue;
+      if (this.session.executions.getHandle(executionId)) continue;
+      const handle = findActiveAgentExecutionHandle(executionId);
+      if (!handle || handle.childStreamId !== streamId) continue;
+      this.session.executions.track(handle);
+      const detachTrace = handle.trace
+        ? this.session.attachRunTrace(handle.trace, streamId)
+        : undefined;
+      void handle.result.finally(() => {
+        detachTrace?.();
+        this.session.executions.untrack(executionId);
+      });
+    }
   }
 
   /**
--- a/packages/desktop/src/main/desktopAgentExecution.ts
+++ b/packages/desktop/src/main/desktopAgentExecution.ts
@@ -51,6 +51,7 @@ import {
   findActiveAgentExecutionHandle,
   getAllActiveExecutionIds,
   SessionHandle,
+  untrackActiveAgentExecution,
 } from '@agent/runtime/SessionHandle';
 import { setProgressViewBridge } from '@agent/runtime/ProgressViewBridge';
 import {
@@ -776,15 +777,31 @@ export class DesktopProgressBridge {
       const executionId = snapshot.executionId ?? allExecutionIds.get(streamId);
       if (!executionId || !activeExecutionIds.has(executionId)) continue;
       if (this.session.executions.getHandle(executionId)) continue;
-      const handle = findActiveAgentExecutionHandle(executionId);
-      if (!handle || handle.childStreamId !== streamId) continue;
+      const found = findActiveAgentExecutionHandle(executionId);
+      if (!found || found.handle.childStreamId !== streamId) continue;
+      const { handle, status } = found;
       this.session.executions.track(handle);
+      // `hydrateRestoredStreams()` already seeded this window's status from
+      // the on-disk snapshot's `lastKnownStatus`, which goes stale the moment
+      // the run transitions while headless (nothing persists it, #8148).
+      // Overwrite with the live status the owning session actually has, so
+      // display and `terminateWaitingHandle`'s WAITING/RESUMING check both
+      // see the truth instead of a possibly-wrong RUNNING carried from disk.
+      if (status) {
+        this.session.status.transition(streamId, status, 'restart-repair', {
+          trace: handle.trace,
+        });
+      }
       const detachTrace = handle.trace
         ? this.session.attachRunTrace(handle.trace, streamId)
         : undefined;
       void handle.result.finally(() => {
         detachTrace?.();
-        this.session.executions.untrack(executionId);
+        // Release from every live session that still owns this handle, not
+        // just this window's -- otherwise the session that originally
+        // launched it (retained post-dispose via `keepActiveExecutions`)
+        // never sees its own registry go idle and leaks in `liveSessions`.
+        untrackActiveAgentExecution(executionId);
       });
     }
   }
--- a/src/agent/runtime/SessionHandle.ts
+++ b/src/agent/runtime/SessionHandle.ts
@@ -41,7 +41,7 @@ import { createChannelTrace } from '@logger';
 import type { StreamTabId } from '@shared/schemas';
 
 import { getRunContextSession, tryUseRunContext } from './RunContext';
-import { ExecutionRegistry } from './executionRegistry';
+import { AgentExecutionHandle, ExecutionRegistry } from './executionRegistry';
 import { ExecutionSubscriptionBinder } from './ExecutionSubscriptionBinder';
 import { StreamStatusMachine } from './StreamStatusService';
 import {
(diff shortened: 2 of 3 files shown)


LionSR:
## Review triage — disposition

Verified every review comment at branch HEAD (single commit `848b9f9e2` reviewed by all bots; PR had no prior fix-up commits).

**Fixed (pushed as `6ea8fddbc`):**

- **codex P1 — "Untrack the original owner when canceling rebound runs" (CONFIRMED).** `rebindActiveExecutions` tracked the handle into the *new* window's `ExecutionRegistry` but only ever untracked it from that same registry on `handle.result` settle (`this.session.executions.untrack(...)`). The original `keepActiveExecutions`-retained session's registry never learns the run finished — its `disposeWhenIdle()` loop polls `this.executions.getActiveIds()` on its *own* map, which stays non-empty forever, so that `SessionHandle` never leaves `liveSessions`. Traced the full path (`ExecutionRegistry.untrackHandle` → `notifyWaiters` is per-registry-instance; `terminateWaitingHandle` on window B's stop path never touches window A's map) and confirmed this is a genuine indefinite leak, not just true for the cancel case codex called out but for *any* termination path once rebound (a handle that finishes naturally is fine because `AgentRunLifecycle`'s own closure still untracks window A, but a `terminateWaitingHandle`-driven stop from the new window bypasses that entirely). Fixed with a new `untrackActiveAgentExecution(executionId)` in `SessionHandle.ts` — a sibling to the existing `getAllActiveExecutionIds`/`findActiveAgentExecutionHandle` cross-`liveSessions` accessors — that releases the execution from *every* live session still holding it, called from the rebind's `.finally()` instead of the single-registry `untrack`.

- **codex P2 — "Seed rebound status from the live owner" (CONFIRMED).** `hydrateRestoredStreams()` runs immediately before `rebindActiveExecutions` and seeds the new window's `session.status` from the **on-disk snapshot's** `lastKnownStatus`. Nothing persists a transition while headless (that's the whole premise of this PR), so if the run went RUNNING→WAITING with no window open, the new window's status stays wrongly RUNNING. Confirmed the user-facing consequence: `ExecutionRegistry.terminateWaitingHandle`'s guard (`status !== WAITING && !resuming`) reads the *new* registry's `streamStatus`, so a stop issued from the reopened window on a legitimately-WAITING run silently no-ops — directly undercutting this PR's own stated goal ("no cancellation route ... the fix restores"). Also confirmed the fix is possible: I traced that `session.status` transitions are applied directly by the flow's own lifecycle code (e.g. `ToolUseWaitNode`'s `streamStatus.transitionToWaiting`) against the *original launching session's* `StreamStatusMachine` object, independent of any bridge/window subscription — so the origin session's `session.status` stays live-accurate even fully headless, unlike the disk snapshot. `findActiveAgentExecutionHandle` now also returns that live status, and `rebindActiveExecutions` reseeds the new window's status with it (cause `'restart-repair'`, matching the existing seeding convention) before returning.

- **claude[bot] nit — filter-then-`toEqual` readability (applied, not blocking).** Simplified to `messagesA.some(...)).toBe(false)` as suggested.

Added two new regression tests in `DesktopAgentExecution.vitest.mts` (verified they fail against the pre-fix code by reasoning through the pre-fix control flow, and pass against the fix):
- P1: after terminal settle, asserts the handle is also untracked from **window A's** registry, not just window B's.
- P2: transitions window A's `session.status` to WAITING directly (mirroring the real lifecycle path) while a stale RUNNING snapshot sits on disk, then asserts the rebound window B shows WAITING, not the stale RUNNING.

**Dismissed:** none — both codex findings and the claude nit were real/actionable.

**Not applicable:** Copilot review declined (quota limit), and the `github-actions`/TeXRA-review bot pass reported "no issues found" (two of its own "observations, no action required" — bare `track()` not seeding status, and trace-subscription accumulation across multiple recreations — are subsumed by the P2 fix above and are otherwise correct as noted).

**Verification:** clean merge onto current `origin/main` (no conflicts), full `npx vitest run` (649 files / 5708 tests, 0 failures), full `npm run typecheck` (all 5 workspace projects, 0 errors), `eslint`/`prettier` clean on all touched files.

Verdict: **recycling-after-fix**, pushed to this branch. Should be re-reviewable/mergeable as-is.

Issue #8161, opened 2026-07-11
Replace StreamSnapshotStore per-key promise-chain write lock with keyed async-mutex


LionSR:
## What

\`src/transcript/StreamSnapshotStore.ts\` (~line 1100, \`pendingWrites\`) serializes per-stream KV writes with the hand-rolled \`prev.then(...)\` promise-chaining pattern (~40 LOC), while the repo already depends on \`async-mutex\` and uses it keyed in \`externalInquiryStorage.ts\` and \`lakeCommands.ts\`.

## Why not a drive-by PR

The chain is interwoven with eviction guards (\`pendingWrites.has(chainKey)\`), version checks, and \`flush\` / \`cancelPendingWritesForStream\` iterating the same Map — a straight swap isn't mechanical. Do this when the file is next touched for real work, not as standalone churn.

## How

Replace the raw chaining with a keyed \`async-mutex\` (matching the existing idiom), keeping the eviction/flush semantics intact and covered by the existing StreamSnapshotStore tests.

Found in a reinvented-wheel audit (2026-07-11); adversarially checked — flagged as borderline, hence issue not PR.

PR #8214, opened 2026-07-12
refactor(transcript): consolidate StreamSnapshotStore's per-stream maps into one record

### LINKED ISSUE #8124: Consolidate StreamSnapshotStore's 17 parallel per-stream maps into one
## Context

Follow-up from #8069 (refactor: fix data-structure complexity findings from codebase audit).

That PR's underlying audit found that `src/transcript/StreamSnapshotStore.ts` splits one logical per-stream snapshot into 17 parallel maps plus 5 shadow overlay maps. The PR fixed 8 lower-risk data-structure findings from the same audit but explicitly deferred this one as "a substantial, behavior-sensitive consolidation into a single `Map`." No follow-up issue had been filed.

## Details

- Consolidate the 17 parallel per-stream maps (plus 5 shadow overlay maps) in `StreamSnapshotStore.ts` into a single `Map<string, StreamSnapshot>` (or equivalent), so each stream's snapshot lives behind one key instead of being smeared across parallel maps kept in sync by hand.
- Behavior-sensitive: audit every mutation/read site to preserve current semantics for all fields before and after consolidation.
- No behavior change expected for the happy path — this is a data-structure-shape refactor, not a feature change.

## References

- PR: #8069

LionSR:
## Issue

Closes #8124. Follow-up from #8069 (refactor: fix data-structure complexity
findings from codebase audit), which fixed 8 lower-risk data-structure
findings from the same audit but explicitly deferred this one: "a
substantial, behavior-sensitive consolidation into a single `Map`."

`src/transcript/StreamSnapshotStore.ts` split one logical per-stream
snapshot into 17 parallel maps/sets kept in sync by hand: 9 accumulator
fields (`outputFiles`, `missingOutputs`, `compileFailures`, `usage`,
`usageUnparsed`, `workPlan`, `meta`, `runDescriptors`, `runConfigs`),
`seeded`/`seedChains` lazy-seeding bookkeeping, 5 overlay fields
(`metaOverlays`, `outputFileOverlays`, `missingOutputsOverlays`,
`compileFailuresOverlays`, `usageOverlays`), and the cached `kvCache`.

Design-sensitive prerequisite reading, both done before touching code:
- **#8014** (`mutateWithOverlay`, merged) — the shared eager-apply +
  overlay-reconcile shape every round/usage mutator now goes through, so an
  eager write racing ahead of its own seed is never clobbered by that seed's
  raw disk read.
- **#7892** (eviction bookkeeping, merged) — introduced `perStreamStores()`
  so `allKnownStreams()`/`evict()`/`evictAll()` derive from one hand-listed
  array instead of three independently hand-maintained ones, after
  `allKnownStreams()` had already drifted once and omitted `metaOverlays`.

## The consolidation

Replaces all 17 maps/sets with one `Map<StreamTabId, StreamRecord>`
(`records`), where `StreamRecord` has one field per former map/set entry.
`streamVersions` (must survive eviction to keep guarding in-flight seed
races) and `pendingWrites` (keyed by the compound `${stream}::${key}`, not a
bare stream id) stay as their own maps — the same two exclusions #7892 had
already carved out of `perStreamStores()`.

**This is the design win the issue asked me to preserve/achieve by
construction, not just move code around:** because every per-stream field
now lives on the *same* object, `evict()`/`evictAll()` no longer need a
hand-maintained field list at all — `perStreamStores()` (17 entries) and
`allKnownStreams()` are deleted outright, replaced by
`records.delete(stream)` / `records.clear()`. Evicting a stream now
evicts *every* field for that stream automatically; the drift class #7892
caught once (`allKnownStreams()` omitting `metaOverlays`) is now
structurally impossible, because there is no second list to omit a field
from.

`mutateWithOverlay`'s overlay-reconcile shape carries over field-for-field:
`overlay: Map<StreamTabId, P>` becomes a pair of `(record) => record.field`
accessor functions (get/set), so each of the 4 call sites
(`addOutputFiles`/`updateMissingOutputs`/`updateCompileFailures`/`addUsage`)
still resolves its own overlay type `P` with full type safety — no `any`,
no loss of the per-accumulator overlay typing #8014 established.
`applyStreamData`'s two-phase read (pre-hydration-await snapshot, then a
fresh re-read of the *same* record after the `await
hydrateRunStateFromMeta(...)` — the exact race #8014 fixed) is preserved:
since records are mutated in place rather than replaced, re-fetching
`this.records.get(stream)` after the await picks up any overlay a
concurrent eager mutation wrote to the record while hydration was in
flight, exactly like the old per-field re-reads did.

One narrow, non-data-affecting behavior nuance from unifying "does this
field have an entry" into "does this stream have a record": `existed` in
`clearMissingOutputs` used to track presence of the `missingOutputs` map
entry specifically; it now tracks presence of the stream's record as a
whole. This can only cause one extra (idempotent) empty-object write of
`missingOutputs.json` in an exceedingly narrow interleaving (some other
field — e.g. `setTodos` — touches the stream first while it's still
unseeded, then `clearMissingOutputs` runs before any seed completes); reads
of an absent vs. `{}` file are already treated identically by
`parsePersistedRoundIndexed`, so this doesn't change any observed
snapshot.

## Net elements (R6)

- **Deleted:** `perStreamStores()`, `allKnownStreams()`, `getOrCreate<T>()`
  (the round-indexed lazy-init helper — no longer needed since a record's
  round-keyed fields are always present, defaulted at record creation), the
  17 separate `private readonly` map/set field declarations.
- **Added:** the `StreamRecord` interface (17 fields, one per former
  map/set) and `getOrCreateRecord()` (replaces `kv()`'s old inline
  lazy-cache-get plus the implicit lazy-creation every mutator used to get
  for free from `Map.set()`).
- Net production diff is +297/-272 lines in `StreamSnapshotStore.ts` — the
  small increase is `StreamRecord`'s field-level JSDoc plus the
  `mutateWithOverlay` accessor-function signature (2 lambda params instead
  of 1 map reference, at 4 call sites) rather than added abstraction layers;
  no new class, port, or facade.
- Test-kernel: one test ("forces the current meta schema version over stale
  cached meta") pokes the store's private internals via `as unknown as` to
  prime an in-memory stale value the on-disk schema would reject outright
  (`schemaVersion` is a Zod literal) — updated to target
  `records.get(STREAM).meta` instead of the old standalone `meta` map;
  behavior asserted is unchanged.

## Consumer counts (R8)

- Grepped every call site of the 17 old map/set names across `src/` and
  `packages/`: all live inside `StreamSnapshotStore.ts` itself. The only
  external private-internals reach-in was the one test above (fixed).
- `mutateWithOverlay`: still exactly 4 call sites
  (`addOutputFiles`/`updateMissingOutputs`/`updateCompileFailures`/`addUsage`),
  unchanged from #8014.
- `applyRoundPatch`: still exactly 3 call sites (output files / missing
  outputs / compile failures) plus the 3 overlay-replay calls in
  `applyStreamData`, unchanged from #8014 — only the first parameter's shape
  changed (accessor function instead of a bare map reference).

## Tests

- `npx vitest run src/test-kernel/transcript/StreamSnapshotStore.vitest.ts`
  — 35/35 pass (includes the #8014 interleaving regression tests for
  `updateMissingOutputs`/`updateCompileFailures`/`clearMissingOutputs` on
  partially-preloaded streams, and the updated internals-poking test).
- `npx vitest run src/test-kernel/transcript/` — 102/102 pass.
- `npx vitest run src/test-kernel/progressView/ src/test-kernel/schemas/StreamDataRoundMaps.vitest.ts src/test-kernel/tools/ExecutionsToolWorkspaceFiles.vitest.ts` — 274/274 pass.
- `npm test` (full suite) — 5673 passed, 4 skipped, 0 failed (645 test files
  passed, 1 skipped).
- `npm run typecheck` — clean across the whole workspace (root, test-kernel,
  extension, CLI, trace-viewer, desktop).
- `npm run lint` — 0 errors, 46 pre-existing warnings, none in touched
  files.

---

> [!NOTE]
> **Medium Risk**
> Large internal refactor of stream snapshot persistence, seeding, overlays, and eviction with race-sensitive behavior; semantics are intended to match prior fixes (#8014/#7892) but warrant regression coverage on resume and partial preload paths.
> 
> **Overview**
> **`StreamSnapshotStore`** replaces **17 parallel per-stream maps/sets** (accumulators, seeding flags, overlays, KV cache) with a single **`Map<StreamTabId, StreamRecord>`**. Eviction is now **`records.delete` / `records.clear`** instead of **`perStreamStores()`** and **`allKnownStreams()`**, which are removed so per-stream fields cannot drift out of sync during eviction.
> 
> **`mutateWithOverlay`** now stores overlays on the stream record via **getter/setter lambdas** instead of separate overlay maps; seed chains, flushing, and **`applyStreamData`** (including the post-hydration re-read for concurrent overlays) are wired through **`StreamRecord`** fields.
> 
> **`clearMissingOutputs`** only persists when **`missingOutputs`** actually had rounds (not merely when a record exists from **`kv()`** or other lazy creation), avoiding spurious **`missingOutputs.json`** writes. One vitest primes stale meta via **`records.get(STREAM).meta`** instead of the old **`meta`** map.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b158117b17de0eb2e2bdc81cf3da8b459c10494. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/transcript/StreamSnapshotStore.vitest.ts
+++ b/src/test-kernel/transcript/StreamSnapshotStore.vitest.ts
@@ -752,15 +752,21 @@ describe('StreamSnapshotStore', () => {
     const store = new StreamSnapshotStore();
     await store.load([]);
 
+    // Consolidated per-stream state (#8124): every field lives on one record
+    // keyed by stream id, so priming a stale in-memory meta means seeding
+    // just the `meta` field of that record rather than a standalone map.
     const internals = store as unknown as {
-      meta: Map<StreamTabId, { schemaVersion: number; description?: string }>;
+      records: Map<
+        StreamTabId,
+        { meta?: { schemaVersion: number; description?: string } }
+      >;
     };
-    internals.meta.set(STREAM, { schemaVersion: 0 });
+    internals.records.set(STREAM, { meta: { schemaVersion: 0 } });
 
     store.setDescription(STREAM, 'Updated session');
     await store.flush();
 
-    expect(internals.meta.get(STREAM)).toMatchObject({
+    expect(internals.records.get(STREAM)?.meta).toMatchObject({
       schemaVersion: RUN_DESCRIPTOR_SCHEMA_VERSION,
       description: 'Updated session',
     });
--- a/src/transcript/StreamSnapshotStore.ts
+++ b/src/transcript/StreamSnapshotStore.ts
@@ -219,75 +219,107 @@ function executionIdFromMeta(
   return meta?.runDescriptor?.executionId ?? meta?.executionId;
 }
 
-export class StreamSnapshotStore {
-  // -- In-memory accumulators (one entry per stream that has emitted) --------
-  // Round-scoped accumulators hold the canonical RoundIndexed record — the
-  // same shape that is persisted and sent over IPC, so no conversion exists
-  // between memory, disk, and the wire.
-  private readonly outputFiles = new Map<
-    StreamTabId,
-    RoundIndexed<OutputFileInfo>
-  >();
-  private readonly missingOutputs = new Map<
-    StreamTabId,
-    RoundIndexed<string>
-  >();
-  private readonly compileFailures = new Map<
-    StreamTabId,
-    RoundIndexed<CompileFailure>
-  >();
-  private readonly usage = new Map<StreamTabId, Map<string, TokenUsageStats>>();
+/**
+ * Every per-stream field this store tracks, keyed by stream id in ONE map
+ * (`records`) instead of 17 independently hand-synced parallel maps/sets (9
+ * accumulator fields, `seeded`/`seedChain` seeding bookkeeping, 5 overlay
+ * fields, and the cached `KVStore` handle). Because every field for a stream
+ * lives on the same object, dropping a stream's memory is one
+ * `records.delete(stream)` — every field disappears with it BY CONSTRUCTION.
+ * The old design needed a hand-maintained `perStreamStores()` list (#7892)
+ * to keep `evict()`/`evictAll()`/`allKnownStreams()` from drifting apart —
+ * and that list had already drifted once (`allKnownStreams()` omitted
+ * `metaOverlays`) before #7892 unified it. A single record makes that whole
+ * drift class structurally impossible: there is no second list to omit a
+ * field from.
+ *
+ * `streamVersions` (must survive eviction to keep guarding in-flight seed
+ * races) and `pendingWrites` (keyed by the compound `${stream}::${key}`, not
+ * a bare stream id) stay as their own maps on the store — the same two
+ * exclusions #7892 already carved out of `perStreamStores()`.
+ */
+interface StreamRecord {
+  // -- Accumulated durable state (mirrors on-disk StreamData) --------------
+  outputFiles: RoundIndexed<OutputFileInfo>;
+  missingOutputs: RoundIndexed<string>;
+  compileFailures: RoundIndexed<CompileFailure>;
+  usage: Map<string, TokenUsageStats>;
   /**
    * Per-run usage values read from disk that failed to parse, preserved
    * verbatim so `writeUsage` can round-trip them back unchanged instead of a
    * lossy read permanently deleting them on the next save (#7464).
    */
-  private readonly usageUnparsed = new Map<StreamTabId, Map<string, unknown>>();
-  private readonly workPlan = new Map<StreamTabId, WorkPlanSnapshot>();
-  private readonly meta = new Map<StreamTabId, StreamTabMeta>();
-  /** Immutable run descriptors parsed/emitted once per execution stream. */
-  private readonly runDescriptors = new Map<StreamTabId, RunDescriptor>();
+  usageUnparsed: Map<string, unknown>;
+  workPlan: WorkPlanSnapshot;
+  meta: StreamTabMeta | undefined;
+  /** Immutable run descriptor parsed/emitted once per execution stream. */
+  runDescriptor: RunDescriptor | undefined;
   /** Current run config, hydrated from executions/{id}/config.json. */
-  private readonly runConfigs = new Map<StreamTabId, AgentConfig>();
+  runConfig: AgentConfig | undefined;
+
+  // -- Lazy seeding: a stream's existing disk data is read into memory BEFORE
+  // the first mutation so an accumulate/merge can't overwrite unloaded disk
+  // data. `seeded` = this stream's memory is current; `seedChain` serializes
+  // refresh/seed/mutate for it.
+  seeded: boolean;
+  seedChain: Promise<void> | undefined;
+
+  // -- Overlays: patches applied eagerly to memory while a seed is in flight,
+  // so `applyStreamData`'s post-seed reconciliation can replay them on top of
+  // the freshly-read disk state — an eager write racing ahead of its own seed
+  // is never clobbered by that seed's raw disk read. See `mutateWithOverlay`.
+  metaOverlay: boolean;
+  outputFileOverlay: OutputFilesPatch | undefined;
+  missingOutputsOverlay: RoundOverlay<string> | undefined;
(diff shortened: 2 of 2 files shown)


LionSR:
## Review triage — disposition

Verified every review comment at branch HEAD (commit `d90ba57eb`, the only commit on the branch) via `gh api repos/LionSR/TeXRA/pulls/8214/comments`.

### Fixed (pushed in `9b158117b`)

1. **codex[bot] P2 — `clearMissingOutputs` existed-check too broad (real bug, confirmed by reading old vs. new code).** Old code gated the disk write on `this.missingOutputs.delete(stream)` — presence of a `missingOutputs`-*specific* map entry. New code gated on `this.records.has(stream)` — presence of *any* record for the stream. Since `kv()` (used by `read()`, `readPersistedExecutionId()`, and the post-`evict()` `kv()` call inside `deleteStream()`) creates a full record purely to cache a `KVStore` handle, and every accumulator field defaults to non-undefined on record creation, this is broader than the PR body's own documented nuance (which only called out another *accumulator* touching the stream first, e.g. `setTodos`) — it also covers pure read-only paths and the post-delete `kv()` call, which could resurrect a `streamData/{id}/` directory for a deleted/never-really-touched stream via `listPersistedStreams()`. Fixed by checking `Object.keys(record.missingOutputs).length > 0` instead of record presence — verified this reproduces the exact old Map-based semantics for every call path (`updateMissingOutputs`, `clearMissingOutputs` itself, and the previously-flagged `setTodos` interleaving all resolve to the same `existed` value as before).

2. **claude[bot] nit — misleading "Re-read fresh" comment.** Confirmed `after = this.getOrCreateRecord(stream)` is the same object reference as `record` (records are mutated in place, not replaced), so "re-read" implied a different data source than what actually happens. Reworded to "re-fetch... same object reference."

3. **claude[bot] note — `writeMergedSidecars` legacyKeys write is broader than old `has()` guards.** Confirmed real (any `data.legacyKeys` entry now gets an unconditional, possibly-empty write, vs. the old per-field guard) but the reviewer's own analysis that it's harmless (`parsePersistedRoundIndexed` treats absent and `{}` identically) checks out — this is a doc gap, not a bug. Added a comment at the `sidecarsToWrite` seed documenting the broader-but-harmless behavior, since the PR body only mentioned the narrower `clearMissingOutputs` case.

### Also required: merge with `origin/main`

Landed on `main` after this branch diverged: #8161 (`refactor(transcript): replace StreamSnapshotStore write chains with keyed async-mutex`, merged just before review), which renamed `pendingWrites` → `writeMutexes` and swapped promise-chaining for `async-mutex` on the exact same lines this PR's `evict()`/`evictAll()`/`flush()` touch. Resolved by keeping this PR's record-consolidation design (single `records` map, no `perStreamStores()`/`allKnownStreams()`) while adopting the `writeMutexes`/`Mutex` rename and semantics from main. Fixed comment prose referencing the old `pendingWrites` name.

### Verification

- `npm run typecheck` — clean across root, test-kernel, CLI, desktop, trace-viewer.
- `npm run lint` — 0 errors (pre-existing warnings elsewhere, none in `StreamSnapshotStore.ts`).
- `npx vitest run` (full suite) — 649 files / 5706 tests passed, 1 skipped (unrelated). One `ENOTEMPTY` temp-dir-cleanup flake in `StreamSnapshotStore.vitest.ts` reproduced as flaky on a clean rerun — not caused by this change (pure `os.tmpdir()` race between parallel test files, no code path in this diff touches directory cleanup).
- Manually traced `existed`'s old (`Map.delete`) vs. new semantics through every `clearMissingOutputs`/`updateMissingOutputs` call path plus the three flagged read-only `kv()` callers to confirm the fix restores parity.

### Verdict: **recycling-after-fix**

Fixed all 3 substantive findings (1 real correctness bug, 2 accurate doc/comment nits) and merged forward past `main`'s concurrent write-mutex rename. No findings dismissed as wrong — codex and claude[bot] were both right here.

PR #8235, opened 2026-07-12
fix(exec): add stream-teardown backstop for aborted array-form commands

### LINKED ISSUE #8217: Add stream/descendant teardown backstop for array-form command abort
## Context

Follow-up from #8202 (fix(exec): delegate array-form abort/force-kill to execa natives).

That PR scoped the hand-rolled `installAbortListener()`/`terminateSubprocess` machinery to the shell/string form of `executeCommand` (`src/utils/system/execUtils.ts`), and delegated array-form abort/force-kill to execa's native `cancelSignal`/`forceKillAfterDelay`. Before the PR, `installAbortListener()` ran unconditionally, so an aborted array-form command also got `terminateSubprocess`'s force-kill-escalation stream-destroy fallback (`subprocess.stdout?.destroy(); subprocess.stderr?.destroy();`) if `await subprocess` didn't unblock on its own.

execa's `cancelSignal`/`forceKillAfterDelay` only signal the single tracked pid (no process group, no descendant killing unless `killDescendants` is enabled). An array-form command that spawns a helper process inheriting stdout/stderr — e.g. `executeCommand(['bash', '-c', 'sleep 60 & wait'], { signal })` — can therefore leave `await subprocess` hanging after abort, since the tracked parent dies but the descendant keeps the inherited streams open.

This was raised independently by two review passes on #8202 (`chatgpt-codex-connector[bot]` P2, and `claude[bot]`'s follow-up analysis), and the PR author's merge-disposition comment flagged it as "follow-up-grade (not blocking, not applied here)": none of the 17 current array-form call sites pass a `signal` together with a shell-spawning/descendant-producing command, so it isn't a regression today, but `executeCommand` is a general-purpose utility and a future caller could hit this.

## Details

- Add a stream-teardown backstop (or enable execa's `killDescendants` where safe) for the array-form abort path in `executeCommand`, so `await subprocess` can't hang indefinitely when an array-form command spawns descendants that inherit stdio.
- Add a regression test exercising an array-form command that backgrounds a child inheriting stdio (e.g. `['bash', '-c', 'sleep 60 & wait']`), aborted mid-flight, asserting the awaited call resolves within a bounded time.
- Confirm the fix doesn't reintroduce process-group semantics for the array form (which intentionally has none) or regress the shell/string form's existing `signalProcessGroup` teardown.

## References

- PR: #8202
- Review threads: https://github.com/LionSR/TeXRA/pull/8202#discussion_r3565263632, https://github.com/LionSR/TeXRA/pull/8202#discussion_r3565349160
- Author's merge-disposition comment: https://github.com/LionSR/TeXRA/pull/8202#issuecomment-4949360690

LionSR:
Closes #8217

## Summary

Follow-up to #8202. execa's `cancelSignal`/`forceKillAfterDelay` only signal the single tracked pid, so an aborted array-form `executeCommand` whose command spawned a descendant inheriting stdout/stderr (e.g. `['bash', '-c', 'sleep 60 & wait']`) left `await subprocess` hanging indefinitely: the tracked parent dies, but the descendant keeps the inherited pipes open so the streams never end. Reproduced empirically against execa 9.6.1 at origin/main HEAD before writing the fix (await still pending 5s after abort with `forceKillAfterDelay: 500`).

**Design choice — no second kill site.** `installAbortListener` stays the single owner of abort-listener lifecycle (install, remove-in-`finally`, pre-aborted replay); it now takes the abort action as a parameter. The array form's action schedules only a stream-destroy backstop on `subprocess.stdout`/`stderr` after `FORCE_KILL_DELAY_MS` — the same last-resort `await`-unblocker the shell form's force-kill escalation already uses — reusing the existing `forceKillTimeoutId` slot so the existing `finally` clears it. No signal is sent from our code for the array form: execa remains the only killer (SIGTERM at abort, SIGKILL at +5s), and the array form intentionally keeps no process-group semantics — the descendant itself is left alone (asserted in the test).

Note: the issue's alternative ("enable execa's `killDescendants` where safe") is not viable — no such option exists in execa 9.6.1 (checked `node_modules/execa/types/arguments/options.d.ts`).

## Regression test

Extended the existing `src/test-kernel/utils/system/SystemUtils.vitest.ts` suite (per one-suite-per-module): array-form `bash -c 'sleep 60 & echo $! > "$PID_FILE"; wait'` aborted mid-flight must resolve within a bounded time (20s test timeout; actually resolves at ~5s = force-kill delay) with `exitCode` 130 / `"Command aborted by user"`, and the descendant must still be alive afterwards (no reintroduced group kill; the test then reaps it). Verified the test fails by timeout against unfixed `execUtils.ts` and passes with the fix.

## Verified

- Reproduced the hang at origin/main HEAD with a standalone execa 9.6.1 script (`bash -c 'sleep 60 & wait'`, cancelSignal aborted at 200ms, `forceKillAfterDelay: 500` → still awaiting at 5s), and verified the stream-destroy backstop resolves it in ~700ms with `isCanceled: true`, which is what `executeCommand` already maps to exitCode 130 / aborted stderr.
- Opened `src/utils/system/execUtils.ts`, `src/tools/bash.ts` (uses the shell/string form — unaffected), and `node_modules/execa/types/arguments/options.d.ts` (no `killDescendants` option exists).
- `npm test -- SystemUtils` — 16/16 pass (including the new regression test); re-ran with the src fix reverted to origin/main: the new test fails by timeout, the other 15 pass.
- `npm test -- BashTool` — 12/12 pass (BashTool + BashToolErrorFeedback consumer suites).
- `npx prettier --check` on both touched files — clean.
- Shell/string form's `signalProcessGroup` teardown untouched; its existing "aborts shell command process groups including children" test still passes.

---

> [!NOTE]
> **Medium Risk**
> Touches shared subprocess abort/teardown in `execUtils`, which many tools rely on; behavior change is narrow (array + signal) and covered by new and existing abort tests.
> 
> **Overview**
> Fixes a hang when **array-form** `executeCommand` is aborted while a child process still holds inherited stdout/stderr (e.g. `bash -c 'sleep 60 & wait'`): execa kills only the tracked PID, so pipes stay open and `await subprocess` never completes.
> 
> On abort, the array path now schedules a **stream-destroy backstop** after `FORCE_KILL_DELAY_MS`, reusing the existing `forceKillTimeoutId` cleanup in `finally`. **No extra signals** are sent—execa still owns SIGTERM/SIGKILL—and **no process-group kill** is added for array commands (background children can outlive abort by design).
> 
> `installAbortListener` is refactored to accept the abort action as a callback; shell/string commands still use `terminateSubprocess` / `signalProcessGroup` unchanged.
> 
> Adds a non-Windows regression test in `SystemUtils.vitest.ts` that abort must resolve with exit 130 while the background descendant remains alive until the test reaps it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70e9dcd845499b0962b2aafb26a57785a01c0521. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/utils/system/SystemUtils.vitest.ts
+++ b/src/test-kernel/utils/system/SystemUtils.vitest.ts
@@ -137,6 +137,43 @@ describe('executeCommand', () => {
     assert.equal(result.stderr, 'Command aborted by user');
     await waitForProcessExit(childPid!);
   });
+
+  it('unblocks aborted array-form await when a descendant holds stdio', async () => {
+    if (process.platform === 'win32') return;
+
+    const dir = mkdtempSync(join(tmpdir(), 'texra-exec-abort-array-'));
+    tempDirs.push(dir);
+    const pidFile = join(dir, 'sleep.pid');
+    const controller = new AbortController();
+    // The backgrounded sleep inherits stdout/stderr; execa's cancelSignal
+    // only kills the tracked bash pid, so without the stream-teardown
+    // backstop this await would hang until the descendant exits.
+    const promise = executeCommand(
+      ['bash', '-c', 'sleep 60 & echo $! > "$PID_FILE"; wait'],
+      {
+        cwd: dir,
+        env: { PID_FILE: pidFile },
+        signal: controller.signal,
+        timeout: 60_000,
+      },
+    );
+
+    await waitForFile(pidFile);
+    const childPid = Number.parseInt(readFileSync(pidFile, 'utf8'), 10);
+    assert.ok(Number.isInteger(childPid) && childPid > 0);
+
+    controller.abort();
+    const result = await promise;
+
+    assert.equal(result.success, false);
+    assert.equal(result.timedOut, false);
+    assert.equal(result.exitCode, 130);
+    assert.equal(result.stderr, 'Command aborted by user');
+    // The array form intentionally has no process-group semantics: the
+    // descendant must survive the abort. Reap it so the test doesn't leak.
+    assert.doesNotThrow(() => process.kill(childPid, 0));
+    process.kill(childPid, 'SIGKILL');
+  }, 20_000);
 });
 
 // ---------------------------------------------------------------------------
--- a/src/utils/system/execUtils.ts
+++ b/src/utils/system/execUtils.ts
@@ -251,12 +251,8 @@ export async function executeCommand(
       }
     };
 
-    const installAbortListener = (): void => {
+    const installAbortListener = (onAbort: () => void): void => {
       if (!options.signal) return;
-      const onAbort = (): void => {
-        shellAborted = true;
-        terminateSubprocess('SIGTERM');
-      };
       options.signal.addEventListener('abort', onAbort, { once: true });
       removeAbortListener = () => {
         options.signal?.removeEventListener('abort', onAbort);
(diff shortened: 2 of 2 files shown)


LionSR:
## Review-thread disposition (finisher pass)

Verified both unresolved review threads against branch tip and re-audited the backstop for leaks.

**1. Stale comment above `terminateSubprocess` (execUtils.ts:232-236) — REAL, fixed in 70e9dcd84.** The comment claimed the array form "delegates straight to execa's own `cancelSignal`/`forceKillAfterDelay` natives" with no hand-rolled machinery, which this PR made inaccurate. Reworded: signalling is still all execa's, but the array form's one hand-rolled piece (the signal-free stream-destroy backstop in its abort listener) is now stated. Comment-only change; `npm run typecheck` clean, `npm test -- SystemUtils` 16/16, prettier clean.

**2. Timeout-without-signal array-form calls keep the descendant-stdio hang — REAL but pre-existing and out of scope here.** Confirmed at tip: `src/tools/wolfram/wolframScriptUtils.ts` and `src/utils/git/worktreeInfo.ts` (the third site now lives in `src/tools/github/index.ts`) pass array-form `timeout` with no `signal`, and the #8235 backstop only arms inside the abort listener. None of those commands spawn stdio-inheriting descendants today, so it's latent, not a live regression, and #8217 is scoped to the abort path. Needs a follow-up issue (searched — none exists yet): extend the stream-destroy backstop to the array-form timeout path (`timeout + FORCE_KILL_DELAY_MS`, cleared in the existing `finally`), same no-signal/no-process-group constraints, plus a `timedOut: true` bounded-resolution regression test. Not fixing in this PR to keep it scoped to the abort teardown.

**Backstop leak audit (fresh):** abort-armed timer lives in `forceKillTimeoutId` and is cleared in the existing `finally` on every exit path; the `{ once: true }` listener is removed via `removeAbortListener` in the same `finally`; the array/shell branches are mutually exclusive per call (`shellTimeout` is only set in the shell branch, so `terminateSubprocess` never runs for arrays — no shared-slot collision); `destroy()` on an already-ended stream is a no-op. Shell/string form's abort body is byte-identical to pre-PR (hoisted to a callback parameter only). No leaks found.

Issue #7444, opened 2026-07-07
Bash-approval timeout collapses into the same shape as a real rejection


LionSR:
## Context

Follow-up from #7327 (fix(cli): clean up subscribeApprovals). Flagged by `claude[bot]` during review, left as a tracked follow-up rather than fixed in that PR (not exploitable today — see below).

Bash-approval timeouts collapse into `{ accepted: false, userMessage: 'Approval request timed out.' }` (`packages/cli/src/chat/tui/state/subscribeApprovals.ts`). `HostBashApprovalResult` (`src/agent/runtime/HostInteractions.ts`) only has `accepted`/`userMessage` fields, unlike `PlanApprovalResult`/`ProposalResult`/`RetryResult`, which #7327 gave a distinct `{ action: 'timeout' }` specifically so a timeout isn't conflated with an explicit user rejection.

Downstream, `bash.ts` feeds `accepted: false` straight into `buildBashApprovalRejectedResult`, which reports `status: 'error'` / "User rejected bash command" to the agent — indistinguishable from a real reject.

## Why not fixed in #7327

Not exploitable today: `requestBashApproval` in `src/tools/approval/bashApproval.ts` never passes `options`/`timeoutMs`, so this path is currently dead code. Fixing it properly means giving `HostBashApprovalResult` a distinct timeout signal (mirroring the other three interaction types) and updating `buildBashApprovalRejectedResult`'s caller to handle it — a small but real contract change, better done as its own focused PR than bundled into #7327's cleanup.

## Details

- Add a way for `HostBashApprovalResult` to express a timeout distinctly from an explicit reject (e.g. an optional `timedOut: true` flag, or restructure similarly to the other three result types).
- Update `bash.ts`'s handling so a timeout doesn't get reported to the agent as "User rejected bash command."

PR #8237, opened 2026-07-12
fix(platform): keep JsonStore.open read-only so secrets reads survive unwritable storage

### LINKED ISSUE #8220: CliSecrets read path chmod's the secrets directory, breaking env-only fallback on unwritable storage
## Context

Follow-up from #8205 (refactor(cli): route CliSecrets through the shared JsonStore).

That PR routed all `CliSecrets` operations — including pure reads (`get`/`getStored`/`listStoredKeys`) — through `JsonStore.open()`. `JsonStore.open()` unconditionally calls `ensureDir()` (`src/platform/defaults/jsonStore.ts`), which does `mkdir(dir, { recursive: true, mode })` and then **always** `chmod(dir, dirMode)`, even when the directory already exists. `chmod` requires ownership; the previous `CliSecrets.readSecrets()` only did a plain `readFile()` and treated `ENOENT` as `{}` without touching the filesystem.

This was flagged independently by `chatgpt-codex-connector[bot]` ("Preserve env-only CLI runs with unwritable storage") and confirmed by a `claude[bot]` review pass. The PR author's merge-disposition comment (https://github.com/LionSR/TeXRA/pull/8205#issuecomment-4949368536) explicitly left it as a follow-up rather than fixing it in the PR, since a clean fix means deciding whether reads should tolerate `ensureDir` failure — a scope decision for `JsonStore`'s general contract (used by 5 other consumers), not a one-line fix local to `CliSecrets`.

## Details

A purely read-only credential check (`get()`/`getStored()`/`hasUsableApiKey()`/`apiKeyExists()`, all reachable via `resolveApiKeyUncached` → `secrets.get()` → `getStored()`) will now throw if the secrets directory can't be created, or — more likely — if `chmod` fails because the process doesn't own an already-existing directory (e.g. a container-mounted state dir owned by a different UID). This breaks the scenario `CliSecrets`'s own doc comment promises: "Environment variables remain the highest-priority source so automation can keep using ephemeral keys" — even when no secret write was ever requested.

Suggested fix direction: don't run `ensureDir`/`chmod` on the read path — e.g. give `JsonStore` a read-only open, or have `readJsonRecord` skip directory creation and only `mkdir`/`chmod` in `flush()` (which already does its own `ensureDir` call). A read should degrade to `{}` on a missing file exactly like before, without requiring write/chmod permission on the containing directory.

Add a regression test exercising a read (`getStored`/`get`) against a secrets path whose containing directory exists but isn't owned/writable by the test process (or is missing and the parent isn't writable), asserting it falls through to the env-var value instead of throwing.

## References

- PR: #8205
- Review thread: https://github.com/LionSR/TeXRA/pull/8205#discussion_r3565350552
- Original report: https://github.com/LionSR/TeXRA/pull/8205#discussion_r3565270254
- Author's triage comment: https://github.com/LionSR/TeXRA/pull/8205#issuecomment-4949368536

LionSR:
Closes #8220

## Summary

`JsonStore.open()` unconditionally ran `ensureDir` — `mkdir(dir, { recursive: true, mode })` plus, when a `mode` is set, an unconditional `chmod(dir, dirMode)`. Since #8205 routed `CliSecrets` reads through `JsonStore.open()`, every pure credential lookup (`get`/`getStored`/`listStoredKeys`, reachable from `resolveApiKeyUncached`) required write/chmod permission on the secrets directory. On unwritable storage (missing dir under a read-only parent, or an existing dir owned by another UID — e.g. a container-mounted state dir) a read threw instead of degrading to "nothing stored", breaking the env-var-only automation `CliSecrets`'s own doc comment promises.

**Design choice (per the maintainer ruling on #8220):** permission hardening belongs on the write path only. `flush()` already runs its own `ensureDir` before every `write-file-atomic` call, so the fix is to delete the open-time `ensureDir` — `open()` is now purely a read (`readJsonRecord` already maps ENOENT to `{}`), and writers keep byte-identical behavior: the directory is created and chmod'd to `0o700` on the first flush, the file lands with `0o600`. No new option, no second open variant, one owner for directory preparation.

Consumer sweep (R8) across all 14 `JsonStore.open` call sites: the only site whose semantics shift is the desktop `projectConfigStore` probe (`packages/desktop/src/main/platform/index.ts`), which previously fell back to the internal store when `.texra/` couldn't be *created*. That fallback's stated purpose — "startup never fails on config" — still holds (open now succeeds instead of throwing), and write failures at flush time were already an anticipated, handled state there (the migration block catches and retries); the comment is updated to match. All other sites (CLI state stores, init-platform config stores, update checker, desktop state/config/secrets stores) either write through the store (flush ensures the dir) or already tolerate open failure. Notably this also fixes CLI startup in a read-only cwd, which previously crashed creating `.texra/` at `initPlatform.ts:256`.

## Changes

- `src/platform/defaults/jsonStore.ts` — remove `ensureDir` from `open()`; document the read-only-open contract on `open()` and `JsonStoreOptions.mode`.
- `packages/desktop/src/main/platform/index.ts` — comment-only: the project-config fallback description now matches the new contract.
- `src/test-kernel/cli/CliSecrets.vitest.mts` — regression test (extend-first, R7): secrets dir uncreatable (parent `0o500`) → `get()` still serves the env var, `getStored()`/`get(miss)` resolve `undefined`, `listStoredKeys()` resolves `[]` instead of throwing.
- `src/test-kernel/desktop/JsonStore.vitest.mts` — contract test in the module's own suite: `open()` under an unwritable parent succeeds and does not create the directory; the first `set()` creates it with `0o700`/file `0o600`.

## Verified

- Opened: `src/platform/defaults/jsonStore.ts`, `packages/cli/src/runtime/cliSecrets.ts`, `packages/desktop/src/main/platform/index.ts`, `packages/desktop/src/main/desktopStreamSnapshot.ts`, `packages/desktop/src/main/platform/electronSecrets.ts`, `packages/cli/src/runtime/cliStateStores.ts`, `packages/cli/src/runtime/initPlatform.ts`, `packages/cli/src/runtime/updateChecker.ts`, `src/platform/defaults/jsonConfigProvider.ts`, both touched test suites.
- `npm run typecheck` — clean (all six sub-projects).
- `npm test -- CliSecrets JsonStore CliStateStores CliInitPlatform ElectronSecrets desktopStreamSnapshot UpdateChecker jsonConfigProvider ConfigProvider` — 9 files, 91 tests, all pass.
- Mutation check: temporarily re-adding the old open-time `ensureDir` makes exactly the two new tests fail with `EACCES, syscall: 'mkdir'` — they are genuine regression guards.
- No TUI rendering touched; headless path unaffected.

---

> [!NOTE]
> **Medium Risk**
> Touches shared JsonStore semantics and desktop config source selection at startup; behavior changes for read-only workspaces and secrets paths but is guarded by new tests and explicit fallback logic.
> 
> **Overview**
> **`JsonStore.open()` no longer creates or chmods parent directories** — that work stays on `flush()` only. Pure reads (including `CliSecrets` credential lookups) succeed on unwritable or missing-dir storage instead of throwing, so env-only secrets and read-only checkouts keep working (#8220).
> 
> **Desktop workspace config** picks project `.texra/config.json` vs the internal store using an explicit **`canCreateOrWrite`** probe and stricter rules: empty project files on read-only trees fall back to internal config so settings aren’t shadowed and later writes don’t fail at flush. Legacy migration still merges internal → project, but the **migrated flag is latched only when the project store is actually the session source** after a successful copy.
> 
> Regression tests cover **CLI secrets on unwritable storage** and **JsonStore open without mkdir until first `set()`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26e8f995d2a8f8680ac44dd3faa35d0392601824. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
## Merge-conflict resolution + bot-finding triage

**Conflict with #8238 (JsonStore rewrite)** — resolved by merging `origin/main` (0a540ab96) and re-applying this PR's intent onto the new JsonStore shape: `open()` keeps main's `resolve(filePath)` but drops the open-time `ensureDir` (flush's read-modify-write + `writeChains` design untouched — `flush()` remains the single owner of directory creation/hardening). The doc comments on `open()`/`JsonStoreOptions.mode` were re-applied onto main's text; `JsonStore.vitest.mts` is main's expanded suite plus this PR's regression test. Gate: `npm run typecheck` clean; `npm test -- CliSecrets JsonStore CliStateStores CliInitPlatform ElectronSecrets desktopStreamSnapshot UpdateChecker jsonConfigProvider ConfigProvider` → 9 files, 95 tests pass. Mutation check re-run on the merged shape: re-adding `ensureDir` to `open()` fails exactly the two regression tests with `EACCES, syscall: 'mkdir'`.

**Cursor Bugbot (medium) “Legacy config shadowed on read-only” + Codex P2 “Preserve desktop fallback”** — REAL, fixed in 372b4254d. On an unwritable tree with no `.texra/config.json`, `open()` now succeeds empty, so the legacy-fallback `catch` never fired and the user's pre-migration internal config was shadowed for the session (with a partial in-memory copy left behind by the failed migration, since `set()` mutates before flushing). Fix is consumer-local (`packages/desktop/src/main/platform/index.ts`): capture whether the project store opened empty; if the migration copy-in then fails, serve the session from the internal store — exactly the pre-#8237 behavior — and retry migration next launch. `JsonStore.open()` stays purely read-only per the #8220 ruling. A readable checked-in project file still wins even when writes fail, matching pre-#8237 semantics there too.

**Cursor Bugbot (low) “Issue tags in code comments”** — WRONG: issue references in code/test comments are established repo idiom, not a violation. Examples on main: `src/agent/runtime/SessionHandle.ts:18` (`#7694`), `:345` (`#8148`), `:349` (`#8229`); `src/agent/runtime/childRunLoop.ts:349` (`#8093`); `src/test-kernel/tools/BashToolErrorFeedback.vitest.ts:158` (“Regression coverage for #7444”); `src/test-kernel/tools/ChildRunDelivery.vitest.ts:163`. The `#8220` tags here follow the same regression-provenance convention.

LionSR:
Addressed both post-372b425 review findings in 26e8f995d by replacing the nested try/catch fallback with a single-owner decision for the session's workspace config source (`packages/desktop/src/main/platform/index.ts`).

**Cursor — "Catch reopens legacy unsafely" (index.ts, old L164-167): real, fixed.** The migration `catch` reopened `legacyWorkspaceConfigPath` unguarded; when the legacy open itself was the failure that entered the `catch` (EACCES/EISDIR — non-strict `JsonStore.open` only swallows ENOENT and `SyntaxError`, `jsonStore.ts` `readJsonRecord`), the reopen rethrew unhandled and `initializeElectronPlatform` rejected, aborting desktop startup where `main` used to warn and continue. Now the internal store is opened exactly once (`index.ts:161-169`) and is never reopened: if it can't open while a project store exists, the session runs on the project store with a warning (`internalConfigStore === undefined` selects the project branch at `index.ts:220`); if there is no project store either, the open failure stays fatal, matching the previously unguarded `?? JsonStore.open(...)` on `main`.

**Claude — fallback only fires when the `try` throws (index.ts, old L165-167): real, fixed.** With an empty legacy store the migration loop performed zero I/O, so on an unwritable tree nothing threw, the migrated flag latched `true`, and the empty project store stayed the session source — every later settings write then rejected at `flush()`'s `ensureDir`. The decision no longer depends on the migration loop throwing: an explicit writability probe (`canCreateOrWrite`, `index.ts:85-99` — deepest existing ancestor of `.texra/config.json` must be W_OK|X_OK; never creates `.texra/`) stands in for the eager-`ensureDir` probe that `JsonStore.open()` performed before the #8220 change. The single decision at `index.ts:216-230`: project store serves the session iff it opened AND (it had content, OR the internal store is unreadable, OR the migration didn't fail and the location is writable). The migrated flag now latches only when the copied-in content actually serves from the project file (`index.ts:231-249`), so a session served from the internal store retries the idempotent merge on the next launch instead of permanently skipping it.

Behavior matrix (project store opened):
- checked-in config, any tree → project store (reads always served from the checked-in file)
- file absent/empty, writable tree → project store; merge lands; flag latches
- file absent/empty, unwritable tree, legacy empty (the no-throw first-run case) → internal store; flag not latched
- file absent/empty, unwritable tree, legacy has keys (the 372b425 case) → internal store via the same decision, no reopen in a catch
- internal store unreadable, project store fine (the crash case) → project store, warn only, startup proceeds

Verified: `npm run typecheck` clean; `npm test -- JsonStore CliSecrets ElectronCompositionRoot` 26/26 passed; eslint + prettier clean on the touched file.

PR #8206, opened 2026-07-11
refactor(cli): adopt clipboardy for text clipboard writes

LionSR:
## What

`packages/cli/src/runtime/clipboardText.ts` (168 LOC) hand-rolled a per-platform clipboard-writer command table (pbcopy / wl-copy / xclip / xsel / powershell) with fallthrough and a manual kill-on-timeout spawn loop — reimplementing exactly what `clipboardy` already does.

Fixes #8169, which asked to pick one of: (a) adopt `clipboardy` and accept module-level mocking in tests, or (b) close as justified divergence. The issue author's own follow-up comment on #8169 said "adapt" (adopt `clipboardy`), so this PR takes path (a).

`clipboardy` is arguably a *safer* implementation than the hand-rolled one: it deliberately avoids `xclip` ("does not properly close stdout/stderr when receiving input via stdin, causing hangs"), bundles fallback binaries for `xsel`/Windows, and already handles Wayland-vs-X11 detection — so the custom per-candidate timeout/kill machinery is no longer load-bearing.

`writeClipboardText()` keeps its `{ ok: true } | { ok: false, reason }` result contract and the Windows CRLF normalization (clipboardy writes raw text as given; the TUI's `ExternalInquiry.tsx` copy-status UI, which only checks `result.ok`, is unaffected). Companion `clipboardImage.ts` is out of scope, per the issue.

## Dependency decision (per repo rule: new deps allowed only if issue-named + small focused lib consistent with existing deps)

The issue itself names `clipboardy` as the library to evaluate. It is a small, focused, single-purpose lib (`execa`-based, from the same `sindresorhus` line as this repo's existing `p-*`/`execa`/`lru-cache`/`delay` dependencies) — consistent with the existing dependency class. Added to `packages/cli/package.json` `dependencies` (the runtime consumer) and to root `package.json` `devDependencies`, mirroring the existing pattern for CLI-only packages needed for module-mocking resolution from the shared root `vitest` test-kernel (e.g. `semver`, which is used only by `packages/cli`/`packages/desktop` but is listed at the workspace root for the same reason).

## Net elements (R6)

- `packages/cli/src/runtime/clipboardText.ts`: 168 → 25 LOC (deleted the per-platform command table, the injectable `spawnCommand` seam, and the manual timeout/kill/stderr-buffering spawn loop; kept CRLF normalization + the result contract).
[...]

PR #8239, opened 2026-07-12
fix(cli): reap wedged clipboard helpers on write timeout

### LINKED ISSUE #8219: Cancel the wedged clipboard subprocess on write timeout, not just the await
## Context

Follow-up from #8206 (refactor(cli): adopt clipboardy for text clipboard writes).

That PR replaced the hand-rolled per-platform clipboard-spawn loop (which `child.kill()`ed a wedged helper on timeout) with `clipboardy`, wrapping `clipboard.write()` in `pTimeout(..., { milliseconds: 5_000 })`. `p-timeout` only races the promise — it has no cancellation signal into `clipboardy`, so it cannot kill the underlying platform helper process (`pbcopy`, `wl-copy`, PowerShell, etc.).

Reviewer feedback (`chatgpt-codex-connector[bot]`, P2, https://github.com/LionSR/TeXRA/pull/8206#discussion_r3565358129) on the final commit pointed this out, and the PR author explicitly acknowledged it as unfixed in the merge comment (https://github.com/LionSR/TeXRA/pull/8206#issuecomment-4949373783): "unlike the old implementation, this doesn't `kill()` the underlying platform process on timeout... Flagging this as a known, minor gap rather than blocking on it."

## Details

When a platform clipboard helper wedges, `writeClipboardText` now reports failure to the caller after 5s (so the TUI's `ExternalInquiry.tsx` Ctrl-Y copy-status UI correctly recovers), but the orphaned subprocess can keep running in the background. Two consequences:

- Repeated Ctrl-Y attempts on a wedged clipboard tool can accumulate multiple stuck background helpers.
- A late write from an orphaned helper can land on the clipboard *after* the UI has already reported `'failed'`, silently overwriting whatever the user copied afterward.

Investigate whether `clipboardy` exposes any cancellation/abort hook for its underlying `execa` calls, or whether the fix needs to track/kill the spawned PID out-of-band. If no clean cancellation path exists, consider documenting the tradeoff explicitly in `clipboardText.ts` rather than leaving it as an implicit gap.

## References

- PR: #8206
- `packages/cli/src/runtime/clipboardText.ts`
- Review comment: https://github.com/LionSR/TeXRA/pull/8206#discussion_r3565358129
- Author acknowledgment: https://github.com/LionSR/TeXRA/pull/8206#issuecomment-4949373783

LionSR:
## What

After #8206 adopted clipboardy, a clipboard write timeout only abandoned the await — `p-timeout` has no cancellation signal into clipboardy, so the wedged platform helper (`pbcopy`, `wl-copy`, `xsel`, PowerShell `Set-Clipboard`, …) lived on. Consequences (per #8219): repeated Ctrl-Y attempts accumulate stuck helpers, and a late write from an orphaned helper can land on the clipboard *after* the UI already reported `failed`, silently clobbering whatever the user copied next.

## Design

**Premise checked first**: clipboardy 5.3.1 (latest on npm as of this PR) exposes no cancellation hook — `write(text)` takes no signal, and internally calls `platformLib.copy({input: text})` with no way to thread execa's `cancelSignal`/`timeout` through. So per the maintainer ruling on this issue, the fix is a **bounded process-scoped cleanup**, not a clipboardy replacement:

- On `TimeoutError` (and only then — an ordinary rejection means the helper already exited), `writeClipboardText` reaps wedged copy helpers **before** returning `{ ok: false }`, so a wedged helper cannot land a stale write after the UI shows `failed`.
- **POSIX**: `ps -Ao pid=,ppid=,comm=`, kill (`SIGKILL`) direct children of `process.pid` whose executable basename is one of clipboardy's copy commands (`pbcopy`, `wl-copy`, `xsel` incl. the bundled fallback, `clip.exe` for WSL, `termux-clipboard-set`).
- **Windows**: one PowerShell `Get-CimInstance Win32_Process -Filter 'ParentProcessId=<pid>'` bounded to `clipboard_*.exe` or `powershell.exe` running `-EncodedCommand` (clipboardy copies via powershell-utils' encoded command; the CLI's only other PowerShell child, `clipboardImage.ts`, uses `-Command`, so it can't be hit).
- The parent-pid bound is load-bearing: a daemonized `wl-copy` from an earlier *successful* copy is reparented away and must stay alive to keep serving the Wayland clipboard — it is never touched.
- The reap itself is best-effort with its own 2s timeout; a reap failure never masks the timeout result.
- Each timeout reaps **all** matching direct children, so strays accumulated from earlier timed-out attempts are also cleaned up.

The write path stays 100% clipboardy — this does **not** reintroduce the hand-rolled per-platform command-table/fallthrough that #8206 deleted. No new dependencies (`p-timeout`'s `TimeoutError` named export, `node:child_process`).

Net: +190 / −7 across the module and its existing test suite (extended, not forked — R7).

## Verified

- Opened: `packages/cli/src/runtime/clipboardText.ts`, `src/test-kernel/cli/ClipboardText.vitest.mts`, `packages/cli/src/chat/tui/modals/ExternalInquiry.tsx` (sole consumer; `{ ok, reason }` contract unchanged), clipboardy 5.3.1 sources (`index.js`, `index.d.ts`, `lib/{macos,linux,wayland,windows,wsl,termux}.js`), `powershell-utils` (`-EncodedCommand` confirmed), `p-timeout` (TimeoutError export; inner promise gets a handler attached, so a post-kill execa rejection cannot become an unhandled rejection), `packages/cli/scripts/build-bundle.mjs` (clipboardy stays external).
- `npx vitest run src/test-kernel/cli/ClipboardText.vitest.mts` — 7/7 passed (3 new: POSIX kill-only-our-helpers, Windows parent-pid-bounded Stop-Process, reap-failure never masks the timeout result).
- `npx vitest run src/test-kernel/cli` — 130 files / 1686 tests passed.
- `npm run typecheck` — clean (root, test-kernel, texra, @texra-ai/cli, trace-viewer, desktop).
- `npx eslint` on both touched files — clean; `corepack pnpm --filter @texra-ai/cli check:architecture` — clean.
- Real-process E2E: spawned a child whose executable basename is `pbcopy` (symlink to `/bin/sleep`) plus an unrelated `sleep` bystander, ran the exact reap code — wedged child died with `SIGKILL`, bystander survived, and `ps -Ao pid=,ppid=,comm=` output format matched the parser on macOS.

Closes #8219

---

> [!NOTE]
> **Medium Risk**
> Process termination on clipboard timeout is inherently sensitive (wrong PID filtering could affect unrelated children), though the design limits kills to direct children and known helper names; TUI copy is the main consumer.
> 
> **Overview**
> Fixes clipboard writes that time out leaving **clipboardy** platform helpers (`pbcopy`, `wl-copy`, `xsel`, PowerShell, etc.) running, which could pile up on repeated copy attempts and still write stale text after the UI shows **failed**.
> 
> **`writeClipboardText`** now runs a **best-effort, parent-PID–bounded cleanup** only on **`TimeoutError`**: POSIX uses `ps` + `SIGKILL` on direct children whose comm matches known copy helpers; Windows runs a bounded PowerShell `Stop-Process` script (with safeguards so the reap process does not kill itself). A **second sweep** after ~300ms catches fallback helpers clipboardy may spawn after the first kill. Reap failures do not change the timeout failure returned to callers.
> 
> **External inquiry** copy (Ctrl-Y) ignores a second trigger while **`copying`**, so concurrent copies do not race the timeout reap and kill the wrong helper.
> 
> Vitest coverage for timeout behavior, selective killing, Windows reap, double sweep, and reap errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2380ea5aafaf8e494a2ac153661418191f58b2dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/tui/modals/ExternalInquiry.tsx
+++ b/packages/cli/src/chat/tui/modals/ExternalInquiry.tsx
@@ -274,6 +274,10 @@ export function ExternalInquiry(
   }
 
   async function copyQuestion(): Promise<void> {
+    // One write in flight at a time: a second Ctrl-Y while a copy is pending
+    // would race the first write's timeout reap, which kills all direct-child
+    // copy helpers -- including the second attempt's.
+    if (copyStatus === 'copying') return;
     setCopyStatus('copying');
     const result = await writeClipboardText(props.payload.question);
     if (result.ok) {
--- a/packages/cli/src/runtime/clipboardText.ts
+++ b/packages/cli/src/runtime/clipboardText.ts
@@ -1,20 +1,89 @@
+import { execFile } from 'node:child_process';
 import { platform as osPlatform } from 'node:os';
+import { basename } from 'node:path';
+import { promisify } from 'node:util';
 
 import clipboard from 'clipboardy';
-import pTimeout from 'p-timeout';
+import pTimeout, { TimeoutError } from 'p-timeout';
 
 import { toErrorMessage } from '@utils/errors/errorMessage';
+
+const execFileAsync = promisify(execFile);
 
 export type ClipboardTextWriteResult =
   { readonly ok: true } | { readonly ok: false; readonly reason: string };
 
 const CLIPBOARD_WRITE_TIMEOUT_MS = 5_000;
+const HELPER_REAP_TIMEOUT_MS = 2_000;
 
 interface ClipboardTextWriteOptions {
   readonly platform?: NodeJS.Platform;
 }
 
-function normalizeClipboardText(text: string, platform = osPlatform()): string {
+// clipboardy (5.3.1) exposes no cancellation hook — `write(text)` takes no
+// signal — so a timed-out await abandons the promise while the platform helper
+// clipboardy spawned lives on (#8219). On timeout we reap wedged helpers
+// out-of-band instead: kill direct children of this process whose executable
+// is one of clipboardy's copy commands. The parent-pid bound keeps this from
+// touching processes we didn't spawn — e.g. a daemonized `wl-copy` from an
+// earlier successful copy has been reparented away and must stay alive to
+// keep serving the clipboard.
+const POSIX_COPY_HELPERS: ReadonlySet<string> = new Set([
+  'pbcopy', // macOS
+  'wl-copy', // Linux (Wayland)
+  'xsel', // Linux (X11), including clipboardy's bundled fallback binary
+  'clip.exe', // WSL
+  'termux-clipboard-set', // Android (Termux)
+]);
+
+async function reapPosixCopyHelpers(): Promise<void> {
+  const { stdout } = await execFileAsync('ps', ['-Ao', 'pid=,ppid=,comm='], {
+    timeout: HELPER_REAP_TIMEOUT_MS,
+  });
+  for (const line of stdout.split('\n')) {
+    const match = /^\s*(\d+)\s+(\d+)\s+(.+)$/.exec(line);
+    if (!match) continue;
+    if (Number(match[2]) !== process.pid) continue;
+    if (!POSIX_COPY_HELPERS.has(basename(match[3].trim()))) continue;
+    try {
+      process.kill(Number(match[1]), 'SIGKILL');
+    } catch {
+      // The helper exited between listing and killing.
+    }
+  }
+}
+
+async function reapWindowsCopyHelpers(): Promise<void> {
+  // clipboardy copies via `powershell.exe -EncodedCommand …` (powershell-utils)
+  // or its bundled `clipboard_<arch>.exe` fallback; both are direct children.
+  const script =
+    `Get-CimInstance Win32_Process -Filter 'ParentProcessId=${process.pid}' | ` +
+    `Where-Object { $_.Name -like 'clipboard_*.exe' -or ` +
+    `($_.Name -eq 'powershell.exe' -and $_.CommandLine -match '-EncodedCommand') } | ` +
+    `ForEach-Object { Stop-Process -Id $_.ProcessId -Force }`;
+  await execFileAsync(
+    'powershell',
+    ['-NoProfile', '-NonInteractive', '-Command', script],
+    { timeout: HELPER_REAP_TIMEOUT_MS, windowsHide: true },
+  );
+}
+
+async function reapWedgedCopyHelpers(
+  platform: NodeJS.Platform,
+): Promise<void> {
+  try {
+    await (platform === 'win32'
+      ? reapWindowsCopyHelpers()
+      : reapPosixCopyHelpers());
+  } catch {
+    // Best effort: a reap failure must not mask the timeout result.
+  }
+}
+
+function normalizeClipboardText(
+  text: string,
+  platform: NodeJS.Platform,
+): string {
   const lf = text.replaceAll(/\r\n?/g, '\n');
   return platform === 'win32' ? lf.replaceAll('\n', '\r\n') : lf;
 }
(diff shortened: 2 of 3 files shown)


LionSR:
Finisher pass on the review findings — all verified against the branch tip and clipboardy 5.3.1 sources before acting. Fixed in 2380ea5aa:

1. **Cursor Bugbot — "Windows reap matches itself" (high): REAL, fixed.** The reap script is itself a direct-child `powershell.exe` whose command line contains the literal `-EncodedCommand` inside its own `-match` pattern, so it satisfied its own filter and could `Stop-Process` itself before reaching the wedged helper. Fixed two ways: `$_.ProcessId -ne $PID` in the filter, and the pattern is now built as `('-Encoded' + 'Command')` so neither this reap's command line nor a concurrent reap's ever contains the contiguous substring it greps for (the `$PID` guard alone would not protect two overlapping reaps from killing each other).

2. **Codex P1 — "fallback spawned after the snapshot": REAL (verified in clipboardy 5.3.1 `lib/linux.js` `xselWithFallback` and `lib/windows.js` `executeWithFallback`), fixed for the harmful half.** Killing the wedged helper rejects clipboardy's first `execa` attempt, and clipboardy then spawns its bundled fallback binary (`fallbacks/linux/xsel`, `clipboard_<arch>.exe`) — after the reap's `ps` snapshot. The reap now runs **two sweeps 300 ms apart**, so a fallback that wedges the same way (the likely case: same broken display / held Windows clipboard lock) is killed by the second sweep instead of surviving until the next Ctrl-Y timeout. The residual — a *healthy* fallback completing the copy — lands within milliseconds of the failure report, i.e. it writes the *intended* text before the user can possibly copy anything else; that is a mislabeled success, not the stale-clobber #8219 is about, and preventing it would require cancellation support inside clipboardy (which 5.3.1 does not expose).

3. **texra-review / Codex P2 — Termux `TASK_COMM_LEN` truncation: REAL, fixed.** Linux `ps -o comm=` reports the kernel task name truncated to 15 chars, so `termux-clipboard-set` surfaces as `termux-clipboar` and never matched. Added the truncated form to `POSIX_COPY_HELPERS` (all other helpers are ≤ 8 chars and unaffected).

4. **texra-review Finding 2 — concurrent Ctrl-Y race: REAL (pre-existing), fixed at the only call site.** `copyQuestion` in `ExternalInquiry.tsx` now ignores Ctrl-Y while `copyStatus === 'copying'`, so a second in-flight write can no longer have its helper killed by the first write's reap.

5. **texra-review Finding 3 — "no file named `clipboardImage.ts` exists": WRONG.** `packages/cli/src/runtime/clipboardImage.ts` exists on this branch (git blob `f8805839`, sibling of `clipboardText.ts`); the reviewer's grep simply missed it. No action.

6. **texra-review Finding 4 — WSL `clip.exe` invisible to Linux `ps`: acknowledged limitation, no change.** Windows interop processes don't appear as reapable Linux children under WSL2; the entry is harmless (safe no-op) and correct in any environment where the comm does surface.

Also re-audited the reaper per the extra-scrutiny ask: it fires only under `error instanceof TimeoutError` (ordinary rejections skip it — covered by an existing test); non-POSIX/exotic `ps` variants degrade to the pre-PR no-reap behavior through the best-effort catch; PID-reuse exposure is the standard list-then-kill window of a few milliseconds on a PID that was a direct child at snapshot time (further narrowed by zombie-holding until libuv reaps) — the accepted idiom for this kind of cleanup.

Gates on 2380ea5aa: `npm run typecheck` clean (all 6 projects); `npx vitest run src/test-kernel/cli` 130 files / 1687 tests passed (suite extended in place: fallback-respawn sweep, truncated-comm kill, self-match exclusion assertions); eslint + prettier clean; `check:architecture` clean.

Note for the merger: the previous head `1c5e4407a` was a format-bot push, so its CI run ended `action_required` (never executed) — that's what made `mergeStateStatus` UNSTABLE. This push is user-triggered and should run CI normally.

Issue #8246, opened 2026-07-12
Progress view: retry/approval requests steal active-stream focus; per-stream retry banners look un-dismissable under subagent fan-out


LionSR:
## Symptoms (user report, GPT-5.6 orchestrator + numerics subagents)

1. Viewing another session/task in the Progress view, the selection keeps jumping back to a failing subagent stream.
2. Clicking **Dismiss** on the "Retry request" banner appears to do nothing — a new "Tool-use call failed" banner reappears (22:28 → 22:30 → 22:31), even after "Retry cancelled by user".

## Root causes (mapped, verified against current main)

**Focus steal:** `requestRetry` unconditionally calls `activateStream(request.streamId)` — `packages/extension/src/progressView/extensionHostInteractions.ts:215` (same guard pattern wraps bash :172, plan :189, proposal :203, userQuestion :229, externalInquiry :241). `activateStream` (:70-78) emits `setActiveStream` **without** `suppressViewSwitch`, and the only backend guard (`ProgressFactApplier.handleSetActiveStream`, `src/controllers/progressView/backend/events/ProgressFactApplier.ts:510-599`) protects the current stream only when *it* has pending permissions (`hasPendingPermissions(currentStream)`). Browsing any idle stream ⇒ every fresh retry/approval from a failing subagent re-activates that subagent's stream.

**"Dismiss doesn't work":** it does work, but per stream. Pending retries are keyed by streamId (`extensionHostInteractions.ts:217-221`, `progressBackendUiConfig.ts:147`); with N failing subagents each raises its own banner, so dismissing one immediately reveals the next — and the failing stream itself may re-emit (replacement path `extensionHostInteractions.ts:84-96`). Combined with the focus steal above, the UX reads as an un-dismissable banner.

## Suggested direction (not prescriptive)

- Don't force-switch the active stream on interaction requests when the user is viewing a different stream; surface a non-stealing indicator instead (e.g. badge on the sessions list / "N pending approvals" chip), or gate `activateStream` on "user is on the requesting stream's session tree".
- Consider aggregating retry banners across streams ("3 subagents hit errors — Retry all / Dismiss all") for fan-out runs.

## Related

The underlying storm-generator (context-window overflow misclassified as retryable, resent unshrunk) is being fixed separately; this issue is only about the Progress-view interaction/focus behavior, which applies to any repeated per-stream failure.

PR #8252, opened 2026-07-12
fix(model-handlers): make the live pre-flight count own OpenAI context decisions

LionSR:
## Problem (user report)

A numerics tool-use subagent on GPT-5.6 hit `Token count of message exceeds context window: 2331803 > 1050000`, then looped failure→retry→failure indefinitely (Progress view banners at 22:28/22:30/22:31), hanging the parent orchestrator. Fable/Anthropic runs were fine (server-side `context_management` compaction).

## Root cause: two divergent owners of one decision

| owner | measurement | power |
|---|---|---|
| `shouldCompact()` | PREVIOUS turn's cumulative usage vs 75% threshold | compaction |
| pre-flight `applyTokenCountLimit` | live count of the CURRENT request | reduce max_output_tokens; throw at 100% |

A single turn that jumps from <75% straight past 100% is invisible to the first owner, and the second owner's throw escaped **outside** the recovery try/catch — so the existing drop-`previous_response_id`-and-compact recovery (`handleCreateResponseError`) was unreachable exactly when needed. The throw also carries no HTTP status, so `normalizeProviderError` fell to "no status → retry for safety" and kept offering retries that resent the identical oversized payload (`RetryState` waits on the prompt with no timeout → the 死循环).

## Fix: one measurement owns every context decision

- **Live-count compaction**: after the pre-flight count, crossing the threshold sets the requested-compaction flag and retries internally — the same mechanism the overflow recovery uses, with a one-shot guard against a compaction that fails to shrink. `shouldCompact()`'s stale-cumulative branch now decides **only** for models that cannot count pre-flight (Codex subscription profile keeps its existing behavior — its capability sets `supportsTokenCounting: false`).
- **Pre-flight overflow routes through `handleCreateResponseError`**: gets the drop-chain + compact + internal-retry recovery instead of dead-ending; the `!supportsTokenCounting` fallback throw (tagged in b5b008009) is covered too.
- **Non-retryable classification**: `formatProviderHttpError` now consults `isContextWindowError` (its doc comment always said "should not be retried" — nothing consumed it), guarded on status code so a retryable provider error that merely mentions tokens (e.g. a 429) keeps its retry affordance. An overflow that survives recovery fails fast with actionable guidance.

## Tests

- New: live-count threshold compaction; pre-flight overflow recovery (chain dropped, compacted transcript sent, no error escapes); tagged + provider-worded overflow classified non-retryable.
- Updated `OpenAIResponseCompaction.vitest.ts` to the live-count semantics (mocks previously returned a 100-token live count while relying on stale cumulative 800 to trigger).
- Full `npm run typecheck` + `npm test` green (5738 passed).

## Verified

Opened: `modelHandlerOpenAIResponse.ts` (compaction phase, count phase, recovery), `ModelHandler.ts` (`applyTokenCountLimit`/`validateTokenLimits`), `providerErrorFormat.ts`, `errorPatterns.ts`, `RetryState.ts`, `OpenAIResponseCompaction.vitest.ts`, `ModelHandlerOpenAIResponse.vitest.ts`, `OpenAIResponseErrors.vitest.ts`.

Related: progress-view focus-steal / banner UX tracked separately in #8246.

---

> [!NOTE]
> **Medium Risk**
> Changes core model-turn context management and retry classification; mistakes could skip compaction, over-compact, or misclassify retryable provider errors.
> 
> **Overview**
> Fixes infinite retry loops when a single turn jumps past the context window without crossing the stale **previous-turn** cumulative threshold used for compaction.
> 
> **OpenAI Responses handler** now treats the **live pre-flight token count** as the single owner for counting-capable models: it still trims `max_output_tokens`, but also **compacts and retries internally** when the count crosses the configured threshold (with a one-shot `compactionRetrySource` guard). `shouldCompact()` no longer uses cumulative usage for those models; cumulative thresholds remain only when pre-flight counting is unavailable. Pre-flight overflow from `applyTokenCountLimit` is caught and routed through **`handleCreateResponseError`**, so drop-chain + compact + internal retry matches API-side overflow. Overflow recovery no longer requires `previous_response_id`—unchained routes can compact alone—and uses **`invalidateChain()`** instead of only clearing the anchor.
> 
> **Provider errors**: `formatProviderHttpError` marks context-window overflows **non-retryable** (with status-code guard so token-related 429s stay retryable) and surfaces guidance when handler recovery did not succeed.
> 
> Tests cover live-count threshold compaction, pre-flight overflow recovery, bounded failed compaction, and error classification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b92a8c6d09ebe6269f9ecf156355b5d7d90a9126. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts
+++ b/src/agent/modelHandlers/openai/modelHandlerOpenAIResponse.ts
@@ -416,12 +416,14 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
   private inFlight = false;
 
   /**
-   * One-shot guard for the live-count compaction retry in
-   * {@link createResponseImpl}: if a compaction attempt fails to shrink the
-   * transcript, the recursive attempt's recount would still be over the
-   * threshold and must not recurse again. Reset per public call.
+   * One-shot guard for the internal compact-and-retry recursions (the
+   * live-count threshold in {@link createResponseImpl} and the overflow
+   * recovery in {@link handleCreateResponseError}): if a compaction attempt
+   * fails to shrink the transcript, the recursive attempt would hit the same
+   * threshold/overflow again and must not recurse a second time. Reset per
+   * public call.
    */
-  private liveCountCompactionAttempted = false;
+  private compactionRetryAttempted = false;
 
   /** DIAGNOSTIC: Pre-flight token estimate for comparison with actual usage */
   private _diagPreFlightTokens: number | null = null;
@@ -682,6 +684,12 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
 
     if (this.supportsTokenCounting) {
       // The live pre-flight count decides for counting-capable models.
+      // Deliberate tradeoff: if the count API soft-fails for a turn, that
+      // turn has no automatic compaction trigger at all (the stale cumulative
+      // figure is not consulted) — the API enforces the window, and an
+      // API-side overflow still recovers via handleCreateResponseError's
+      // compact-and-retry. Costs one extra round-trip in that rare failure
+      // mode; keeps the live count the single decision owner.
       return false;
     }
 
@@ -1461,7 +1469,7 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
       );
     }
     this.inFlight = true;
-    this.liveCountCompactionAttempted = false;
+    this.compactionRetryAttempted = false;
     try {
       return await run();
     } finally {
@@ -1735,7 +1743,7 @@ export class ModelHandlerOpenAIResponse extends ModelHandler<
     if (
       preFlightTokens !== undefined &&
       !compactedThisCall &&
-      !this.liveCountCompactionAttempted &&
+      !this.compactionRetryAttempted &&
       this.getCompactionThresholdPercent() > 0 &&
       preFlightTokens > this.getCompactionTokenThreshold() &&
       this.canCompactRoute()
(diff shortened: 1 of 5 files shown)


LionSR:
Addressed the review findings in b8fc8f7b5 (+ format commit):

- **Overflow skips compact without chain (cursor high, confirmed by claude)** — real regression, fixed. The recovery now enters for `hasPreviousResponseId() || canCompactRoute()`, so an overflow after `invalidateChain()` or on a non-chaining route recovers via compaction alone. Widening the condition removed the natural terminator (the nulled `previous_response_id`), so both internal compact-and-retry sites now share one per-call one-shot guard (`compactionRetryAttempted`). New tests: unchained overflow recovery, and failed-compaction termination (asserts no oversized request reaches the API and no infinite recursion).
- **Count failure disables compaction (cursor medium / claude note)** — deliberate tradeoff, now documented at the `shouldCompact()` gate: a count-API soft-failure leaves the turn to the API's own enforcement, and an API-side overflow still recovers through `handleCreateResponseError` (one extra round-trip in a rare failure mode, in exchange for keeping the live count the single decision owner rather than re-adding the stale-cumulative path as a second one).
- **createResponseImpl keeps growing (cursor low)** — declining the extraction: the threshold block has a single caller, and repo guardrails (CLAUDE.md § abstraction-cost) ban single-caller extractions; the policy lives next to the count it consumes.

Model-handler suites 446 passed; full typecheck green.

LionSR:
Addressed all three review findings on the rebased head:

- separated threshold-triggered compaction from overflow recovery so a failed threshold compaction cannot suppress the one bounded overflow recovery;
- invalidated the response chain bookkeeping before retrying, so failed compaction resends the complete local transcript instead of a sliced unchained delta;
- included body-inferred HTTP status when deciding whether a context-worded error is retryable.

Validation: focused Vitest 37/37, targeted ESLint, root typecheck, test-kernel typecheck, and `git diff --check` all pass. The existing failed-compaction test was refactored to cover both recursion bounds and history preservation rather than adding another broad scenario.

PR #8265, opened 2026-07-12
fix(review): add execution-owned stop action

### LINKED ISSUE #8260: Agent Review: add an execution-owned stop path
## Problem

Agent Review can remain in the running state indefinitely when its active model/tool cycle wedges. The extension then reports that an agent review is already running and blocks subsequent reviews, while the UI exposes no stop command.

`AgentReviewService.runReview()` owns the `running` state but does not capture the execution handle available from `runAgent({ onRun })`. `clear()` only disconnects results and advances the generation; it does not stop the execution. `stopAfterCycle` is not sufficient because it is observed only after the active cycle reaches the wait node.

This matches the reproduced UI symptom where Agent Review never receives a stop signal and keeps blocking the review surface.

## Proposed design

- Capture the active `ExecutionHandle` through `runAgent({ onRun })` as the service's single cancellation owner.
- Add a Stop Review command/action that is visible only while review is running.
- Route both explicit stop and clear-while-running through the same execution termination path.
- Clear the stored handle identity-safely when the owning run settles so a stale run cannot stop a newer one.

## Evidence

- `packages/extension/src/frontend/review/AgentReviewService.ts`: `runReview()` gates on `running`, launches without `onRun`, and `clear()` does not terminate execution.
- `packages/extension/package.json`: no Agent Review stop command is contributed.
- `src/agent/implementations/flows/tooluse/nodes/ToolUseWaitNode.ts`: stop-after-cycle is handled only after a cycle reaches the wait node.

LionSR:
## Summary
- give Agent Review one active run token that owns its session and execution handle
- route Stop and clear-while-running through the shared execution registry stop API
- replace Find Issues with a Stop action while review is running
- preserve partial findings on Stop while Clear still discards stale results
- latch stops that arrive after lifecycle tracking but before an interrupt capability attaches

## Design
`AgentReviewRunController` is the single source of truth for review activity, stop intent, and exact cancellation identity. `AgentExecutionHandle` owns the narrow pre-setup stop window: lifecycle launches explicitly enable it, the first attached flow/handler replays a pending stop, and the window closes once setup returns so suspended and post-detach handles keep their existing semantics.

## Tests
- stop requested before review handle capture is replayed once the handle binds
- repeated Stop is idempotent
- stale finalization cannot clear or stop the newer review run
- lifecycle stop from `onRun` replays once tool-use interruption attaches
- existing execution-registry waiting/teardown coverage remains green
- command catalog and extension package invariants

## Validation
- focused Vitest: 4 files, 50 tests passed
- root and test-kernel TypeScript checks
- full repository ESLint (0 errors; existing warnings only)
- direct extension-host esbuild bundle

`compile:fast` through pnpm is blocked only by isolated-worktree dependency symlinks; the underlying extension esbuild succeeds and CI runs from a clean install.

Closes #8260

---

> [!NOTE]
> **Medium Risk**
> Changes agent run lifecycle and cancellation semantics (pending interrupt window), but scope is narrow and covered by new controller and lifecycle tests.
> 
> **Overview**
> Adds a **Stop Agent Review** command and swaps the SCM panel’s “Find Issues” action for **Stop** while a review is running (`texra.agentReview.running`).
> 
> **`AgentReviewRunController`** owns the single active review run: it claims a run slot, binds the live execution handle from `runAgent`’s `onRun`, and routes stop/clear through `session.executions.stopAgentStream`. **Stop** keeps partial findings and clears pending commit re-runs; **Clear** still discards results but now also stops an in-flight reviewer.
> 
> The agent runtime gains a **pending interrupt** window on `AgentExecutionHandle` so a stop requested between lifecycle tracking and tool-flow attach is latched and replayed once interruption is wired; lifecycle skips the RUNNING transition when a pre-setup stop already landed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67290c3455dbfad56fbb28802d322f5c8ba403f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/package.json
+++ b/packages/extension/package.json
@@ -1234,6 +1234,13 @@
         "icon": "$(ellipsis)"
       },
       {
+        "command": "texra.agentReview.stop",
+        "title": "Stop Agent Review",
+        "shortTitle": "Stop",
+        "category": "TeXRA",
+        "icon": "$(debug-stop)"
+      },
+      {
         "command": "texra.agentReview.fixAllIssues",
         "title": "Fix All Agent Review Issues",
         "shortTitle": "Fix All Issues",
@@ -1502,12 +1509,17 @@
         },
         {
           "command": "texra.agentReview.run",
-          "when": "texra.activated && view == texra.scmAgentReviewView",
+          "when": "texra.activated && view == texra.scmAgentReviewView && !texra.agentReview.running",
+          "group": "navigation@1"
+        },
+        {
+          "command": "texra.agentReview.stop",
+          "when": "texra.activated && view == texra.scmAgentReviewView && texra.agentReview.running",
           "group": "navigation@1"
         },
         {
           "command": "texra.agentReview.runWithOptions",
-          "when": "texra.activated && view == texra.scmAgentReviewView",
+          "when": "texra.activated && view == texra.scmAgentReviewView && !texra.agentReview.running",
           "group": "navigation@2"
         },
         {
@@ -1560,11 +1572,15 @@
       "commandPalette": [
         {
           "command": "texra.agentReview.run",
-          "when": "texra.activated"
+          "when": "texra.activated && !texra.agentReview.running"
         },
         {
           "command": "texra.agentReview.runWithOptions",
-          "when": "texra.activated"
+          "when": "texra.activated && !texra.agentReview.running"
+        },
+        {
+          "command": "texra.agentReview.stop",
+          "when": "texra.activated && texra.agentReview.running"
         },
         {
           "command": "texra.agentReview.fixAllIssues",
--- a/packages/extension/src/commands/review/agentReviewCommands.ts
+++ b/packages/extension/src/commands/review/agentReviewCommands.ts
@@ -39,6 +39,7 @@ const CHANNEL = 'AgentReview';
 const agentReviewCommands = {
   run: 'texra.agentReview.run',
   runWithOptions: 'texra.agentReview.runWithOptions',
+  stop: 'texra.agentReview.stop',
   fixAll: 'texra.agentReview.fixAllIssues',
   fixIssue: 'texra.agentReview.fixIssue',
   dismissIssue: 'texra.agentReview.dismissIssue',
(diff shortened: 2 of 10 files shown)


LionSR:
Addressed the review findings in `a7fea2c92`:

- early Stop now latches on lifecycle-owned `AgentExecutionHandle` until the first real interrupt capability attaches; a focused lifecycle test reproduces the `onRun` timing
- Stop during diff collection exits before changing previous findings
- repeated Stop is idempotent
- the pending-interrupt window is explicitly lifecycle-only and closes after runner setup, preserving WAITING/post-detach semantics

LionSR:
Addressed the latest setup feedback in `b125e3c97`: pending interruption now replays synchronously when the real tool-use flow or interrupt handler attaches, so `runToolUseFlow`'s immediate setup guard observes cancellation before persisted resume reads. Public controller methods now have concise ownership JSDoc. The same 4 focused files / 50 tests, both TypeScript checks, targeted lint, and direct extension bundle pass.

LionSR:
Addressed the ordering findings in `2e6d93b9f`:

- review generation is now claimed before the initial async context-key update, so Clear cannot be overwritten by a later `executeReview` start
- the pre-attach interrupt window closes in `finally` around the runner, including throw-before-attach paths
- the interrupt-window JSDoc now matches its actual runner-exit boundary

Focused 4 files / 50 tests, targeted lint, both TypeScript checks, and the extension bundle pass.

LionSR:
Addressed the status resurrection finding in `3fe21dfc9`: lifecycle startup now preserves the existing CANCELLED status when the handle already owns a pending pre-attach interrupt. The focused onRun race test asserts the stream is still CANCELLED before attaching the tool-use flow, then verifies synchronous interrupt replay. All 50 focused tests and validation gates pass locally.

LionSR:
Addressed latest-main review feedback in `67290c345`:

- Agent Review Stop now uses the shared `detachSubagentsOnStop()` workspace policy
- Clear JSDoc explicitly states the run slot remains held until settlement
- the lifecycle mock now returns a real resolved Promise from `switchModel`

The `stopAgentStream` no-target diagnostics suggestion is not applicable here: the controller first identity-checks the exact registry handle, passes that handle's runtime host, and calls the registry synchronously without an intervening await, so `no_target` and `missing_runtime_host` cannot occur on this path. The only remaining outcomes are the existing successful interruption/marking semantics.

PR #8285, opened 2026-07-12
fix(progress): stop retry/approval requests from stealing active-stream focus

### LINKED ISSUE #8246: Progress view: retry/approval requests steal active-stream focus; per-stream retry banners look un-dismissable under subagent fan-out

LionSR:
Closes #8246

## Problem

1. **Focus steal**: every retry/approval/user-question/external-inquiry request emitted `setActiveStream` without `suppressViewSwitch`, so a failing subagent yanked the Progress view back to its stream on every request cycle while the user was inspecting another stream. The only backend guard (`ProgressFactApplier.handleSetActiveStream`) protects the *current* stream only when it has pending permissions of its own.
2. **"Dismiss does nothing"**: dismiss works — it cancels that stream's retry cycle (`CANCEL_RETRY_REQUEST` → `resolve(kind:'retry', action:'cancel')` → `RetryState` → stream CANCELLED, flow stops). But under subagent fan-out, each failing stream raises its own banner and the focus steal immediately surfaced the *next* failing stream's banner, so the UX read as one un-dismissable banner.

## Fix: interaction requests surface per-stream, never switch focus

Interaction-request emitters now pass `suppressViewSwitch: true` — the existing applier vocabulary for "register the stream (state, logs, hints, content sync) but do NOT switch the active tab" (same contract background child streams already use in `childStream.ts`). The switch policy stays owned by `ProgressFactApplier.handleSetActiveStream`; no new wire surface, no new controls.

What the user sees instead of a yank: the requesting stream's row carries the existing pending-approval indicator (`pendingApprovalIds$` → `has-pending-approval` orange rail), the request panel is already stream-scoped (`filterPermissionsForStream`), and the Progress view is still revealed via `requestEnsureProgressView`. Selecting the badged row shows the panel; dismissing it cancels exactly that stream's cycle and nothing steals focus afterwards.

Applied to all four emitters of the same pattern:

- `packages/extension/src/progressView/extensionHostInteractions.ts` — `activateStream` → `revealStream` (bash, plan, proposal, retry, userQuestion, externalInquiry)
- `packages/desktop/src/main/desktopHostInteractions.ts` — same helper, same contract
- `src/tools/approval/toolEditApproval.ts` — `emitToolEditApprovalPrompt` (shared by the VS Code and desktop tool-edit surfaces; CLI does not use it)
- `src/tools/inquiry/ExternalInquiryTool.ts` — the ask-time registration emission. CLI unaffected: the TUI's inquiry presentation goes through `enqueueTuiApproval`, whose `onPresent` deliberately emits its own (un-suppressed) activation when the modal is actually presented; headless drops inquiries entirely.

Not done (deliberately): no "Retry all / Dismiss all" aggregation — the issue marked it non-prescriptive, and per-panel actions remain the single home for each action (duplicate-UI-controls rule).

## Verified

Opened: `extensionHostInteractions.ts`, `desktopHostInteractions.ts`, `toolEditApproval.ts`, `ExternalInquiryTool.ts`, `ProgressFactApplier.ts` (`handleSetActiveStream` suppress path :510-599), `progressBackendUiConfig.ts`, `RetryState.ts` (dismiss → `userCancelled` → CANCELLED), `ProgressViewMessageHandler.ts` (`handleCancelRetryRequest`), `RetryRequestPanel.ts`, `RequestPanels.ts`, `BaseStreamContent.ts` (stream-scoped panels), `StreamTabs.ts` + `StreamTab.styles.ts` + `progressState.ts` (existing per-row pending badge), `desktopSessionProgressBridge.ts` (`requestEnsureProgressView` still routes to progress), CLI `subscribeApprovals.ts`/`subscribeRuntimeHost.ts`/`approvalAdapter.ts` (on-present activation, headless parity).

- `npm run typecheck` clean (all 6 projects)
- Full `npm test`: 5768 passed, 4 skipped (652 files), exit 0
- Regression test: `ExtensionHostInteractions.vitest.mts` "surfaces retry requests without stealing active-stream focus (#8246)"; updated the payload-exact assertions in `DesktopHostInteractions.vitest.mts`, `DesktopToolEditApproval.vitest.mts`, `HumanPromptProgressEvents.vitest.ts` (+ the `progressTestUtils` recording fake) to the new contract
- Lint: 0 errors; remaining warnings pre-existing in unrelated files

---

> [!NOTE]
> **Medium Risk**
> Changes how all interaction requests drive Progress tab focus and category filtering; behavior is user-visible across extension, desktop, and shared approval paths, though logic reuses existing `suppressViewSwitch` handling.
> 
> **Overview**
> Fixes **#8246**: retry, approval, user-question, and similar prompts no longer force the Progress view onto the requesting stream while you are inspecting another tab.
> 
> **Host emitters** (extension `revealStream`, desktop equivalent, `emitToolEditApprovalPrompt`, external inquiry) now emit `setActiveStream` with **`suppressViewSwitch: true`** and **`ensureVisible: true`** instead of switching the active tab. Progress is still opened via `requestEnsureProgressView`; the stream row can show the pending badge and stream-scoped panels.
> 
> **Backend** (`ProgressFactApplier`): adds schema flag **`ensureVisible`**. When both flags are set, a restrictive **agent category filter** widens to `all` so the stream tab is not hidden, and filter changes trigger a full stream metadata sync while **keeping** the current `activeStream`.
> 
> Tests and recording fakes were updated to the new payload contract, including a retry focus-steal regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c458c513e1aec9b14fd723a2c0d97f129011fadb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/desktopHostInteractions.ts
+++ b/packages/desktop/src/main/desktopHostInteractions.ts
@@ -92,7 +92,7 @@ class DesktopHostInteractions implements HostInteractions {
       streamId: streamId ?? '',
     };
     return this.showPending(requestId, { kind: 'bash', streamId }, (settle) => {
-      this.activateStream(streamId);
+      this.revealStream(streamId);
       this.options.getApprovalHandlers().bash.show(payload);
       return settle;
     });
@@ -105,7 +105,7 @@ class DesktopHostInteractions implements HostInteractions {
       request.approvalId,
       { kind: 'plan', streamId: request.streamId },
       (settle) => {
-        this.activateStream(request.streamId);
+        this.revealStream(request.streamId);
         this.options.getApprovalHandlers().planApproval.show(request);
         return settle;
       },
@@ -119,7 +119,7 @@ class DesktopHostInteractions implements HostInteractions {
       request.proposalId,
       { kind: 'proposal', streamId: request.streamId },
       (settle) => {
-        this.activateStream(request.streamId);
+        this.revealStream(request.streamId);
         this.options.getApprovalHandlers().agentProposal.show(request);
         return settle;
       },
@@ -138,7 +138,7 @@ class DesktopHostInteractions implements HostInteractions {
       request.requestId,
       { kind: 'userQuestion', streamId },
       (settle) => {
-        this.activateStream(streamId);
+        this.revealStream(streamId);
         this.options.getApprovalHandlers().userQuestion.show(request);
         return settle;
       },
@@ -235,12 +235,23 @@ class DesktopHostInteractions implements HostInteractions {
     });
   }
 
-  private activateStream(streamId: StreamTabId | undefined): void {
+  // Interaction requests surface per-stream (pending badge on the stream
+  // row); `suppressViewSwitch` registers the stream without yanking the
+  // active tab away from whatever the user is inspecting — matches the
+  // extension host contract (#8246).
+  private revealStream(streamId: StreamTabId | undefined): void {
     this.options.runtimeHost.emit('requestEnsureProgressView', {});
     if (streamId) {
       this.options.session.events.emit({
         scope: 'session',
-        event: { type: 'setActiveStream', payload: { streamId } },
+        event: {
+          type: 'setActiveStream',
+          payload: {
+            streamId,
+            suppressViewSwitch: true,
+            ensureVisible: true,
+          },
+        },
       });
     }
   }
--- a/packages/extension/src/progressView/extensionHostInteractions.ts
+++ b/packages/extension/src/progressView/extensionHostInteractions.ts
@@ -67,12 +67,23 @@ export function createExtensionHostInteractions(
 
   const handlers = () => options.getApprovalHandlers();
 
-  const activateStream = (streamId?: StreamTabId | null) => {
+  // Interaction requests surface per-stream: the request panel is scoped to
+  // the viewed stream and the requesting stream's row shows a pending badge.
+  // `suppressViewSwitch` registers the stream so that row exists without
+  // yanking the active tab away from whatever the user is inspecting (#8246).
+  const revealStream = (streamId?: StreamTabId | null) => {
     options.runtimeHost.emit('requestEnsureProgressView', {});
     if (streamId) {
       options.session.events.emit({
         scope: 'session',
-        event: { type: 'setActiveStream', payload: { streamId } },
+        event: {
+          type: 'setActiveStream',
+          payload: {
+            streamId,
+            suppressViewSwitch: true,
+            ensureVisible: true,
+          },
+        },
       });
     }
   };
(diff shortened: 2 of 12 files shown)


LionSR:
Addressed the blocking feedback in `c458c513e`:

- shared tool-edit approval now emits `requestEnsureProgressView`, restoring desktop visibility
- interaction stream registration carries `ensureVisible: true` alongside `suppressViewSwitch`
- the backend widens a restrictive category filter to `all` only for these interaction requests, preserves the current active stream, and sends one full metadata refresh so the pending row/badge is reachable
- background child registration keeps its existing suppressed/no-filter-change behavior

Validation: 6 focused files / 65 tests, targeted ESLint, root + test-kernel TypeScript, extension esbuild, and diff checks.

PR #8295, opened 2026-07-12
fix(cli): preserve context after interrupted response

LionSR:
## Summary

- preserve the established tool-use flow record when Escape interrupts an active CLI response
- prepare both normally returned and provider-rejected cancellations at the same resumable follow-up boundary
- route subsequent user messages through one controller-owned interrupted-execution admission path
- retain messages submitted during teardown, queue handoff, and explicit-resume replacement
- batch concurrent post-interruption submissions into one ordered resume
- transfer accepted batches to manual or launcher resume instead of competing with them
- retain an explicit recoverable-stream identity and requeue accepted input when resume preparation fails
- add regression coverage for persistence, provider rejection, concurrent admission, ownership transfer, retry, cancellation, and rehydration failure

## Root cause

Escape correctly stopped the active stream, but ordinary mid-run cancellation deleted the persisted flow record. The terminal transcript remained visible, while the next message was admitted as a fresh root execution. This made the displayed conversation diverge from the context sent to the model.

The repair gives `ChatSessionController` one atomic admission operation and gives the session an explicit interrupted-stream identity. The controller owns interruption detection, captures the stream before teardown clears active state, preserves message order, and gives every accepted message exactly one owner. Before the resumed queue is ready, messages belong to the interruption batch; afterward, they belong to the shared follow-up queue. If an explicit manual or launcher resume supersedes the batch, its ordered messages are transferred to that resume. If preparation fails, they are requeued with restored recovery state. At the flow layer, both returned and thrown interruption paths persist `shouldSkipCycle` and reset the cursor to the flow entry point before later continuation. The TUI only handles the controller's admission result.

## User impact

Pressing Escape now cancels the active response without silently discarding the preceding conversation. Messages entered immediately afterward are retained through teardown and every resume handoff; concurrent submissions cannot compete for separate resumes; provider cancellation cannot leave a preserved flow at a mid-cycle cursor; and explicit resume cannot drop an already accepted message or attach later input to an obsolete interrupted stream.

## Validation

- `corepack pnpm run lint` (no errors; existing repository warnings remain)
- `corepack pnpm test`: 5,901 passed, 4 skipped; the unrelated `RunChatSignalOwnership` test reached its 10-second timeout under full-suite concurrency and passed when rerun alone
- 177 focused tests on the contributor-revised branch head
- final ownership-transfer revision: full type check and 28 controller tests; 67 directly affected tests passed in the preceding combined run

---

> [!NOTE]
> **Medium Risk**
> Touches CLI session lifecycle, resume orchestration, and persisted tool-use flow teardown — incorrect ordering could still drop follow-ups or start duplicate root runs, but behavior is heavily regression-tested.
> 
> **Overview**
> **Escape during an active CLI response no longer silently starts a new conversation** — the interrupted root stream is tracked, the tool-use flow record is kept, and the next user message is routed through resume instead of a fresh root run.
> 
> On **stop**, the session records `interruptedStreamId` before tearing down the active stream. **`ChatSessionController.admitInterruptedFollowUp`** batches messages sent during teardown into one ordered resume, waits for the prior run to finish, then calls **`tryResumeStream`** with **`extraFollowUps`**. Failed resumes restore recoverability; explicit **`resume`** / **`tryResumeStream`** supersedes stale interruption state and can transfer pending messages (manual resume passes them as **`drainedFollowUps`**).
> 
> In **`runToolUseFlow`**, user interruption now **preserves the persisted flow** (new preservation reason), and **`ToolUsePersistedFlow.prepareForFollowUp`** sets **`shouldSkipCycle`** and resets the cursor so continuation is valid — including when **`pf.run`** throws after interruption (e.g. provider abort).
> 
> **`resumeQueuedToolUse`** adds **`onFollowUpQueueReady`**, invoked after the queue is marked resuming but before drain, so late follow-ups can attach at the right boundary. The TUI submit path consults admission before **`startSession`**. Changelog and regression tests cover persistence, batching, handoff, and failure/retry paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc3de53132ab72f19e12cbb5ec95753f8440f80e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,9 @@ All notable changes to this project will be documented in this file.
 
 #### Bug Fixes
 
+- **Interrupted chats retain their earlier context** — pressing Escape during
+  an active CLI response now preserves the resumable conversation, and the next
+  message continues that conversation instead of silently starting a new one.
 - **Model access choices apply to the launched session** — switching access at
   startup overrides an earlier command-line mode, and selecting ChatGPT turns
   off OpenRouter routing so requests use the chosen subscription.
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -83,6 +83,37 @@ import {
 } from './tui/state/transcript';
 import { projectStreamTranscript } from './tui/state/transcriptProjection';
 
+interface InterruptedFollowUp {
+  readonly text: string;
+  readonly mediaFiles?: readonly string[] | undefined;
+  readonly displayText?: string | undefined;
+}
+
+type InterruptedFollowUpAdmission =
+  | { readonly kind: 'not_interrupted' }
+  | {
+      readonly kind: 'accepted';
+      readonly streamId: StreamTabId;
+      readonly completion: Promise<boolean>;
+    };
+
+interface InterruptedContinuationBatch {
+  readonly streamId: StreamTabId;
+  readonly followUps: InterruptedFollowUp[];
+  completion: Promise<boolean>;
+  superseded: boolean;
+}
+
+interface SupersededInterruptedRecovery {
+  readonly streamId: StreamTabId;
+  readonly followUps: readonly InterruptedFollowUp[];
+}
+
+interface AutoResumeOptions {
+  readonly extraFollowUps?: readonly InterruptedFollowUp[];
+  readonly onFollowUpQueueReady?: () => void;
+}
+
 // ---------------------------------------------------------------------------
 // Reusable config builders & execution-registration helpers
 // (moved from runChatTui.tsx — host-neutral, no Ink dependency)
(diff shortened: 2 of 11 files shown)


LionSR:
this is very serious we need to make sure we disentangle the spaghetti

LionSR:
Real PTY validation found and fixed two gaps that the original mocks missed:

- provider cancellation can reject `PersistedFlow.run()`, so preservation now keys off a started flow plus an actual interruption rather than a fulfilled run promise;
- cancelled flows previously persisted a terminal cursor, so automatic resume completed without consuming the queued follow-up. Cancelled flows now rewind with `shouldSkipCycle`, reaching the wait boundary before the next model round.

Also removed overlapping controller/state assertions. Validation: 186 focused tests, full typecheck, lint (0 errors; existing warnings only), formatting, CLI bundle, and an actual `texra-local chat` PTY run where Escape stopped round 1 and the immediate follow-up appeared and completed as round 2 with preserved context.

PR #8461, opened 2026-07-15
fix(relay): gate openai/anthropic/google to Ultra tier

LionSR:
## Summary
- Free and Max relay users no longer get server-side-key access to openai, anthropic, or google models; they receive a 403 pointing at their own API key or an Ultra upgrade.
- The block is enforced on the URL's `:provider` segment (route level), not by model name, so it can't be bypassed by an unrecognized/future model string. `PROVIDER_CONFIGS`/`ProviderKey` and all routing code are untouched — Ultra still needs full access.
- xai, dashscope, minimax, glm, deepseek, and moonshot are unaffected at every tier.
- Public `/relay/tier-config` model lists updated to match, so client model pickers stay consistent with what's actually enforced. `FREE_TIER_SUGGESTED_MODEL` moves from `gemini35f` to `deepseek` since gemini no longer qualifies for free/Max.
- Already deployed to production (`supabase functions deploy relay --no-verify-jwt`); this PR lands the source-of-truth change in the repo.

## Test plan
- [x] `deno check models.ts` — clean
- [x] `deno check index.ts` — only a pre-existing, unrelated `bodyToSend` type error (confirmed present on unmodified `main` via `git stash`)
- [x] Local `deno eval` smoke test: free/Max tier model lists contain no gemini/gpt/claude entries; `deepseek` resolves as free-tier-allowed; Ultra still resolves `gpt-5`
- [x] Live check against deployed `/relay/tier-config`: free and Max model lists confirmed free of gemini/gpt/claude/haiku/opus/sonnet entries

---

> [!NOTE]
> **Medium Risk**
> Changes tier access and public `/relay/tier-config` behavior for a large share of relay traffic; enforcement is early and explicit but misconfiguration could surprise Max users expecting those providers on relay.
> 
> **Overview**
> **Relay no longer proxies `openai`, `anthropic`, or `google` for free and Max users** using server-side keys. Those tiers get a **403** on `/relay/:provider/*` when the URL provider segment is one of those three, with guidance to use personal API keys or upgrade to Ultra. **Ultra** is unchanged.
> 
> The restriction is enforced **on the route `provider` parameter**, not on model names, so odd or future model strings cannot bypass it. **`models.ts`** defines `ULTRA_ONLY_PROVIDERS` / `ULTRA_ONLY_PROVIDER_SET`, strips those providers from free and Max model lists in **`TIER_CONFIG`** (Max moves from `models: '*'` to an explicit list), and documents that per-model checks for Max stay permissive because provider gating runs earlier. **`FREE_TIER_SUGGESTED_MODEL`** changes from `gemini35f` to **`deepseek`** so tier-denial hints point at a model still allowed on relay.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00b62dfe2cd35f2cff68b4bc2548c8f372b60fa2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/supabase/functions/relay/index.ts
+++ b/supabase/functions/relay/index.ts
@@ -79,7 +79,7 @@ import {
   isRetiredModelRequest,
   isModelAllowedForTier,
   FREE_TIER_SUGGESTED_MODEL,
-  ULTRA_ONLY_PROVIDERS,
+  ULTRA_ONLY_PROVIDER_SET,
   ULTRA_TIER,
   FREE_TIER,
   MAX_TIER,
@@ -134,8 +134,6 @@ const OPENAI_GPT5_REASONING_EFFORT_CAPS: Record<string, ReasoningEffortCap> = {
   [FREE_TIER]: 'medium',
   [MAX_TIER]: 'high',
 };
-
-const ULTRA_ONLY_PROVIDER_SET = new Set<string>(ULTRA_ONLY_PROVIDERS);
 
 // Hop-by-hop and auth headers stripped before forwarding to the upstream
 // provider (auth is re-added from the server-side key).
--- a/supabase/functions/relay/index.ts
+++ b/supabase/functions/relay/index.ts
@@ -37,9 +37,9 @@
  * ============================================================================
  *
  * TIER HIERARCHY (model access):
- * - free: input <= $1.5/M AND output <= $9/M
- * - Max: every model
- * - Ultra: every model
+ * - free: input <= $1.5/M AND output <= $9/M, excluding openai/anthropic/google
+ * - Max: every model except openai/anthropic/google
+ * - Ultra: every model, including openai/anthropic/google (Ultra-only via relay)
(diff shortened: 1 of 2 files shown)


LionSR:
Addressed the two robustness findings in 08efb6f84: exported `ULTRA_ONLY_PROVIDER_SET` from `models.ts` instead of duplicating it in `index.ts`, and lowercased `m.provider` before the Set lookup in the free/Max filters.

On finding #1 (`FREE_TIER_SUGGESTED_MODEL = 'deepseek'` possibly not a real shortName): verified against the live `llm-zoo` registry — `deepseek` (provider `deepseek`, fullName `deepseek-v4-flash`) is a real entry with that exact shortName, and `TIER_CONFIG.tiers.free.models.includes('deepseek')` returns `true`. No change needed there.

PR #8504, opened 2026-07-15
fix(approvals): resolve bash/tool-edit bypass across stream ancestry

LionSR:
## Summary
- Bash/tool-edit/super-YOLO bypass ("YOLO") was a one-shot snapshot copied at the instant a new stream was created, with no live link back to whichever stream the user actually toggled it on — so it appeared to reset after one round of conversation.
- Extension: a delegated subagent stream copied its parent's bypass once at creation, so toggling YOLO afterward (or spawning a later sibling subagent) never picked it up.
- CLI: every chat round mints a brand-new root stream id, so bypass had nothing to attach to on the next turn.
- Fix: `SessionApprovals` now tracks a shared per-session ancestry map (`registerStreamParent`); `isBypassed()` walks up from a stream to its parent whenever the stream has no explicit bypass value of its own, instead of defaulting straight to `false`. An explicit toggle at any level still wins over an inherited one. Delegated subagents register the orchestrator as parent; each new CLI chat round registers the previous round's stream as parent.

## Test plan
- [x] `tsc --noEmit` (root, `tsconfig.test-kernel.json`, and `packages/cli`) — clean
- [x] `ChildStreamYoloInheritance.vitest.ts` — 6/6 pass, including 2 new regression tests: bypass toggled *after* the child stream already exists, and round-to-round inheritance via the session-level ancestry link (with explicit-override-wins coverage)
- [x] Full approval-suite sweep (`vitest -t "pproval"`) — 224/224 pass

---

> [!NOTE]
> **Medium Risk**
> Changes how auto-approval is resolved for bash, edits, and delegation across streams; behavior is security-sensitive but scoped and heavily tested.
> 
> **Overview**
> **Approval bypass (bash, tool-edit, super-YOLO) now follows a per-stream ancestry chain** instead of a one-shot copy when a new stream is created. `isBypassed()` walks parents when a stream has no explicit value; **toggle** flips the resolved state so inherited bypass can be turned off on the first press.
> 
> `SessionApprovals` adds **`registerStreamParent`** / **`forgetStreamAncestry`** with **separate graphs per bypass kind** (bash ancestry does not imply tool-edit or proposal). Stream teardown calls **`forgetStreamAncestry`** so children do not keep inheriting from cleared parents.
> 
> **CLI:** each new chat round links its root stream to the previous round in `onStreamResolved`. **Delegated subagents:** `inheritBashBypassOnChildStream` registers bash-only parent links instead of copying bypass at creation time.
> 
> Regression coverage expands in `ChildStreamYoloInheritance.vitest.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 620c1736a7ebb86020f28c36ffa9a4f698044d3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -426,6 +426,23 @@ export function createChatSessionController(
             lifecycleStarted = true;
           },
           onStreamResolved: (resolvedStreamId) => {
+            // Each chat round mints a fresh root StreamTabId (new
+            // executionId), so bash/tool-edit/super-YOLO bypass — which is
+            // keyed per stream — would otherwise reset every round even
+            // though the user is continuing the same conversation. Link the
+            // new round's stream to the previous one so bypass resolution
+            // (see `registerStreamParent`) falls through to whatever the
+            // prior round had, unless this round sets its own explicit value.
+            const previousRootStreamId = rootStreamId.get();
+            if (
+              previousRootStreamId &&
+              previousRootStreamId !== resolvedStreamId
+            ) {
+              defaultSession().approvals.registerStreamParent(
+                resolvedStreamId,
+                previousRootStreamId,
+              );
+            }
             session.streamId = resolvedStreamId;
             publishChatTuiRunState(session);
             rootStreamId.set(resolvedStreamId);
--- a/src/agent/runtime/streamApprovalQueue.ts
+++ b/src/agent/runtime/streamApprovalQueue.ts
@@ -35,6 +35,12 @@ export type ApprovalBypassEvent =
  * Single implementation behind the tool-edit, bash, and proposal (super-YOLO)
  * bypass toggles, so set/toggle/clear semantics and UI notification stay
  * uniform across approval kinds.
+ *
+ * A stream with no explicit bypass value of its own defers to its ancestor
+ * chain (see `resolveParent`) rather than defaulting straight to `false` —
+ * this is what lets a delegated subagent stream, or the next round of a CLI
+ * conversation, inherit a bypass its predecessor turned on, without a
+ * one-shot copy that misses toggles made after the child/round was created.
  */
 export interface StreamApprovalBypass {
   isBypassed(streamId: StreamTabId): boolean;
@@ -57,6 +63,7 @@ export interface StreamApprovalBypass {
 
 export function createStreamApprovalBypass(
   event: ApprovalBypassEvent,
+  resolveParent: (streamId: StreamTabId) => StreamTabId | undefined,
 ): StreamApprovalBypass {
   const byStream = new Map<StreamTabId, boolean>();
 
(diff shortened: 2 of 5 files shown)


LionSR:
### Possible privilege escalation: the child-stream ancestry link is bash-only by contract but grants all three bypass kinds

`inheritBashBypassOnChildStream` documents itself as linking a child *"for bash-bypass resolution, **independent of any tool-edit auto-approval**"*, and the child tool-edit grant is deliberately gated as a **separate** call:

```ts
// subagentExecution.ts:176-178
inheritBashBypassOnChildStream(resolvedStreamId, orchestratorStreamId); // unconditional — every child
if (options?.enableYoloOnChild) {
  enableYoloOnChildStream(resolvedStreamId);                            // gated
}
```

But `inheritBashBypassOnChildStream` registers the link through `session.approvals.registerStreamParent`, and in `createSessionApprovals` a **single shared** `parentOf`/`resolveParent` is fed to **all three** controllers (`toolEdit`, `bash`, `proposal`). So the unconditional bash link at line 176 also makes `toolEdit.isBypassed(child)` and `proposal.isBypassed(child)` walk up to the parent.

Consequences:

- **The `enableYoloOnChild` gate (line 177) is defeated for tool-edit.** Even when it's `false`, a child auto-approves edits whenever the parent has tool-edit YOLO on — inherited via the shared ancestry from line 176 — including a parent toggle made *after* delegation (the exact "propagates via live link, not one-shot copy" property, applied to the wrong kind).
- **super-YOLO leaks to children.** Delegation-proposal (super-YOLO) bypass is never intentionally propagated to a child by any wiring, yet the child inherits it from the parent through the same link, so a child auto-approves *its own* further delegations (the `proposal.isBypassed(child)` check on the delegation-proposal path).

This contradicts `inheritBashBypassOnChildStream`'s own doc contract ("bash… independent of the tool-edit YOLO flag; a parent with AUTO-BASH but edits still gated still propagates bash").

Note the CLI round-forwarding path (`onStreamResolved` → `registerStreamParent` for successive rounds) genuinely *does* want all three kinds carried forward, so the fix isn't "stop sharing the map" — it's to **scope the child link per bypass kind**: have `inheritBashBypassOnChildStream` register only a bash ancestry link, keep child tool-edit inheritance owned by the gated `enableYoloOnChildStream`, and let the CLI round path register all three (e.g. a per-kind `parentOf` or a `registerStreamParent(child, parent, kinds)` argument).

The existing `ChildStreamYoloInheritance` test only covers the parent-tool-edit-*unset* case, so it doesn't catch this — a regression test where the parent has tool-edit YOLO (or super-YOLO) on while `enableYoloOnChild` is `false`, asserting the child stays gated for those two kinds but still inherits bash, would lock the intended semantics in.

_(Reviewed against `2c73682`; the `toggleBypass` first-press-no-op point from an earlier revision is already resolved on this commit.)_

---

PR #8652, opened 2026-07-16
feat(storage): cross-host execution liveness heartbeat guards deletes

LionSR:
First half of #8625 (follow-up to #8624's shared `~/.texra` storage root).

## Problem

With all three hosts sharing one storage root, a delete in one host could remove an execution another host was actively writing. Nothing on disk distinguishes *running* from *crashed mid-run*: `ExecutionMeta.terminalStatus` is write-once at the end, and every host's delete guard consulted only its own in-process registry — the CLI's `history delete` had no guard at all.

## Changes

- **Heartbeat**: `ExecutionRegistry` touches `executions/{id}/heartbeat.json` for every active handle from one shared interval (10s, `unref`'d so it never keeps a headless CLI process alive), immediately on track, stopped when the last handle untracks. The body stamps the owning host (`extension`/`desktop`/`cli`, set once at each host's init) and pid for guard messages and diagnosis; freshness is carried by mtime.
- **Classification** (`executionLiveness.ts` in `@agent/storage`): live ⇔ meta exists ∧ no `terminalStatus` ∧ heartbeat fresher than 3 intervals. Crashed runs go stale within ~30s and stay deletable; pre-heartbeat executions read as not live.
- **Guards**: extension and desktop single-delete refuse live runs with the owning host named ("running in the TeXRA CLI"); clear-all excludes live ids alongside locally-active ones; CLI `history delete` now refuses live runs (text: stderr + exit 1; JSON/NDJSON gain `live`/`liveHost`, `--all` gains `skippedLive` and reports actual deletions when any were skipped).
- `heartbeat` registered as a reserved KV key so storage-dir walkers recognize it.

Internal post-kill cleanup paths are intentionally unguarded — a killed run's `terminalStatus` is written at finalize, so it classifies as not live immediately.

The storage-watcher live-refresh half of #8625 follows as a separate PR.

## Verification

- New `ExecutionLiveness` suite (6 tests: fresh/stale/missing heartbeat, terminal status, owner stamping, live-id listing) and `ExecutionRegistryHeartbeat` suite (2 tests: immediate + interval touches, stop-on-last-untrack) 
- Full affected set green: 41 files / 373 tests across agent/storage, agent/runtime, settings + desktop history handlers
- `tsc --noEmit` clean at root, test-kernel, extension, desktop, CLI; eslint clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
[...]

PR #8658, opened 2026-07-16
feat(history): live history refresh when another host writes executions

### LINKED ISSUE #8625: Cross-host shared storage follow-ups: guard destructive ops against live runs; live history refresh
Follow-up to #8622 / PR #8624, which pointed all three hosts (CLI, desktop, VS Code extension) at the shared \`~/.texra\` storage root.

The normal read/write paths are safe without locks: write-file-atomic rename semantics prevent torn reads, and each execution is only written by the host that started it (single-writer-per-execution). Two gaps remain:

## 1. Destructive ops can race a live run in another host

\`deleteExecution\` / \`deleteAllExecutions\` / housekeeping \`clean\`/\`pack\` in one host can remove an execution another host is actively writing (its debounced StreamLogStore save then recreates a partial dir or errors). Industry-standard fix, in order of weight:
- liveness check before delete: skip (or confirm) executions whose stream liveness marker is fresh;
- or an advisory per-execution lockfile (proper-lockfile-style, stale TTL) taken only by destructive ops — never by the normal read/write path.

## 2. History/progress views don't live-update across hosts

PR #8624 rescans on explicit history refresh, but a run started in the CLI while the extension history tab is open won't appear until re-open. Standard answer is fs watching (chokidar/fs.watch on \`executions/\`) with debounce, feeding the existing listing invalidation — read-side only, no cross-writes.

LionSR:
Second half of #8625 (follows the heartbeat delete guard, #8652). Closes #8625.

## Problem

With all three hosts sharing the `~/.texra` executions root (#8624), a run started or finished in one host only appeared in another host's history tab after closing and reopening it — nothing observed the storage directory.

## Changes

- **Extension**: `SettingsViewMessageHandler` registers a `FileSystemWatcher` over the shared executions directory. The directory lives outside the workspace, so the watcher uses a `RelativePattern` rooted at the storage path (a bare glob would never fire). Events re-send the history list to the active settings webview via the existing `withActiveWebview` + `sendHistoryData` path; watcher lifetime is the extension's (`context.subscriptions`), matching the `MainViewProvider` watcher pattern.
- **Desktop**: the main process watches the same directory with `fs.watch({recursive: true})` (created best-effort next to `DesktopHistoryHandlers`, closed in the window teardown block) and re-posts through the existing `postHistoryData`.
- Both hosts **ignore `heartbeat.json` events** — the #8652 heartbeat touches it every 10s per live run, which would otherwise re-post history continuously — and debounce the remaining launch/terminal write bursts (`DEBOUNCE_OPTIONS_MS`) into a single refresh.
- The CLI needs nothing: each invocation re-scans disk already.
- Progress board intentionally out of scope: it renders this process's in-memory stream snapshots, not the on-disk history (cross-host entries don't belong there).

## Verification

- Typechecks clean (root, test-kernel, extension, desktop, CLI); eslint clean; dead-code ratchet at baseline
- settings + desktop suites: 68 files / 537 tests green

---

> [!NOTE]
> **Low Risk**
> Read-only filesystem observation with existing history refresh paths; no auth or execution logic changes.
> 
> **Overview**
> **Live cross-host history updates** so the settings history list refreshes when the shared `~/.texra` executions directory changes from the CLI, extension, or desktop—without reopening the tab.
> 
> **Desktop** adds a recursive `fs.watch` on that directory (created if missing), debounces refreshes, ignores `heartbeat.json` events, and closes the watcher on window teardown. Refreshes reuse `historySettingsController.postHistoryData()`.
> 
> **Extension** registers a `FileSystemWatcher` with a `RelativePattern` rooted at the storage path (outside the workspace), re-registers on workspace-folder changes, applies the same heartbeat filter and debounce, and pushes updates via `historyHandlers.sendHistoryData` to the active settings webview.
> 
> Watch setup is best-effort; failures only disable automatic refresh—manual refresh still works.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5daacb3ed302bf26bce0b4ddb9b9f9f3dfc76300. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/index.ts
+++ b/packages/desktop/src/main/index.ts
@@ -928,7 +928,14 @@ function createWindow(options: {
       executionsDir,
       { recursive: true },
       (_event, filename) => {
-        if (filename?.endsWith('heartbeat.json')) return;
+        // filename may be a Buffer (or null) per fs.watch's contract; the
+        // separator anchor avoids matching an unrelated *heartbeat.json name.
+        if (
+          filename != null &&
+          /(?:^|[\\/])heartbeat\.json$/.test(String(filename))
+        ) {
+          return;
+        }
         void debouncedHistoryRepost();
       },
     );
--- a/packages/desktop/src/main/index.ts
+++ b/packages/desktop/src/main/index.ts
@@ -1,4 +1,4 @@
-import { existsSync } from 'node:fs';
+import { existsSync, mkdirSync, watch, type FSWatcher } from 'node:fs';
 import { writeFile } from 'node:fs/promises';
 import { spawn } from 'node:child_process';
 import { dirname, join, resolve as resolvePath } from 'node:path';
@@ -13,6 +13,7 @@ import {
 } from 'electron';
 
 import { platform } from '@platform/platform';
+import { RUNS_STORAGE_DIR } from '@platform/defaults/workspaceStorage';
 import { SHUTDOWN_PHASE, type LifecycleHost } from '@platform/interfaces';
 import { LatexConfigPersistenceController } from '@controllers/settingsView/LatexConfigPersistenceController';
 import { LatexToolingController } from '@controllers/settingsView/LatexToolingController';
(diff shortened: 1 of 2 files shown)


LionSR:
All four review points addressed: the extension watcher re-registers on `onDidChangeWorkspaceFolders` (storage path follows the workspace) and ensures the executions dir exists before watching; both hosts' heartbeat filters are separator-anchored; the desktop callback handles `fs.watch`'s `Buffer`/null `filename` contract. Typechecks, lint, and the settings+desktop suites (537 tests) green.

PR #8683, opened 2026-07-17
refactor: resolve overlapping-responsibility findings from architecture survey

LionSR:
## Summary

- route CLI and desktop workspace-containment checks through the canonical `isPathWithin` and `isStrictlyWithin` helpers
- replace three inline basename-without-extension calculations with the shared dotfile-safe `getFileStem`
- rename the misleading extension command folder from `history` to `taskFormState`
- correct contributor documentation for progress-view ownership and the separate `AppSignals` bus

## Design

Before this change, workspace-boundary predicates were independently implemented in the CLI and desktop host despite an existing shared owner in `pathCore.ts`. File-stem extraction was also repeated in three host call sites. These copies increased the number of places that future path hardening or dotfile corrections would need to change.

After this change, the host call sites use the existing shared helpers directly. No wrapper or compatibility export is added. The command-directory rename removes a naming collision between task-form restoration and the unrelated execution-history feature.

A proposed `FileList.parsePath` conversion was deliberately removed during current-main review: that component needs a lossless directory/basename split, while generic `getBasename` intentionally treats a trailing slash as the preceding segment. Keeping the local split preserves `foo/` exactly and avoids forcing unlike semantics through one helper.

## Net elements (R6)

| What | Added | Removed |
| --- | --- | --- |
| Exports | None | None |
| Functions | None | private `isContainedRelativePath` |
| Files | None | None; two command files renamed in place |
| Lines | 25 | 33 (-8 net) |

## Verification

- rebased onto current `main`
- 12 focused CLI, desktop, extension, and progress-view suites (189 tests)
- root, test-kernel, CLI, extension, and desktop TypeScript checks
- ESLint on every changed TypeScript file
- prior full repository CI, build, and webview smoke checks passed before the current-main rebase; fresh exact-head CI is running


LionSR:
@codex review exact head 27fc7633bc857cccef7eff3a534e8d272cadec89

LionSR:
Rebased onto current main and removed the mismatched FileList/getBasename conversion after self-review showed that a lossless path split has different trailing-slash semantics. The remaining diff is 10 files, -8 net lines, with 189 focused tests and all relevant type/lint/format checks passing. @codex review exact head f1d936b358245c68d84434bb812ac05474caf269

PR #8721, opened 2026-07-17
refactor(cli): delete the vestigial workflow-script progress projection

### LINKED ISSUE #8713: Delete the bespoke workflow-script CLI progress projection once the run is a child stream
Follow-up to the workflow-script async-detach issue (#8712). Do this AFTER #8712 lands.

## Goal

Once `delegate_workflow_script` runs as its own detached child execution/stream, the bespoke CLI tool-row projection added in #8672 becomes vestigial. Delete it and let the workflow run render through the **generic child-stream viewport** that already renders `delegate_agent` / `delegate_workflow` children. This is a pure simplification — remove a special case, reuse the general path.

## What to delete / rework

- `packages/cli/src/chat/tui/state/workflowScriptProgress.ts` (`applyWorkflowScriptProgressEvent`, `startWorkflowScriptOwnership`, `finishWorkflowScriptOwnership`, `patchWorkflowScriptOwner`) — the whole projection is built around finding-and-patching the owning `DELEGATE_WORKFLOW_SCRIPT_TOOL_NAME` tool row keyed to `logId`. With the run as a stream, its phase/log lines route into a `StreamSlice` like any child stream.
- The `workflowScriptFacts` / `workflowScriptOutcome` plumbing on the conversation entry (`cliState.ts`), the early-promotion special-case in `subscribeStreamLog.ts` (`isSettledEntry` treating `workflowScriptFacts !== undefined` as settleable), and the `workflowScriptFact` / `workflowScriptCompletion` Static item kinds in `StaticConversationTranscript.tsx` — all removable if the run is a normal stream.
- Confirm the `log`/`stage.start` handling added to `subscribeRuntimeHost.ts` for the workflow case is either subsumed by generic stream rendering or removed.

## Guardrail

The workflow run's phase stages (`kind: 'phase'`) + log lines must render in the focused-child viewport. Verify the focused-child renderer handles a stream that emits phase stages + info/error log lines (the extension already does via the transcript plane; confirm CLI parity). If the generic viewport does NOT render phases today, the smaller-surface fix is to teach the generic viewport about phase stages once — still net simpler than the per-tool projection.

## Verify

A running workflow-script shows its phases/agent-lines/cost by focusing its stream (Tab / Esc-s), same as any delegated child. Net-LoC must be clearly negative — this issue is a deletion. Report the count.

LionSR:
## Summary

After `delegate_workflow_script` became a detached child execution in #8716, its phase and log events no longer belonged to the orchestrator's tool row. This change removes the parallel tool-row projection and renders the run from its own stream trace through the ordinary focused-child transcript.

Workflow-script streams retain a distinct identity: their child header reads `workflow script`, rather than `subagent`, using the dispatch tool name already present in the child roster summary.

Closes #8713.

## Issue acceptance gates

- Deletes `workflowScriptProgress.ts` and its tool-row ownership transitions.
- Deletes `workflowScriptFacts`, `workflowScriptOutcome`, `activeWorkflowScript`, and the two workflow-specific static transcript item kinds.
- Removes the workflow-only `log`, `tool.start`, and `tool.end` fact subscription. The general `stage.start` subscription remains because round staging also uses it.
- Adds `workflow-script#` streams to the full child-log transcript set, so phase groups and `Running:`/`Finished:` log lines, including cost text, appear when the stream is focused.
- Gives workflow-script child streams a distinct header without adding a second transcript renderer or state projection.
- Replaces the projection test with a child-stream transcript test covering phase rows, agent log rows, final scrollback, cost text, and the distinct header.

## Single source of truth

The workflow run's `RunTrace` is the source of transcript content. The existing child-stream roster summary is the source of the child kind and display label. `StaticConversationTranscript` only transforms those two existing inputs for display.

The change removes the workflow-only event projection, three stored state fields, and two workflow-only static item variants. It adds no wrapper, manager, state field, or renderer. The only general change is that `subscribeStreamLog.ts` admits the workflow child-stream prefix to the existing full-log path.

## Net elements (R6)

- Files: 1 added, 2 deleted, 4 modified.
- Exported declarations: 0 added, 3 deleted (`WorkflowScriptProgressFact`, `ActiveWorkflowScriptInvocation`, and `applyWorkflowScriptProgressEvent`).
- Classes and enums: 0 added, 0 deleted. Interfaces: 0 added, 1 deleted.
- Lines: 167 inserted, 917 deleted; net −750.

## Consumer counts (R8)

- `applyWorkflowScriptProgressEvent` had one production caller, `applyDirectTuiRunEvent`; both the caller and definition are removed.
- The projected fields were confined to four production modules: their state declaration, producer, stream-log preservation path, and static renderer. All four paths are removed together.
- The removed `log`, `tool.start`, and `tool.end` entries belonged to one direct TUI run-fact subscription. Workflow content now arrives through the run stream's transcript subscription.

## Host impact

Extension: None.

Desktop: None.

CLI: Workflow scripts render from their child stream and use a `workflow script` child header. The orchestrator tool row no longer mirrors their progress.

## Scheduled refactoring

None.

## Validation

- `npm run format` (commit hook)
- `npm run typecheck`
- `npm run compile:fast`
- `npm test` (738 files passed, 1 skipped; 6765 tests passed, 4 skipped)
- Focused transcript suite: 56 tests passed
- `git diff --check`

---

> [!NOTE]
> **Medium Risk**
> Large deletion of transcript projection logic changes where users see workflow progress (child focus vs parent tool row); behavior is intentional but worth verifying in CLI UX. Core change is display/state only, not auth or persistence.
> 
> **Overview**
> **Workflow script progress no longer mirrors onto the parent’s `delegate_workflow_script` tool row.** The deleted `workflowScriptProgress` bridge, related CLI state (`workflowScriptFacts`, `workflowScriptOutcome`, `activeWorkflowScript`), and the extra static transcript item kinds are gone; the parent stream only keeps the normal tool entry.
> 
> **Focused workflow runs use the same path as other detached child logs.** `subscribeStreamLog` treats `workflow-script#…` like bash/claude/codex children so phase labels and `Running:` / `Finished:` lines (including cost text) come from that stream’s `RunTrace`. The child header reads **`workflow script`** instead of **`subagent`** when the roster summary’s tool name is `delegate_workflow_script`.
> 
> **Runtime wiring is simpler:** TUI run-fact handling no longer subscribes to workflow-only `log` / `tool.start` / `tool.end` for that projection. Tests move from the old projection suite to a child-stream transcript test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d4f8a1fd5488b84726213611ec57a5daceda9eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
The PR description now uses the repository template and includes the exact `## Net elements (R6)` and `## Consumer counts (R8)` sections. The amended commit also distinguishes workflow-script children in the shared header using the existing roster `toolName`; no parallel renderer or state field was added.

PR #8723, opened 2026-07-17
fix(runtime): restart retained child stream status

### LINKED ISSUE #8719: Fix stream status transition on deterministic workflow-script relaunch
## Context

Follow-up from #8716 (feat(workflow): detach delegate_workflow_script into an async child execution).

When a workflow script run is relaunched with the same `meta.name` (the primary resume path built to support #8666), the deterministic `runExecutionId` produces an identical `childStreamId` to the prior attempt. `createChildStream` → `trackAgentExecution` transitions `streamStatus` with cause `'lifecycle'`, but `canTransitionStreamPhase` (`src/shared/streams/streamStatus.ts:151`) only allows terminal→RUNNING for cause `RESUME`. Since the prior attempt already wrote a terminal phase to that `streamStatus` entry, the transition is silently rejected — the new run starts for real, but `streamStatus.get(childStreamId)` keeps reporting the old terminal phase. `executions action=wait/view` then reports the relaunch as already finished with stale results while the new run is still in flight.

## Details

- `createChildStream` (`src/tools/childStream.ts:86`) generates `childStreamId` from `streamPrefix + executionId`. For a deterministic relaunch, `executionId` is identical, so `childStreamId` collides with the prior attempt.
- `untrackAgentExecution` / `untrackHandle` do not clear `streamStatus` — the old terminal phase persists.
- `trackAgentExecution` transitions with cause `'lifecycle'`, but `canTransitionStreamPhase` rejects terminal→RUNNING for that cause.
- Either `trackAgentExecution` should use cause `'RESUME'` when the stream already exists in a terminal state, or `createChildStream` should clear/update the phase when it detects a pre-existing stream.

## References

- PR: #8716
- Review comment: https://github.com/LionSR/TeXRA/pull/8716#discussion (claude[bot] review thread on `WorkflowScriptTool.ts:180` and `WorkflowScriptTool.ts:170`)
- PR review: https://github.com/LionSR/TeXRA/pull/8716#issuecomment-5007656689
- Related: #8666 (durable named checkpoints), #8712 (design issue closed by #8716)

LionSR:
## Summary

Named workflow-script relaunches deliberately reuse their execution and child-stream identifiers. The preceding run leaves a terminal phase in the session status table, so the new handle's former lifecycle transition to `running` was rejected and callers could observe the preceding result while the relaunched workflow was active.

`ExecutionRegistry` now classifies an initial `running` status as a resume when the retained stream phase is terminal. This keeps the transition rule in the registry that already owns initial handle status, and it applies to any deterministic child relaunch.

Closes #8719.

## Issue acceptance gates

- A workflow-script child can be finalized and relaunched with the same execution identifier.
- The retained stream changes from `completed` to `running` for the second handle.
- The published status event records cause `resume` and the preceding terminal phase.
- The active-child roster reports the new run as active rather than returning the preceding terminal result.

## Single source of truth

`ExecutionRegistry.trackAgentExecution` remains the owner of the initial stream-status transition for a newly tracked agent handle. It derives the transition cause from the existing `StreamStatusMachine` phase and the requested initial phase.

No workflow-specific status write, option, or wrapper was added. The fix adds one cause derivation at the existing registry boundary and one regression test through `createChildStream`.

## Net elements (R6)

- Files: 0 added, 0 deleted, 2 modified.
- Exported declarations, classes, interfaces, and enums: 0 added, 0 deleted.
- Lines: 72 inserted, 1 deleted; net +71, of which 64 lines are the end-to-end regression test.

## Consumer counts (R8)

n/a. No export or event subscription is deleted or rerouted. The existing status transition call remains in place and now chooses its cause from retained state.

## Host impact

Extension: Relaunched deterministic child executions report their current running phase.

Desktop: Relaunched deterministic child executions report their current running phase.

CLI: `executions view` and `executions wait` no longer see the preceding terminal phase while a named workflow-script relaunch is active.

## Scheduled refactoring

None.

## Validation

- `npm run format` (commit hook)
- `npm run typecheck`
- `npm run compile:fast`
- Focused runtime suites: 42 tests passed
- Isolated `SystemUtils.vitest.ts`: 18 tests passed
- `git diff --check`
- `npm test` did not complete cleanly in the local environment. The first run passed 6773 tests and failed only the unrelated process-group PID assertion, which passed immediately when rerun in isolation. A second run was stopped after the known onboarding failure from #8720 and several unrelated tests exceeded the shared 10-second timeout. CI is required for the full-suite result.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to initial status transition cause derivation in the execution registry, with a focused regression test and no new public APIs.
> 
> **Overview**
> Deterministic child-stream relaunches (e.g. named **workflow-script** runs that reuse the same execution and stream IDs) could leave the session status stuck on the **previous terminal phase** because `trackAgentExecution` always published initial `running` with cause `lifecycle`, which the status machine rejects from `completed`/`failed`/`cancelled`.
> 
> **`ExecutionRegistry.trackAgentExecution`** now reads the retained phase for that stream and, when the requested initial phase is `running` and the prior phase is terminal, uses transition cause **`resume`** instead of **`lifecycle`**. Status events then record `resume` with the prior terminal phase, and active-child rosters reflect the new run as in flight.
> 
> An end-to-end regression test covers finalize-then-relaunch via `createChildStream` for workflow-script delegation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16a66d068d3f731081da297f8259e6b7561fa388. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/runtime/executionRegistry.ts
+++ b/src/agent/runtime/executionRegistry.ts
@@ -33,6 +33,7 @@ import {
 import {
   isActivePhase,
   isInFlightPhase,
+  isTerminalOutcomePhase,
   projectRunOutcome,
 } from '@shared/streams/streamStatus';
 import { formatDuration } from '@utils/core';
@@ -295,10 +296,16 @@ export class ExecutionRegistry {
     options: TrackAgentExecutionOptions = {},
   ): void {
     if (options.status) {
+      const previousStatus = this.streamStatus.get(handle.childStreamId);
+      const cause =
+        options.status === STREAM_PHASE.RUNNING &&
+        isTerminalOutcomePhase(previousStatus)
+          ? 'resume'
+          : 'lifecycle';
       this.streamStatus.transition(
         handle.childStreamId,
         options.status,
-        'lifecycle',
+        cause,
         this.streamStatusEmitOptions(handle),
       );
     }
--- a/src/test-kernel/agent/followUp/ChildStreamProgressEvents.vitest.ts
+++ b/src/test-kernel/agent/followUp/ChildStreamProgressEvents.vitest.ts
@@ -19,6 +19,7 @@ import {
   type ExecutionId,
   type StreamTabId,
 } from '@shared/schemas';
+import { DELEGATE_WORKFLOW_SCRIPT_TOOL_NAME } from '@shared/constants/delegationTools';
 import { createChildStream } from '@tools/childStream';
 import {
   createRecordingHost,
@@ -44,6 +45,8 @@ const normalizedErrorExecutionId = 'c11117' as ExecutionId;
 const normalizedErrorChildStreamId = 'codex#c11117' as StreamTabId;
 const noProjectionAutoCloseExecutionId = 'c11118' as ExecutionId;
 const noProjectionAutoCloseChildStreamId = 'bash#c11118' as StreamTabId;
+const workflowRelaunchExecutionId = 'c11119' as ExecutionId;
+const workflowRelaunchChildStreamId = 'workflow-script#c11119' as StreamTabId;
 const config = {
   agentCategory: AgentCategory.ToolUse,
   model: 'test-model',
@@ -92,6 +95,7 @@ describe('child stream progress events', () => {
       failedChildStreamId,
       normalizedErrorChildStreamId,
       noProjectionAutoCloseChildStreamId,
+      workflowRelaunchChildStreamId,
     ]) {
       clearStreamStatusForTest(defaultSession().status, streamId);
     }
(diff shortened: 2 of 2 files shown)


LionSR:
The differing cause derivation in `updateAgentExecutionStatus` is intentional. `trackAgentExecution` is registering a new handle for a retained stream identity, so terminal → running is a relaunch. `updateAgentExecutionStatus` operates on an already tracked handle; if that stream has become terminal, an explicit stop or terminal finalization has won, and the existing method contract says that loop updates must not revive it. Its handle-identity guard and lifecycle cause preserve that rule.

PR #8784, opened 2026-07-18
refactor(storage): unify deletion recovery state

### LINKED ISSUE #8725: StreamSnapshotStore: make deletion recovery a single discriminated state
## Problem

`StreamSnapshotStore` represents one stream-deletion transaction with three independent state dimensions:

- membership in `stagedDeletions`
- membership in `failedRollbacks`
- a mutable `phase: 'live' | 'transitioning' | 'staged' | 'unavailable'`
- an optional mutable `recovery` promise

Both maps store the same mutable `StagedDeletionState` object. A rollback failure first moves that object into `failedRollbacks` and later removes it from `stagedDeletions`; release code must compare object identity and delete from both maps. Every write must query the two maps in priority order and interpret `phase` together with the presence of `recovery`.

The behavior is well tested and was introduced to preserve writes across difficult filesystem failures. The concern is not missing validation at the boundary; it is that map membership has become an implicit part of the state machine. Several combinations are representable even though only a subset has a coherent meaning, and the valid transition graph is distributed across staging, rollback, write, reconciliation, flush, and release methods.

## Evidence

- `StagedDeletionState` has a mutable phase and optional recovery promise, while ownership is stored separately in two maps.
- `releaseStagedOwnership` must inspect and possibly delete the same object from both maps.
- `stageDeleteStream` checks both maps and recursively retries after awaiting different settlement paths.
- `write` checks both maps, then has a special branch for `failedRollback.phase === 'live' && !failedRollback.recovery`.
- `reconcileStagedDeletions`, `flush`, and `evictAll` each enumerate one or both maps with different lifecycle assumptions.

## Desired design

Use one per-stream deletion-state map whose value is a discriminated state. The discriminant should state who owns the stream and what operations are legal, for example:

- active staging transaction
- rollback recovery idle
- rollback recovery in progress

Namespace location (`live`, `transitioning`, `staged`, `unavailable`) should remain explicit where the filesystem genuinely makes it uncertain, but it should not be combined freely with two independent ownership maps and an optional promise. A recovery-in-progress state should carry its promise as a required field; an idle recovery state should not have one.

Keep the state machine inside `StreamSnapshotStore`. This issue does not justify extracting a single-caller coordinator or adding pass-through methods.

## Acceptance criteria

- One map owns deletion/recovery state for each stream.
- The state type makes active staging, failed rollback, and recovery-in-progress mutually exclusive.
- Callers do not infer ownership from membership in two maps or compare a state object against both maps.
- `write`, `stageDeleteStream`, reconciliation, flush, and release consume the same discriminated vocabulary.
- All existing rollback, namespace-loss, concurrent-write, flush, and restart-reconciliation tests remain intact.
- Add transition-focused tests only for distinctions not already covered; avoid duplicating the extensive filesystem matrix.
- Report the number of maps, state variants, mutable optional fields, branches, and net lines before and after. The refactor should remove state combinations or branch sites, not merely move code into another file.

## Priority

Medium, after the recent deletion-recovery changes have stabilized. This is a complexity and future-correctness risk, not evidence of a current data-loss defect.

LionSR:
## Summary

Represent each stream deletion transaction in one discriminated state map. Active staging, idle rollback recovery, and recovery in progress are now mutually exclusive, and only the recovery-in-progress variant carries a recovery promise.

## Issue acceptance gates

Closes #8725.

- One map owns deletion and recovery state per stream.
- The state discriminant controls whether settlement or recovery is legal.
- Stage, write, reconcile, flush, drain, and release paths use the same state vocabulary.
- Namespace authority remains an explicit four-value phase because filesystem failures can make it genuinely uncertain.
- Existing rollback, namespace-loss, concurrent-write, flush, and restart-reconciliation coverage remains intact.

Before to after metrics:

- ownership maps: 2 to 1
- representable ownership and recovery shapes, excluding namespace phase: at least 7 to 4 including the absent state
- explicit ownership variants: 0 to 3
- mutable optional lifecycle fields: 1 to 0
- namespace variants: 4 to 4
- conditional sites in the affected lifecycle and write region: 49 to 50; the added condition is an explicit stale-owner guard
- net lines: +35

## Single source of truth

StreamSnapshotStore.deletionStates now owns the entire per-stream deletion lifecycle. Its discriminated value says whether the stream is actively staging, waiting for rollback recovery, or currently being recovered.

Removed the stagedDeletions and failedRollbacks ownership split, cross-map object-identity cleanup, priority-ordered map lookups, and the optional recovery promise. The added file-local union encodes the lifecycle invariant; no coordinator, pass-through method, exported contract, or new runtime module was added.

## Net elements (R6)

- Files: 0 added, 0 deleted, 1 modified
- Exported symbols: 0 added, 0 deleted
- Runtime classes or enums: 0 added, 0 deleted
- File-local interfaces: 3 added, 0 deleted; these are the three discriminated lifecycle variants
- Net LoC: +35 (128 additions, 93 deletions)

## Consumer counts (R8)

n/a. No export or event path changed; all affected state is private to StreamSnapshotStore.

## Host impact

Extension: No host-specific behavior change; snapshot deletion and recovery keep the same contract.

Desktop: No host-specific behavior change.

CLI: No host-specific behavior change.

## Scheduled refactoring

None.

## Validation

- npm run format
- npm run compile:safe
- npm run lint (0 errors; one pre-existing replaceAll warning)
- npm run check:dead-code-ratchet (228 current, 228 baseline)
- CI=true npx vitest run src/test-kernel/transcript/StreamSnapshotStore.vitest.ts (62 passed), including after rebasing onto current origin/main
- npm run typecheck after rebasing onto current origin/main
- CI=true npm test: all 735 other files passed in repeated runs; full-suite load hit unrelated 10-second CLI timeouts in ToolUseRowPatch once and RunChatSignalOwnership twice. Both exact files pass alone (5/5 and 2/2); RunChatSignalOwnership takes 9.29 seconds alone against a 10-second timeout.

---

> [!NOTE]
> **Medium Risk**
> Touches filesystem staging, rollback, and buffered-write replay in snapshot deletion—failure modes are subtle, though behavior is intended to stay the same with existing tests.
> 
> **Overview**
> **`StreamSnapshotStore`** now tracks each stream’s reversible snapshot deletion in a single **`deletionStates`** map instead of separate **`stagedDeletions`** and **`failedRollbacks`** maps. Each entry is a discriminated union: **`staging`** (active stage with settlement), **`rollback-idle`** (buffered writes after a failed rollback), or **`rollback-recovering`** (the only variant that holds an in-flight recovery promise).
> 
> Stage, reconcile, flush, write buffering, and ownership release all branch on **`kind`** and shared **`writes`/`phase`** fields. **`runRollbackRecovery`** uses **`p-defer`** to dedupe concurrent recovery and transitions back to idle when recovery finishes; **`markRollbackFailed`** replaces cross-map promotion when staging fails. **`stageDeleteStream`** waits on prior deletion state in a loop before starting a new stage. Buffered writes during **`rollback-idle`** with **`phase === 'live'`** still mirror to disk and clear ownership when the write map drains, with a stale-owner guard on **`writes`** identity.
> 
> Public snapshot deletion behavior and the four-value namespace **`phase`** are unchanged; this is an internal state-machine cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b07338ee8cad12819583aa014aa7bc3958ef4698. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/transcript/StreamSnapshotStore.ts
+++ b/src/transcript/StreamSnapshotStore.ts
@@ -18,6 +18,7 @@
  */
 
 import { Mutex } from 'async-mutex';
+import pDefer from 'p-defer';
 import pMap from 'p-map';
 import { z } from 'zod';
 
@@ -295,15 +296,30 @@ interface StreamRecord {
 type StagedDeletionPhase = 'live' | 'transitioning' | 'staged' | 'unavailable';
 type StagedRecoveryOutcome = 'discarded' | 'restored' | 'unchanged';
 
-interface StagedDeletionState {
+interface DeletionStateBase {
   writes: Map<string, unknown>;
   /** Namespace authority and whether failed ownership may mirror writes. */
   phase: StagedDeletionPhase;
-  /** One recovery owns namespace repair and buffered-write replay at a time. */
-  recovery?: Promise<StagedRecoveryOutcome>;
+}
+
+interface StagedDeletionState extends DeletionStateBase {
+  kind: 'staging';
   settled: Promise<void>;
   resolveSettled: () => void;
 }
+
+interface IdleRollbackState extends DeletionStateBase {
+  kind: 'rollback-idle';
+}
+
+interface RecoveringRollbackState extends DeletionStateBase {
+  kind: 'rollback-recovering';
+  /** Owns namespace repair and buffered-write replay. */
+  recovery: Promise<StagedRecoveryOutcome>;
+}
+
+type RollbackState = IdleRollbackState | RecoveringRollbackState;
+type DeletionState = StagedDeletionState | RollbackState;
 
 export class StreamSnapshotStore {
   private readonly records = new Map<StreamTabId, StreamRecord>();
@@ -313,14 +329,7 @@ export class StreamSnapshotStore {
    * transaction boundary: commit discards them, rollback replays them only
    * after the live namespace has been restored.
    */
-  private readonly stagedDeletions = new Map<
-    StreamTabId,
-    StagedDeletionState
-  >();
-  private readonly failedRollbacks = new Map<
-    StreamTabId,
-    StagedDeletionState
-  >();
+  private readonly deletionStates = new Map<StreamTabId, DeletionState>();
 
   // -- Per (stream, category) serialized write locks -------------------------
   private readonly writeMutexes = new Map<string, Mutex>();
@@ -945,9 +954,10 @@ export class StreamSnapshotStore {
     for (const stream of this.records.keys()) this.bumpStreamVersion(stream);
     this.records.clear();
     this.writeMutexes.clear();
-    for (const state of this.stagedDeletions.values()) state.resolveSettled();
-    this.stagedDeletions.clear();
-    this.failedRollbacks.clear();
+    for (const state of this.deletionStates.values()) {
+      if (state.kind === 'staging') state.resolveSettled();
+    }
+    this.deletionStates.clear();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


LionSR:
Addressed the review suggestion in b07338ee8c: the recovering owner is now constructed and installed before the recovery callback starts, backed by a required deferred promise. This removes the before-declaration closure without reintroducing an optional lifecycle field. Focused tests, typecheck, and lint remain green.

Issue #8832, opened 2026-07-18
Phase 3: workflow-script agent({ schema })


LionSR:
Part of #8829. Depends on Phase 1.

## Scope

Expose structured output to workflow scripts via `agent(prompt, { schema })`.

## Work

- `src/agent/workflowScript/runWorkflowScript.ts`: delete the reject-guard (currently throws on `schema`/`outputSchema`); accept a plain JSON-Schema object; add `schema` to `WorkflowAgentCallOptions` (`types.ts`). Journal identity already hashes options via stable-stringify, so resume stays correct with the schema included.
- `src/tools/delegation/workflowScriptAgentRunner.ts`: when `options.schema` is present, `resolveStructuredOutput(jsonSchema)` then `buildTerminalTool`, inject into the child run; the validated object surfaces as `.structured` on the returned envelope.

## Verification

- Engine test with a fake `runAgent` that returns a typed object; assert `(await agent(p, { schema })).structured` reaches the script.
- Assert the journal key includes the schema (resume replays with the schema).
- `npm run typecheck` + the workflow-script suites.

LionSR:
Implemented and merged in #8840.

Issue #8831, opened 2026-07-18
Phase 2: final-synthesis turn + native forcing


LionSR:
Part of #8829. Depends on Phase 1.

## Scope

Add native forcing as an accelerator under the Phase 1 guarantee, plus the final-synthesis-turn orchestration that makes forcing usable for multi-turn agents.

## Work

- Final-synthesis turn in the tool-use flow terminate path: run the normal unforced tool loop to natural termination (`end_turn` / no tool call), then issue exactly one additional turn with `tool_choice` forced to the terminal tool. Single-shot runs collapse to just that one forced turn.
- Add `finalTool?: { name: string }` to `CreateResponseOptions` (`src/agent/types/ModelHandlerContracts.ts`).
- Each `createResponseImpl`, when `finalTool` is set and the provider supports forcing, maps it to native forced tool_choice, replacing the `'auto'` literal on that turn:
  - Anthropic `tool_choice: { type: 'tool', name }`
  - OpenAI Chat / Responses `tool_choice`
  - OpenRouter `toolChoice`
  - Google `function_calling_config: { mode: ANY, allowed_function_names: [name] }`
  - VS Code LM: no forcing, uses the unforced floor
- Per-handler `supportsForcedToolChoice` overridable getter (default false; native providers override true). Local, an optimization not a correctness gate.

## Verification

- Per-provider mocked-client tests asserting `finalTool` maps to the native forced tool_choice, and VS Code LM takes the unforced floor.
- A mocked multi-turn run: exploration turns are unforced, the final turn forces the terminal tool.
- `npm run typecheck` + suites.

Issue #8829, opened 2026-07-18
Structured output (Zod-native) via terminal tool: tracking


LionSR:
## Goal

Add a single Zod-native structured-output contract to TeXRA: any agent run can declare an output schema and is guaranteed a validated object of that type, on every provider. The value surfaces as `.structured` on the agent result envelope; workflow scripts get it via `agent(prompt, { schema })`.

This design is the synthesis of a read-only design tournament (4 approaches x 3 adversarial lenses). Two findings shaped it:

1. **The guarantee is host-side `z.parse` + one schema-guided repair retry**, which the tool layer already implements (ZodError to ToolResult, model self-corrects). Native forcing is a pure accelerator, never the guarantee.
2. **The mechanism is a synthetic terminal tool** built from the schema, reusing the whole existing tool spine (`convertToolSchema` + `defineTool` Zod-parse + `SdkToolCall` extraction), with a final-synthesis turn for multi-turn agents.

## Locked decisions

- `agent(prompt, { schema })` resolves to the envelope plus a `.structured` field (not a bare object).
- `strict: false`. TeXRA SSOT schemas use `.int()/.min()/.nullish()` which violate provider strict mode; gate native use per-call with `isStrictNativeCompatible(spec)` and fall through to validate+repair.
- Sandbox authors a JSON Schema object; host lifts it with `z.fromJSONSchema` (Zod 4.4.3). Journaled via stable-stringify.
- Review keeps per-finding streaming: `report_review_issue` stays; not folded into a monolithic `findings[]`.
- Capability gating is local (per-handler overridable getter), not an llm-zoo release.

## Sequential phases (each a PR; disjoint lanes where possible)

- [x] #8830 Phase 1: structured-output contract + terminal tool (foundational, blocks the rest)
- [x] #8831 Phase 2: final-synthesis turn + native forcing across handlers
- [x] #8832 Phase 3: workflow-script agent({ schema })
- [x] #8833 Phase 4: Nunjucks render() sandbox primitive
- [x] #8834 Phase 5: review alignment onto the shared spine

## Verification

Mocks and synthetic state only, no slow live models. `npm run typecheck` + targeted `npm test` per phase.

LionSR:
Complete. Zod-native structured output is on main end to end.

[...]

PR #8840, opened 2026-07-18
feat: structured output via terminal tool (contract + tool-use consumer)

### LINKED ISSUE #8830: Phase 1: structured-output contract + terminal tool
Part of #8829. **Foundational, blocks Phases 2-5.**

## Scope

Add the structured-output contract and the terminal-tool builder, plus the `.structured` result field. No model-handler edits in this phase: the guarantee comes from the existing tool-input Zod-parse + repair loop, which already works on every provider via ordinary (unforced) tool calls.

## Work

- New `src/tools/structuredOutput.ts`:
  - `resolveStructuredOutput(input: z.ZodType | JSONSchemaObject, name): StructuredOutputSpec` where the spec is `{ name, jsonSchema, zodSchema, validate(raw): Result<unknown> }` (mirrors `ToolDefinition`'s dual shape). Host callers pass live Zod; sandbox callers pass a JSON Schema object lifted with `z.fromJSONSchema`.
  - `buildTerminalTool(spec): ToolDefinition` (name `submit_output`, `zodSchema = spec.zodSchema`), reusing `defineTool` / `convertToolSchema`.
  - `isStrictNativeCompatible(spec): boolean` (false when the schema uses numeric/length/optional keywords).
- `.structured?: unknown` on BOTH members of `AgentFinalResultSchema` (`src/agent/runtime/AgentFinalResult.ts`, workflow and toolUse), populated in `buildAgentFinalResult` from a run-scoped capture slot modeled on `setReportReviewIssueSink`.

## Constraints

- Single validation site: route through the existing tool-input `z.parse`; do not add a parallel validator.
- `.structured` must be JSON-serializable (crosses the sandbox bridge unchanged).

## Verification

- `resolveStructuredOutput` round-trips a JSON Schema via `z.fromJSONSchema` and validates/rejects sample payloads.
- `isStrictNativeCompatible` rejects `.int()/.min()/.nullish()` schemas.
- Mocked run where the model's first terminal-tool call is invalid, then the repair produces a valid `.structured` on the second; assert it appears on both envelope variants.
- `npm run typecheck` + the new suites.

LionSR:
Adds provider-independent structured output through a synthetic terminal tool, including its first runtime and workflow-script consumers. Tracking #8829.

## Design

- `normalizeStructuredOutputSchema(...)` is the single boundary for live Zod schemas and sandbox JSON Schema. It converts both to the canonical object-root tool schema, rejects unsupported roots early, and blocks regex-bearing sandbox schemas before they can compile into host `RegExp` instances.
- `buildTerminalTool(...)` creates a run-local `submit_output` tool through the existing `defineTool` / `BaseTool` validation spine. Successful submission is single-shot, JSON-safe, and explicitly ends the tool-use turn.
- The tool-use flow installs the terminal tool through a non-mutating registry overlay, moves a completed result into persisted shared state for interrupt/resume durability, and refuses to complete a configured structured-output run without a submission.
- Tool dispatch owns terminal-result semantics: it pairs the terminal result, stops later calls, and completes the turn. The workflow runner no longer rechecks the contract after child-run persistence.
- `BaseTool` preserves synchronous validation timing for ordinary schemas and falls back to async parsing only for schemas that require it.
- Final-result schemas enforce that `.structured` is persistable JSON on both workflow and tool-use envelopes.
- Workflow scripts accept `agent(prompt, { agentName, schema })`, normalize the schema at the sandbox boundary, include it in journal identity, route the call through a tool-use agent, and return the captured value as `.structured`.

The unused speculative native-forcing predicate and contract wrapper were removed. Native forcing remains an optional follow-up accelerator in #8831; this PR's ordinary tool-call path owns correctness.

## Validation

- `npm run format`
- `npm run typecheck`
- `npm run lint` (zero errors; one unrelated existing warning)
- `npm run compile:fast`
- `npm test` — 740 files passed, 1 skipped; 6,860 tests passed, 4 skipped

Part of #8829. Closes #8830, #8832.

---

> [!NOTE]
> **Medium Risk**
> Changes core tool-use flow, dispatch ordering, and subagent delegation paths with new completion invariants, but behavior is gated on outputSchema/schema options and covered by extensive new tests.
> 
> **Overview**
> Introduces **provider-independent structured output** for tool-use agents: when config carries `outputSchema`, the run exposes a run-scoped **`submit_output`** terminal tool (via a non-mutating registry overlay), validates submissions through the normal tool/Zod path, and surfaces the captured value as **`structured`** on flow results, final envelopes, and persisted shared state (resume-safe).
> 
> **Tool dispatch** now honors an optional **`endTurn`** flag on executed tool results so a successful terminal submission stops further calls in the batch and completes the model turn instead of continuing the tool loop.
> 
> **Workflow scripts** replace the old rejection of `schema`/`outputSchema` with first-class **`agent(prompt, { schema, agentName })`**: normalized object-root JSON Schema (regex patterns blocked for sandbox safety) routes to a tool-use subagent with `outputSchema`, exempts the empty-input-files guard, and includes schema in journal identity; callers read **`.structured`** on the returned envelope.
> 
> Supporting changes: optional **`outputSchema`** on tool-use agent config and trace schema; **`BaseTool`** falls back to async Zod parse when needed; docs/tool descriptions updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 375f9e883314ed1fe006e0eba4f5d5b46c2bf03f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/resources/docs/agent-creation/tool_catalog.md
+++ b/packages/extension/resources/docs/agent-creation/tool_catalog.md
@@ -63,8 +63,11 @@ recommended groups at the bottom are a good starting point.
   workflow-agent calls with predetermined branching and fan-out. Pass a default
   `agent`, the complete `script`, and optional JSON `args`. The script begins
   with `export const meta = { name, description }` and can use `agent`, `phase`,
-  `log`, `parallel`, `pipeline`, and `concat`. It is intentionally absent from
-  the recommended orchestrator set.
+  `log`, `parallel`, `pipeline`, and `concat`. `agent(prompt, { schema })`, where
+  `schema` is a JSON Schema object, runs a tool-use agent (name one via
+  `agentName`) that finishes by calling `submit_output`; the call resolves to an
+  envelope whose `.structured` is the validated object. It is intentionally
+  absent from the recommended orchestrator set.
 - `delegate_agent` — delegate to another tool-use agent. Pass `agent`,
   `model`, and `instruction` for a fresh run, or `execution_id` +
   `instruction` to resume a WAITING subagent.
--- a/packages/trace-viewer/src/traceDataSchema.ts
+++ b/packages/trace-viewer/src/traceDataSchema.ts
@@ -65,6 +65,7 @@ const TraceAgentConfigSchema = z
     }),
     TraceAgentConfigSharedFieldsSchema.extend({
       agentCategory: z.literal(AgentCategory.ToolUse),
+      outputSchema: z.record(z.string(), z.unknown()).nullish(),
     }),
   ])
   .superRefine((config, ctx) => {
--- a/src/agent/core/definition/AgentConfig.ts
+++ b/src/agent/core/definition/AgentConfig.ts
@@ -55,6 +55,13 @@ const WorkflowAgentConfigFieldsSchema = AgentConfigSharedFieldsSchema.extend({
 
 const ToolUseAgentConfigFieldsSchema = AgentConfigSharedFieldsSchema.extend({
   agentCategory: z.literal(AgentCategory.ToolUse),
+  /**
+   * JSON Schema (a plain object) describing a structured output the agent must
+   * submit through the synthetic `submit_output` terminal tool. Serializable by
+   * design: a live Zod schema or ITool must never travel through config. Absent
+   * for ordinary tool-use runs.
+   */
+  outputSchema: z.record(z.string(), z.unknown()).nullish(),
 });
 
 const AgentConfigFieldsSchema = z.discriminatedUnion('agentCategory', [
--- a/src/agent/core/flows/toolUseRound/ToolUseDispatchNode.ts
+++ b/src/agent/core/flows/toolUseRound/ToolUseDispatchNode.ts
@@ -69,6 +69,10 @@ interface ToolExecutionResult {
   logRef: { logId: string | undefined; groupId: string | undefined };
 }
 
+function endsToolUseTurn(result: ToolExecutionResult | null): boolean {
+  return result?.result.status === 'executed' && result.result.endTurn === true;
+}
+
 /**
  * Dispatches tool calls with ordering barriers.
  *
@@ -191,6 +195,7 @@ export class ToolUseDispatchNode<C> extends Node<
             batchController.signal,
           );
           i += 1;
+          if (endsToolUseTurn(results[index])) break;
           continue;
         }
         // Contiguous run of parallel-safe calls executes concurrently.
@@ -212,6 +217,7 @@ export class ToolUseDispatchNode<C> extends Node<
             }),
           ),
         );
+        if (run.some((index) => endsToolUseTurn(results[index]))) break;
       }
       this.fanOutDuplicateResults(calls, results);
       this.rejectUnsafeDuplicates(calls, results);
@@ -631,6 +637,11 @@ export class ToolUseDispatchNode<C> extends Node<
     if (interrupted) {
       shared.shouldStop = true;
     }
+    const endTurn = allResults.some((result) => endsToolUseTurn(result));
+    if (endTurn) {
+      shared.shouldStop = true;
+      shared.endTurn = true;
+    }
 
     const assistantText = shared.text ?? '';
 
@@ -691,6 +702,6 @@ export class ToolUseDispatchNode<C> extends Node<
 
     shared.toolCalls = [];
 
-    return FlowTransition.CONTINUE;
+    return endTurn ? FlowTransition.COMPLETE : FlowTransition.CONTINUE;
   }
 }
--- a/src/agent/implementations/flows/tooluse/ToolUseServices.ts
+++ b/src/agent/implementations/flows/tooluse/ToolUseServices.ts
@@ -4,7 +4,7 @@ import type { IToolRegistry } from '@agent/core/tools/ToolTypes';
 import type { IToolUseSession } from '@agent/core/flows/IToolUseSession';
 import type { SubagentProgressUpdate, TodoItem } from '@shared/schemas';
 import type { TaskRunFileService } from '@utils/files';
-import type { PreparedShared } from './nodes/types';
+import type { PreparedShared, ToolUseRunShared } from './nodes/types';
 
 export interface ToolUseServices<C = unknown> extends BaseFlowContextInit<C> {
   readonly setting: AgentToolUseSetting;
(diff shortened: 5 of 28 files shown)


LionSR:
Fixed the Bugbot finding in 0d907b1: `isStrictNativeCompatible` now treats any object node (`type: 'object'` or a `properties` key) as requiring `additionalProperties: false`, and rejects unconstrained schemas (a bare `{}`, including the `convertToolSchema` null-fallback). Added tests for a bare `{ type: 'object' }` and an empty `{}`.

LionSR:
Thanks for the thorough pass. Triage of the findings on this contract-only PR:

**Fixed here**
- **Non-object root schema (codex P1):** `buildTerminalTool` now throws a clear error when the schema root is not an object, since provider tool inputs and native structured output require an object root. Wrapping a scalar/array in a property is the caller's job. Added a test (`14cdab8`).
- **Open/empty object passing the strict check (cursor, codex P2):** already addressed in `0d907b1` — `isStrictNativeCompatible` now treats any object node (`type: 'object'` or a `properties` key) as requiring `additionalProperties: false`, and rejects unconstrained schemas (bare `{}` / the `convertToolSchema` null-fallback). Tests cover `{ type: 'object' }` and `{}`.

**Not a bug**
- **`z.fromJSONSchema` "not public":** it is present and public in the installed `zod@4.4.3` (`z.toJSONSchema` and `z.fromJSONSchema` are both exported from the main entry). The round-trip test exercises it: a JSON Schema object is lifted to a validator that accepts a good payload and rejects a bad one. Verified against `node_modules/zod`.

**Deferred to the phase that owns the behavior (all tracked under #8829)**
- **Single-shot / terminal `submit_output` (codex P1):** enforcing exactly one terminal call belongs to the final-synthesis turn in Phase 2 (#8831), which runs the tool loop to natural termination and then issues one forced terminal turn. This PR only builds the tool; it is not yet wired into a run.
- **Native strict forcing, array-constraint rejection, dictionary schemas (codex P2):** `isStrictNativeCompatible` is the gate these feed; the actual native `tool_choice` / `response_format` mapping and its per-provider guards land in Phase 2 (#8831).
- **JSON-safe values in the persisted envelope (codex P2):** `.structured` is `z.unknown()`; the value only ever originates from a tool-call argument (already JSON) or the JSON-only sandbox bridge, so it is JSON-safe by construction at the points that set it. Phase 3 (#8832) wires those and will assert it.

**On exports without a runtime consumer (claude, codex P1):** fair point, and rather than argue it I am folding the first runtime consumer into this PR: the workflow-script `agent(prompt, { schema })` path (Phase 3, #8832), which calls `resolveStructuredOutput` + `buildTerminalTool` in the delegation runner. The updated PR will then be "contract + workflow-script structured output," so every export has a real caller here, not only in tests. `isStrictNativeCompatible` is exercised by the native-forcing gate that follows in Phase 2 (#8831); I will keep it exported with its tests and note the consumer lands there.

LionSR:
Expanded this PR to include the first runtime consumer, directly addressing the "exports need an in-PR consumer" feedback.

`agent(prompt, { schema })` in workflow scripts now routes to a tool-use run that injects a synthetic `submit_output` terminal tool (via a per-run registry overlay, no module globals) and surfaces the validated call as `.structured`. Floor only; native `tool_choice` forcing and the final-synthesis turn remain in #8831.

Now closes #8830 and #8832. 108 tests, typecheck green across all projects.

LionSR:
Addressed the review round in `15b0b21`:

- **Schema call without `agentName` (P1):** now throws a clear `WorkflowRunAbortError` in `prepare()` explaining that structured-output calls must name a tool-use agent. No silent fallback to a workflow-default agent (which has no tool-use counterpart). A general `assistant` tool-use agent exists but isn't guaranteed to be in the run's delegation scope, so erroring is the honest contract.
- **Floor could return `structured: undefined` (P1):** a schema call whose run completes without a captured value now fails clearly ("agent did not call submit_output; no structured result was produced"). This makes the floor a clear-failure guarantee; reliable forcing/repair is the accelerator in #8831.
- **Tool contract didn't advertise `schema` (P1):** documented `agent(prompt, { schema })` in the `delegate_workflow_script` description, `workflowScript/README.md`, and the tool catalog (JSON Schema object, runs a tool-use agent via `agentName`, finishes with `submit_output`, result carries `.structured`).
- **`buildAgentFinalResult` ignored `flowResult.structured` (Low):** the flowResult branch now surfaces it (`source.structured ?? result.structured`).
- **Export surface:** `SUBMIT_OUTPUT_TOOL_NAME` is now file-local (no external runtime consumer); the test hardcodes the name.

121 tests, typecheck green.

LionSR:
Hardened the untrusted sandbox->host schema boundary in `930b79b`:

- **`schema: {}` / non-object schema crashed mid-run (Medium):** now validated at the script boundary in `normalizeAgentOptions` — a schema whose root is not an object (`type: "object"` or a `properties` key) is rejected up front with a clear `agent() option "schema"` error, before any run launches.
- **Missing output could persist as completed (P1, durability):** moved the guarantee into the flow. When `outputSchema` is set and `submit_output` was never called, `runToolUseFlow` reports a `failed` outcome, so the child never checkpoints as an empty success and a resume retries the call. The runner check remains only as a clearer message.
- **Untrusted schema regex = host ReDoS (P1, security):** added `assertSafeSandboxSchema`, applied ONLY to sandbox-supplied schemas (never to trusted host Zod callers). Walking the schema it rejects `pattern`/`format` (the `z.fromJSONSchema` RegExp-compilation ReDoS vector), `$ref`/`$dynamicRef`, prototype-polluting property keys (`__proto__`/`constructor`/`prototype`), and trees past a 12-level / 1000-node cap.

117 tests, typecheck green across all projects.

LionSR:
On the two latest Bugbot findings:

- **Tuple `items` skip the regex scan — not applicable to the committed code.** The cited `assertNoHostRegex` does not exist; the actual guard is `assertSafeSandboxSchema` in `runWorkflowScript.ts`, which walks arrays element-by-element (`Array.isArray(node)` branch, lines 84-86) AND recurses into every value (`for (const value of Object.values(record)) walk(value, ...)`, lines 119-121). So a `pattern` nested inside a tuple `items: [{ pattern: ... }]` is reached and rejected. Covered by a test.
- **Captured output not persisted across resume — real, fixing now.** `capturedStructured` is a flow-local closure, so a run resumed after `submit_output` was already in the transcript would not re-fire the capture. Adding transcript recovery: at completion, if the value is uncaptured but the transcript contains a valid `submit_output` call, recover and re-validate it (via the same Zod parse) rather than failing the run.

LionSR:
Resume-capture fixed in `f536b82`: at completion, if the structured value wasn't captured live but the transcript contains a valid `submit_output` call, it is recovered and re-validated through the same Zod parse (a new provider-agnostic `findLastToolCallInput` handles all six message shapes). The FAILED floor still fires only when no valid call exists. `SUBMIT_OUTPUT_TOOL_NAME` is re-exported since the recovery scan is now a real runtime consumer. Typecheck green; 25 structured/delivery tests pass.

LionSR:
On the last minor note: the completion guards already key off `structuredSpec` consistently (the recovery block and the FAILED floor both use it; `outputSchema` only appears deriving `structuredSpec`). That comment anchored to a shifted line in the resume-fix diff. The two re-listed Bugbot items are the now-fixed resume recovery (`f536b82`) and the tuple false-positive (the sanitizer walks arrays + all values). I believe this is ready.

PR #9010, opened 2026-07-20
fix(cli): bound transcript memory in multi-agent runs

LionSR:
## Summary

- serialize dirty transcript writes so only one complete transcript is encoded at a time
- keep complete projected entries only for the focused CLI session, while preserving exact background summaries and restoring transcripts when focused or printed
- release dormant transcript entries through the existing persistent store lifecycle and reload them before follow-up or persisted-session resume
- retain only the scalar exit code needed by the CLI instead of a second raw tool payload
- remove the separate child-response summary cache and use the existing stream description

## Root cause

Large multi-agent runs held each complete transcript twice: once in the transcript store and once in the CLI signal state. A save burst also serialized every dirty transcript concurrently, causing several large JSON strings to coexist. With roughly fifty subagent sessions, these two effects could exhaust Node's 4 GB V8 heap.

## User impact

Inactive subagent transcripts no longer accumulate in the terminal's live state. Their row summaries remain visible, and focusing a session or requesting full output restores the exact transcript from the existing store. No setting or persisted-data format changes.

## Validation

- `npm run format`
- `npm run compile:fast`
- `npm run typecheck`
- `npm test` (6,919 passed; 4 skipped)

---

> [!NOTE]
> **Medium Risk**
> Changes core transcript projection, lazy release/reload, and persistence flush ordering; regressions could show empty subagent transcripts, stale summaries, or missed entries when switching streams or printing.
> 
> **Overview**
> **Large multi-agent runs use substantially less memory** by not keeping every subagent’s full projected transcript in CLI signal state when that stream isn’t focused.
> 
> For **unfocused, non-active** streams, `syncStreamLog` now keeps only synthetic entries plus a **`description`** (latest assistant reply after the last user message, or the in-flight instruction) and calls **`releaseEntries`** on the transcript store so heavy log data can be dropped from memory. The **focused** stream (or **`forceFull`**, e.g. print transcript) still gets the full entry list; switching tabs **`ensureLoaded`**s the target stream and syncs the previous one. Print and load failures surface via **transient notices**.
> 
> **`NormalizedToolUse`** drops the **`parsed`** raw payload; **`exitCode`** is normalized once in **`normalizeToolUseData`** and used by CLI tool renderers. The subagent list reads **`slice.description`** instead of scanning entries via **`latestChildResponseSummary`** (removed).
> 
> **Tool-use resume/follow-up** paths **`await transcripts.ensureLoaded`** before continuing. **`StreamLogStore`** flushes dirty streams **sequentially** (one serialized JSON at a time) instead of **`Promise.allSettled`** in parallel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65e0a9fa3e87c141c81d407f70615efbe2d36c85. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR (commented):
## Audit

**Verdict: sound — approving.**

The design is the right shape: single-copy ownership (CLI signal state now materializes full projected entries only for the focused stream; the transcript store stays the one authoritative copy — removes the "held twice" duplication behind the heap growth), reuses the existing `releaseEntries`/`ensureLoaded`/`pendingRelease` lifecycle in `StreamLogStore` rather than inventing a new eviction layer, and includes two genuine deletions (`NormalizedToolUse.parsed` → scalar `exitCode`; the `latestChildResponseSummary` WeakMap render-cache → `slice.description` computed at sync time, moving a render-time cache into the data model per the TUI stateless-renderer convention). Sequential flush in `executeWrite` bounds peak memory to one serialized transcript while preserving per-stream failure isolation.

### Bot findings — checked against head `5006b0014e`

- Bugbot "stale focus in print callback": fixed by `c5567a8a` ("read current focus after transcript load") — `printStreamOutput` now reads `activeStreamIdSignal.get()`/`streamsSignal.get()` after the load resolves.
- Bugbot "aborted focus load retains transcript": refuted by the store's own lifecycle. On focus A→B→C mid-load, the handler syncs B (previous); `releaseEntries(B)` queues into `pendingRelease` because `pendingLoads.has(B)` is true, and `ensureLoaded`'s completion honors the queued release (`StreamLogStore.ts:362-370`), deleting the log. No leak.
- texra-review's exit-code regex nit: already the tighter `/^\d+$/` in `normalizedExitCode` at head.

### Notes (non-blocking)

- Focus-switch to a large dormant child briefly shows compact state until the disk read lands — acceptable, covered by the new `TuiStateAndFocus` tests for correctness rather than latency.
- Active RUNNING streams (and the focused/root transcript) still grow unboundedly by design — this PR's win is largest once subagent sessions finish (WAITING/final → released), not while a large fleet is concurrently running.

Filing two follow-ups for failure modes this PR doesn't touch (idle-websocket unhandled-rejection crash on the 60-min Responses limit; unclassified HTTP/2 `UND_ERR_INFO` stream-timeout fetch failures) — unrelated to this diff.

## Verified

- Full diff (978 lines) and all 3 commits at head `5006b0014e`.
- `src/transcript/StreamLogStore.ts:255-420` (release/ensureLoaded/pendingRelease semantics predate this PR; queued-release honored at load completion).
- `packages/cli/src/chat/tui/state/subscribeStreamLog.ts`, `App.tsx`, `SubagentList.tsx`, `toolRenderers.tsx`, `childControls.ts`.
- `src/shared/toolUse.ts`, `src/shared/schemas/progressView/data.ts`, `packages/extension/src/progressView/frontend/formatters/logFormatters/toolFormatters.ts`.
- New/changed tests: `TuiStateAndFocus.vitest.mts`, `StreamLogStoreLoad.vitest.ts`, `NormalizeToolUse.vitest.ts`, `SubscribeStreamLogDispose.vitest.mts`.

PR #9029, opened 2026-07-20
fix: bound five unrelated memory-usage findings from an audit sweep

LionSR:
## Summary

A follow-up to the PR #9010 audit: ran a 10-agent memory-leak/waste sweep across every package **not** already covered by #9010 (desktop, extension webviews/host, trace-viewer, deeper agent/core/storage, tools/subprocess handling, CLI runtime, platform/eventBus, plus a cross-cutting "same data held twice" pass). All 18 candidate findings were independently re-verified against the real current code; 5 are fixed here. Each is independent, isolated, and changes no user-visible behavior.

- **`packages/desktop/src/renderer/main.ts`** — `wireRailTabs()`/`wireConversation()` bind DOM event listeners unconditionally. Normal startup wires once, but a bootstrap-recovery retry (its own failure re-showing the same fallback UI, whose button re-invokes recovery) could re-wire a second time on the same long-lived elements, stacking duplicate handlers. Added a guard matching the existing `shellWatcherInstalled` idempotency pattern already used a few lines away for the same recovery path.
- **`LogList.ts` / `webviewStorage.ts` / `streamLifecycleSlice.ts`** — every stream ever visited created a `logListState:<streamId>` entry in the shared webview-persisted-state blob, with no deletion path; the entry survived the stream's own DOM-cache eviction (LogList bounds that to 5) and even a full stream deletion. Added `delete()` to `StateStorage`/`PersistedState` (both backend and webview implementations) and wired the actual deletion into `DELETE_STREAM`/`DELETE_ALL` — the one point where a stream's toggle state genuinely has no future consumer.
- **`inlineCriticism.ts`** — `pushManualCriticism()` (backing the `diagnostics add` tool command) unconditionally appended every call to a per-file diagnostics array with no dedup and no cap; the array is only ever cleared by the whole-feature enable/disable toggle. A critique agent re-flagging the same line/message across repeated rounds would stack unbounded duplicate squiggles. Now skips re-adding an exact duplicate (same code/line/message).
- **`packages/trace-viewer/src/main.ts`** — the singlefile export mode's raw trace JSON sits on `window.__TEXRA_TRACE__` for the whole page lifetime after `parseTraceData` consumes it; nothing else reads it. Deleted after read.
- **`OutputFileProcessor.ts`** — `captureXmlSummary()` stored the model's extracted document text twice in the same `RoundOutput.xmlSummary` (bare in `tagContents[OUTPUT_DOCUMENTS_TAG]`, and again re-wrapped in XML tags in `documents`) — confirmed nothing downstream reads either field. Since `persistedFlow.ts` `structuredClone`s the entire flow shared state (which embeds every round's `xmlSummary`) on every single node step, this duplication's clone/serialize cost compounded across every round of every run. Stopped building the redundant `documents` copy; kept the one canonical `tagContents` representation.
- **`TextEditorTool.ts`** — `fileHistory` (undo history) is a process-wide singleton Map, keyed per execution+file, that only shrinks via explicit `undo_edit` calls. A long execution editing the same file many times without ever undoing retained every historical full-file snapshot forever. Capped the per-key array at 50 entries (oldest evicted first); `undo_edit` only ever pops the most recent entry, so this never changes that behavior.

## Investigated but not changed (reported as follow-ups)

- **`executionLease.ts`'s `fencedOwnershipKeys`** — genuinely unbounded (never pruned for abandoned/undurable executions that are never resumed), but the fence specifically exists to catch *lagging* async writers after abandonment; there's no safe "definitely done, nothing will reference this executionId again" moment to hook without either a TTL or reference-counting scheme. Fixing it risks reintroducing the exact cross-process race it guards against. Left as-is; needs a design decision, not a free fix.
- The cross-execution accumulation of `fileHistory` *keys* (as opposed to unbounded array length, fixed above) has the same "needs an execution-completion hook that doesn't exist yet" shape, but is now a much smaller residual per key.

## Validation

- `npx tsc --noEmit` — repo root, test-kernel, extension, trace-viewer, and desktop renderer tsconfigs all clean
- `npx eslint` on every changed/new file — clean
- `npx prettier --check` — clean
- New/extended tests: `TextEditorToolHistoryCap.vitest.ts` (new — cap behavior + undo still works), `ProgressViewMessageDispatcher.vitest.ts` (+2 — DELETE_STREAM/DELETE_ALL clear persisted LogList state)
- `npx vitest run` across `progressView`, `tools`, `agent/output`, `schemas/OutputSummaryTransforms`, `shared` — 172 files / 1290 tests passed

---

> [!NOTE]
> **Low Risk**
> Targeted retention and deduplication fixes with no intended behavior change; covered by new/extended tests for storage cleanup and TextEditor undo cap.
> 
> **Overview**
> Follow-up to a memory audit: **five independent fixes** that trim retained data without changing user-visible behavior.
> 
> **Desktop** — `wireRailTabs` / `wireConversation` now register DOM listeners only once, so bootstrap recovery retries cannot stack duplicate handlers on the same long-lived elements.
> 
> **Progress view storage** — `StateStorage` gains `delete()`; per-stream `logListState:*` keys are removed on `DELETE_STREAM` / `DELETE_ALL` via a shared `logListStateKey()` helper (tests added).
> 
> **Extension** — inline criticism from the diagnostics tool skips exact duplicate squiggles on the same line/message.
> 
> **Trace viewer** — drops `window.__TEXRA_TRACE__` after the inline trace is parsed so the raw JSON is not kept for the page lifetime.
> 
> **Agent / tools** — `captureXmlSummary` stops duplicating document text in `xmlSummary.documents` (only `tagContents` is populated, cutting clone cost in persisted flows). **TextEditorTool** caps per-file undo snapshots at 50 while preserving undo of the latest edit (new vitest).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3621a724891c91fa01fc4d5fa116bb4005f0fb7d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/renderer/main.ts
+++ b/packages/desktop/src/renderer/main.ts
@@ -703,7 +703,16 @@ window.addEventListener('message', (event) => {
 // Wire <stream-tabs> + <stream-conversation> events to shared handlers
 // =============================================================================
 
+// Each guard below protects its wiring function against double-registration:
+// a bootstrap recovery attempt that itself fails re-renders the same
+// fallback UI, whose button re-invokes recoverFromBootstrapFallback()
+// against these same module-level railTabs/conversationView elements, which
+// never get recreated or torn down for the life of the renderer.
+let railTabsWired = false;
+
 function wireRailTabs(): void {
+  if (railTabsWired) return;
+  railTabsWired = true;
   railTabs.addEventListener('stream-switch', ((e: CustomEvent) => {
     handleStreamSwitch(e);
     // Switching to a stream pulls the user out of the launcher view.
@@ -720,7 +729,11 @@ function wireRailTabs(): void {
   railTabs.addEventListener('delete-all', handleDeleteAll as EventListener);
 }
 
+let conversationWired = false;
+
 function wireConversation(): void {
+  if (conversationWired) return;
+  conversationWired = true;
   conversationView.addEventListener(
     'stream-switch',
     handleStreamSwitch as EventListener,
--- a/packages/extension/src/frontend/latex/inlineCriticism.ts
+++ b/packages/extension/src/frontend/latex/inlineCriticism.ts
@@ -178,6 +178,18 @@ export function pushManualCriticism(entry: ManualCriticismEntry): boolean {
   );
 
   const existing = collection.get(uri) ?? [];
+  // This list is only ever cleared by the whole-feature enable/disable
+  // toggle, so a critique agent re-flagging the same line/message across
+  // repeated rounds would otherwise stack an unbounded number of identical
+  // squiggles. Skip re-adding an exact duplicate instead.
+  const isDuplicate = existing.some(
+    (d) =>
+      d.code === CODE_TOOL &&
+      d.range.start.line === range.start.line &&
+      d.message === diag.message,
+  );
+  if (isDuplicate) return true;
+
   collection.set(uri, [...existing, diag]);
   return true;
 }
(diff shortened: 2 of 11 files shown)


LionSR:
are we over engineering and being pre optimization is the source of all evil here

PR #9122, opened 2026-07-24
feat: route workflow calls by model

LionSR:
## Summary

- add a declarative optional `model` field to every workflow-script `agent()` call
- route workflow-agent and structured tool-use calls through the existing delegation availability policy using that requested model
- include the normalized model in durable call identity, so changing a model invalidates the corresponding journal replay
- fail the workflow explicitly when a declared model is unavailable instead of allowing fan-out to hide the configuration error
- document per-call routing for inexpensive routine work and stronger difficult steps

This is stacked on #9121 so the model-routing change remains separate from launch-file support. After #9121 merges, this PR can be retargeted to `main` without changing its own commit.

## Design

Model choice remains data: `agent(prompt, { model: "..." })`. The script receives no imperative dispatcher or model registry. Omission preserves the established delegation policy, including parent-model inheritance and access-list fallback. Explicit values use the same availability check as ordinary delegation.

The field belongs to the common call options, so it applies uniformly to file-editing workflow agents and structured tool-use agents. It is normalized once at the sandbox boundary and participates automatically in the existing journal hash.

## Validation

- `npm run format`
- `npm run typecheck`
- `npm run lint`
- `npm run compile:fast`
- `npm test` — 7,413 passed, 4 skipped
- focused workflow engine and runner tests — 98 passed

---

> [!NOTE]
> **Medium Risk**
> Touches multi-agent delegation and model selection for orchestrated workflows; misconfiguration now aborts the whole script (intentionally), but scope is limited to the opt-in workflow-script path.
> 
> **Overview**
> Workflow scripts can now pass an optional **`model`** short name on each **`agent()`** call so routine steps can use a cheaper model and harder steps keep a stronger one; omitting it keeps the existing delegation policy.
> 
> The sandbox normalizes **`model`** like **`id`** (non-empty string), types and docs cover workflow and structured **`schema`** tool-use calls, and the production runner forwards **`requestedModel`** through **`selectAvailableDelegationModel`**. When a script **declares** a model that is unavailable, the run fails with **`WorkflowRunAbortError`** instead of resolving to **`null`** inside **`parallel()`** / **`pipeline()`**; undeclared selection failures behave as before.
> 
> Journal identity includes the normalized model so changing it invalidates replay for that call. Changelog and tool descriptions are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5984b56360e2bd9ea8b865af92708b193cb569a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Addressed the review observation about failure scope in commit 5984b56360. An explicitly declared but unavailable model remains a run-fatal workflow configuration error; when `model` is omitted, the runner now preserves the pre-existing delegation failure unchanged. A regression test covers the omitted-model path.

Issue #8546, opened 2026-07-16
Refactor: make stream-content sync a strict union


LionSR:
## Problem

`SYNC_STREAM_CONTENT` is a generic optional property bag that permits contradictory states.

`src/shared/ipc/progressView/outbound.ts` combines an optional action, an empty-string stream sentinel, workflow-only outputs, tool-use plan/todos/controls, and a generic category. The producer constructs those combinations with mutable locals and conditional spreads. The frontend then reconstructs intent with multiple `has*` booleans, kind guards, and casts in `packages/extension/src/progressView/frontend/state/slices/syncSlice.ts`.

Clear operations also send fake empty values, and `ProgressViewState.contextState` is initialized but never assigned while still being emitted. This is an invalid-state and dead-plumbing problem, not a naming problem.

## Correct ownership

The IPC schema should describe the actual states: clear or render, with render variants discriminated by stream kind. A pure projector should be the single producer of those variants.

## Proposed refactoring

1. Replace the property bag with strict `clear` and `render` variants.
2. Split render payloads by workflow/tool-use capabilities.
3. Add a pure `buildStreamContentSync` projector.
4. Switch the frontend reducer exhaustively over the union.
5. Remove dead `contextState` storage and emission if revalidation confirms it has no writer.

## Acceptance gates

- Schema tests reject contradictory field combinations.
- Clear needs no fake todos/plan/follow-up values.
- Frontend handling is exhaustive and contains no stream/draft casts.
- Progress backend sync, work-plan, process-output, and frontend state tests remain green.

Risk: medium because this changes a shared IPC contract across extension and desktop consumers.

PR #9172, opened 2026-07-25
refactor(progress): delete the hand-mirrored stream content projector

### LINKED ISSUE #9143: refactor(progressView): delete the hand-mirrored streamContentSync projector
> **Re-scoped 2026-07-25 against `bf2ae8f7f1` (post-#9167).** Half (b) of this issue —
> the inert `forceRebuild` flag — **shipped** in #9160 (`803fc357ff`, 4 files, +14/−33).
> Half (a) — deleting the hand-mirrored `streamContentSync` projector — is what remains,
> and it is now unblocked: #9167 (`bf2ae8f7f1`, merged 2026-07-25T07:56Z) landed parts 2
> and 3 of #9139, which was the only sequencing dependency. Every line number, count and
> LOC figure below was re-measured at `bf2ae8f7f1`; the original body's numbers were taken
> at `60908967ae` and are all stale. **This is now a one-PR item.**

## Problem

`src/shared/streams/streamContentSync.ts` (**109 lines** at `bf2ae8f7f1`) is a hand-written
flat mirror of a Zod-derived wire shape. It declares five interfaces plus a union alias
(`:20-58`) that restate `SYNC_STREAM_CONTENT`'s payload field-for-field, and
`buildStreamContentSync` (3 overloads + 1 implementation, `:61-109`) does nothing but
re-nest those flat fields into the schema's `outputs` / `workPlan` / `controls` groupings —
renaming `missingOutputs` → `missing` on the way through. No validation, no defaults, no
derivation, no class construction, no captured context.

**The governing rule is Zod-as-SSOT, not the factory rule.** The original body cited
AGENTS.md "Discouraged factory patterns" and quoted the justification list as *"meaningful
logic, class construction, or captured context"* — **omitting the first bullet**, which is
`AGENTS.md:526` "**Called from multiple locations (DRY)**". This function has two call
sites, so the factory rule on its own does **not** condemn it. That argument is withdrawn.

What does condemn it is `AGENTS.md:132-133` under "Schemas as the single source of truth":
*"Define schemas first, then derive TypeScript types using `z.infer<typeof Schema>`"* and
*"Use schema composition (`.extend()`, `.pick()`) instead of duplicating field
definitions."* `ActiveStreamContentSync`, `StreamContentSyncBase`,
`WorkflowStreamContentSync` and `ToolUseStreamContentSync` are duplicated field definitions
of `SyncStreamContentMessageSchema`, derived from nothing.

Its sibling in the same directory is the counter-example: `buildStreamMetadata`
(`src/shared/streams/streamMetadata.ts:17-21`) derives its input type from
`StreamMetadataSchema.partial().extend(...)` and applies real defaults, and its own
docstring says it is *"derived … rather than hand-duplicated, so it can't drift from the
wire schema."* `buildStreamContentSync` is the same author's pattern without the derivation.

## Verified symptom: 2 of 2 subsequent edits were pure mirror tax

The file was created by #8554 (`8cc3ec87a4`, 2026-07-16). `git log -- src/shared/streams/streamContentSync.ts`
returns exactly three commits. **Both** post-creation commits edited the same three files in
lockstep, and in both the mirror edit carried zero information — it only restated the schema
change:

| Commit | `outbound.ts` (schema) | `streamContentSync.ts` (mirror) | `syncSlice.ts` (reader) |
| --- | --- | --- | --- |
| `a1fa9fad4d` #9155 | ±12 | ±6 | ±6 |
| `bf2ae8f7f1` #9167 | −11 | −1 | −1 |

#9167's entire edit to this file is deleting `processes: ActiveChildInfo[];` from
`ActiveStreamContentSync.badges` — the same line it deleted from `outbound.ts` and read
from in `syncSlice.ts`. Three edits where two would do, twice in nine days.

**The compiler does not catch mirror drift — empirically verified.** The claim that the
projector's schema-derived return type pins the interfaces to the schema is false. Standalone
repro (`tsc --strict`, exit 0, no diagnostics): a `Mirror` interface carrying a stale
`processes` field, spread into a return value annotated as the wire type, compiles clean —
because the value reaches the returned literal through a variable spread
(`streamContentSync.ts:80`, `...(activeState ? { activeState } : {})`), and TypeScript does
not apply excess-property checking across a spread of a variable. So the mirror can and does
silently disagree with `outbound.ts`.

**Inlining makes it strictly stronger.** `WebviewUpdater.sendSyncStreamContent`
(`src/controllers/progressView/backend/WebviewUpdater.ts:347`) takes
`payload: SyncStreamContentPayload`. Passing an object literal directly there gets
contextual typing and excess-property checking against the Zod-derived discriminated union
at the call site — the check the hand mirror currently absorbs.

**The three overloads buy nothing.** `ProgressFactApplier` passes each narrowed result
straight into `sendSyncStreamContent`, which widens it back to `SyncStreamContentPayload`;
`replayTrace` assigns to `syncPayload` and spreads it into a `SyncStreamContentMessage`.
Neither consumer reads the narrowed type.

## Element accounting (recomputed at `bf2ae8f7f1`)

LOC figures are **measured, not estimated**: the inline was applied to scratchpad copies of
both call-site files and diffed.

| Element (counting unit) | Today | After | Δ |
| --- | ---: | ---: | ---: |
| Files in `src/shared/streams/` | 7 | 6 | **−1** |
| Hand-written type declarations mirroring the wire shape (`streamContentSync.ts:20-58`) | 6 (5 interfaces + 1 union alias) | 0 | **−6** |
| Signatures of `buildStreamContentSync` (`:61-72`) | 4 (3 overloads + 1 impl) | 0 | **−4** |
| Exported symbols across `streamContentSync.ts` + `outbound.ts` render-payload aliases | 5 | 1 | **−4** |
| Files edited in lockstep for one stream-content wire field change | 3 | 2 | **−1** |
| Edit sites to add one field to `outputs` (schema → mirror decl → mirror body → 2 producers → reader) | 6 | 4 | **−2** |
| Indirection hops between canonical state and the wire payload | 1 | 0 | **−1** |
| `missingOutputs` → `missing` rename hops on the send path | 1 | 0 | **−1** |
| **Grows:** call sites that name the `outputs`/`workPlan`/`controls`/`action:'render'` wire groupings | 4 (all in the projector) | 8 (4 per package × 2) | **+4** |
| **Grows:** production LOC at the two call sites (measured) | — | — | **+9** |
| Production LOC, net (measured) | — | — | **−108** |

Measured breakdown of the −108: `streamContentSync.ts` **−109** (whole file);
`ProgressFactApplier.ts` **+2** (+20/−18, including 2 new comment lines);
`replayTrace.ts` **+7** (+22/−15); `outbound.ts` **−8** (the alias collapse below).

The `+4` row is the honest cost and the reason this needed re-checking: the nesting is
written once today and twice after. It is outweighed because the *edit-site* metric — the
one that actually predicts maintenance — falls from 6 to 4, and because the 4 duplicated
groupings replace 6 interfaces restating 19 field names.

## Scope: what changes

One PR. Four files.

### Delete

**`src/shared/streams/streamContentSync.ts`, lines 1-109: the entire file.**
- `ClearStreamContentSync` (`:20-22`), `ActiveStreamContentSync` (`:24-31`),
  `StreamContentSyncBase` (`:33-37`), `WorkflowStreamContentSync` (`:39-44`),
  `ToolUseStreamContentSync` (`:46-55`)
- `export type StreamContentSyncSnapshot` (`:57-58`)
- `export function buildStreamContentSync` — 3 overload signatures (`:61-69`) and the
  implementation (`:70-109`)

**`src/shared/schemas/progressView/outbound.ts:347-358`** — three exported aliases collapse
to one:

```ts
export type StreamContentRenderPayload = Extract<
  SyncStreamContentPayload,
  { action: 'render' }
>;
```

This deletes `WorkflowStreamContentPayload` and `ToolUseStreamContentPayload`, whose **only**
cross-file consumer is the deleted projector (verified: `grep -rn` for both names over
`src packages` returns hits in exactly `streamContentSync.ts` and `outbound.ts` itself).
`StreamContentRenderPayload` is kept — `packages/extension/src/progressView/frontend/slices/syncSlice.ts:7,16`
imports it, and the collapsed `Extract` is type-identical to the union it replaces.
`AgentCategory` stays imported in `outbound.ts` (still used at `:307` and `:317`).

### Edit

**`src/controllers/progressView/backend/events/ProgressFactApplier.ts`** (measured +20/−18)
- `:42` — drop `import { buildStreamContentSync } from '@shared/streams/streamContentSync';`
- `:674-676` — `sendSyncStreamContent(buildStreamContentSync({ action: 'clear' }))` →
  `sendSyncStreamContent({ action: 'clear' })`
- `:695-699` — `shared` gains `action: 'render' as const`, and `activeState,` **must become
  the conditional spread** `...(activeState ? { activeState } : {})` (see Blockers)
- `:701-711` — workflow arm builds the wire literal directly:
  `{ ...shared, kind, outputs: { files: …getOutputFiles(stream), missing: …getMissingOutputs(stream), compileFailures: …getCompileFailures(stream) } }`
- `:716-733` — tool-use arm builds
  `{ ...shared, kind, workPlan: { todos, plan, queuedFollowUps }, controls: { toolEditBypass, superYoloBypass, goal } }`.
  Note the local `const controls = this.getStreamControls(stream)` at `:715` shadows nothing —
  `controls: { toolEditBypass: controls.toolEditBypass, … }` is fine.
- Leave `toActiveStreamContentSync` (`:736-746`) alone; it is still the `activeState` producer.

**`packages/trace-viewer/src/replayTrace.ts`** (measured +22/−15)
- `:14` — drop the `buildStreamContentSync` import
- add `type SyncStreamContentPayload` to the existing `@shared/schemas` import block (`:2-11`)
- `:195-217` — `syncSnapshot` gains `action: 'render' as const`; **annotate**
  `const syncPayload: SyncStreamContentPayload = …` and replace the two
  `buildStreamContentSync({…})` calls with the wire literals. The annotation is **required**,
  not stylistic: without a contextual type, `goal: { active: false }` widens `false` to
  `boolean` and no longer matches `GoalStateSchema`'s discriminated union
  (`src/shared/schemas/goal.ts:46`).

## Blockers and traps

1. **Get an explicit maintainer ruling before starting.** This is a partial revert of
   #8554 (`8cc3ec87a4`, 2026-07-16, closing #8546), whose acceptance gate read "Add a pure
   `buildStreamContentSync` projector" — nine days old. The argument for the PR description:
   **#8546's actual goal — strict, exhaustive, kind-specific transport variants — was
   delivered by the Zod half of that change** (`outbound.ts:279-338`: `StreamContentRenderFields`,
   three `strictObject` variants, two discriminated unions, `WithoutCommand`-derived payload
   types) and survives untouched. The projector was scaffolding, and #9155 and #9167 have
   since shown it costs a lockstep edit per schema change while catching no drift.

2. **The knip ratchet trap.** Deleting the projector orphans `WorkflowStreamContentPayload`
   and `ToolUseStreamContentPayload`; knip reports them as new `types` findings and
   `config/ratchets/knip-baseline.json` (18 findings) has no `progressView` entries, so
   `npm run check:dead-code-ratchet` fails. The original filer reproduced this and reported a
   non-obvious subtlety: keeping the aliases and merely *using them internally* inside
   `outbound.ts` does **not** clear the finding. The `Extract<SyncStreamContentPayload,
   { action: 'render' }>` collapse above is the verified remedy. Consistent with `knip.json`
   (root workspace `project: src/**/*.{ts,mts}`, so `outbound.ts` is analyzed).
   **Re-verify at implementation time; do not "fix" this by adding baseline entries.**

3. **Preserve the conditional `activeState` spread.** `ProgressFactApplier.ts:695-699`
   currently builds `shared` with `activeState` present-but-possibly-`undefined`, and
   `streamContentSync.ts:80` converts that to `...(activeState ? { activeState } : {})`, so
   the key is *omitted* rather than sent as `undefined`. The desktop revalidates every
   outbound message through `ProgressViewOutboundMessageSchema` over structured clone
   (`packages/desktop/src/main/hostBridge.ts:48`, `packages/desktop/src/renderer/main.ts:241`),
   which preserves `undefined`. Reproduce the spread exactly — a naive inline that keeps
   `activeState,` changes what goes on the wire.

4. **`packages/trace-viewer` is a separate workspace and is *not* in `knip.json`'s
   `workspaces`** (`['.', 'packages/extension', 'packages/cli', 'packages/desktop']`). knip
   will not catch a dangling import there — only `pnpm --filter @texra/trace-viewer run
   typecheck` will. Run the full six-tsconfig `npm run typecheck`, not just root `tsc`.

5. **Tests: zero edits expected.** No test imports `buildStreamContentSync` (verified by grep
   over `src/test-kernel`). Two suites are the behavioral guards and **must stay green
   unmodified** — if either goes red the inline changed behavior:
   - `src/test-kernel/progressView/StreamContentSync.vitest.ts` — drives
     `ProgressFactApplier.syncStreamContent` and asserts the emitted `SyncStreamContentPayload`.
   - `src/test-kernel/traceViewer/replayTrace.vitest.ts` — exercises the real dispatcher + slices.
   `src/test-kernel/desktop/DesktopAgentExecution.vitest.mts` asserts `activeState` payload
   shape and is a third guard.

## Considered and rejected

- **Keep the projector but derive its input types from the schema** (the `buildStreamMetadata`
  pattern), e.g. `type WorkflowStreamContentSync = Omit<WorkflowStreamContentPayload, 'action' | 'outputs'> & WorkflowStreamContentPayload['outputs']`.
  This fixes the SSOT violation and avoids the `+4` duplication, but it only nets ≈ −30 LOC,
  keeps the file, and leaves a zero-logic re-nesting hop whose sole remaining job is undoing
  the `missing` → `missingOutputs` rename that the derivation itself deletes. Strictly
  dominated by the inline.
- **Deleting the `activeState` channel** (`outbound.ts:295-307`,
  `ProgressFactApplier.toActiveStreamContentSync`, the `includeActiveState` option,
  `syncSlice.activeStateFields`). It is not duplication, it is the repair path: incremental
  round-stage/progress/badge sends are gated on the stream being active, so a backgrounded
  parent's badges go stale and `activeState` is the only carrier that repairs them on tab
  activation — documented at `outbound.ts:295-296` ("Active state is all-or-nothing").
  Removing it would require *adding* a per-tab `UPDATE_STREAM_METADATA` send.
- **Collapsing `StreamMetadata` into `activeState`** — two different repair scopes (tab list
  vs active tab), not two spellings of one record.
- **Moving `syncFullView` / `setActiveStream` / `revealStream` into `ProgressBackend`** — the
  premise is false; `ProgressViewProvider.syncFullView` reads `vscode.window.activeColorTheme`
  and calls `vscode.window.showErrorMessage`, and `src/controllers` is a lint-enforced
  VS Code-free zone, so the move deletes 3 port members and adds 4.
- **`docs/prds/2026-02-21-progress-view-placement.md:412` and
  `docs/prds/2026-02-21-prd-progressview-streaming-cleanup.md:336-338`** still describe the
  now-deleted `syncFullView({ forceRebuild: true })` / `_pendingUpdateOptions`. Dated PRDs are
  point-in-time records and `docs/prds` is outside `check:guidance-refs` scope. Leave them.

## Acceptance criteria

- [ ] `src/shared/streams/streamContentSync.ts` no longer exists.
- [ ] `grep -rn "buildStreamContentSync\|StreamContentSyncSnapshot\|streamContentSync" --include='*.ts' --include='*.tsx' --include='*.md' . ` (excluding `node_modules`, `dist`, `out`) returns **zero** matches.
- [ ] `src/shared/schemas/progressView/outbound.ts` exports exactly two stream-content payload types: `SyncStreamContentPayload` and `StreamContentRenderPayload`.
- [ ] `packages/trace-viewer/src/replayTrace.ts` annotates `syncPayload` with `SyncStreamContentPayload`.
- [ ] `ProgressFactApplier.syncStreamContent` emits `activeState` as an **omitted key** (not `undefined`) when `includeActiveState` is false.
- [ ] `StreamContentSync.vitest.ts`, `replayTrace.vitest.ts` and `DesktopAgentExecution.vitest.mts` pass **without modification**.
- [ ] `config/ratchets/knip-baseline.json` is unchanged (18 findings) and `npm run check:dead-code-ratchet` passes.
- [ ] Net production-line change is negative (measured target **≈ −108**); state the actual `git diff --stat` figure in the PR description.
- [ ] The PR description states why #8546's acceptance criterion is being reverted (blocker 1).

## Verification

```bash
npm run typecheck              # six tsconfigs: root, tsconfig.test-kernel.json, texra,
                               # @texra-ai/cli, @texra/trace-viewer, @texra/desktop.
                               # @texra/trace-viewer is the ONLY gate on replayTrace.ts —
                               # knip does not analyze that workspace.
npm test
npm run lint                   # includes the src/controllers VS Code-free-zone rule
```

Targeted while iterating:

```bash
npx vitest run --config vitest.config.mjs \
  src/test-kernel/progressView/StreamContentSync.vitest.ts \
  src/test-kernel/traceViewer/replayTrace.vitest.ts \
  src/test-kernel/desktop/DesktopAgentExecution.vitest.mts
npx knip --reporter json | python3 -c "import json,sys; [print(i['file'],c,[x['name'] for x in i[c]]) for i in json.load(sys.stdin)['issues'] for c in ('exports','types','files') if i.get(c)]"
```

Manual smoke (both hosts): open the progress view, switch between a workflow tab and a
tool-use tab, confirm output files / missing-output list / compile failures render on the
workflow tab and todos / plan / queued follow-ups / YOLO toggles / goal banner render on the
tool-use tab, and confirm subagent badges still repair on tab activation for a backgrounded
parent. Then open a trace in the standalone trace viewer and confirm the same panels populate.

---

## Changelog of this issue

- **2026-07-25** — Re-verified at `bf2ae8f7f1`. Half (b) (`forceRebuild`) removed from scope:
  landed in #9160; `grep -rn "forceRebuild\|_pendingUpdateOptions" src packages` now returns
  zero, and `ProgressBackendLifecycleOptions.rebuildRenderedStreams` is already
  `(options: { syncActiveStream: boolean })` with both call sites passing it
  (`ProgressBackend.ts:53`, `:213`, `:294`). Sequencing blocker on the dead-process wire PR
  removed: #9167 merged. File length corrected 112 → **109**. Withdrew the
  "Discouraged factory patterns" argument (it misquoted `AGENTS.md:526`, which lists
  "Called from multiple locations (DRY)" as a justification, and this function has two
  callers); replaced it with the Zod-as-SSOT rule plus an empirical drift proof and the
  2-of-2 lockstep-edit history. LOC figures re-measured by applying the inline rather than
  estimated (−115 → **−108**). Now a one-PR item.

LionSR:
## Summary
- delete the hand-written `streamContentSync.ts` mirror and inline the schema-shaped payload literals at its two producer sites
- preserve `activeState` omission semantics with the existing conditional spread
- collapse the orphaned workflow/tool-use render aliases into one schema-derived `StreamContentRenderPayload`

This intentionally retires the projector acceptance criterion from #8546. That issue's actual correctness goal—strict, exhaustive, kind-specific transport variants—remains owned by the Zod schemas and derived `SyncStreamContentPayload` union in `outbound.ts`. The projector only mirrored those fields, did no validation or derivation, and incurred lockstep edits on both subsequent schema changes while TypeScript could not detect stale fields hidden behind its variable spread.

## Net elements (R6)
- 4 production files changed: 42 insertions, 152 deletions, net **−110 lines**
- deletes 1 file, 6 hand-written mirror declarations, 4 function signatures, and 4 exported symbols
- adds no new files or exported symbols
- tests and the knip baseline are unchanged

## Consumer counts (R8)
- `SyncStreamContentPayload`: remains the schema-derived payload SSOT and now contextually checks trace replay directly
- `StreamContentRenderPayload`: one frontend consumer; now extracted directly from `SyncStreamContentPayload`
- `buildStreamContentSync`, `StreamContentSyncSnapshot`, and `streamContentSync`: zero remaining references

## Validation
- targeted unchanged behavioral suites: 3 files, 91 tests passed
- full suite: 774 files / 7,521 tests passed; 1 file / 4 tests skipped
- full six-project typecheck passed
- lint passed
- dead-code ratchet passed (18 findings, all baselined; baseline unchanged)
- Prettier and `git diff --check` passed
- independent code review: no findings

Closes #9143

---

> [!NOTE]
> **Low Risk**
> Refactor-only: same wire shapes and omission semantics, with types enforced by existing Zod schemas at producers.
> 
> **Overview**
> Removes the hand-maintained `streamContentSync` projector and builds **`SYNC_STREAM_CONTENT`** payloads directly at the two producer sites (`ProgressFactApplier.syncStreamContent` and trace `replayTrace`).
> 
> Payloads now match the Zod transport shape in place: workflow renders use nested **`outputs`** (`files`, `missing`, `compileFailures`); tool-use renders use **`workPlan`** and **`controls`** instead of flat top-level fields. Clear actions send `{ action: 'clear' }` without a helper; **`activeState`** is still omitted unless included via the existing conditional spread.
> 
> Type exports in **`outbound.ts`** drop the separate workflow/tool-use render aliases in favor of a single schema-derived **`StreamContentRenderPayload`** (`action: 'render'` on **`SyncStreamContentPayload`**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5cffc57c1aea63385cc446f393dd540b3302da3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
I independently implemented #9143 in parallel (#9174, now closed as a duplicate — our diffs were byte-identical apart from an import-ordering nit). Two things I verified empirically while doing it are worth recording here, because both correct the **issue body**, not this PR's code. This PR is correct as merged; nothing below asks for a change.

**1. Trap 3 in #9143 is overstated — `activeState: undefined` does not break the desktop revalidation.**

The issue asserted that a naive inline keeping `activeState,` "changes what goes on the wire" in a way the desktop's `ProgressViewOutboundMessageSchema` revalidation would catch. I probed it directly against current `main`:

```
clone keeps key:            true
parse with undefined key:   true
parse with omitted key:     true
```

`structuredClone` does preserve the explicit `undefined` key, but `activeState` is declared `.optional()`, so Zod accepts it either way. The frontend consumer (`syncSlice.activeStateFields`) tests `if (!data.activeState) return {}`, which handles both, and the VS Code webview path JSON-serializes, which drops the key anyway.

So the conditional spread is **wire hygiene and faithfulness to the deleted projector's behavior, not a correctness cliff**. Keeping it (as this PR does) is still the right call — it's what the projector did, and sending phantom `undefined` keys across three hosts is worth avoiding — but a future reader shouldn't believe there's a schema guard behind it, because there isn't. I nearly committed a code comment claiming there was, which is why I checked.

**2. The `SyncStreamContentPayload` annotation on `syncPayload` genuinely is load-bearing — and it self-enforces.**

Confirmed by deleting it and running `pnpm --filter @texra/trace-viewer typecheck`:

```
src/replayTrace.ts(229,9): error TS2322: ...
  The types of 'controls.goal' are incompatible between these types.
    Type '{ active: boolean; }' is not assignable to type '{ active: false; } | { active: true; status: ...; objective: string; }'
```

Since `tsc` fails loudly, this needs no protective comment — the build is the guard. (This is also the one workspace knip doesn't analyze, so the six-tsconfig `typecheck` is the only thing standing behind it.)

**One real gap, pre-existing and not introduced here.**

`StreamContentSync.vitest.ts` asserts the emitted payload with `toMatchObject`, which passes whether `activeState` is absent *or* present-and-`undefined`, and it only exercises the `includeActiveState: true` path. So the omission behavior both this PR and the issue treat as important is not actually pinned by any test. I verified it holds on merged `main` with a throwaway probe — `Object.hasOwn(payload, 'activeState') === false` with `includeActiveState` false, still absent after `structuredClone`, and the full message parses clean — but nothing would catch a future regression.

Given finding 1, that gap is lower-stakes than it first looks (no host actually misbehaves on an `undefined` key). I'd file it as a small test-hardening issue rather than anything urgent — happy to do that if a maintainer wants it tracked.

PR #9174, opened 2026-07-25
refactor(progressView): delete the hand-mirrored streamContentSync projector

### LINKED ISSUE #9143: refactor(progressView): delete the hand-mirrored streamContentSync projector

LionSR:
## Summary

Deletes `src/shared/streams/streamContentSync.ts` (109 lines) and inlines the projection into its two production call sites.

The file was a flat, hand-written mirror of a Zod-derived wire shape: five interfaces plus a union alias restating `SYNC_STREAM_CONTENT`'s payload field-for-field, and `buildStreamContentSync` (3 overloads + 1 impl) that only re-nested those flat fields into the schema's `outputs` / `workPlan` / `controls` groupings, renaming `missingOutputs` → `missing` on the way. No validation, no defaults, no derivation, no class construction, no captured context.

**Why it goes (and why this reverts one of #8546's acceptance criteria).** This is a partial revert of #8554 (`8cc3ec87a4`, 2026-07-16), whose gate read "Add a pure `buildStreamContentSync` projector". **#8546's actual goal — strict, exhaustive, kind-specific transport variants — was delivered by the Zod half of that change and survives untouched**: `outbound.ts` still has `StreamContentRenderFields`, the three `strictObject` variants, both discriminated unions, and the `WithoutCommand`-derived payload type. The projector was scaffolding on top of that.

The governing rule for the projector is Zod-as-SSOT (AGENTS.md "Schemas as the single source of truth": derive types with `z.infer`, compose with `.extend()`/`.pick()` instead of duplicating field definitions), **not** the factory rule — the function has two callers, so the factory rule alone would not condemn it. Its sibling in the same directory is the counter-example: `buildStreamMetadata` derives its input type from `StreamMetadataSchema.partial().extend(...)` and its own docstring says it is "derived … rather than hand-duplicated, so it can't drift from the wire schema."

**The mirror caught no drift and cost a lockstep edit per schema change.** Both post-creation commits (#9155, #9167) touched `outbound.ts`, the mirror, and `syncSlice.ts` together, and in both the mirror edit carried zero information — #9167's entire edit to the file was deleting the same `processes` line it deleted from the schema.

**Inlining strengthens typing rather than weakening it.** `WebviewUpdater.sendSyncStreamContent(payload: SyncStreamContentPayload)` gives the inlined object literals contextual typing and excess-property checking against the Zod-derived discriminated union — the check the hand mirror absorbed, because TypeScript does not excess-property-check across a spread of a variable (exactly the shape at the deleted `streamContentSync.ts:80`). I verified that empirically rather than assuming it: adding a bogus `bogusStaleField: 1` to the inlined `outputs` literal in `ProgressFactApplier` now fails `tsc` with `TS2353` — precisely the class of drift the old mirror let compile clean.

## Issue acceptance gates

Closes #9143. Every gate in the issue body is satisfied:

- [x] `src/shared/streams/streamContentSync.ts` no longer exists.
- [x] `grep -rn "buildStreamContentSync\|StreamContentSyncSnapshot\|streamContentSync"` over `*.ts`/`*.tsx`/`*.mts`/`*.md` (excluding `node_modules`/`dist`/`out`) returns **0** matches.
- [x] `outbound.ts` exports exactly two stream-content payload types: `SyncStreamContentPayload` and `StreamContentRenderPayload`.
- [x] `replayTrace.ts` annotates `syncPayload` with `SyncStreamContentPayload`.
- [x] `ProgressFactApplier.syncStreamContent` emits `activeState` as an omitted key (not `undefined`) when `includeActiveState` is false — verified empirically, see Validation.
- [x] `StreamContentSync.vitest.ts`, `replayTrace.vitest.ts`, `DesktopAgentExecution.vitest.mts` pass **without modification** (91 tests).
- [x] `config/ratchets/knip-baseline.json` unchanged (18 findings); `check:dead-code-ratchet` passes.
- [x] Net production-line change is negative: **−104** (issue's prototype measured −108; delta explained below).
- [x] PR description states why #8546's acceptance criterion is being reverted (above).

Half (b) of the original issue (the inert `forceRebuild` flag) was already shipped by #9160 and is correctly out of scope here — re-verified: `grep -rn "forceRebuild\|_pendingUpdateOptions" src packages` returns zero.

## Single source of truth

`SyncStreamContentMessageSchema` (`src/shared/schemas/progressView/outbound.ts`) is now the sole owner of the `SYNC_STREAM_CONTENT` wire shape. Before this PR that fact had two owners — the schema and the hand-written interfaces in `streamContentSync.ts` — with nothing pinning them together, since TypeScript does not excess-property-check across a variable spread. Producers now write object literals checked directly against the schema-derived type at the `sendSyncStreamContent` boundary.

**Removed:** six hand-written type declarations (5 interfaces + 1 union alias) restating 19 field names already declared in Zod; four `buildStreamContentSync` signatures (3 overloads + 1 impl); two type aliases whose only cross-file consumer was the projector; one zero-logic indirection hop; one `missingOutputs` → `missing` rename hop on the send path.

**Added:** no new abstraction. The honest cost is that the `outputs` / `workPlan` / `controls` / `action: 'render'` groupings are now named at 2 call sites instead of 1 (the issue's `+4` row). That is outweighed because the edit-site count for adding one field falls from 6 to 4 (schema → mirror decl → mirror body → producer → reader collapses to schema → producer → reader), and because 4 duplicated groupings replace 6 interfaces restating 19 field names.

The three overloads were not recreated: neither caller read the narrowed type — `ProgressFactApplier` passed each result straight into `sendSyncStreamContent`, which widens back to `SyncStreamContentPayload`, and `replayTrace` spread it into a `SyncStreamContentMessage`.

## Net elements (R6)

Diffstat (`git diff --numstat` against `origin/main` at `bf2ae8f7f1`):

```
 25	15	packages/trace-viewer/src/replayTrace.ts
 21	18	src/controllers/progressView/backend/events/ProgressFactApplier.ts
  2	10	src/shared/schemas/progressView/outbound.ts
  0	109	src/shared/streams/streamContentSync.ts
 4 files changed, 48 insertions(+), 152 deletions(-)
```

| Element | Before | After | Δ |
| --- | ---: | ---: | ---: |
| Files | 4 | 3 | **−1** |
| Net production LoC | — | — | **−104** |
| Exported symbols (`^[+-]export` over the diff: 4 `buildStreamContentSync` signatures + `StreamContentSyncSnapshot` + `WorkflowStreamContentPayload` + `ToolUseStreamContentPayload` + `StreamContentRenderPayload` removed; `StreamContentRenderPayload` re-added) | 8 | 1 | **−7** |
| Interfaces | 5 | 0 | **−5** |
| Type aliases | 4 | 1 | **−3** |
| Functions | 1 (4 signatures) | 0 | **−4 signatures** |
| Classes / enums | 0 | 0 | 0 |
| Files edited in lockstep for one stream-content wire field change | 3 | 2 | **−1** |

**On the −104 vs the issue's measured −108:** the −4 delta is 4 lines of explanatory comment I deliberately added at the two spots that are easy to "simplify" back into bugs — the `activeState` conditional spread and the `syncPayload` annotation. Both are non-obvious correctness constraints, so I'd rather carry the 4 lines than hit the predicted number.

## Consumer counts (R8)

No emit path changed. Export deletions, each grepped over `src packages` (`*.ts`/`*.tsx`/`*.mts`) before removal:

| Deleted export | Cross-file consumers | Disposition |
| --- | ---: | --- |
| `buildStreamContentSync` | 2 (`ProgressFactApplier.ts`, `replayTrace.ts`) | both inlined |
| `StreamContentSyncSnapshot` | 0 | deleted |
| `WorkflowStreamContentPayload` | 1 (`streamContentSync.ts` only) | orphaned by the deletion → collapsed |
| `ToolUseStreamContentPayload` | 1 (`streamContentSync.ts` only) | orphaned by the deletion → collapsed |

`StreamContentRenderPayload` is **kept** — `packages/extension/src/progressView/frontend/slices/syncSlice.ts:7,16` consumes it — and redefined as the type-identical `Extract<SyncStreamContentPayload, { action: 'render' }>`. `syncSlice` narrows on `data.kind` itself, so it loses nothing. `AgentCategory` stays imported in `outbound.ts`; still used at `:307` / `:317`.

Per the issue's verified trap, keeping the two aliases and merely using them internally inside `outbound.ts` does **not** clear the knip finding; the `Extract` collapse is the remedy, and `config/ratchets/knip-baseline.json` is untouched.

## Host impact

**Extension:** none observable. `ProgressFactApplier` emits byte-identical payloads; `syncSlice.ts` is unchanged.

**Desktop:** none observable, but this is the host the change is most delicate for. The desktop revalidates every outbound message through `ProgressViewOutboundMessageSchema` over a structured clone (`packages/desktop/src/main/hostBridge.ts:48`, `packages/desktop/src/renderer/main.ts:241`), which preserves `undefined` — so the conditional `activeState` spread is reproduced exactly and commented in place. `DesktopAgentExecution.vitest.mts` passes unmodified.

**CLI:** none — no CLI code path touches this wire message.

The standalone trace viewer (`packages/trace-viewer`) is affected and is the one workspace knip does not analyze, so it is gated only by the six-tsconfig `npm run typecheck`, which was run in full.

## Scheduled refactoring

None required for this change. One optional follow-up worth a maintainer's call, surfaced below.

## Validation

All four gates green locally on this branch, and all CI checks pass:

- `npm run typecheck` — all six tsconfigs (root, test-kernel, `texra`, `@texra-ai/cli`, `@texra/trace-viewer`, `@texra/desktop`)
- `npm test` — 774 files / 7521 tests passed, 1 file + 4 tests skipped. Nothing flaky, nothing modified, no reruns needed.
- `npm run lint` — clean, including the `src/controllers` VS Code-free-zone rule
- `npm run check:dead-code-ratchet` — 18 current vs 18 baselined, `unusedTypes=0`
- `npx prettier --check` on all three edited files

Targeted guards, run separately and passing **unmodified**: `StreamContentSync.vitest.ts`, `replayTrace.vitest.ts`, `DesktopAgentExecution.vitest.mts` (91 tests).

Two claims in this PR were verified empirically rather than asserted:

1. **Excess-property checking really does apply at the inlined call sites.** Temporarily adding `bogusStaleField: 1` to the inlined `outputs` literal fails with `TS2353: Object literal may only specify known properties`. Reverted before commit.
2. **`activeState` is an omitted key, not `undefined`.** With `includeActiveState` false: `Object.hasOwn(payload, 'activeState') === false`, the key is still absent after `structuredClone`, and the full message parses clean through `ProgressViewOutboundMessageSchema`.

### One thing worth a reviewer's eye

While writing verification (2) I found the existing guard does not actually cover it — `StreamContentSync.vitest.ts` asserts with `toMatchObject`, which passes whether the key is absent or present-and-`undefined`, and it only exercises the `includeActiveState: true` path. So I verified it with a throwaway probe rather than a committed test, to keep the diff the pure deletion the issue scoped.

The coverage gap predates this PR, but this PR is exactly the kind of edit that could have silently broken it. Happy to land that probe as a real regression test in a follow-up if you'd like the behavior pinned.

Manual smoke across both hosts (progress view tab switching, trace viewer panels) was **not** run — the change is a mechanical inline with byte-identical output verified by the three behavioral guards plus schema revalidation, but flagging it since the issue listed it.


LionSR:
Closing as a duplicate of #9172, which merged this same work at `3cf91f2ba9` and closed #9143.

Timing: #9172 was opened at 08:12:46Z and merged at 08:21:48Z; I opened this PR at 08:21:17Z, 31 seconds before that merge, having branched off `bf2ae8f7f1` before #9172 existed. My mistake was not re-checking for an open PR against the issue immediately before pushing — the branch-time check was clean, but that's the wrong moment to check.

For the record, the two implementations are byte-identical apart from an import ordering nit and two explanatory comments I'd added. I diffed my branch against merged `main`: `outbound.ts` and the file deletion are character-for-character the same, and the two call sites differ only by those comments. Nothing here is worth salvaging into a follow-up — I checked, and one of my two comments was actually wrong (details below).

Two findings from verifying this work are worth recording, since they correct the issue body rather than the merged code. Posting them on #9172.

PR #9158, opened 2026-07-25
refactor(runtime): retire the dead ProcessExecutionHandle pipeline

LionSR:
**Part 1 of 3 for #9139** — the runtime + tools layer. Closes nothing; the issue
closes with PR 3.

## What this deletes

`ProcessExecutionHandle` had **zero production constructions** at HEAD: since
`3b6483e6e5` (#2967) a background bash is a normal child stream, and
`createChildStream` builds an `AgentExecutionHandle`. Everything hanging off that
constructor was therefore structurally unreachable — and the dead branches kept
shipping bugs (#8155, where `killBackgroundProcesses()` swept only the type that
never exists).

| Deleted | Where |
| --- | --- |
| `ProcessExecutionHandle` class; `\| 'process'` on `ExecutionHandle.category` | `ExecutionHandle.ts` |
| `isChildExecution`'s "not an agent handle ⇒ true" fallback | `ExecutionHandle.ts` |
| `ProcessOutputPoller.ts` (310 lines) + its test (248 lines) | whole files |
| `executions/processOutputFormat.ts` (20 lines) | whole file |
| all 5 `instanceof ProcessExecutionHandle` branches; `getActiveProcessOutputSnapshots`; the `processOutput` field, ctor param and `dispose()` call; the `process.output` emitter wiring in `attachSessionEvents` | `executionRegistry.ts` |
| `getActiveChildren()`'s always-empty `processes` key (now returns `ActiveChildInfo[]`); `emitChildActivity`'s `kind` param; `collectChildSummary`'s `ctor` param; `detachActiveChildren`'s now-unused `visited` param | `executionRegistry.ts` |
| the dead `/executions/{id}/output` implementation, the structurally-always-zero `bgCount`, and the endpoint's two tool-description lines | `ExecutionsTool.ts` |
| `formatListingHeader`'s unreachable `", N background processes running"` suffix | `executions/summaryFormat.ts` |
| `formatPostCompactionContext`'s `processes` parameter and the always-empty `<active-background-bash>` block | `subagentResults.ts` |
| `replayActiveProcessOutput()` and the `processes` child-activity seed | `desktopAgentExecution.ts` |

`StreamProcessOutput` in `streamMetaReducer.ts` lost its last cross-module
consumer, so it is now file-local (`config/ratchets/knip-baseline.json` is
unchanged — nothing was baselined).

## Element accounting vs the issue's table

| Row | Target | This PR |
| --- | --- | --- |
[...]

Issue #9139, opened 2026-07-24
refactor: retire the dead ProcessExecutionHandle background-process pipeline


LionSR:
## Problem

A background OS process used to be modelled as a `ProcessExecutionHandle` whose stdout/stderr were
tailed out of temp files by a `ProcessOutputPoller`. Commit `3b6483e6e5` ("feat: stream background
bash in child tabs", #2967) replaced that with a normal child stream: `src/tools/bash.ts:275-286`
calls `createChildStream(..., { runKind: 'process', toolName: 'bash' })`, and
`src/tools/childStream.ts:157-164` builds an **`AgentExecutionHandle`** (category
`AgentCategory.ToolUse`). Nothing from the old model was deleted.

At HEAD, `new ProcessExecutionHandle(...)` appears in exactly five places, **all of them test
files**:

```
src/test-kernel/desktop/DesktopAgentExecution.vitest.mts:1532
src/test-kernel/desktop/DesktopAgentExecution.vitest.mts:3887
src/test-kernel/agent/runtime/ProcessOutputPoller.vitest.ts:75
src/test-kernel/agent/runtime/ExecutionRegistry.vitest.ts:215
src/test-kernel/agent/runtime/ExecutionRegistry.vitest.ts:290
```

Zero production constructions. Everything downstream of that constructor is therefore structurally
unreachable: the poller is never handed a source, the `process.output` trace event and the
`child.activity{kind:'processes'}` arm are never emitted, the `UPDATE_PROCESS_OUTPUT` wire message
is never sent, the webview's `processOutputs` map is permanently empty, and the CLI TUI's
`StreamSlice.activeProcesses` is permanently `[]`. Six dedicated production files (1,298 lines) plus
process-only branches in ~45 more files across five packages exist to serve a code path that no
execution can reach.

Because the dead surface looks alive, the repo keeps investing in it and keeps getting bitten by it.

## Verified symptom today

**1. It has already shipped one bug of exactly this class (#8155).**
`ExecutionRegistry.killBackgroundProcesses()` swept only `instanceof ProcessExecutionHandle` handles
— of which there are none — so background bash survived host shutdown. The fix bolted an
`AgentExecutionHandle` branch on underneath and left the dead branch on top, plus a 13-line doc
comment whose entire job is to warn the reader that the obvious branch is the wrong one:

`src/agent/runtime/executionRegistry.ts:548-569`
```ts
  /**
   * Kill only background OS processes (bash, codex) ...
[...]

Issue #9138, opened 2026-07-24
Tracking: macro debt tournament (top 6) + workflow-script progress UI


LionSR:
Umbrella for a macroscopic tech-debt tournament over the workflow-script system, delegation, agent runtime, progress-view backend and frontend, CLI TUI, wire contracts, and persistence/desktop — plus the workflow-script progress-UI work that came out of reviewing what those hosts actually render.

## The governing constraint

Every child issue must **reduce elements**. Each one carries an explicit element-accounting table (files, exported symbols, types, schemas, state fields, branches — before → after). Where an item does not reduce, it says so plainly and justifies it in one line rather than hiding it.

This matters because this repo has measured the opposite happening: a prior calibration audit found 22 self-described "reduction" PRs net-**added** 5,046 production lines. Framing is not evidence.

## How these were selected

Nine parallel scouts mapped the areas above and produced 25 raw candidates, merged to 10 with every citation re-verified against current HEAD. Each of the 10 then faced two hostile refuters with distinct lenses — deletion-enumeration completeness, and symptom reality plus net-element honesty — aimed at this repo's two documented audit failure modes. Three judges then scored every candidate independently on maintainer cost, correctness risk, and landability.

**Eight of the ten were refuted.** In nearly every case the refutation had the same shape: the "duplicate" turned out to be a justified divergence (host capability gaps, two required durability contracts, two renderer lifecycles, a persisted legacy read shim). The conclusion is that the remaining duplication in this codebase is largely load-bearing, and the genuinely recoverable debt is **dead code, not clones**.

Several items below therefore ship only as a **carve-out** of the original proposal. Each such issue names the dropped halves and why. Do not widen scope back to the original candidate — that is how this gets merged wrong.

## The dominant failure mode: deferred deletion

A path gets replaced, the replacement ships, and the old path stays wired into every host — and then people keep investing in it. #9019 added +663 lines on 2026-07-20 to replay output that nothing emits.

## Tech-debt items

- [x] #9139 — refactor: retire the dead ProcessExecutionHandle background-process pipeline  
  `hard` → Opus 5 · 3 PR(s) · risk:medium
- [x] #9140 — refactor(runtime): drop ExecutionRegistry's unread stop-result union and runtime-host thread  
  `hard` → Opus 5 · 1 PR(s) · risk:medium
- [x] #9141 — refactor(delegation): collapse the native workflow strategy clone and fix in-band lease durability  
[...]

PR #9175, opened 2026-07-25
feat(tools): let the agent read a running background command's live output

### LINKED ISSUE #9169: feat(tools): let the agent read a running background command's live output
## This is a net-ADD, and it is outside #9138's governing constraint

#9138's standing rule is that every child issue **reduces elements**. This one does not: it adds a
tool route, adds a projection, and adds tests. Roughly **+130–170 production lines**. It is filed
under the tracker because it is the one user-facing capability the dead-process retirement leaves
missing, not because it pays down debt. Nothing here is dressed up as a refactor.

What it does *not* do is bring back any element #9139 deleted — see "What this must NOT do" below.
The mechanism is entirely different: the data already exists, in a store that #9158/#9167 never
touched.

Everything below is described against the state **after #9167 merges** (parts 2 and 3 of #9139,
+232/−3,255 over 82 files).

---

## Problem

An agent can launch a long-running command (`bash` with `run_in_background: true`) but has **no way
to look at its output until it finishes**. There is no "tail the build log" capability. For a 20-minute
`make` or a long test run, the agent is blind until the end — and even at the end it only sees a
20-line preview.

### Verified current behavior

**`/executions/{id}/output` is a stub.** `src/tools/ExecutionsTool.ts:250-262` unconditionally
returns `Output for {id} is no longer available (ephemeral output is cleaned up on completion).`
Its own comment says the endpoint "has nothing of its own to read." #9158 removed it from the
advertised path list (`ExecutionsTool.ts:170-192`) and from the unknown-path error's list
(`:284-293`), leaving it as an undocumented compatibility route. Before #9158 its live branch
required `handle instanceof ProcessExecutionHandle` and nothing in production ever constructed one,
so it had never worked.

**What the agent *can* do today:**

- `/executions/{id}` summary — for a tracked run, status + elapsed only
  (`ExecutionsTool.ts:396-414`, `buildRunningSummaryLines`). No output.
- `/executions/{id}/report` — `ExecutionsTool.ts:648-657`; written only at completion
  (`bash.ts:506-510`, `finalizeAndReport`).
- `/executions/{id}/result` — `ExecutionsTool.ts:663-675`, documented at `:179` as "process result
  for background commands"; `readResultMeta` returns `null` until the run ends, and it carries
  `{exitCode, wallTimeMs, success, timedOut, command}` (`bash.ts:632-639`) — no output text at all.
- `action: 'wait'` — `ExecutionsTool.ts:338-349`; waits for a **status** change, not output.
- `action: 'subscribe'` — `ExecutionsTool.ts:585-605`. The follow-ups it produces are status
  transitions and a terminal "finished" line, formatted in
  `src/agent/runtime/ExecutionSubscriptionBinder.ts:141-156`. A background bash child goes RUNNING →
  terminal with no intermediate transitions, so subscribing to one yields exactly one message, at the
  end. **No output ever flows through this path.**

**The completion follow-up is smaller than it looks.** `bash.ts` keeps a 12,000-char tail and a
1,000-char head per stream (`bash.ts:67-78`, `:443-470`), but `formatBashDelivery` delivers
`lastNLines(stdout.tail, OUTPUT_PREVIEW_LINES)` with `OUTPUT_PREVIEW_LINES = 20`
(`src/tools/subagentResults.ts:356`, `:399-400`, `:535-538`) — i.e. **the last 20 lines**, plus the
≤1,000-char head only when truncation actually occurred. So the middle of a long log is unreachable
*even after the command completes*.

Meanwhile the child stream log retains up to **200,000 characters** of that same output
(`BACKGROUND_LOG_CAP_CHARS`, `bash.ts:80`). Today nothing can read it back.

---

## Where the data actually lives (the design premise, verified)

A background bash is **not** a `ProcessExecutionHandle`, and its output is **not** in a temp file. It
is already durable, and already keyed to the execution.

1. `bash.ts:382-391` calls `createChildStream(executionId, parentStreamId, { streamPrefix:
   'bash@tool', runKind: 'process', toolName: 'bash', … })`.
2. `childStream.ts:94` derives `childStreamId = ` `` `${streamPrefix}#${executionId}` `` — i.e.
   `bash@tool#<executionId>`, deterministic from the execution id.
3. `childStream.ts:100-105` builds `createRunTrace(childStreamId, session.transcripts, …)`;
   `childStream.ts:158-177` constructs an **`AgentExecutionHandle`** (not a process handle) and
   registers it with `session.executions.trackAgentExecution(handle, { status: RUNNING })`, with
   `handle.toolName = 'bash'` at `:171`.
4. `runTrace.ts:50` acquires a `TranscriptWriter` from that `StreamLogStore`; `:70-74` attaches
   `attachTranscriptRecorder(trace, writer)`.
5. `bash.ts:443-470` routes **every** stdout chunk through `logChunk(chunk, 'info')` and every stderr
   chunk through `logChunk(chunk, 'warn')` (`bash.ts:420-431`), which call
   `logger.info` / `logger.warn` on that child trace (`childStream.ts:180`).
6. `TexraTranscriptRecorder.ts:240-252` handles `type: 'log'`, and `:215-234` appends a
   `STREAM_LOG_ENTRY_TYPES.LOG` row with the chunk text, the level, and `messageType: DEFAULT`.
   `shouldEmit` (`:58-62`) passes `info`/`warn`.
7. `StreamLog.append` (`StreamLog.ts:142-173`) stores it in order; `StreamLog.toJSON()` (`:428-430`)
   reads them back.

**So: yes, a running background bash's output is durably readable mid-run, by execution id.**

- **Live, in-memory:** `currentSession().executions.getHandle(id)` (`executionRegistry.ts:334-336`)
  → `AgentExecutionHandle.childStreamId` (`ExecutionHandle.ts:107-120`) →
  `currentSession().transcripts.get(streamId)` (`StreamLogStore.ts:262-264`). The log is guaranteed
  resident while the command runs: `requestEviction` refuses to drop a stream that has an open
  writer (`StreamLogStore.ts:307-333`), and bash's writer stays open until `finalize`
  (`childStream.ts:202-209`, `runTrace.ts:123-141`).
- **Persisted:** the store saves on a 300 ms debounce (`StreamLogStore.ts:32`, `:200-203`) to
  `streamLogs/{streamId}.json`, so the rows survive completion and process restart. Nothing deletes
  them on `autoClose` (`childStream.ts:292-300` only emits `removeStream`; no
  `transcripts.delete` / `requestEviction` call exists on that path).
- **Cold lookup by id already exists:** `resolvePersistedStreamIdForExecution`
  (`src/transcript/executionStreamResolver.ts:134-185`) resolves an execution id to its stream id via
  cached meta, sidecar meta, or the `#${executionId}` suffix match (`:42-48`) — which matches
  `bash@tool#<id>` exactly.

No new plumbing, no new event, no new store. The read is the missing piece.

---

## Proposal

Restore **`/executions/{id}/output`** as a real, documented, pull-only route, sourced from the child
stream's `StreamLog`.

**Shape** (`action: 'view'`, the default):

```
Output for 0a1b2c3d (bash, running, 4m12s elapsed) — 8,312 of 200,000 logged chars.
Rows are in arrival order; `err:` marks a row logged from stderr.

  ...  (default window: last N lines)
  err: warning: unused variable 'x'
  [ 84%] Building CXX object src/foo.cc.o

[still running — re-read for more, or action='wait' on /executions/{id} to block until it finishes]
```

**Resolution, two branches, both already available:**

1. **Tracked** (running now, this session): handle → `childStreamId` → `transcripts.get(streamId)`.
2. **Not tracked** (finished, or launched by an earlier session):
   `resolvePersistedStreamIdForExecution(id, { streamLogStore: currentSession().transcripts })` →
   `transcripts.ensureLoaded(streamId)` (`StreamLogStore.ts:448-458`) → `get(streamId)`.

Branch 2 is not scope creep — it is the same read, and it retires the misleading "no longer
available" text for free. It is what makes the middle of a finished 200k-char log reachable at all,
given the 20-line delivered preview.

**Projection:** LOG-type rows with text; level `warn` → stderr-marked, otherwise stdout. Rows render
**in append order**, which preserves stdout/stderr interleaving — deliberately unlike the deleted
`formatProcessOutputSections`, which split them into two `<stdout>`/`<stderr>` blocks and destroyed
ordering.

**Windowing:** default to the last N lines (suggest 200) so a chatty build cannot dominate the
context window; `view_range: [start, end]` pages the full projection using the already-imported
`applyViewRange` (`src/tools/executions/conversationFormat.ts:33-40`) and the existing
`ViewRangeSchema` (`src/tools/formatting.ts:5-9`). Restore the `view_range` mention at
`ExecutionsTool.ts:188` that #9158 removed.

**Guard:** only for process-kind executions. `registerExecution(..., 'process')` (`bash.ts:370-376`)
persists `meta.category = 'process'` (`src/agent/storage/executionLifecycle.ts:132-136`), and a live
handle carries `toolName === 'bash'`. For anything else, return a one-line pointer to
`/executions/{id}/conversation` rather than dumping a subagent's model transcript as if it were
command output.

**No new module.** The route body belongs in `ExecutionsTool.ts` beside `showConversation`
(`:711`). A separate single-caller formatter file is banned by CLAUDE.md's abstraction guardrails
(code-review § 13) — and `src/tools/executions/processOutputFormat.ts` was exactly that shape before
#9158 deleted it.

---

## What this must NOT do

It must reintroduce **nothing** from the retired pipeline. Explicitly, none of:

Deleted by #9158 (`a0d4c6b126`):
- `src/agent/runtime/ProcessOutputPoller.ts` (310 lines) or any timer-driven polling of process output
- the `ProcessExecutionHandle` class, its `outputPaths`, or any `handle instanceof
  ProcessExecutionHandle` branch
- ephemeral **temp-file** stdout/stderr capture and the `platform().fs.readFile` reads over it
- `src/tools/executions/processOutputFormat.ts` / `formatProcessOutputSections`
- the background-process count in the `/executions` listing header

Deleted by #9167:
- the `process.output` `AgentEvent` arm and its `RUN_FACT_EVENT_TYPES` entry
  (`src/agent/trace/events.ts`), its `ProgressFactApplier` handler, or its
  `TexraTranscriptRecorder` case
- `UpdateProcessOutputPayload` / `UpdateActiveProcessesPayload`
- the `child.activity` `kind: 'subagents' | 'processes'` discrimination
  (`executionRegistry.ts:658-663` now emits an unkinded event)
- the `processes` roster in `BackendOwnedFieldsSchema`, `BackgroundTasksPanel`, the CLI child list,
  or the `badges.processes` bucket in `src/shared/streams/streamContentSync.ts`
- `streamMetaReducer.ts`, `TaskDetailView.tsx`, `taskDetailScroll.ts`, `completedProcessTranscript.ts`

Additional constraints:
- **Pull-only.** No new push channel to any host. No new `AgentEvent` type, no new wire payload, no
  new webview/CLI surface. The agent asks; nothing streams to a UI.
- **No `bus.emit` from a VS Code-free zone** (CLAUDE.md platform-decoupling rule).
- **No `ActiveChildInfoSchema` changes** — #9167 deliberately left that two-arm union alone as the
  program's one persisted-data trap.
- **Do not raise `BACKGROUND_LOG_CAP_CHARS` in this PR.** If 200k proves too small, that is a
  separate, separately-justified change.

---

## Size estimate

One PR. ~**+130–170** production LOC (essentially all in `src/tools/ExecutionsTool.ts`: the route
body, the projection, the description/error-list lines), plus ~**+100–140** test LOC. No new
production files. `check:dead-code-ratchet` impact: none expected (no new exports).

---

## Risks and open decisions

1. **Source cap.** Output stops being logged at 200,000 chars (`bash.ts:80`, `:420-431`), after which
   a single `[Stream log truncated at 200k chars …]` warn row is written and nothing more. The route
   must state the cap in its header rather than implying the log is complete. A command that emits
   more than 200k chars still has an unreadable middle — accepted for v1.
2. **Response size.** Every tool result already passes `checkToolResultTextLimit`
   (`src/agent/modelHandlers/utils/toolAttachmentUtils.ts:231`, `:132-165`), which head/tail-truncates
   at 4,000/50,000 chars (`contextManagementConstants.ts:30-31`). Relying on that backstop would burn
   context; hence the default last-N-lines window plus `view_range`.
3. **Polling cost vs `wait`/`subscribe`.** Neither `wait` nor `subscribe` carries output
   (`ExecutionsTool.ts:338-349`; `ExecutionSubscriptionBinder.ts:141-156`), so neither covers this
   gap — but both remain the right way to detect *completion*. The route's footer should say so, to
   discourage a re-read loop. **Open decision: v1 should NOT add a wait-for-new-output variant.**
   That would need a new change signal on the stream log and is a much larger change than the read.
4. **Row purity.** The child stream also carries a handful of non-chunk rows: `finalizeChildStream`'s
   `Completed in …` (`childStream.ts:243-244`) and error line (`:240-242`), plus bash's cap warning.
   v1 renders them inline in order. **Rejected alternative:** tagging chunk rows with a new
   `MessageType` — that union is exhaustively switched in `completedRunArchive.ts:298-338`
   (`assertNever`) and consumed by the trace viewer, so a new member ripples well beyond this
   feature for a cosmetic gain.
5. **Permissions / trust boundary.** The `executions` tool is not approval-gated
   (`ExecutionsTool.ts:167-193` sets only `slow: true`), and no route restricts reads by ownership —
   `/report`, `/result`, and `/conversation` already expose other executions' content. `/output` is
   the same boundary, not a wider one. Worth a sentence in review, not a blocker.
6. **Ephemeral stores.** Headless/test sessions may hold an ephemeral `StreamLogStore`
   (`StreamLogStore.ts:246-255`); `ensureLoaded` is a no-op there (`:449`). The in-memory read still
   works; only the cold cross-process branch degrades. Must not throw.

---

## Acceptance criteria

- [ ] `/executions/{id}/output` on a **currently running** background bash returns its output so far,
      in arrival order, with stderr rows marked — verified by a test that holds a mocked
      `executeCommand` open (the `run_in_background` harness already exists in
      `src/test-kernel/tools/BashTool.vitest.ts:603-686`, and the default test session supplies an
      in-memory store via `src/test-kernel/support/defaultSessionTestSetup.ts`).
- [ ] The same route on a **finished** background bash returns its retained log instead of
      "no longer available".
- [ ] `view_range` pages the projection; the default response is bounded to the last N lines and says
      so.
- [ ] The header reports run status and the logged-chars/cap figures.
- [ ] A non-process execution id returns a pointer to `/conversation`, not a model transcript.
- [ ] The route is restored to the advertised list in the tool description and to the unknown-path
      error message.
- [ ] `grep -rn "ProcessExecutionHandle\|ProcessOutputPoller\|process\.output\|outputPaths\|
      formatProcessOutputSections" src packages` returns nothing new.
- [ ] Element accounting stated in the PR body, honestly, as a net add.

## Verification

```
npm run typecheck
npm test
npm run lint
```

---

Related: #9138 (tracker), #9139 / #9158 / #9167 (the retirement this fills a hole left by).

🤖 Generated with [Claude Code](https://claude.ai/code)

LionSR:
Closes #9169.

## What

`/executions/{id}/output` was an unconditional stub returning `Output for {id} is no longer available`. It is now a real, advertised route that projects the child stream's transcript log, so an agent can read a long `bash run_in_background` build/test log **while it is still running** — the one user-facing capability the dead-process retirement (#9139 / #9158 / #9167) left missing.

Verified chain the read sits on (re-checked at `bf2ae8f7f1`, not taken on faith from the issue): `bash.ts` `createChildStream(runKind:'process')` → deterministic `bash@tool#{executionId}` → `createRunTrace` → `attachTranscriptRecorder` → every stdout chunk `logger.info` / stderr chunk `logger.warn` → `STREAM_LOG_ENTRY_TYPES.LOG` rows in `StreamLog`. Residency while running is guaranteed by `requestEviction` refusing a stream with an open writer.

## Shape

```
Output for d297922456c6 (process, running (1s elapsed)) — 109 of 200,000 logged chars, 3 lines.
Lines are in arrival order; `err:` marks one written to stderr.
Showing lines 1-3 of 3.

[ 42%] Building CXX object src/foo.cc.o
err: warning: unused variable 'x'
[ 84%] Building CXX object src/bar.cc.o

[still running — re-read for more output, or use action='wait' on /executions/d297922456c6 to block until it finishes]
```

- **Arrival order preserved** — stdout and stderr interleave as they actually arrived, unlike the deleted `formatProcessOutputSections` which split them into two blocks and destroyed ordering. Consecutive same-channel rows are concatenated before the line split, so a line straddling two chunk boundaries stays one line (covered by a test).
- **Bounded** — default window is the last `200` lines and the header says so; `view_range` pages the full projection through the already-imported `applyViewRange`, clamped to `1,000` lines per read with a "continue from view_range: [N, …]" hint. Lines split on `\r\n | \r | \n`, so a bare-CR progress bar (curl, pip, docker) can't collapse a whole download into one enormous "line" that defeats the window.
- **Resolution, two branches** — live handle → `childStreamId`; once untracked → `resolvePersistedStreamIdForExecution` + `ensureLoaded`. A finished-but-retained log therefore still reads; only a genuinely gone log returns a pointer to `/report`, which retires the "no longer available" lie.
- **Guard** — restricted to process executions (`meta.category === 'process'`, persisted by `registerExecution(..., 'process')`, which only `bash` passes). I chose the restriction rather than "any child stream with a log" because an agent child stream's rows are a model transcript that `/conversation` already renders properly; dumping it as if it were command output would be actively misleading. Non-process ids get a one-line pointer to `/conversation`.
- **Discovery** — restored to the tool description path list, to the unknown-path error list, and to `view_range`'s doc line (all three removed by #9158). `getAvailablePaths` already advertised `output` for the `process` category, so the `/executions/{id}` summary was pointing at the stub. The background launch result now names the route too.

## Element accounting — honest net add

This is deliberately **outside #9138's reduce-elements constraint**, as the issue states.

| | |
|---|---|
| Production LOC | **+214 / −15** (net **+199**) |
| Test LOC | +307 (1 new test file, 8 cases) |
| New production files | 0 |
| New exports | 1 constant (`BASH_BACKGROUND_LOG_CAP_CHARS`), *moved* from `tools/bash.ts` to `@shared/toolUse` because the writer and the reader must agree on the figure the header reports — `check:dead-code-ratchet` unchanged at 18/18 |

The projection is a module-level non-exported function in `ExecutionsTool.ts` beside the route body — no single-caller formatter module (CLAUDE.md § abstraction guardrails; `executions/processOutputFormat.ts` was exactly that shape before #9158 deleted it).

## Nothing retired comes back

```
$ grep -rn "ProcessExecutionHandle\|ProcessOutputPoller\|process\.output\|outputPaths\|
   formatProcessOutputSections\|UPDATE_PROCESS_OUTPUT\|UpdateProcessOutputPayload\|
   UpdateActiveProcessesPayload" src packages
(no matches)
```

Pull-only: no new `AgentEvent`, no wire payload, no push channel, no `bus.emit`, no `ActiveChildInfoSchema` change, no temp-file polling, and `BASH_BACKGROUND_LOG_CAP_CHARS` is unchanged at 200,000.

## Tests

New `src/test-kernel/tools/ExecutionsToolOutput.vitest.ts` (8 cases) drives a **real** background `bash` launch whose mocked process emits chunks and then stays open, so the mid-run read is genuinely mid-run:

1. mid-run read — output so far, stderr marked, interleaving preserved, `[still running]` footer
2. chunk-split line rejoined + `view_range` paging (`Showing lines 2-3 of 5`)
3. out-of-range `view_range` → "No lines in the requested range"
4. bare-CR progress redraw → separate lines, not one collapsed line
5. 1,500-line log → default tail window `1301-1500` + the cap biting at 1,000 lines on a wide `view_range`
6. finished execution → retained log, `[finished]` footer, never "no longer available"
7. process execution with no retained log → pointer to `/report`
8. non-process execution → pointer to `/conversation`; unknown id → `Execution not found`

## Verification

All four gates run locally on this branch:

- `npm run typecheck` — clean (all six tsconfigs; needed the documented `COREPACK_ENABLE_PROJECT_SPEC=0 CI=true` workaround, and `pnpm-lock.yaml` was restored afterwards)
- `npm test` — **775 files passed | 1 skipped; 7,529 tests passed | 4 skipped**, no failures, nothing flaky (re-run in full after the `07db7c1` follow-up)
- `npm run lint` — clean
- `npm run check:dead-code-ratchet` — `18 current vs 18 baselined`, OK

---

> [!NOTE]
> **Medium Risk**
> Adds substantive executions-tool behavior and transcript projection logic agents rely on during long runs; bounded reads limit blast radius, but incorrect projection or stream resolution could mislead debugging.
> 
> **Overview**
> **Replaces the `/executions/{id}/output` stub** with a real route so agents can read a background `bash` run’s stdout/stderr **while it is still running**, from the child stream’s transcript log (the same path `run_in_background` already writes).
> 
> Projection keeps **arrival order** (stdout/stderr interleaved), prefixes stderr lines with `err:`, rejoins lines split across chunks, and splits on `\r` so progress redraws don’t become one giant line. Responses default to the **last 200 lines**, honor **`view_range`** with a **1,000-line cap** per read, and report progress against the shared **`BASH_BACKGROUND_LOG_CAP_CHARS`** (200k), now defined in `@shared/toolUse` for `bash.ts` and the reader.
> 
> Non-process executions get a pointer to `/conversation`; missing logs point to `/report`. Tool docs, error help, background launch text, and CHANGELOG are updated. New integration tests cover mid-run reads, paging, finished runs, and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 07db7c18fcb707e3b38b97da96566cb4ce47aaa1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ All notable changes to this project will be documented in this file.
   available for demanding research and engineering tasks, replacing Opus 4.8
   as the default Anthropic model. Opus 5 keeps the full 1M context window and
   adaptive thinking.
+- **Agents can read a background command's output while it runs** — the
+  `executions` tool's `/executions/{id}/output` path returns a long build or
+  test run's stdout and stderr in arrival order, updating as the command
+  proceeds instead of only at completion. Reads default to the last 200 lines
+  and accept `view_range` to page back through the retained log, which stays
+  readable after the command finishes.
 
 #### Bug Fixes
 
--- a/src/shared/toolUse.ts
+++ b/src/shared/toolUse.ts
@@ -113,7 +113,7 @@ export function normalizeToolUseData(data: unknown): NormalizedToolUse | null {
 }
 
 // ============================================================================
-// Default tool timeouts
+// Shared tool limits
 // ============================================================================
 //
 // Shared between the tool implementations and the host UIs that display a
@@ -123,6 +123,14 @@ export function normalizeToolUseData(data: unknown): NormalizedToolUse | null {
 
 /** Default `bash` tool timeout (ms) when the model omits `timeout`. */
 export const BASH_TOOL_DEFAULT_TIMEOUT_MS = 120_000;
+
+/**
+ * Max chars a background `bash` run logs to its child stream before it stops
+ * logging and writes a single truncation notice. Lives here because the writer
+ * (`tools/bash.ts`) and the reader (`/executions/{id}/output`) must agree on
+ * the figure the output header reports.
+ */
+export const BASH_BACKGROUND_LOG_CAP_CHARS = 200_000;
 
 /** Default `executions wait` timeout (seconds) when the model omits `timeout`. */
 export const EXECUTIONS_WAIT_DEFAULT_TIMEOUT_SECONDS = 300;
--- a/src/test-kernel/tools/ExecutionsToolOutput.vitest.ts
+++ b/src/test-kernel/tools/ExecutionsToolOutput.vitest.ts
@@ -139,7 +139,7 @@ describe('ExecutionsTool /executions/{id}/output', () => {
 
     assert.equal(result.status, 'executed');
     const output = result.output ?? '';
-    assert.match(output, /^Output for \S+ \(bash, running/);
+    assert.match(output, /^Output for \S+ \(process, running/);
     assert.ok(output.includes('of 200,000 logged chars'));
     assert.ok(output.includes('[ 42%] Building CXX object src/foo.cc.o'));
     assert.ok(output.includes("err: warning: unused variable 'x'"));
@@ -182,6 +182,24 @@ describe('ExecutionsTool /executions/{id}/output', () => {
     const past = await readOutput(run.executionId, [900, 950]);
     assert.equal(past.status, 'executed');
     assert.ok((past.output ?? '').includes('No lines in the requested range'));
+
+    await run.finish();
+  });
+
+  it('treats a carriage-return progress redraw as separate lines', async () => {
+    // curl/pip/docker redraw with `\r` and no newline. Splitting on LF alone
+    // would make the whole progress bar one enormous line, so the bounded
+    // window would still hand back the entire log.
+    const run = await launchBackgroundRun((sink) => {
+      sink.onStdout?.('  0%\r 50%\r100%\r\ndownload complete\n');
+    });
+
+    const result = await readOutput(run.executionId);
+    const output = result.output ?? '';
+
+    assert.ok(output.includes('Showing lines 1-4 of 4.'));
+    assert.ok(output.includes('  0%'));
+    assert.ok(output.includes('download complete'));
 
     await run.finish();
   });
--- a/src/test-kernel/tools/ExecutionsToolOutput.vitest.ts
+++ b/src/test-kernel/tools/ExecutionsToolOutput.vitest.ts
@@ -1,0 +1,289 @@
+// Test composition imports
+import '@test/support/defaultSessionTestSetup';
+
+// Node imports
+import { strict as assert } from 'node:assert';
+
+// Third-party imports
+import { afterEach, describe, it, vi } from 'vitest';
+
+// Local imports
+import { registerExecution } from '@agent/storage';
+import { AgentCategory } from '@agent/core/definition/AgentDataclass';
+import { AgentConfigSchema } from '@agent/core/definition/AgentConfig';
+import { FileInteractionState } from '@agent/core/state/AgentWorkspaceState';
+import { withToolEnvironment } from '@agent/followUp/ToolFileInteractionContext';
+import * as toolUseFollowUp from '@agent/followUp/ToolUseFollowUp';
+import { defaultSession } from '@agent/runtime/SessionHandle';
+import type { StreamTabId } from '@shared/schemas';
+import type { ExecResult } from '@shared/schemas/opResults';
+import { setupPlatform } from '@test/support/setupPlatform';
+import { ExecutionsTool } from '@tools/ExecutionsTool';
+import { BashTool } from '@tools/bash';
+import { generateExecutionId } from '@utils/core';
+import * as execUtils from '@utils/system/execUtils';
+
+// Local file imports
+import {
+  createRecordingHost,
+  recordSessionEvents,
+} from '../agent/progressTestUtils';
+
+const PARENT_STREAM_ID = 'executions-output-parent' as StreamTabId;
+
+type ExecChunkSink = Pick<
+  Parameters<typeof execUtils.executeCommand>[1] & object,
+  'onStdout' | 'onStderr'
+>;
+
+interface BackgroundRun {
+  readonly executionId: string;
+  /** Settle the mocked process and wait for its completion follow-up. */
+  readonly finish: () => Promise<void>;
+}
+
+/**
+ * Launch a real background `bash` run whose mocked process emits `chunks`
+ * synchronously and then stays open until `finish()` — the only way to
+ * observe a mid-run read of the child stream's transcript log.
+ */
+async function launchBackgroundRun(
+  emit: (sink: ExecChunkSink) => void,
+): Promise<BackgroundRun> {
+  let release!: (result: ExecResult) => void;
+  const processExit = new Promise<ExecResult>((resolve) => {
+    release = resolve;
+  });
+
+  vi.spyOn(execUtils, 'executeCommand').mockImplementation(
+    async (_command, options = {}) => {
+      emit(options);
+      return processExit;
+    },
+  );
+  const followUp = vi
+    .spyOn(toolUseFollowUp, 'sendFollowUp')
+    .mockResolvedValue({ status: 'sent' });
+
+  const { host } = createRecordingHost();
+  const recorded = recordSessionEvents(defaultSession().events);
+  let launched;
+  try {
+    launched = await withToolEnvironment(
+      {
+        run: { runtimeHost: host, streamId: PARENT_STREAM_ID },
+        call: { tracker: new FileInteractionState() },
+      },
+      () =>
+        new BashTool().call({
+          command: 'make build',
+          run_in_background: true,
+        }),
+    );
+  } finally {
+    recorded.detach();
+  }
+
+  assert.equal(launched.status, 'executed');
+  const executionId = /Execution ID: (\S+)/.exec(launched.output ?? '')?.[1];
+  assert.ok(executionId, 'Background launch should report its execution ID');
+
+  return {
+    executionId,
+    finish: async () => {
+      release({
+        success: true,
+        stdout: null,
+        stderr: null,
+        timedOut: false,
+        exitCode: 0,
+      });
+      await vi.waitFor(() => {
+        assert.ok(
+          followUp.mock.calls.length > 0,
+          'Background run should deliver a completion follow-up',
+        );
+      });
+    },
+  };
+}
+
+function readOutput(
+  executionId: string,
+  viewRange?: [number, number],
+): Promise<{ status: string; output?: string; error?: string }> {
+  return new ExecutionsTool().call({
+    path: `/executions/${executionId}/output`,
+    ...(viewRange ? { view_range: viewRange } : {}),
+  });
+}
+
+describe('ExecutionsTool /executions/{id}/output', () => {
+  setupPlatform({
+    workspacePath: '/workspace',
+    config: { 'texra.toolUse.requireBashApproval': false },
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('returns a running background command output so far, with stderr marked', async () => {
+    const run = await launchBackgroundRun((sink) => {
+      sink.onStdout?.('[ 42%] Building CXX object src/foo.cc.o\n');
+      sink.onStderr?.("warning: unused variable 'x'\n");
+      sink.onStdout?.('[ 84%] Building CXX object src/bar.cc.o\n');
+    });
+
+    const result = await readOutput(run.executionId);
+
+    assert.equal(result.status, 'executed');
+    const output = result.output ?? '';
+    assert.match(output, /^Output for \S+ \(bash, running/);
+    assert.ok(output.includes('of 200,000 logged chars'));
+    assert.ok(output.includes('[ 42%] Building CXX object src/foo.cc.o'));
+    assert.ok(output.includes("err: warning: unused variable 'x'"));
+    assert.ok(output.includes('[ 84%] Building CXX object src/bar.cc.o'));
+    assert.ok(output.includes('[still running'));
+    // stdout rows must not be marked as stderr.
+    assert.ok(!output.includes('err: [ 42%]'));
+    // Arrival order, not stdout-then-stderr sections.
+    assert.ok(
+      output.indexOf("err: warning: unused variable 'x'") <
+        output.indexOf('[ 84%] Building CXX object src/bar.cc.o'),
+      'stdout/stderr interleaving must survive the projection',
+    );
+
+    await run.finish();
+  });
+
+  it('joins chunks that split a line and pages the projection with view_range', async () => {
+    const run = await launchBackgroundRun((sink) => {
+      // A single logical line delivered across two stdout chunks.
+      sink.onStdout?.('line1\nline2\nli');
+      sink.onStdout?.('ne3\nline4\nline5\n');
+    });
+
+    const full = await readOutput(run.executionId);
+    assert.ok(
+      (full.output ?? '').includes('line3'),
+      'A line split across chunks must be rejoined, not broken in two',
+    );
+
+    const paged = await readOutput(run.executionId, [2, 3]);
+    const output = paged.output ?? '';
+    assert.equal(paged.status, 'executed');
+    assert.ok(output.includes('Showing lines 2-3 of 5.'));
+    assert.ok(output.includes('line2'));
+    assert.ok(output.includes('line3'));
+    assert.ok(!output.includes('line1'));
+    assert.ok(!output.includes('line4'));
+
+    const past = await readOutput(run.executionId, [900, 950]);
+    assert.equal(past.status, 'executed');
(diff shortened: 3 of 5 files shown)


LionSR:
Follow-up commit `07db7c1` responds to the two review findings, plus one gap I found on my own re-read.

**Bot finding 2 (low) — header hardcoded `bash`: fixed.** The header now reports `(process, …)`. The category is what the route's guard actually verified; the tool name was an assumption that only holds while `bash` is the sole registrant of a process-kind execution.

**Bot finding 1 (medium) — category guard when `meta` is `null`: not changing, deliberately.** The suggested `meta && meta.category !== 'process'` inverts the guard's failure direction. `meta === null` with a live handle is the only case it changes, and in that case the route has *no* evidence about the execution's kind — so it would start projecting a subagent's model transcript as if it were command output, which is precisely what the guard exists to prevent. The review itself notes the race is "effectively impossible" (`registerExecution` awaits its meta write before `createChildStream` constructs the handle, and it throws if that write fails). Failing closed on an unknown kind is the right default here, and keeping the decision sourced from persisted `meta.category` alone keeps it single-owner rather than a `??` chain across the handle and the store.

**Bugbot — `Completed in Xs` / error rows rendered inline:** this is issue #9169's explicitly accepted v1 decision (risk 4). The alternative it considered — tagging chunk rows with a new `MessageType` — was rejected there because that union is exhaustively switched with `assertNever` in `completedRunArchive.ts` and consumed by the trace viewer, so a new member ripples well past this feature for a cosmetic gain.

**My own finding — bare `\r` defeated the line window.** Progress bars (curl, pip, docker) redraw with a bare `\r` and no newline. Splitting on LF alone collapsed an entire download into one enormous "line", so the bounded window handed back the whole log however small the request was — the one case where the "never return an unbounded buffer" requirement actually failed. Lines now split on `\r\n | \r | \n`; a CRLF pair still counts as one break, and same-channel rows are still concatenated before the split so a break straddling two chunks survives. Covered by a new test (`treats a carriage-return progress redraw as separate lines`), bringing the file to 8 cases.

Re-verified after the change: `npm run typecheck` clean, `npm test` **775 files / 7,529 tests passed**, `npm run lint` clean, `npm run check:dead-code-ratchet` 18 vs 18 baselined. Nothing flaky.

Issue #9177, opened 2026-07-25
Workflow-script run progress: phase on the run row, phase-grouped agent rows when focused, per-task cost


LionSR:
## Problem

A workflow-script run is the one place TeXRA fans out to N agents at once, and it is the place the
progress surfaces read worst. Three concrete gaps:

1. **From the parent's viewport**, a workflow run is one row that says `running` and nothing else. A
   3-phase, 40-minute run looks identical at minute 2 and minute 38 — in the CLI's child list
   (`SubagentList.tsx` `SessionRow`, :72-165) and in the progress view's Background Tasks panel
   (`BackgroundTasksPanel.ts:373-440`).
2. **Focusing the run** gives a flat list of `agent()` grandchildren
   (`● writer running · claude-sonnet-5 … 1m 41s · 7 tool calls · ↓12k`) with nothing saying which
   phase each belongs to — even though the transcript directly above them is already grouped by
   `◆ Phase (i/n)` dividers.
3. **The per-task cost is run-cumulative**, not per task. Every card and every transcript line reports
   the run total at the moment that task settled. See [Per-task cost](#per-task-cost).

**Stated plainly: this is a net-ADD feature.** It sits outside tracking issue #9138's reduce-elements
constraint and is not dressed as cleanup. Honest estimate: **≈ +300 production LOC / +150 test LOC
across ~24 files**, minus ~15 LOC bought back by retiring the `getLiveCostUsd` option.

This body is the *middle* design: between the reference (Claude Code's Background-tasks panel) and a
strict minimum. Everything cut is listed under [Explicitly out of scope](#explicitly-out-of-scope)
with its reason, not silently dropped.

## What already exists (do not rebuild)

| Surface | Already on `main` |
| --- | --- |
| Shared copy, single owner | `formatWorkflowTaskMetadataParts` (`workflowTask.ts:11-30`), `workflowPhaseTaskProgress` (`:38-45`), `workflowTaskDetail` (`:52-63`), `formatWorkflowTaskLine` (`:68-74`). Three consumers: `workflowScriptRun.ts:190` (run log / headless), `workflowTaskFormatter.ts:47,58` (webview card), `subscribeStreamLog.ts:380` (CLI transcript row). Extend these; never fork them. |
| CLI phase dividers | `◆ {label} ({i}/{n})` (`transcriptEntryLayout.ts:238-246`), per-status task markers `WORKFLOW_TASK_STATUS_STYLE` (`:104-114`). |
[...]

Issue #9182, opened 2026-07-25
Fix workflow task rows to report per-task cost


LionSR:
## Context

Stage 1 of #9177, split out because it is the smallest independently shippable part of that larger staged UX design.

## Problem

`workflowScriptRun.ts` stamps each task's terminal metadata with the workflow run's cumulative live spend. Two calls costing $0.05 each therefore render `$0.05 total` and `$0.10 total` on per-task rows, even though readers naturally interpret a number on a task row as that task's cost.

The strategy already retains per-invocation cost observations keyed by invocation identity for run-total reconciliation; the progress path discards only the per-call granularity and reads a cumulative scalar.

## Work

- Accumulate per-call cost by `invocation.index` alongside the existing journal-reconciliation map; do not replace or alter the existing identity-keyed accounting.
- Replace `getLiveCostUsd?: () => number` with `getCallCostUsd?: (index: number) => number | undefined`.
- Use the terminal event's index for settled tasks.
- Preserve the index for tasks finalized from the "run ended while still running" path, mirroring the existing phase map.
- Delete the now-dead cumulative `liveCostUsd` progress accumulator.
- Change the shared task metadata copy from `$X total` to `$X`, so the run log, CLI task line, and webview card stay consistent through the existing formatter.
- Add a changelog entry.

Retries under the same engine index accumulate into the logical task's cost, including discarded attempts. Cached replay behavior remains unchanged and is out of scope.

## Acceptance

- Two calls costing $0.05 each report `0.05` and `0.05`, not `0.05` and `0.10`.
- No workflow task line or card renders the word `total`; `formatWorkflowTaskMetadataParts` remains the single copy owner.
- `getLiveCostUsd` and the strategy's progress-only `liveCostUsd` accumulator no longer exist.
- Existing run-total journal reconciliation remains unchanged and green.
- No unrelated phase-row/grouping work from #9177 is included.

PR #9184, opened 2026-07-25
Stage 1 of 4 for #9177: report a workflow task's own cost, not the run total

LionSR:
Part of #9177 — the first of four stages.

## What was wrong

`WorkflowTaskProgress.totalCostUsd` was `getLiveCostUsd()`, the run-cumulative
live spend at the instant that task settled. An N-task run where every call cost
the same still read `$0.05`, `$0.10`, `$0.15` down the list. The ` total`
qualifier made the copy technically honest, but a number on a per-task row is
read as that row's.

## What changed

- The strategy's existing `onCost` handler folds the same attempts a second way,
  by the engine's 0-based call index (`observedCallCosts`).
  `runPersistedWorkflowScriptWithProgress` takes
  `getCallCostUsd(index): number | undefined` in place of `getLiveCostUsd()`.
  `agent:end` carries `event.index` directly; the cleanup path (a task still
  `running` when the run tears down) resolves its index from the call it never
  settled.
- **`observedCosts` and its `${index}:${key}` identity are untouched.**
  `sumCurrentWorkflowRunCost` consumes them at `workflowScriptStrategy.ts:212`
  and `:227` for journal reconciliation — that is a live consumer, and the run's
  cumulative figure still settles onto the parent through `ports.recordCost`.
  The new map is kept beside it rather than derived from it for exactly that
  reason.
- Retried and skipped attempts of one logical call share an engine index, so a
  task bills the work it discarded — the same reading the run's own total takes.
- Two same-key maps in the projector collapse into one: `callPhases` plus the
  index the cleanup path needs become `liveCalls`, holding both facts with the
  same set-at-`agent:start` / delete-at-`agent:end` lifecycle. (Net effect on
  the `has()` vs `get()` distinction that separates "recorded undefined phase"
  from "not recorded" is unchanged — the value is always an object when set.)
- The copy change lives in `formatWorkflowTaskMetadataParts`, still the only
  owner, so the webview card, the CLI transcript row, and the run log the
  invoking model sees all move together: `$0.020 total` → `$0.020`.
  Changelog entry added, since this is user-visible.

## Acceptance criteria this stage satisfies

- [x] Two `agent()` calls costing $0.05 each report `0.05` and `0.05`; the
      pinning test was rewritten with the opposite expectation and renamed to
      *"bills each task its own call's cost, not the run total so far"*.
- [x] No workflow task line or card renders the word `total`;
      `formatWorkflowTaskMetadataParts` remains the only owner of that copy.
- [x] `getLiveCostUsd` and the strategy's `liveCostUsd` accumulator no longer
      exist.

## Deliberately left to later stages

Everything else in #9177: `phaseStage` on the run's row (stage 2), the roster
`workflowPhase` join (stage 3, already open as #9183 — no file overlap with this
PR), and phase-grouped rows in the focused CLI list (stage 4). Also untouched:
the `cached`-replay under-reporting the issue lists as *"Not fixed here"* — a
cached arm carries no terminal metadata and reports no cost, exactly as before.

## Scope correction vs. the issue

The issue estimated stage 1 at 4 files / 2 test files. It is 9 files / **5** test
files. Two extra test files assert the rendered copy and were not listed:

- `src/test-kernel/progressView/TaskGroupListIndex.vitest.ts` (webview card text)
- `src/test-kernel/cli/WorkflowScriptStreamTranscript.vitest.mts` (CLI transcript
  row) — missed by the issue's grep and by my first sweep because it is `.mts`;
  the full suite caught it.

The `getLiveCostUsd` → `getCallCostUsd` rename touched the six call sites the
issue listed in `WorkflowScriptProgressBridge.vitest.ts` (`:523, :544, :566,
:611, :729, :775`) plus its two assertion flips, and no other assertion in that
file needed an edit. Three stubs were made index-aware (`(index) => index === 0
? … : undefined`) so they actually pin the index lookup rather than passing on a
constant — notably the abandoned-call cleanup path, whose index resolution is
the non-obvious one.

One test was added to `WorkflowScriptStrategy.vitest.ts` covering the criterion
through the real production wiring (`createWorkflowScriptStrategy`): two calls at
$0.05 report `[0.05, 0.05]` on their cards while `recordCost` still receives
`0.10` for the run. The bug lived in that wiring, so the run-level test alone
would not have caught it.

## Verification

`npm run typecheck`, `npm test` (7551 passed / 1 skipped file), `npm run lint`,
`npm run check:dead-code-ratchet` — all green on `df99b4b` (rebased onto current
`main`). No flakes; the one failure during the run was the genuine `.mts` copy
assertion above, fixed and re-verified in isolation and in the full suite.
`pnpm-lock.yaml` was reverted after the `CI=true` typecheck workaround and is
confirmed absent from the commit.

---

> [!NOTE]
> **Low Risk**
> Display and progress wiring change only; run-level cost rollup via recordCost and journal reconciliation maps are explicitly preserved.
> 
> **Overview**
> Workflow-script **task cards and finish lines** used to show **cumulative run spend** at the moment each task settled, so identical `$0.05` calls read as `$0.05`, `$0.10`, `$0.15` down the list. Each task now gets **`totalCostUsd` for its own `agent()` call** (by engine call index), and shared copy drops the **`total` suffix** (`$0.020 total` → `$0.020`) everywhere hosts render metadata.
> 
> **`runPersistedWorkflowScriptWithProgress`** takes **`getCallCostUsd(index)`** instead of **`getLiveCostUsd()`**; **`agent:end`** uses **`event.index`**, and tasks still **running** when the run tears down resolve cost from a **`liveCalls`** map (phase + index). The strategy folds the same **`onCost`** attempts into **`observedCallCosts`** by index while **`observedCosts`** / **`ports.recordCost`** still settle the **full run total** on the parent.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5af4654466dcfbd736f3dc7f2167f0835c61108a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ All notable changes to this project will be documented in this file.
 
 #### Bug Fixes
 
+- **A workflow task's cost is now that task's own cost** — the figure on a
+  workflow-script task card and task line reported the whole run's spend at the
+  moment the task settled, so it grew down the list even when every task cost
+  the same. Each task now reports what its own `agent()` call spent, and the
+  now-redundant `total` qualifier is gone (`$0.020 total` → `$0.020`). The run's
+  cumulative cost still rolls up to the invoking agent's usage unchanged.
 - **Large foreground shell commands use bounded memory** — command results
   retain useful diagnostics from the beginning and end without buffering the
   complete output in memory.
--- a/src/shared/copy/workflowTask.ts
+++ b/src/shared/copy/workflowTask.ts
@@ -7,6 +7,8 @@ import { formatCompactDuration, formatCostUsd } from '@utils/text/stringUtils';
 
 /**
  * Canonical metadata copy for terminal workflow-task progress on every host.
+ * The cost is this task's own spend — a per-task row is read as that row's, so
+ * it carries no `total` qualifier and does not grow down a phase's card list.
  */
 export function formatWorkflowTaskMetadataParts(
   task: WorkflowTaskProgress,
@@ -25,7 +27,7 @@ export function formatWorkflowTaskMetadataParts(
       : formatCompactDuration(task.durationMs),
     task.totalCostUsd === undefined
       ? undefined
-      : `${formatCostUsd(task.totalCostUsd)} total`,
+      : formatCostUsd(task.totalCostUsd),
   ].filter((part): part is string => part !== undefined);
 }
 
--- a/src/test-kernel/agent/WorkflowScriptProgressBridge.vitest.ts
+++ b/src/test-kernel/agent/WorkflowScriptProgressBridge.vitest.ts
@@ -505,24 +505,26 @@ return await agent('Review the argument', { id: 'review-task' })`;
     }
   });
 
-  it('renders the running total on live finish lines only', async () => {
+  it("bills each task its own call's cost, not the run total so far", async () => {
     const store = getExecutionStore(executionId);
     const script = `${meta}
 await agent('First')
 return await agent('Second')`;
     const { trace, events } = recordingTrace();
-    let liveCostUsd = 0;
+    const callCosts = new Map<number, number>();
     await runPersistedWorkflowScriptWithProgress(trace, {
       store,
       checkpointId: 'live-cost',
       script,
-      runAgent: async () => {
-        liveCostUsd += 0.05;
+      runAgent: async ({ index }: WorkflowAgentInvocation) => {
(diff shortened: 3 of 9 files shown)


LionSR:
Closing as a concurrent duplicate of merged PR #9186, which implements the same Stage 1 per-task-cost correction, preserves run-total reconciliation, updates the same shared copy surfaces, and closed the focused sub-issue #9182.

PR #9185, opened 2026-07-25
Stage 2 of 4 for #9177: show a running workflow's current phase on its row

LionSR:
Part of #9177. Stage 2 of 4 — `phaseStage` on the run's row, both hosts.

## The problem

A workflow-script run's row says `running` and nothing else, so a 3-phase,
40-minute run looks identical at minute 2 and minute 38 from the parent's
viewport. The data is already emitted and dropped: `trace.openStage(title,
{ kind: 'phase', index, total })` (`workflowScriptRun.ts`) produces a
`stage.start` fact carrying label/kind/index/total, and the projections all
bail on `if (event.kind !== 'round') return`.

## What this does

Adds a `phaseStage` sibling to `roundStage` on stream metadata, mirroring
`roundStage` hop for hop: schema → backend state → wire → merge → slice →
renderer. One shared `formatPhaseStageLabel` (`streamStatusDisplay.ts`) owns
the copy for both hosts.

```
   ● workflow-script running · Reduce 2/3 · claude-sonnet-5    4m 06s · ↓41k
```

- **CLI** — takes the slot `SubagentList.tsx` already gives
  `formatRoundStageLabel`, which is always empty for a workflow run (workflow
  runs emit phases, not rounds). The row stays one line.
- **Webview** — `BackgroundTasksPanel` gets the phase per row. See "One
  deviation" below for how, and why the route in the issue body does not work.
- **Formatting** — `Reduce 2/3` with a declared position and total, `Reduce 2`
  with only a position, and the bare title for a dynamically opened phase
  (`phase()` outside `meta.phases` carries neither index nor total). No
  truncation in the formatter — the CLI row already `truncate-end`s and the
  webview span already ellipsizes (open question 1 in the issue).

## Acceptance criteria this stage satisfies

- [x] A running workflow's row shows `{phase} {i}/{n}` in the CLI child list
      and in the Background Tasks panel, on one line, at every terminal width;
      a reflection stream's row still shows `r{i}/{n}`, and no row shows both.
      Asserted by an ink render test at 120 / 80 / 64 / 56 columns plus a
      both-fields-set row, and by a jsdom render test of the panel.
- [x] The CLI status bar does **not** gain a phase segment —
      `statusBarDisplay.ts` and `StatusBar.tsx` are not in the diff. When the
      run is focused, the workflow band already shows the phase *with*
      `done/total` in the same viewport.
- [x] `sessionProgressSubscription.ts` is untouched (`git diff --stat` does not
      list it), and headless output is byte-identical: that module is reached
      only from `runExecution.ts`, while `subscribeRuntimeHost.ts` is imported
      only by `chatSessionController.ts` → `runChatTui.tsx`.
- [x] The webview `<wa-details>` phase groups, their persisted open state, and
      `renderGroupProgress` are the same code path as before — `TaskGroupList.ts`
      is not in the diff.

## Deliberately left to later stages

- **Stage 1** (#9184, open) — per-task cost and retiring `getLiveCostUsd`. No
  file overlap with this branch.
- **Stage 3** (#9183, open) — the roster join (`workflowPhase` on
  `ActiveChildInfo`). Touches a different region of `streamState.ts`.
- **Stage 4** — phase-grouped grandchild rows in the focused CLI list,
  `streamTreeEntries` ordering, `ChildListValue` widening.

## One deviation from the issue body, and why

The issue proposed passing `.streamStates=${streamStates$.get()}` into
`<background-tasks-panel>` from `WorkflowStreamContent` / `ToolUseStreamContent`.
**That would render the field but never update it.** Those two components
extend `BaseStreamContent`, which is a plain `LitElement` — not a
`SignalWatcher` — so it re-renders only when `streamStateContext` changes, and
that context is fed by `activeStreamState$`, documented as *"Only changes when
the ACTIVE stream's state changes, not any stream"* (`progressState.ts`). A
workflow run's row lives in its **parent's** panel, so the run is precisely not
the active stream; its phase would land in `appState` and never reach the DOM.

Instead the panel consumes a narrow `phaseStagesContext`
(`Map<streamId, PhaseStage>`) provided by `StreamConversation`, which *is* a
`SignalWatcher` and already provides `streamByIdContext` to this same panel for
exactly the same reason. The backing `phaseStages$` selector omits streams
without a phase and returns a shared empty map when none has one, so a session
with no workflow-script run keeps a stable identity and re-renders exactly as
often as before this PR.

Same reason on the backend: `updateRoundStage` is gated to
`state.activeStream === streamId`, which would never fire for a child run, so
the phase push rides the existing `updateStreamMetadata` per-stream patch (no
new outbound command; phases advance a handful of times per run).

## Open questions from the issue, as answered here

1. **Truncate the phase label?** No — left to each host's existing truncation.
   The webview span additionally caps at `10rem` with a `wa-tooltip` carrying
   the full title, matching the standardized hover-hint idiom (#9166/#9178).
2. **Keep the last phase after a run settles?** Kept, mirroring `roundStage`,
   which is likewise not cleared on settle. `resetPerRunChildState` clears both
   when a *new* run starts on the same stream.
3. Retry/header-count semantics — stage 4.

## Verification

All four gates green on this branch:

```
npm run typecheck                # clean (workspace + 5 projects)
npm test                         # 777 files passed | 1 skipped, 7569 tests passed
npm run lint                     # clean
npm run check:dead-code-ratchet  # 18 current vs 18 baselined, no new findings
```

Nothing was flaky; the only failure at any point was the one expected
`buildStreamMetadata` `toEqual` needing `phaseStage: null`, updated in this PR.
`pnpm-lock.yaml` was restored after the `CI=true` typecheck workaround and is
not part of either commit.

### Follow-up commit: `phaseStages$` memo

Review found that the selector reused the shared empty map only when *no*
stream had a phase — so once a run was live (the case it exists for) it
allocated a fresh `Map` per recompute, and it recomputes on any field of any
stream. `bdbc9823af` caches the prior map and returns it when the phases are
unchanged, mirroring the `pendingApprovalIds$` stable-Set memo directly above
it, with a reset in `resetProgressState()`. Comparison is by content, not
reference, because an unchanged phase arrives as a fresh object every time a
metadata patch crosses postMessage. The regression test fails without the memo
(`expected Map to be Map // Object.is equality`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ All notable changes to this project will be documented in this file.
   proceeds instead of only at completion. Reads default to the last 200 lines
   and accept `view_range` to page back through the retained log, which stays
   readable after the command finishes.
+- **A running workflow script shows its current phase from the parent's view**
+  — the run's row now reads `Reduce 2/3` beside its status, in the CLI's child
+  list and in the progress view's Background Tasks panel, so a long multi-phase
+  run is no longer indistinguishable at minute 2 and minute 38. Tool-use rows
+  keep showing their round (`r2/3`); no row shows both.
 
 #### Bug Fixes
 
--- a/packages/cli/src/chat/tui/panes/SubagentList.tsx
+++ b/packages/cli/src/chat/tui/panes/SubagentList.tsx
@@ -7,6 +7,7 @@ import { useMemo } from 'react';
 // Local imports - shared stream state
 import type { StreamTabId } from '@shared/schemas';
 import {
+  formatPhaseStageLabel,
   formatRoundStageLabel,
   formatStreamStatusLabel,
 } from '@shared/streams/streamStatusDisplay';
@@ -105,7 +106,11 @@ function SessionRow({
   // this truncate-end text sheds inline elapsed (narrow mode only), the round,
   // and last the pending-approval kind. The metadata column never shrinks.
   const approvalSuffix = pendingApprovalRowSuffix(pendingKinds);
-  const roundLabel = formatRoundStageLabel(session.slice?.roundStage);
+  // A workflow-script run advances through phases, a tool-use run through
+  // rounds; one slot carries whichever this stream has.
+  const stageLabel =
+    formatPhaseStageLabel(session.slice?.phaseStage) ??
+    formatRoundStageLabel(session.slice?.roundStage);
   // The resolved model is per-agent identity (a workflow run's grandchildren
   // can each resolve a different model); the list-root row is the conversation
   // itself, whose model already rides the status bar. A background bash stream
@@ -146,7 +151,7 @@ function SessionRow({
           {session.label}
           {statusLabel ? ` ${statusLabel}` : ''}
           {approvalSuffix ? ` · ${approvalSuffix}` : ''}
-          {roundLabel ? ` · ${roundLabel}` : ''}
+          {stageLabel ? ` · ${stageLabel}` : ''}
           {modelLabel ? ` · ${modelLabel}` : ''}
           {!metadataColumn && elapsed ? ` · ${elapsed}` : ''}
         </Text>
--- a/packages/cli/src/chat/tui/state/cliState.ts
+++ b/packages/cli/src/chat/tui/state/cliState.ts
@@ -13,6 +13,7 @@ import {
   type MessageType,
   type NormalizedToolUse,
   type OutputFileInfo,
+  type PhaseStage,
   type Plan,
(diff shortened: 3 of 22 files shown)


LionSR:
Addressed the selector-identity finding in bdbc9823af: `phaseStages$` now compares phase content and reuses its prior map across unrelated stream ticks, with reset handling and regression coverage in `StreamMetaState.vitest.ts`. The content comparison is deliberate because unchanged metadata may arrive as fresh objects across postMessage.

PR #9223, opened 2026-07-26
refactor(progressView): route hand-rolled wa-icon markup through waIcon()

LionSR:
## Summary

A scheduled UI-consistency audit found that the progress-view **formatters** build `<wa-icon>` markup by hand in ~14 places instead of calling the shared `waIcon()` helper (`@shared/wa/webAwesomeIcons`), even though most of the same files already import and correctly use `waIcon()` elsewhere. This routes all of them through the single helper and tightens the icon-name types that had regressed to plain `string` along the way.

## Issue acceptance gates

No tracked issue — surfaced by a scheduled codebase-consistency review, not a bug report. Acceptance gate is self-defined and **scoped to `formatters/` + `ContextManagement.ts`** (not the whole `progressView/frontend/` tree): every hand-authored `<wa-icon library=${TEXRA_ICON_LIBRARY} name=...>` template in `packages/extension/src/progressView/frontend/formatters/` and `components/ContextManagement.ts` now goes through `waIcon()`, and every `iconName`/`icon` field that fed one of those templates is typed `TeXRAIconName` (or `TeXRAIconName | typeof SPINNER_ICON_NAME`) instead of `string`.

**Not in scope / follow-up candidates:** 14 files under `packages/extension/src/progressView/frontend/components/` still use the same hand-rolled `<wa-icon library=${TEXRA_ICON_LIBRARY}>` pattern and are untouched by this diff: `TodoList.ts`, `UsagePanel.ts`, `WorktreeChip.ts`, `LatexdiffResults.ts`, `TaskGroupList.ts`, `BackgroundTasksPanel.ts`, `ExternalInquiryPanel.ts`, `ProposalRequestPanel.ts`, `QueuedFollowUps.ts`, `FileList.ts`, `RequestPanels.ts`, `StreamTabs.ts`, `UserMessage.ts`, `WorkflowToolUseFollowupSection.ts`. A future pass should cover these the same way.

## Single source of truth

`waIcon()` in `src/shared/wa/webAwesomeIcons.ts` is already the canonical `<wa-icon>` constructor (per AGENTS.md). This PR removes the parallel hand-rolled construction path that had grown up alongside it in the formatters layer — no new module introduced.

Removed: ~14 duplicated inline `<wa-icon library=... name=... aria-hidden="true">` constructions across 8 formatter/component files, all collapsed onto the existing `waIcon()` call. No new abstraction added.

## Net elements (R6)

12 files changed, 72 insertions(+), 84 deletions(-) — net negative LoC. No exported symbols added or removed; two exports changed signature only (`TOOL_ICON_MAP`'s value type, `getToolIconName`'s return type), both narrowed from `string` to `TeXRAIconName`.

## Consumer counts (R8)

n/a — no emit path or export was deleted or re-routed, only retyped/inlined at existing call sites.

## Host impact

Extension: `packages/extension/src/progressView/frontend/formatters/**` and `components/ContextManagement.ts` — icon rendering behavior is unchanged (same icon names, same `aria-hidden`/library attributes via `waIcon()`'s equivalent output); this is a construction-path consolidation, not a visual change.

Desktop: none (desktop reuses these same extension components; not touched directly).

CLI: none.

## Scheduled refactoring

Follow-up: apply the same `waIcon()` conversion to the 14 `components/*.ts` files listed above (currently untouched, same hand-rolled pattern). No issue filed yet.

## Validation

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run src/test-kernel/progressView` — 52 files / 430 tests passed
- `npx vitest run` (full suite) — 780 passed / 2 failed; both failures are pre-existing, timing-sensitive tests unrelated to this change (QuickJS memory-exhaustion timing, process-group abort timing), also called out as flaky in the just-opened #9216.

### DIFF
--- a/packages/extension/src/progressView/frontend/components/ContextManagement.ts
+++ b/packages/extension/src/progressView/frontend/components/ContextManagement.ts
@@ -12,21 +12,21 @@ import '@awesome.me/webawesome/dist/components/tooltip/tooltip.js';
 
 // Local imports - shared styles
 import { designTokens, commonViewStyles } from '@shared/styles';
-import { TEXRA_ICON_LIBRARY } from '@shared/wa/webAwesomeIcons';
+import { waIcon, type TeXRAIconName } from '@shared/wa/webAwesomeIcons';
 
 // Local imports - progress view helpers
 import { buildDetailsSummary } from '../formatters/htmlBuilders';
 
 /** Stat item to display */
 export interface StatItem {
-  icon: string;
+  icon: TeXRAIconName;
   label: string;
   value: string;
 }
 
 /** Action configuration */
 export interface ActionConfig {
-  icon: string;
+  icon: TeXRAIconName;
   label: string;
   color: string;
 }
@@ -141,12 +141,7 @@ export class ContextManagement extends LitElement {
             (item) => item.label,
             (item, index) => html`
               <span id="context-stat-${index}" class="stat-item">
-                <wa-icon
-                  library=${TEXRA_ICON_LIBRARY}
-                  name=${item.icon}
-                  aria-hidden="true"
-                ></wa-icon>
-                ${item.value}
+                ${waIcon(item.icon)} ${item.value}
               </span>
               <wa-tooltip for="context-stat-${index}">${item.label}</wa-tooltip>
             `,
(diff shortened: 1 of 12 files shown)


LionSR:
The review note about scope is addressed in the current PR description: the acceptance gate now names only `formatters/` plus `components/ContextManagement.ts`, and the 14 remaining component files are listed explicitly as follow-up candidates. The current head has no hand-authored `wa-icon` markup or `TEXRA_ICON_LIBRARY` reference within that stated scope.

PR #9228, opened 2026-07-26
fix(auth): construct ServerSideKeyService lazily, delete the mandatory bootstrap step

### LINKED ISSUE #9241: Construct ServerSideKeyService lazily without mandatory host bootstrap
## Problem

The default configured-credential path reaches `getServerSideKeyService()`, which currently requires each host or embedder to have called a separate `initializeServerSideKeyAccess()` bootstrap first. A minimal external embedder with its own provider key can therefore fail before its first model run even though every bootstrap option is optional and already available from platform state.

## Scope

- Construct the singleton lazily on first access.
- Preserve state, logging, notification, and host behavior for extension, desktop, and CLI.
- Degrade stateless embedders safely without silently enabling relay access.
- Delete redundant host bootstrap wiring and update embedder documentation.
- Preserve explicit test injection/reset seams.

Implementation is proposed in PR #9228; this issue tracks exact-head review, validation, and merge.

LionSR:
Closes #9241

## Problem

`getServerSideKeyService()` threw unless a host had already called
`initializeServerSideKeyAccess()`:

```
ServerSideKeyService not initialized. Call initializeServerSideKeyAccess() first.
```

That accessor is reached unconditionally by the default
`resolveClientCredential(selection='configured')` path — `ModelHandler.ts:502-507,958`,
`support/ProxyConfigResolver.ts:120,158`, `runtime/ModelFactory.ts:558`,
`model/computeModelOptions.ts:365`, `model/providerCapabilities.ts:206`,
`model/setupCredentialAccess.ts:39` — so it was the **only mandatory `throw` on a
minimal agent-run path**. An external embedder holding its own Anthropic/OpenAI
key could not complete a single run without first calling a bootstrap function
whose three options are all optional.

## Fix

Construct the service on first use. Everything the three hosts were passing is
already reachable host-agnostically, so the bootstrap step carries no
information and is deleted outright. The embedder guide now describes this lazy
construction, the safe stateless fallback, and the requirement to initialize a
stateful platform before first use when relay access is desired.

| what the host passed | where it comes from lazily | equivalent? |
| --- | --- | --- |
| `state` | `tryGlobalState()` | **Yes.** Extension passed `context.globalState` — the *same object* it hands `initPlatform()` at `extension.ts:235`. Desktop passed `platform().globalState`. CLI passed `tryPlatform()?.globalState`. |
| `logger` | `@logger/logUtils` | **Yes.** Extension passed that module verbatim (`import * as logger from '@logger/logUtils'`) — byte-identical. Desktop passed the global `console` (`index.ts:1330`: `initializeDesktopServerSideKeyAccess(console)`); logUtils' default sink is `console.info`, and `installDesktopAppLog()` patches `console[level]` to mirror into the app log, so the destination is unchanged (and now timestamped + secret-redacted). CLI: **every** `initCliPlatform` / `initInteractiveCliPlatform` / `initLocalCliPlatform` call site hardcodes `quietLogs: true`, which makes both the old `cliPlatformLog` (`logAt` early-returns) and the logUtils sink (noop factory) silent — no delta. |
| `notifyIncludedModelAccessChanged` | `appSignals.emit('includedModelAccessChanged', …)` in the lazy construct | **Yes.** This callback had real behavior and is preserved, not dropped — the extension's closure was exactly this emit. The only subscribers are `extension.ts:651` and `MainViewProvider.ts:133`, both extension-side, so CLI/desktop are unaffected. |

`setServerSideKeyService` is **kept** — 5 live test call sites across 4 suites
(`DesktopSupabaseAuth`, `ModelFactoryRouting` ×2, `CodexSubscriptionFallback`,
`ComputeModelOptions`).

### Why no ordering hazard

The singleton is memoized on first call, so a pre-`initPlatform()` call would
pin a stateless instance. That path cannot exist today: it would have *thrown*
before this PR. Every host also constructs after `initPlatform()` —
extension `initPlatform` at :235 vs old init at :315; desktop at `index.ts:1330`,
after `platform().secrets`; CLI inside the `if (!tryPlatform()) { … }` block's
successor.

## Behavior ruling: included access is OFF when there is no state store

**This is a deliberate change, and it is strictly safer than today.**

`ServerSideKeyService.ts:102-103` read
`this.globalState?.get(USE_INCLUDED_ACCESS_KEY, true) ?? true` — with a null
state, included relay access defaulted **ON**. Lazily constructing without
touching that would make a stateless embedder silently phone
`https://remote.texra.ai` on its first model call, routing its traffic through
TeXRA's servers without ever having asked.

So the constructor now branches:

```ts
this.useIncludedModelAccess = this.globalState
  ? (this.globalState.get<boolean>(USE_INCLUDED_ACCESS_KEY, true) ?? true)
  : false;
```

A host that cannot read, persist, or surface the toggle must not opt into it.
`canUseServerSideKeys()` short-circuits on `!this.useIncludedModelAccess`
*before* any network call, so a stateless embedder makes zero relay requests.

Behavior for the three real hosts is **unchanged** — all three supply a state
store, so they take the first branch, which is the old expression. Locked by a
new test: `keeps included access off when there is no host state store`.

Per the `silent degradation is a defect` rule, the miss is **loud**:
`getServerSideKeyService()` emits a `logger.warn` naming the cause and both
remedies when `tryGlobalState()` returns null. It fires once (the instance is
memoized immediately after).

## LAY-1 baseline: the new `auth->eventBus` edge

Second commit adds exactly one line to
`config/ratchets/architecture-edges-baseline.json`:

```diff
     { "from": "auth", "to": "common", "kind": "value" },
+    { "from": "auth", "to": "eventBus", "kind": "value" },
     { "from": "auth", "to": "logger", "kind": "value" },
```

**This edge is essential and semantically correct, not incidental.** CLAUDE.md
defines `src/eventBus/` as AppSignals only — "cross-cutting app-lifecycle
signals (**auth**, subscriptions, tool availability, workspace-file writes)" —
and states that the event-channel restrictions "do not restrict
`appSignals.emit(...)` on the separate `AppSignals` bus within its documented
scope." An auth subsystem emitting an auth-lifecycle signal is the textbook case
that bus exists for. The signal has to be emitted from where the state changes.
The edge was absent before only because the dependency was being **laundered
through a closure the extension supplied** — collapsing that indirection is the
point of this PR, and LAY-1 tracks intra-`src` subsystem edges whose baseline is
meant to be extended consciously (its own failure message says exactly that).

Contrast with the R-b host-import ratchet on a sibling PR, where the opposite
call was correct: there the growth was **incidental and avoidable** (a type
annotation that could be re-sourced from an already-imported module), and R-b
guards the SDK program's host→core boundary. Different ratchet, different
purpose, different ruling.

Verified before bumping:

1. **No host-coupling violation.** `src/eventBus/` is a single file
   (`AppSignals.ts`) whose only import is `node:events`. No `vscode`, no host
   coupling. The arrow points from a VS Code-*allowed* zone (`src/auth/`) into a
   VS Code-*free* one (`src/eventBus/`), which is always the safe direction.
   Precedent already in the baseline: `{ "from": "tools", "to": "eventBus",
   "kind": "value" }`.
2. **Exactly one edge added, no others, no `value-escalation`.** The ratchet's
   own collector reports a single violation:
   `[{ edge: 'auth->eventBus', reason: 'new-edge', currentKind: 'value' }]`.
   The two other new imports create no new pairs — `auth->logger` and
   `auth->platform` value edges already existed (baseline lines 18-19).
3. **Both subscribers still receive the signal.** The
   `notifyIncludedModelAccessChanged(value)` invocation inside
   `setUseIncludedModelAccess` is **untouched** by this PR — same `if (changed)`
   guard, same position after `clearAllCaches`, same try/catch. Only the
   supplied callback moved, and its body is character-for-character the
   extension's old closure. Both `extension.ts:651` and `MainViewProvider.ts:133`
   subscribe to the same `appSignals` module singleton. Covered by the passing
   `continues notifying after a listener fails` test.
4. **The stateless-default-off ruling is unaffected** — the baseline bump touches
   only a JSON file; the constructor branch and its test are unchanged and green.

## Net-element accounting

**Deleted**

- `initializeServerSideKeyAccess()` — exported function (`src/auth/serverKeys/index.ts`)
- its inline 3-field options type `{ state?, logger?, notifyIncludedModelAccessChanged? }`
- `initializeDesktopServerSideKeyAccess()` — exported function (desktop)
- the unconditional `throw new Error('ServerSideKeyService not initialized…')`
- 3 host registration sites + 1 desktop indirection call site
- 6 now-unused import bindings (`initializeServerSideKeyAccess` ×3,
  `initializeDesktopServerSideKeyAccess` ×1, `platform` in `desktopSupabaseAuth.ts`,
  and the `StateStore` / `SupabaseSessionLog` type imports in `serverKeys/index.ts`)

**Added**

- 0 new files, 0 new exports, 0 new types, 0 new parameters, 0 wrappers, 0 ports
- 3 import bindings in `serverKeys/index.ts` (`appSignals`, `logger`, `tryGlobalState`)
- 1 `logger.warn` call
- 1 test locking in the stateless-default-off ruling
- 1 baseline JSON line
- 1 embedder-guide correction replacing the deleted bootstrap instructions

**Exported symbols: −2, +0.**

**Lines**

| | added | deleted | net |
| --- | ---: | ---: | ---: |
| production | 42 | 64 | **−22** |
| tests | 11 | 2 | +9 |
| ratchet baseline | 1 | 0 | +1 |
| documentation | 46 | 53 | −7 |
| **total** | **100** | **119** | **−19** |

One existing test changed: `continues notifying after a listener fails` builds a
service with `state = null` and previously asserted the `true → false`
transition. With the stateless default now `false`, `true` is the transition
that fires; the assertion flipped accordingly. The test's subject (a throwing
listener does not suppress later ones) is untouched.

## Verification

All green:

| command | result |
| --- | --- |
| `npm run typecheck` (`tsc --noEmit` ×4 projects) | **exit 0, clean** |
| `npx tsc --noEmit` | **exit 0, clean** |
| `npm run lint` | **exit 0, clean** |
| `npm run check:dead-code-ratchet` | **OK** — 18 current vs 18 baselined, no new findings |
| `npx vitest run src/test-kernel/architecture/` | **9 files, 58 tests, all pass** (LAY-1, R-b, QA-2 included) |
| `npx vitest run src/test-kernel/auth/ src/test-kernel/model/` | **25 files, 209 tests, all pass** |
| documentation formatting, public/internal boundary, and guidance references | **pass** |

Pre-existing failures elsewhere, confirmed unrelated by running the identical
file sets on this branch and at `origin/main` (detached HEAD) and comparing:

| file set | this branch | `origin/main` |
| --- | --- | --- |
| 10 CLI files | 9 failed / 11 tests failed / 86 passed | **9 failed / 11 tests failed / 86 passed** |
| 9 desktop files | 6 failed / 142 passed | **6 failed / 142 passed** |

Identical on both sides. The two desktop files that actually cover this diff —
`DesktopSupabaseAuth.vitest.mts` and `ElectronCompositionRoot.vitest.mts` — pass
on both. `CliSupabaseAuth.vitest.mts` and `InitPlatformSignalHandlers.vitest.mts`
(which cover `initPlatform.ts`) pass 10/10 on this branch in isolation.

<sub>An earlier revision of this PR body reported 12 typecheck errors and ~80
failing test files. Those were an artifact of installing with
`COREPACK_ENABLE_PROJECT_SPEC=0`, which pins pnpm 10 against a pnpm-11 lockfile
and leaves packages unresolvable. Re-verified with `CI=true corepack pnpm install`
(pnpm 11.5.2); no lockfile changes are committed.</sub>

https://claude.ai/code/session_01XZ8Z6rhypTfkVq728XCcCd

---

> [!NOTE]
> **Medium Risk**
> Touches default credential resolution and included-model relay policy on the model path; behavior for shipped hosts with state is intended unchanged, but early singleton creation before initPlatform() pins relay off for the process.
> 
> **Overview**
> Removes **`initializeServerSideKeyAccess()`** and host wrappers so **`getServerSideKeyService()`** builds the singleton on first use from **`tryGlobalState()`**, **`@logger/logUtils`**, and **`appSignals.emit('includedModelAccessChanged')`**. CLI, desktop, and extension no longer call a bootstrap step before model runs.
> 
> **Included (relay) access** no longer defaults **on** when there is no state store: without **`globalState`**, the constructor starts with relay off (avoids silent relay traffic for stateless embedders). First use without a platform logs a **warning** and memoizes that policy for the process.
> 
> Docs in **`embedding-the-agent-runtime.md`** describe lazy construction and BYOK policy instead of mandatory Step 2 bootstrap. Architecture ratchet adds **`auth → eventBus`**. Tests cover the stateless default and adjust the listener-failure case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8baac80f0d0acfd515a45d9f2d6886eb9573607e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
The exact-head review’s documentation suggestion is addressed in `eb2e020`: the embedder guide no longer imports or calls the deleted initializer, and now explains lazy construction, the stateful BYOK policy choice, safe stateless degradation, and the first-use ordering constraint. Documentation formatting, boundary, and guidance-reference checks passed locally.

PR #9396, opened 2026-07-29
Stop bundling the Codex and Claude Code CLIs in the desktop app

LionSR:
The desktop installer shipped a full copy of both native CLIs: the
@openai/codex platform packages are 310-409 MiB each and the
@anthropic-ai/claude-agent-sdk platform packages 245-262 MiB, and the
macOS universal build carried two of each. That was well over half of a
~1.4 GiB packaged app.

Nothing required them to be there. Both SDKs' JavaScript is inlined into
dist/main by esbuild, and the binary resolvers in codexImport.ts and
claudeAgentImport.ts already probe the global npm prefix and PATH, which
is how the VS Code extension has always worked — it ignores node_modules
entirely. Installing the CLI once per machine now serves every TeXRA
surface, and the Integrations card's "Install in Terminal" action already
drives that flow on the desktop.

Moving both SDKs to devDependencies is what does the work: electron-builder
copies the node_modules of production dependencies only. With the payload
gone, the asarUnpack rules, the mac x64ArchFiles override, and the afterPack
hook that pruned non-matching platforms all become dead config, so this
deletes them along with the payload-budget baselines they were measured
against.

verify-desktop-package.mjs flips from "the expected platform package is
present and within budget" to "no Codex or Claude Code payload is present
at all", covering both the direct and pnpm-store node_modules layouts
inside and outside app.asar, so the binaries cannot creep back in.

Claude-Session: https://claude.ai/code/session_01TBVMZHqN3GFjHd8QvA2nag

---

> [!NOTE]
> **Medium Risk**
> Desktop breaking change for users who relied on bundled CLIs without a separate install; packaging and detection changes are broad but guarded by verify-desktop-package and tests.
> 
> **Overview**
> **Breaking (desktop):** The installer no longer ships bundled Codex or Claude Code native CLIs (~500+ MiB per platform). Integrations that relied on the in-app copy will show **Not Found** until the user installs each CLI once (Dashboard → Integrations → Install in Terminal). VS Code and users who already have the CLIs on PATH are unchanged.
> 
> The desktop app moves `@openai/codex-sdk` and `@anthropic-ai/claude-agent-sdk` to **devDependencies** so electron-builder does not copy their platform packages; esbuild still inlines the SDK JS. Packaging drops Codex **asarUnpack**, the **afterPack** prune hook, and payload-budget tooling; **verify-desktop-package** now **fails** if any Codex/Claude native CLI payload appears (direct or pnpm layout). Universal mac builds add `!dist-packaged/**` and narrow **x64ArchFiles** to **node-pty** prebuilds only.
> 
> **Shared:** **Install in Terminal** picks **brew**/**winget** when available instead of global npm; desktop PTY **PATH** uses `extendEnvPath` so Dock-launched shells can run suggested install commands. Detection adds **`~/.local/bin`**, and on Windows PATH resolution requires **`.exe`** so npm shims are not treated as installed. Docs/changelog document the desktop breaking change and install flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5fe948171e8f61b336b1c5d6b1426e4c87ce1a06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
## Review disposition

All six open findings are addressed in `5fe9481`; one earlier finding was already fixed in `bed670c`. Details below, then one finding I could not reproduce.

### Fixed

| Finding | Disposition |
|---|---|
| Provide a usable Windows Claude installer (P1) | `bed670c`. Windows now offers `winget install Anthropic.ClaudeCode`, and the PATH probe requires a real `.exe` so npm's `claude` / `.cmd` / `.ps1` shims are rejected instead of resolving to something unspawnable. Codex keeps npm on Windows: its install command and its global-prefix probe name the same package, so the probe resolves `codex.exe` inside `@openai/codex-win32-x64` without consulting PATH. |
| Put Homebrew on the terminal PATH (P1) | `5fe9481`. Confirmed: `ptyEnvironment()` copied the raw GUI environment while detection probes through `commandEnv()` → `extendEnvPath()`. The pty now extends PATH the same way, overwriting the key in place since Windows spells it `Path`. This mattered for npm too — a Dock-launched app may not see the user's node either. |
| Probe the native install directory (P1) | `5fe9481`. `getExtraDirs()` added `$HOME/bin` but not `~/.local/bin`, where the installer Anthropic's docs recommend puts `claude`. Added on macOS and Linux. |
| Select a package manager that has an install command (P2) | `5fe9481`. Replaced `detectPackageManager()` at this call site with a new `hasPackageManager(name)`, so only managers a command map actually names are probed — a Linux host with apt *and* Linuxbrew now gets the brew command instead of falling through to npm. `detectPackageManager()` reuses it and now caches misses, which also removes a repeated three-process probe on every call. |
| Mark the CLI removal as a breaking change (P1) | `5fe9481`. The desktop entry moved to **Breaking Changes** and states the migration explicitly: anyone who used either integration without installing its CLI sees **Not Found** until they install it once. |
| Use a permitted changelog category (P1) | `5fe9481`. `#### Improvements` is not one of the categories AGENTS.md permits; the entries are now **Breaking Changes** and **Bug Fixes**. |

### Not reproduced

**"Install tests break with Homebrew"** — this does not reproduce, and the stated mechanism cannot occur. No subprocess runs under vitest at all: `setupFakePlatform` (the global `setupFiles` entry) points every command at a workspace directory that does not exist, so `execaSync` fails before PATH is consulted. Verified by stubbing a `brew` on PATH and removing the mock — `CliTools.vitest.mts` still passed — and by observing that `executeCommandSync(['node', '--version'])` fails identically inside a test.

The underlying fragility is real, though: the assertion reads live definitions whose command *is* host-dependent by design, and only that unrelated stub keeps it stable. So the mock stays, with a comment describing the actual reason rather than the reported one.

### Verification

Run locally on this branch: `typecheck` clean, `lint` clean, `check:dead-code-ratchet` clean (17 vs 17 baselined — the three new `toolUtils` exports all have consumers), full `vitest` suite **7947 passed / 9 skipped**. The one failure in that run was `WorkflowScriptEngine > contains guest memory exhaustion inside the QuickJS runtime`, a 3-second wall-clock assertion that took 7.1s because I had eslint's 6 GB-heap process running concurrently; the file passes 92/92 on its own and is untouched by this PR.

The `Desktop Package` workflow is also green end-to-end on all three platforms for the first time since 2026-07-11 — `check:desktop-package` reports `- no bundled Codex or Claude Code CLI payload` against the real packaged app.

---

LionSR:
## Second review round

Three further findings arrived on `5fe9481`; all are addressed in `47b0fa4`.

| Finding | Disposition |
|---|---|
| Restrict the Claude Homebrew cask command to macOS (P2) | Correct — Homebrew Cask does not exist on Linuxbrew, so the `brew` probe passing on Linux handed over a cask command that cannot work. Both Homebrew commands are now macOS-only, which also matches what the install guides already claimed. |
| Probe winget before offering its install command (P2) | Correct — Windows Server and managed images ship without App Installer. winget is now probed like any other package manager, and when absent Claude Code offers **no** command instead of one that fails on arrival: the card and `texra tools install` fall back to the install guide, which names the native installer. |
| Cache the absence of a package manager (P2) | Already fixed in `5fe9481` — `hasPackageManager()` caches misses as well as hits, so the repeated three-process probe is gone. Thread was marked outdated. |

Expressing the first two needed a shape change rather than more branches. Install commands are now an ordered candidate list, each candidate carrying its own platform and package-manager guard, with an optional result:

```ts
preferredInstallCommand([
  { command: 'brew install --cask claude-code',           requires: 'brew',   platforms: ['darwin'] },
  { command: 'winget install Anthropic.ClaudeCode',       requires: 'winget', platforms: ['win32'] },
  { command: 'npm install -g @anthropic-ai/claude-code',                      platforms: ['darwin', 'linux'] },
])
```

"Windows without winget" then falls out as no button at all rather than a special case, which also retired the `win32` precedence branch and the `IS_WINDOWS` import.

One deliberate non-change: `winget` stays out of the set that drives `detectPackageManager()`. That set feeds the LaTeX tab's `DEPENDENCY_INSTALL_COMMANDS`, which has no winget entries, so widening it would have silently cost Windows users their `scoop` command.

### Verification

`typecheck` clean, `eslint` clean, `check:dead-code-ratchet` clean (17 vs 17, after un-exporting two symbols and adding `PackageManagerName`). Install-command coverage is now five branches — macOS + Homebrew, Linuxbrew (cask *not* offered), Windows + winget, Windows without winget (no command), and the npm fallback — 15 tests passing.

---

PR #9463, opened 2026-07-30
refactor: extract shared persistTerminalExecution helper

### LINKED ISSUE #9418: Tech debt: extract the duplicated terminal-persistence tail (ExecutionRegistry / AgentRunLifecycle)
**What/where:** `finishWaitingTermination` (executionRegistry.ts:958-997) and `finalizeRunTerminal` (AgentRunLifecycle.ts:151-181) duplicate the `finalizeExecution` try/catch with failed-stage logging, `markOwnedExecutionLeaseUndurable`, and the `terminalStatusPersisted` return.

**Why it's debt:** Amplification in the raciest runtime code. Note the original finding's bigger framing was refuted: the two-arm split is a designed, guarded peer arrangement (documented at AgentRunLifecycle.ts:124-135, enforced by `claimTerminalFinalize`), and its ordering differences are deliberate and tested.

**Clean end-state:** One `persistTerminalExecution(executionId, agentName, outcome, flowRecord)` helper in `AgentRunLifecycle.ts` with exactly two callers. The registry keeps its publish-before-persist ordering, `untrackIfCurrent`, `synchronizeAgentResultOutcome`, and root-lease follow-ups. Do not reroute the waiting arm through `finalizeRunTerminal` and do not delete the injection seams.

**Win:** roughly -30 net LoC, 1 element cut. Zero behavior change; existing vitest ordering/lease tests stay valid verbatim.

**Risk/hazards:** None if scoped exactly as above; every larger version is behavior-changing.

---
Filed from the 2026-07-30 multi-agent tech-debt audit (8 subsystem scouts, dedup/rank, per-item adversarial verification). Numbers are the verifier's corrected estimates.

LionSR:
## Summary

Closes #9418.

`finishWaitingTermination` (`executionRegistry.ts`) and `finalizeRunTerminal`
(`AgentRunLifecycle.ts`) each ran the identical `finalizeExecution` try/catch:
call `finalizeExecution`, mark the owned lease undurable and warn on either a
`'failed'` result or a rejected promise, and hand back whether the terminal
status reached disk. Extracted that block into a single
`persistTerminalExecution` helper in `AgentRunLifecycle.ts` with exactly two
callers.

The two call sites differ in two ways that existing tests assert verbatim:
the warn-log message text, and whether the log `data` includes an
`agentIdentifier` field. So the helper takes `failedMessage` /
`rejectedMessage` and the caller's own channel-trace `logger` as parameters
instead of hardcoding either — this keeps the log channel origin and text
unchanged rather than silently rerouting both call sites' warnings through
`AgentRunLifecycle`'s `agentRunLifecycle` channel.

Scope kept deliberately narrow per the issue: the registry keeps its
publish-before-persist ordering, `untrackIfCurrent`,
`synchronizeAgentResultOutcome` follow-up, and root-lease release in its own
`finally` around the helper call — none of that moved into the helper. The
`claimTerminalFinalize` two-arm terminal split was not touched.

## Net elements (R6)

- +1 export: `persistTerminalExecution` (plus its two small param/result
  interfaces) in `AgentRunLifecycle.ts`.
- -1 duplicated `finalizeExecution` try/catch block (was inlined at both call
  sites, ~24 and ~30 lines respectively).
- Net: one shared implementation instead of two near-identical ones; the two
  call sites shrank from a 20-30 line inline block each to a single call.

## Consumer counts (R8)

- `persistTerminalExecution`: 2 callers — `finalizeRunTerminal` in
  `AgentRunLifecycle.ts` (same file) and
  `ExecutionRegistry.finishWaitingTermination` in `executionRegistry.ts`.
  Both callers ship in this PR.

## Verification

- `npm run typecheck` — full chain (root tsc, test-kernel tsc, agent build,
  cli/trace-viewer/desktop typecheck) passes clean, no new errors.
- `npx vitest run src/test-kernel/agent/runtime/AgentRunLifecycle.vitest.ts src/test-kernel/agent/runtime/ExecutionRegistry.vitest.ts` —
  62/62 pass, unmodified (including the exact-message warn-log assertions
  this refactor had to preserve).
- `npx eslint src/agent/runtime/AgentRunLifecycle.ts src/agent/runtime/executionRegistry.ts` — clean.
- `npm run check:dead-code-ratchet` — 17 findings, unchanged vs baseline.

---

> [!NOTE]
> **Low Risk**
> Refactor-only with behavior preserved via parameterized log messages and caller-owned follow-up; touches durable execution finalization but does not change control flow at either call site.
> 
> **Overview**
> Duplicates the same **`finalizeExecution`** tail—project outcome to terminal status, mark the owned lease undurable and warn on failure or rejection, return whether the terminal status persisted—into **`persistTerminalExecution`** in a dedicated module so **`ExecutionRegistry`** can use it without importing **`AgentRunLifecycle`** (circular dependency).
> 
> **`finalizeRunTerminal`** and **`finishWaitingTermination`** now call the helper instead of inlined try/catch blocks. Call sites still pass their own trace **`logger`**, warn message strings, and optional **`agentName`** for log **`data`**, so existing test assertions on log channel and text stay unchanged. Registry-only behavior (**`synchronizeAgentResultOutcome`**, root lease release, publish/untrack ordering) remains at the call sites; the helper never throws.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 251a11f735ba3eaeff410157ed252df294ca2ba0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/runtime/AgentRunLifecycle.ts
+++ b/src/agent/runtime/AgentRunLifecycle.ts
@@ -25,6 +25,7 @@ import {
   STREAM_PHASE,
   buildRunDescriptor,
   toRetryErrorInfo,
+  type ExecutionId,
   type RunOutcome,
   type StreamTabId,
 } from '@shared/schemas';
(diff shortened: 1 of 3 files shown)


LionSR:
Addressed both review findings and rebased onto current main.

**Circular dependency (finding 1).** Agreed, and taken as more than theoretical: the reviewer's suggested fix is what shipped. `persistTerminalExecution` now lives in its own module, `src/agent/runtime/persistTerminalExecution.ts`, which both `AgentRunLifecycle.ts` and `executionRegistry.ts` import. Neither caller depends on the other, so the graph no longer closes and the TDZ risk from a future move to a synchronous path is gone.

**Truthiness guard (finding 2).** Changed `agentName ? ... : {}` to `agentName !== undefined ? ... : {}`, so an empty agent name can no longer silently drop `agentIdentifier` from the warn data.

Verification after the change: `npm run typecheck` clean, eslint and prettier clean, and the 62 tests in `AgentRunLifecycle.vitest.ts` and `ExecutionRegistry.vitest.ts` still pass unmodified. Those tests staying untouched was the agreed tripwire for scope creep on a change that is meant to be zero behavior change.

Also rebased onto main to pick up #9464, #9465, and #9466.

LionSR:
Round 2 findings addressed.

**Dead imports (finding 1).** Confirmed all three independently: `finalizeExecution`, `markOwnedExecutionLeaseUndurable`, and `projectRunOutcome` each appeared exactly once in `AgentRunLifecycle.ts`, on their own import line, with no remaining call site. All removed; `FinalizeExecutionInput` stays as a type-only import since lines 74 and 85 still use it, and `onOwnedExecutionLeaseLost`/`captureOwnedExecutionLeaseIfPresent` stay because they are still called. Worth noting eslint did not flag these, so the review pass is what caught them.

**Unused `type ExecutionId` (finding 2).** Also confirmed unused and removed rather than justified with a comment. It was added defensively and earns nothing.

Re-verified after the removal: `npm run typecheck` clean, eslint clean, and the 62 tests in `AgentRunLifecycle.vitest.ts` and `ExecutionRegistry.vitest.ts` still pass unmodified.

PR #9459, opened 2026-07-30
fix: unify memory directory traversal

LionSR:
## Summary

- use the symlink-safe memory walker for settings, tool listings, and pinned-memory counting
- preserve the model-facing directory listing format and depth limit
- stop pinned-memory counting at the configured cap
- cover symlink cycles and the exact listing output

## Verification

- `npm run typecheck`
- `npm run lint`
- `npm run compile:fast`
- `npx vitest run src/test-kernel/tools/MemoryTool.vitest.ts src/test-kernel/settings/MemoryFileSystemConcurrency.vitest.ts`

Closes #9428

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Shared filesystem traversal affects pin limits and all memory listings; behavior changes (prefix reads, symlink skip) are intentional but touch multiple user-facing paths.
> 
> **Overview**
> **Unifies memory tree walking** so settings UI, `MemoryTool` directory listings, and pinned-memory limits all use the same `walkMemoryDirectory` async generator in `memoryFileSystem`, instead of separate recursive walkers (including the old full-file-read pin counter in `memoryMeta`).
> 
> The shared walker **skips symlinks**, supports **`maxDepth` / `includeDirs`**, reads **frontmatter via bounded prefix streams** with **concurrency limits**, and **tolerates unreadable files** without failing the whole listing. **`countPinnedMemories`** moves here and **stops scanning once the limit is reached**.
> 
> **`MemoryTool`** drops its private directory walk and builds model-facing listings from walker metadata (preserving depth limit, attribution, and `[pinned]` tags). **Imports** in settings wiring switch to the filesystem module. **Tests** cover concurrency, early pin-count exit, exact listing output, depth bounds, and symlink cycles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 982eb21e98a1c709ec3593e5219b7319eadbf5e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9467, opened 2026-07-30
refactor: collapse SettingsMemoryController pass-through deps and dedupe pin-with-cap

### LINKED ISSUE #9426: Tech debt: collapse SettingsMemoryController pass-through deps and dedupe pin-with-cap logic
**What and where**

- `src/controllers/settingsView/SettingsMemoryController.ts`
- `src/controllers/settingsView/SettingsViewHost.ts`
- `src/tools/memory/MemoryTool.ts`

Verified at SettingsViewHost.ts:76-100: every one of the controller's 9 non-host deps is the core symbol passed through unchanged (memoryStoragePath: resolveMemoryStoragePath, storage: StorageFS, parseMemoryFile: parseFrontmatter, buildMemoryFile: buildFile, setPinnedMeta, countPinnedMemories, maxPinnedMemories: MAX_PINNED_MEMORIES) — a core-to-core injection vocabulary with exactly one production construction site, in a VS Code-free zone where direct imports are legal. Meanwhile the actual pin-with-cap business logic exists twice (controller vs MemoryTool.pin/unpin) with two copies of the cap-reached message that can and do drift.

**Refactor**

Extract one core mutation setMemoryPinned(storagePath, pinned) -> 'changed'|'already'|'cap-reached' in src/tools/memory (two callers, so not a banned single-caller extraction); MemoryTool and the controller both call it and format only their own surface message. Collapse the controller deps to { prompt, isMemoryEnabled, setMemoryEnabled }, import the rest directly, delete SettingsMemoryStorage/SettingsMemoryMeta and the SettingsViewHost wiring block. Tests move onto FakePlatform like the rest of core.

**Estimated win:** roughly -75 net LoC, 12 elements cut.

**Risk:** Low. Single production caller for the deps object; SettingsViewHost.vitest.ts needs rewriting onto FakePlatform; prompt semantics stay injected per host.

---
Filed from the 2026-07-30 seam-focused tech-debt audit (6 scouts over the runtime/SDK surface, runtime-to-UI and runtime-to-persistence couplings, memory/storage, a repo-wide projection sweep, and wasted-effort archaeology; merged and ranked by a judge that spot-checked claims).

**Not adversarially verified.** The audit's per-item verification stage did not run (usage credits exhausted), so the evidence and estimates here are scout-and-ranker level. Confirm the cited file:line claims against current main before starting work.

LionSR:
Closes #9426.

Two separate collapses in the same module.

**Pass-through injection.** Every non-host dependency `SettingsViewHost` injected into `SettingsMemoryController` was a core symbol handed over unchanged: `resolveMemoryStoragePath`, `StorageFS`, `parseFrontmatter`, `buildFile`, `setPinnedMeta`, `countPinnedMemories`, `MAX_PINNED_MEMORIES`. There is one production construction site, and the controller lives in a VS Code-free zone where importing those directly is legal, so the injection bought nothing. The controller now imports them and takes only what is genuinely host-specific: `{ prompt, isMemoryEnabled, setMemoryEnabled }`.

**Duplicated business rule.** The pin-with-cap rule existed twice, once in the controller and once in `MemoryTool.pin`/`unpin`, including two copies of the cap-reached message that could drift apart. Both now call one `setMemoryPinned(storagePath, pinned)` returning `'changed' | 'already' | 'cap-reached'`, and each caller formats only its own surface message.

## Net elements (R6)

Removed: `SettingsMemoryStorage`, `SettingsMemoryMeta`, the `SettingsViewHost` wiring block, and 7 injected dependency fields. Added: one `setMemoryPinned` function and its result type. Net -151/+254 lines across 6 files, with the controller dropping from 85 lines of plumbing.

## Consumer counts (R8)

`setMemoryPinned`: 2 consumers (`MemoryTool` pin/unpin, `SettingsMemoryController`). Not a single-caller extraction.

## Notes

- Prompt semantics stay injected per host; only the core symbols were de-injected.
- This builds on #9459, which unified the memory walkers earlier today, so `countPinnedMemories` is already the shared symlink-guarded implementation and the cap check keeps its early exit.

## Verification

`npm run typecheck` clean. Targeted tests pass: `SettingsMemoryController.vitest.ts`, `SettingsViewHost.vitest.ts`, `MemoryTool.vitest.ts`, 15 tests. Rebased onto current main.

---

> [!NOTE]
> **Low Risk**
> Refactor with behavior preserved via shared `setMemoryPinned`; risk is limited to pin-cap edge cases and settings memory CRUD wiring, covered by existing targeted tests.
> 
> **Overview**
> **Settings memory wiring** is simplified: `SettingsMemoryController` no longer takes injected loaders, storage, path resolution, or frontmatter helpers—it calls `loadMemoryItems`, `loadMemoryPreview`, `resolveMemoryStoragePath`, `StorageFS`, and `setMemoryPinned` directly. Dependencies are reduced to host-only `{ prompt, isMemoryEnabled, setMemoryEnabled }`, and `SettingsViewHost` no longer wires the removed pass-throughs. `SettingsMemoryStorage` / `SettingsMemoryMeta` are removed.
> 
> **Pin/unpin behavior** is unified in `memoryFileSystem` via new `setMemoryPinned(storagePath, pinned)` returning `'changed' | 'already' | 'cap-reached'` (with post-pin count on success). `MemoryTool` pin/unpin and the settings controller both use it instead of duplicating read/cap-check/write logic; each surface still formats its own messages.
> 
> **Tests** for `SettingsMemoryController` switch to Vitest module mocks for the memory filesystem and storage; host tests only construct the slimmer controller deps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 24fab5a2346660e68250af7ff4b413c48a6513ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/controllers/settingsView/SettingsMemoryController.ts
+++ b/src/controllers/settingsView/SettingsMemoryController.ts
@@ -1,39 +1,23 @@
 // Local imports
 import type { PromptHost } from '@hosts/uiHosts';
+import { resolveMemoryStoragePath } from '@platform/defaults/workspaceStorage';
 import { SETTINGS_VIEW_COMMANDS } from '@shared/ipc';
 import type {
   MemoryPreview,
   MemoryViewItem,
 } from '@shared/schemas/settingsViewMessages';
-import type { MemoryFileMeta } from '@tools/memory/memoryMeta';
-
-export interface SettingsMemoryStorage {
-  delete(path: string, options?: { recursive?: boolean }): Promise<void>;
-  read(path: string): Promise<string>;
-  write(path: string, content: string): Promise<void>;
-}
-
-export type SettingsMemoryMeta = MemoryFileMeta;
+import { MAX_PINNED_MEMORIES } from '@tools/memory/constants';
+import {
+  loadMemoryItems,
+  loadMemoryPreview,
+  setMemoryPinned,
+} from '@tools/memory/memoryFileSystem';
+import { StorageFS } from '@utils/files';
 
 export interface SettingsMemoryControllerDeps {
   prompt: Pick<PromptHost, 'confirm' | 'warning'>;
-  loadMemoryItems(): Promise<MemoryViewItem[]>;
-  loadMemoryPreview(storagePath: string): Promise<MemoryPreview>;
   isMemoryEnabled(): boolean;
   setMemoryEnabled(enabled: boolean): Promise<void>;
-  memoryStoragePath(storagePath: string): string;
-  storage: SettingsMemoryStorage;
-  maxPinnedMemories: number;
-  parseMemoryFile(raw: string): {
-    meta: SettingsMemoryMeta | null;
-    content: string;
-  };
-  buildMemoryFile(content: string, meta: SettingsMemoryMeta | null): string;
-  setPinnedMeta(
-    meta: SettingsMemoryMeta | null,
-    pinned: boolean,
-  ): SettingsMemoryMeta;
-  countPinnedMemories(maxCount: number): Promise<number>;
 }
 
 export type SettingsMemoryMessage =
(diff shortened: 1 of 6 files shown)


LionSR:
We should do more like this and clean deeper

LionSR:
Both findings addressed.

**Finding 1, double directory walk.** Confirmed as a real regression this PR introduced, and taken seriously rather than waved through: splitting the cap check into `setMemoryPinned` left `MemoryTool.pin` calling `countPinnedMemories()` again for its display count, so a successful pin walked the tree twice, the second time over every file with no early exit. The pre-refactor code walked once and derived the count arithmetically.

Fixed the way the review suggested: `setMemoryPinned` now returns `{ status, pinnedCount }` instead of a bare string, so the display count comes from the cap-check walk it already performs. The constraint is documented on the function itself, since the failure mode is invisible from the call site:

> `pinnedCount` on a successful pin is the post-write total, derived from the cap-check walk this function already performs. Callers must not walk the tree again to display it: a second `countPinnedMemories()` without a limit loses the early exit and rescans every file.

That also left `countPinnedMemories` as an unused import in `MemoryTool`, now removed.

**Finding 2, untested `'already'` branch.** Added. The controller does treat `'already'` and `'changed'` identically, so nothing was broken, but the contract was unpinned. The new test asserts it refreshes and does not warn.

Re-verified: `npm run typecheck` clean, 17 tests pass across `SettingsMemoryController.vitest.ts`, `MemoryTool.vitest.ts`, and `MemoryFileSystemConcurrency.vitest.ts`.

Issue #9413, opened 2026-07-30
Tech debt: merge the twin extension/desktop HostInteractions pending-map implementations


LionSR:
**What/where:** `packages/extension/src/progressView/extensionHostInteractions.ts` (304 LoC) and `packages/desktop/src/main/desktopHostInteractions.ts` (306 LoC): logic-identical `submit*`/`request*`/`revealStream`/`cancel`/`approvePendingDelegatedWork` over `ApprovalRequestHandlerSet`, synced only by "Mirrors the extension's..." comments.

**Why it's debt:** Layering and amplification. Duplicated approval-broker logic in the raciest UI seam; every interaction change is written twice. The rest of the scouted surface (secondTierActions, controller factories, follow-up appliers) is the deliberate host port #9388 shaped the day before the audit; merging it just renames the seam.

**Clean end-state:** One `createProgressHostInteractions(options)` in `src/controllers/progressView/backend` (VS Code-free, both hosts already import that directory), with a toolEdit port and host extras: extension keeps `notifyUnavailableTools`/`readDiagnostics`/`addCriticism` (its only vscode uses) and native toolEdit; desktop keeps `showInfoMessage` and its toolEdit controller port.

**Win:** roughly -150 net LoC, 2 elements cut.

**Risk/hazards:** Triage two real asymmetries explicitly before merging: `openExternalInquiry` revealStream (extension yes, desktop no) and the `bash-` vs `desktop-bash-` request-id prefixes. Files were churned by #9388 and #9407 this week; let them settle first.

---
Filed from the 2026-07-30 multi-agent tech-debt audit (8 subsystem scouts, dedup/rank, per-item adversarial verification). Numbers are the verifier's corrected estimates.

LionSR:
Resolved by 96dbb2c (#9489, "implement the approved cross-host consistency plan").

Verified at HEAD:

- `packages/extension/src/progressView/extensionHostInteractions.ts` is now **41 lines** (was 304) and `packages/desktop/src/main/desktopHostInteractions.ts` is **22 lines** (was 306). Both are thin wrappers over the new `src/controllers/progressView/backend/progressHostInteractions.ts` (269 lines, `createProgressHostInteractions`, no `vscode` import).
- Host-specific extras land exactly as the issue proposed: extension keeps `readDiagnostics`/`addCriticism`/`notifyUnavailableTools`, desktop keeps `showInfoMessage`.

Both flagged asymmetries were explicitly triaged rather than glossed over:

[...]

PR #9468, opened 2026-07-30
refactor: make settlements the host-interaction result type

### LINKED ISSUE #9429: Tech debt: approval settlement/result dual vocabulary and the SDK's hand-copied kind union
**What and where**

- `src/agent/runtime/HostInteractions.ts`
- `src/agent/runtime/hostInteractionResultMappers.ts`
- `packages/cli/src/runtime/approvalAdapter.ts`
- `packages/agent/src/index.ts`

Bash and userQuestion speak two vocabularies for one decision ({action,feedback} settlements vs {accepted,userMessage} results) bridged by hostInteractionResultMappers.ts (whole 31-line file) at 6 host call sites, while the CLI re-implements the identical conversion a third time (approvalAdapter.ts toBashResult) — and the three sibling kinds (plan/proposal/retry) prove the seam unnecessary by passing settlements straight through; the mapper's own comment records the ext/desktop drift the dual shape already caused. Relatedly, packages/agent/src/index.ts:31-42 hand-copies the 7-member PendingInteractionKind union verbatim (verified), so the published SDK type silently drifts when the next kind lands.

**Refactor**

Owner: the settlement union becomes the result type (HostInteractionResultByKind.bash/userQuestion become action-discriminated, matching retry); move the reject-only userMessage gating to the single agent-facing consumer (buildBashApprovalRejectedResult side); delete hostInteractionResultMappers.ts, both Result types, and the CLI's third copy; rewrite ~8 consumer sites from .accepted to action === 'approve'. In the same territory, replace the SDK's inline union with export type { PendingInteractionKind, ... } from '@agent/runtime/HostInteractions' (the dts alias-rewrite pipeline already supports it). MUST sequence after #9413 lands — that issue rewrites both host files into one factory (which shrinks this change to one call site) and its body explicitly asks for post-#9388/#9407 settling time.

**Estimated win:** roughly -55 net LoC, 7 elements cut.

**Risk:** Low once sequenced after #9413. Preserve the mapper file's documented alignment semantics (userMessage only on reject) exactly; verified the result shape has only in-process consumers, nothing persisted or on the frozen wire.

---
Filed from the 2026-07-30 seam-focused tech-debt audit (6 scouts over the runtime/SDK surface, runtime-to-UI and runtime-to-persistence couplings, memory/storage, a repo-wide projection sweep, and wasted-effort archaeology; merged and ranked by a judge that spot-checked claims).

**Not adversarially verified.** The audit's per-item verification stage did not run (usage credits exhausted), so the evidence and estimates here are scout-and-ranker level. Confirm the cited file:line claims against current main before starting work.

LionSR:
Closes #9429.

Bash and userQuestion spoke a second vocabulary for one decision, `{accepted, userMessage}` and `{submitted, answers, feedback}`, which a dedicated mapper module translated to and from the `{action, feedback}` settlements every other kind already used. The three sibling kinds (plan, proposal, retry) pass settlements straight through, which is the evidence the seam was unnecessary, and the mapper's own comment recorded the extension/desktop drift the dual shape had already caused.

The settlement union is now the result type for all five kinds. `hostInteractionResultMappers.ts` is deleted, along with the CLI's separate third copy of the same conversion in `approvalAdapter`. The reject-only `userMessage` gating is preserved; it moves to the single agent-facing consumer instead of being enforced by the shape.

## Net elements (R6)

Deleted: `hostInteractionResultMappers.ts` (whole module), `HostBashApprovalResult`, `HostUserQuestionResult`, and the CLI's `toBashResult`. Two entries drop out of `config/ratchets/host-agent-import-baseline.json` because the module they pointed at is gone, so the ratchet tightens rather than loosens. Net -257/+128 across 30 files, most of them call sites and tests.

## Consumer counts (R8)

`BashSettlement` and `UserQuestionSettlement` were already the settlement types used by the cancellation factories; they now additionally serve as the result types, so no new exports were introduced. Consumer sites rewritten from `.accepted` to `action === 'approve'`: 8.

## The issue was wrong about one thing

#9429 also proposed replacing the hand-copied 7-member `PendingInteractionKind` union in `packages/agent/src/index.ts` with a re-export from `HostInteractions.ts`. **That is not possible as written, and the duplication there is load-bearing rather than sloppy.** Any re-export from that file drags `@anthropic-ai/sdk` into the published package's type surface through `ModelHandlerContracts.ts`, which the package's own provider-leak validation gate rejects. Confirmed with isolated before-and-after builds: 680 declarations validate clean today, and the re-export produces an explicit "Provider type leaked" failure.

That part is therefore reverted and out of scope; `packages/agent` is untouched by this PR. The honest fix is to extract `PendingInteractionKind` into a provider-independent module first, which deserves its own issue rather than riding along here.

## Sequencing

#9413 (merge the twin extension/desktop HostInteractions implementations) was expected to land first and would have shrunk this to a single call site. It has not, so this proceeds independently and #9413 will need to rebase over it.

## Verification

`npm run typecheck` clean. Targeted tests pass across core, CLI, desktop and extension: `HostInteractionSettlements`, `SessionInteractions`, `ApprovalCleanupScope`, `BashToolErrorFeedback`, `TuiApprovalRetry`, `TuiHostInteractionsCancelForStream`, `DesktopHostInteractions`, `ExtensionHostInteractions`. 108 tests. Rebased onto current main.

---

> [!NOTE]
> **Medium Risk**
> Touches approval plumbing across CLI, desktop, extension, and tool execution; behavior should be equivalent but any missed `.accepted`/`.submitted` site would break approvals at runtime.
> 
> **Overview**
> **Unifies bash and user-question host interaction results** on the same `{ action, feedback? }` settlement shapes already used for plan, proposal, and retry, instead of parallel `{ accepted, userMessage }` / `{ submitted, answers }` result types.
> 
> `HostBashApprovalResult` and `HostUserQuestionResult` are removed from `HostInteractions.ts`; `HostInteractionResultByKind` and the public `requestBashApproval` / `askUserQuestion` promises now use **`BashSettlement`** and **`UserQuestionSettlement`**. **`hostInteractionResultMappers.ts` is deleted**, and desktop/extension no longer call `toBashApprovalResult` / `toUserQuestionResult` when completing approval handlers—they pass settlements through directly.
> 
> CLI paths (TUI `subscribeApprovals`, headless `approvalAdapter`, `askUserQuestionDenial`) build settlements at the edge. **Tools** (`bashApproval`, `bash`, `UserQuestionTool`, agent CLI prelude, etc.) branch on `action === 'approve'` / `action === 'submit'` and map reject feedback via `feedback` instead of `userMessage`. Cancellation defaults for bash/user-question align with explicit rejections (`action: 'reject'`). The host-agent import ratchet drops `@agent/runtime/hostInteractionResultMappers` from desktop and extension baselines.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d25bba2f6701956f93de706862a99ecc94eba51e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Addressed the naming-drift note: `buildBashApprovalRejectedResult`'s second parameter is renamed from `userMessage` to `rejectionFeedback`, since every caller now passes `approval.feedback`. That name was the last residue of the dual vocabulary this branch removes, so leaving it would have shipped a PR that unifies a vocabulary while keeping its vestigial name.

Thanks for the independent confirmations on the load-bearing claims: zero remaining references to the deleted mapper, the ratchet tightening rather than loosening, and `action !== 'submit'` catching exactly what `!result.submitted` did.

Targeted tests still pass after the rename (`BashToolErrorFeedback`, `HostInteractionSettlements`).

PR #9473, opened 2026-07-31
refactor: module-level simplification sweep

LionSR:
## What

A second simplification pass over the whole tree, structural rather than line-level. Round 1 (#9472) worked in fixed-size slices and did the intra-file work: inlining single-caller wrappers, collapsing conditionals, deleting dead branches. This round re-partitions the tree by **module** — each agent got a whole directory subtree (99 batches, median ~4.9k LoC, max ~14k, all 2,434 files covered exactly once) specifically so it could see across files.

Agents were told the small stuff was already done and to look instead for dual systems, the same helper reimplemented across a module, logic sitting in the wrong layer, shallow modules whose interface rivals their implementation, data-model problems patched downstream instead of at the source, and if/else chains that are really tables. Fleet was 61 Opus on core and large modules, 38 Sonnet 5 on the rest.

**Net -1,836 lines across 310 files.** Fewer files touched than round 1 and a smaller delta, which is the expected shape: this pass was looking for a smaller number of larger structural wins, and a good number of modules were correctly left alone.

## Not in scope

No feature changes, no API changes, no new abstractions, no coverage removed. Structural problems whose fix reached outside a batch were reported rather than acted on.

## Verification

| gate | result |
| --- | --- |
| `npm run typecheck` | clean except one pre-existing error, see below |
| `npm test` | 7,949 passed, 0 refactor-caused failures |
| `npm run lint` | pass |
| `npm run check:dead-code-ratchet` | 17 findings vs 17 baselined, no new ones |

**The one typecheck error is not from this PR.** `modelHandlerGoogleInteractions.ts:1890` fails with `'temperature' does not exist in type 'GenerationConfig_2'`, because #9450 bumped `@google/genai` and the Interactions `GenerationConfig` dropped `temperature`. The offending code is byte-identical on main, and main's own CI is red on 4b3e8e6b0a (`static checks`, `validate`). Deciding what replaces `temperature` in the Interactions request is an API-behavior call and belongs in its own PR, not buried in a 310-file refactor.

**Test failures are pre-existing.** The full suite fails a handful of files per run, but the failing *set* changes between runs and every one is a hardcoded-timeout expiry (a 30s `beforeAll` import, a 20s test) on a slow machine. Each was re-run in isolation and passed; for each, both the test file and its target module are untouched by this diff. One of them, `AgentsCommand.vitest.mts`, carries a comment already acknowledging the problem: "warm the module graph so the first test isn't charged the import cost, which can exceed the per-test timeout on slow machines."

## Regressions caught and fixed before this landed

Independent adversarial review ran per area, weighted toward this round's characteristic failure mode — bad consolidation, where two call sites get merged onto a helper whose semantics differ from one of them. Eight real items were found and fixed:

- **`listDesktopWorkspaceFiles` swapped a falsy guard for a default parameter.** Callers inject a `getWorkspacePath` returning `undefined` to mean "no workspace"; a default parameter fires on `undefined`, so that answer was discarded and the process-wide workspace read instead. Latent today (no production caller injects), but it inverted the contract of the very injection point that exists to express it.
- **`getGitInfo` tightened its git probe.** It moved onto the shared `isGitRepository`, which additionally requires `stdout === 'true'`. `git rev-parse --is-inside-work-tree` exits 0 and prints `false` in a bare repo and inside `.git`, so those workspaces silently lost their branch/dirty block and the agent started being told to avoid git history checks in a repo where `git log` works. Restored the exit-status-only probe; the `timeoutMs` parameter added to `isGitRepository` for this caller is reverted with it, since nothing else passes one.
- **A latex notification switch lost its `default`.** The absorbed implementation fell back to showing the message for an unrecognized severity; the consolidated switch returns `Promise<void>`, so a non-exhaustive switch is not a type error and a future fourth severity would render nothing at all.
- **Two auth log lines were merged into one prefix.** A security-relevant "session user mismatch" in the device-code flow started logging under `[AUTH]` instead of `[DEVICE]`, so filtering on `[DEVICE]` no longer surfaced it and the two flows became indistinguishable in logs. The flow's log tag is now threaded through the shared mint function.
- **`SUPABASE_SERVICE_ROLE_KEY` was newly exported with no consumer** outside its own module — the most sensitive value in the file, widened for nothing. Back to module-local.
- **An `it.each` merge weakened a negative assertion.** Two cases asserting `plan.rootAgent` is undefined became `plan.rootAgent?.name` toBe `undefined`, which also passes when a root *was* inferred but happens to have no name — exactly the regression those rows exist to catch.
- **A test stopped testing a production default.** `PlanApproval`'s feedback placeholder was exercised through the production default; after consolidation the test declared its own copy of the string, so changing the real constant would leave the test green. The constant is now exported and imported.

Two findings were reviewed and deliberately not acted on: `groupByKind` and two `appInteractionPolicy` helpers are single-caller by the letter of the abstraction rule, but each is a file-local named function sitting beside existing peers in the same style, and inlining a documented loop back into `willUpdate` would cost more clarity than the rule buys.

---

> [!NOTE]
> **Medium Risk**
> Large cross-cutting refactor with several consolidation fixes already called out in the PR (workspace path injection, git probe, logging tags); behavior-sensitive areas include TUI approvals, session run start, and desktop IPC routing.
> 
> **Overview**
> This is a **module-scoped structural refactor** (round 2 after #9472): same behavior and APIs, fewer duplicated paths and shallow wrappers.
> 
> **CLI chat / TUI** consolidates repeated run-start setup via `beginRunContext`, moves approval-row grouping and workflow-child control checks into `appInteractionPolicy`, and drops slash-command `formName` in favor of `formComponent` only. List pickers (`/agent`, `/model`) share `useAsyncPickerForm` and shared key-hint helpers; approval modals share `confirmCardFeedbackRows` and `KeyHint` typing. Shell completion scripts pull agent/model listings from centralized `COMPLETION_SOURCES`; auth and related commands use `booleanArg` and `cliProgressWriter`.
> 
> **Desktop** unifies renderer overlay posting (`tryShowInRenderer`), confirmation dialogs, workspace file listing, and inbound `message` routing into a single route table; Ink wizards use `async`/`await` instead of manual `Promise` wrappers. E2E helpers gain `createIsolatedProfile` and settings-tab waits on the active panel.
> 
> **Extension** adds `runGuardedLatexCommand`, simplifies pack/clean missing-output events around one `FileOpTarget`, and inlines command ID strings at registration sites.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42d238a26d63eec154c873d1b4fbef8f8cf8bac9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Bugbot's finding is correct and is fixed in 42d238a26d.

The deleted local `basename` helper ended in `?? filePath`, so `/` normalized to an empty string and fell back to `/`. `getBasename('/')` returns `''` with no fallback, which would have rendered a blank project name in the sidebar. `workspaceName` now falls back to the path itself.

Worth noting this is exactly the failure mode this round was most likely to produce — two implementations that look interchangeable until you diff their edge cases — so thanks for catching it.

LionSR:
On the second Bugbot finding (`AgentListForm` hardcoding `hasItems`): real inconsistency, but not a regression from this PR, so I'm leaving it.

On main, `AgentListForm` rendered `props.selectable ? KeyHints[navigate, select] : KeyHints[...]` with no empty-list check at all. Hardcoding `hasItems` reproduces that behavior exactly. The divergence from `ModelListForm` (which passes `hasItems={items.length > 0}`) predates this change; the consolidation onto `PickerKeyHints` just made it visible.

Changing it would alter behavior in a PR whose contract is that it doesn't, so it belongs in a follow-up rather than here.

PR #9477, opened 2026-07-31
refactor: subsystem-level simplification sweep

LionSR:
## What

Third simplification pass, and the first one above the module line. Rounds 1 (#9472) and 2 (#9473) worked in fixed-size slices and then in single modules. This round re-partitions the tree into **34 subsystems** (median ~21k LoC, max ~26k, all 2,435 files covered exactly once) so each agent spans many modules at once. Fleet was 30 Opus / 4 Sonnet 5, all at high reasoning effort.

**Net -593 lines across 194 files**, with 10 new files — the shape you'd expect from a round that relocates code to better homes rather than only deleting it.

## The point of the wider span

Round 2's agents were each asked, at the end, to describe any structural problem whose fix reached outside their batch. **93 of 99 had one.** Those reports were harvested and attached to whichever round-3 subsystem now owns the relevant code, so every one of the 34 batches started from a concrete list rather than cold. Leads were framed as unverified observations, with agents told to confirm each against current code, discard the wrong ones, and say which were wrong — the refutations being useful signal about the previous round's judgment.

A representative confirmed lead: `MemoryConfigProvider` in the published `@texra-ai/agent` package hand-rolled the same key-resolution loop as `JsonConfigProvider`, and round 2 could not fix it because the shared helper lived across a package boundary. Round 3 confirmed the lead's technical claim (a non-generic arrow property cannot satisfy a generic `get<T>` method slot), wrote the adapter that closes the gap, and consolidated the two onto one lookup algorithm.

Equally useful were the refusals. Several subsystems reported "already coherent, here is the evidence" and changed nothing — for instance `nodePlatform()` vs `createNodePlatform()`, which look like a dual system but are not: one builds a deliberately ephemeral, filesystem-free platform for standalone npm consumers, the other requires host-supplied stores, and they already share every genuinely identical part by import.

## Not in scope

No feature changes, no API changes, no coverage removed.

## Verification

| gate | result |
| --- | --- |
| `npm run typecheck` | pass |
| `npm test` | 794 files, 7,986 tests, 0 failures |
| `npm run lint` | pass |
| `npm run check:dead-code-ratchet` | 17 findings vs 17 baselined, no new ones |

All three gates passed on their first convergence round, including the test gate — an improvement over round 2, where separating real breakage from this machine's pre-existing import-timeout flakes took a dedicated investigation.

## Regressions caught and fixed before this landed

Adversarial review ran per area with a new `bad-move` category, since relocating code can change *when* it runs rather than only where it lives. Seven real items were found and fixed:

- **A host guard was deleted, again.** An agent removed the `isDesktopHost` gate around `<latexdiffs-section>` in `MainApp.ts` — the identical regression round 1 produced. On the desktop host this renders a LaTeX-diff panel wired to `LATEXDIFF`/`LATEXDIFFVC` commands that `packages/desktop` implements no handler for, so the controls silently do nothing. No test covers the gating. Reverted. That two independent rounds converged on the same deletion suggests the guard reads as dead code from inside the file; a desktop-gating test would be worth adding separately.
- **`settingsUi.openExternal` was pointed at `previewHost.openExternal`.** Not the same function: the preview host wraps the call, shows an error dialog on failure, and rethrows a *rewrapped* error. The sole consumer does not catch. Restored the plain wrapper.
- **A VS Code disposable was dropped as redundant.** `context.subscriptions.push({ dispose: () => UsageLogService.dispose() })` was deleted in favor of the shutdown-phase hook, but those are not equivalent: VS Code skips `deactivate()` when `activate()` throws, while subscriptions are disposed regardless — and the activation body after `initialize()` can throw. Restored.
- **A test leaked a session-hub subscription.** `recorded.detach()` sat inside the `try` instead of the `finally`, on a hub shared by every test in the file, so a throwing run would leave the subscriber attached for the rest of the suite. The sibling test in the same file already did this correctly.
- **A shared JSDOM harness was not a superset of what it replaced.** Two suites migrated onto `useLitComponentTestDom`, whose `domGlobalKeys` omits `DocumentFragment` that their hand-rolled setup installed. Added, to both the key list and the replacements map.
- **A shared skill fixture flattened per-skill bodies to one constant**, which would make a loader regression that attributed one package's content to another indistinguishable. Body restored at the call site.

Deliberately not acted on: four reviewers independently flagged that replacing minimal trace stubs (`{ emit: vi.fn() }`) with the fuller `spiedTrace()` removes an implicit tripwire — an unstubbed trace call used to throw loudly and now no-ops. No asserted behavior regresses, and the fixture consolidation is otherwise sound, so this is noted rather than reverted.

---

> [!NOTE]
> **Medium Risk**
> Touches CLI slash dispatch, follow-up batch consumption/acknowledgment, and model-handler concurrency guards—behavior-sensitive paths—but the PR is refactor-only with tests updated and no intentional API changes.
> 
> **Overview**
> This pass **deduplicates and recenters** behavior that had drifted across hosts, without changing product features.
> 
> **CLI chat** drops the giant `switch` in `handleTuiSlashCommand` in favor of registry **`handler`** callbacks (renamed from `argHandler`); session commands like `/help`, `/status`, and `/compact` move into `sessionCommands.ts` and register beside the rest in `registerBuiltins`. One-shot Ink flows (login picker, init, onboarding, orchestration, config TUI) share **`renderCliPrompt`**, including first-resolution-wins and visible vs scrollback cleanup.
> 
> **Node agent platform** maps in-memory config stores through shared **`ConfigKeyValueStore`** / **`firstStoredValue`** instead of a private lookup loop on `MemoryConfigProvider`.
> 
> **Agent runtime** centralizes YAML tool resolution in **`resolveAgentSettingTools`**, follow-up append + transcript logging in **`applyFollowUpBatch`** (used from tool-use round prep and wait nodes), and single-turn **`withSingleTurnGuard`** on OpenAI Response / Google Interactions handlers. Output prefill/scratchpad handling moves into **`ModelHandler`**; **`max_tokens_reduced`** logging has one **`logMaxTokensReduced`** path. Chat export attachment labels align with **`ExportAttachmentType`**. **`runUsage` / `sessionUsage`** sit on the base stream state schema for both workflow and tool-use kinds.
> 
> **Desktop** routes settings via **`postDesktopSettingsRoute`** and **`buildDesktopSetRouteMessage`**; command-palette shortcuts come from **`desktopCommandPaletteShortcut`**. Embedded browser external handoff is renamed/clarified vs app navigation policy.
> 
> **Extension** folds **`runGuardedLatexCommand`** into **`activeFileGuards`** as **`runGuardedFileCommand`**; compare/accept arg schemas live with registry handlers; Overleaf clone uses **`platform().secrets`**. Progress view side-effect imports register Lit components; settings send helpers are narrowed to private where appropriate. Agent prompt contract tests merge into **`AgentPromptContracts.vitest.mts`**.
> 
> **Misc:** `relayTierConfigUrl`, `getVisibleModels` colocated with model options computation, skills import split (`runtimeSkills` / `skillSources`), test **`installPlatform`** helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b65da141406786be1cc1c7f0abfdf9aaa4876b44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Bugbot's finding is correct and is fixed in the latest commit.

`renderCliPrompt` used `resolved ??= value`, which does not latch when the first resolution is `undefined` — and `undefined` is exactly how callers signal cancel (`loginProviderPicker` and `runInitWizard` both do). A later stray resolve during exit could therefore overwrite a cancel with a committed choice, contradicting the function's own doc comment about first-wins. Replaced with an explicit `hasResolved` latch.

PR #9475, opened 2026-07-31
refactor: consolidate duplicate logic and adopt native/shared helpers

LionSR:
## Summary

A scheduled sweep of the codebase for consolidation opportunities: duplicate logic across 2+ call sites, and hand-rolled implementations of things a native method or an existing shared helper already covers. Four parallel scans (`src/agent`, `src/tools`/`src/model`/`src/utils`/etc., `packages/extension`, `packages/cli`/`packages/desktop`/`packages/trace-viewer`) turned up eight concrete findings, all applied here. `src/agent/**` came back clean — no findings survived the "real, multi-call-site impact" bar there; that scope is excluded from this PR.

- Add `groupBy()` to `@utils/core` and use it at 5 hand-rolled `Map` bucketing sites, including `groupPendingApprovalsByRow` in `packages/cli/src/chat/tui/appInteractionPolicy.ts` (rebased onto main's own `pendingApprovalsForRows` extraction — see rebase note below), `state/streamViews.ts`, `panes/StaticConversationTranscript.tsx`, and `packages/desktop/src/renderer/desktopCommandPalette.ts`.
- `packages/desktop/src/desktopTaskShell.ts`: drop the local `clamp()` reimplementation, use the shared one from `@utils/core` (already used elsewhere in the same package).
- `packages/cli/src/chat/tui/state/subscribeRuntimeHost.ts`: replace the hand-rolled array-equality check (`sameStringList`) with `node:util`'s `isDeepStrictEqual`, matching the existing pattern in the sibling file `subscribeStreamLog.ts`.
- `packages/extension/src/commands/latex/{figCommands.ts,imageCommands.ts}`: replace manual `.slice(0, 100) + '...'` truncation with the existing grapheme-aware `truncateWithEllipsis()`. **Minor observable change** (flagged by review): the shared helper uses a single-character Unicode ellipsis (`…`) instead of three ASCII dots, and only appends it when the string is actually truncated (the old code appended `...` unconditionally, even to short strings). For `figCommands.ts` it also avoids splitting a multi-byte grapheme mid-truncation. Both are improvements, not regressions, but they are technically observable output changes in a debug log / quick-pick detail string — not pure no-ops.
[...]

PR #9481, opened 2026-07-31
refactor: unify cross-host handling of shared concerns

LionSR:
## Summary

Fourth simplification pass. The first three partitioned the tree by directory — files, then modules, then subsystems — so no agent could ever see how the CLI, the VS Code extension and the Electron desktop app each solve the same problem. This round partitions by **concern**: 12 concerns, 25 batches, each holding the CLI, extension, desktop and core code for one concern together (1,181 files, 278k LoC, no overlap).

The diff is deliberately small. Most cross-host duplication turned out to be either legitimate product difference or blocked by a seam costing more to move than the duplication costs to keep. The larger deliverable is the census below.

## Issue acceptance gates

No issue. This is a scheduled refactoring round, not issue-driven work. Follow-ups it produced are filed as #9482, #9483, #9484, #9485.

## Single source of truth

Each change moves a decision to the module that already owned it, rather than creating a new owner:

- **`lineChanges` / `userPatch` on a tool-edit approval** — `finalizeApprovalResult` (`src/tools/approval/toolEditApproval.ts`) is now the sole producer. The extension pre-filled both before returning, so core's own derivation never ran.
- **Disabled tool IDs** — `getDisabledToolIds` / `setToolEnabled` (`src/utils/config/constants.ts`) own the read and the matching write against one store.
- **Execution list ordering** — `listExecutions()` (`src/agent/storage/executionListing.ts`) already returns newest-first; `historySlice.ts`'s client-side re-sort was dead re-derivation.
- **Agent config validation on history load** — `ExecutionKVStore.readConfig()` validates via `readValidated`; `historyHandlers.ts` re-parsed the same schema.
- **Webview post failures** — `ProgressBackend.sendMessage` and `WebviewBridge.deliver` each already handle rejection; `ProgressViewProvider`'s third `try/catch` swallowed the failure they now surface.

**No abstraction was added.** The round's hard constraint was that reconciling N host implementations must never produce an N+1st thing wrapping them. Permitted outcomes were: delete implementations and move callers, extract a pure leaf with no policy or dispatch, or change nothing and describe the problem. Reviewers were given a dedicated `middle-layer-added` category to catch violations; **zero were found**.

Duplication removed: host-local recomputation of approval line-change stats and user patches (two hosts), a client-side re-sort duplicating storage ordering, a redundant schema re-parse, a redundant third error-handling layer, and `optionalAction`/`requiredAction` collapsed into one `action` in `desktopCommandSurface.ts` now that every action is required.

Six problems were identified whose only available fix would have been a middle layer. They were **recorded, not built** — the largest is filed as #9485.

## Net elements (R6)

| | |
| --- | --- |
| Files changed | 103 (101 modified, 2 deleted, 0 added) |
| Net LoC | **-274** (871 insertions, 1,145 deletions) |
| Exported symbols | **-7** (19 added, 26 removed, by `^[+-]export` over the diff) |
| Classes / interfaces / enums | 0 net added |

The commit message's "-284" was stale; **-274** is authoritative and matches `git diff --shortstat origin/main HEAD`.

## Consumer counts (R8)

Grepped over `packages/` and `src/`, excluding `dist/`:

| symbol | action | remaining consumers |
| --- | --- | --- |
| `selectBaseFile` | deleted | **0** |
| `refreshInputFiles` | deleted | **0** |
| `refreshBaseFiles` | deleted | **0** |
| `metadataToStreamStatePartial` | un-exported | **4**, all inside its defining file (`streamStateMerge.ts`), 0 external |
| `getStreamState` | removed from `progressState.ts` | **19**, none referring to the deleted export — 14 in `ProgressBackend.vitest.ts` and 5 in `src/controllers/progressView/` + desktop, all resolving to the surviving owner |

No emit path or subscribe surface was deleted or re-routed; no `bus.emit` sites were added or removed.

## Host impact

**Extension:** stops pre-computing `lineChanges`/`userPatch` before returning approval results, so core's derivation now runs. This is a **behavior change**, disclosed below. Loses a redundant `try/catch` around `webview.postMessage` whose failure the two callers already handle.

**Desktop:** file-dialog filter now strips the leading dot from extensions, so the "Supported files" filter actually matches — Electron matches literally and `getFileListConfig` returns `.tex`-style entries, so it previously matched nothing. This is a **behavior change**, disclosed below. Tool toggles now read and write the same store. `SELECT_MULTIPLE_FILES` for `output` now warns instead of silently reporting the message handled.

**CLI:** no behavior change. Participated in the census; its divergences were either fixed in core or recorded.

## Scheduled refactoring

Filed from this round's census:

- #9482 — transcript secret redaction happens only in the CLI (security-relevant)
- #9483 — CLI agent resolution can launch the wrong agent when a custom workflow agent shares a builtin name
- #9484 — delegated child model selection has two owners in core disagreeing on the empty and unavailable cases
- #9485 — post-credential-change invalidation encoded three times, one per host

Remaining: 174 unfixed bad surfaces and 156 `drift-left` divergences catalogued but not filed.

## The census

The primary output. Agents classified every host divergence found, whether or not they touched it:

| verdict | count |
| --- | --- |
| by-design | **145** |
| drift-left | **160** |
| drift-fixed | 31 |

Plus **216 bad surfaces**, 38 fixed here, 6 flagged as middle-layer-only.

The `by-design` count matters as much as the fixes: 145 differences an agent could have "unified" and correctly did not. The CLI keeping AUTO-BASH and AUTO-APPROVE separate while the extension has one YOLO shield is a product decision, and the shared machinery already models it — `streamApprovalQueue` tracks a separate ancestry graph per bypass kind so a parent with bash granted and edits gated propagates that split to delegated children.

## Behavior changes, kept deliberately

Two changes are genuine behavior changes rather than pure refactors. Both are the drift-fixes this round existed to find, so they are called out rather than buried:

1. **`lineChanges` on an approval the user edited before accepting** now reports original-vs-applied on all three hosts. The extension alone previously reported original-vs-*proposed*, i.e. the model's numbers rather than what was written.
2. **The desktop file-dialog filter** now actually filters, as described under Host impact.

## Regressions caught and fixed before merge

- **A reachable guard was dropped from a shared path.** `requestEnsureProgressView` was moved out of its `if (streamId)` branch, so a streamless approval would fire it — and desktop handles that event by routing the window to the progress view, yanking the user off their current route. Restored.
- **`null` and `undefined` became two meanings of "absent".** `getDisabledToolIds`/`setToolEnabled` gained `store: StateStore | null = tryGlobalState()`. A default parameter fires only on `undefined`, so `undefined` wrote to the ambient store while `null` silently dropped the write, both returning success. `null` is out of the type.
- **A silent drop was enshrined by a new test.** Desktop reported `SELECT_MULTIPLE_FILES` for `output` as handled and did nothing; a new test asserted exactly that. It now warns.
- **A core import was replaced by a local copy**, orphaning `REVEAL_TIMEOUT_MS` and tripping the dead-code ratchet. Restored.

## Validation

| check | result |
| --- | --- |
| `npm run typecheck` | pass |
| `npm test` | 793 files, 8,008 tests, 0 failures |
| `npm run lint` | pass |
| `npm run check:dead-code-ratchet` | 17 vs 17 baselined, unmodified |
| host `@agent/*` deep-import ratchet | pass |
| `src/test-kernel/architecture/` | 10 files, 68 tests, pass |

All gates passed on their first convergence round. None of the anticipated cross-host breakages materialised: no `vscode` import reached a VS Code-free zone, no new `@agent/*` deep import appeared in a host package, no import cycle was created.

Rebased onto `bb2137a2d3` (#9475), taking main's `createSessionBypassAccessors` on the one conflicting hunk.


LionSR:
Both review points are addressed.

**Template compliance.** You were right, and this was wrong on all four PRs in this series, not just this one. The body now carries every required section: `## Issue acceptance gates`, `## Single source of truth`, `## Duplication and abstraction budget`, `## Net elements (R6)`, `## Consumer counts (R8)`, `## Host impact`, `## Scheduled refactoring`, `## Validation`.

R6, measured rather than asserted: 103 files (101 modified, 2 deleted, 0 added), **-274 net LoC**, **-7 exported symbols** (19 added / 26 removed by `^[+-]export` over the diff), 0 net classes/interfaces/enums.

R8, grepped over `packages/` and `src/` excluding `dist/`, for exactly the symbols you named:

| symbol | action | remaining consumers |
| --- | --- | --- |
| `selectBaseFile` | deleted | 0 |
| `refreshInputFiles` | deleted | 0 |
| `refreshBaseFiles` | deleted | 0 |
| `metadataToStreamStatePartial` | un-exported | 4, all inside its defining file, 0 external |
| `getStreamState` | removed from `progressState.ts` | 19, none referring to the deleted export |

You also caught the **-274/-284 inconsistency**: -274 is authoritative (`git diff --shortstat origin/main HEAD`); the commit message figure was stale and the body now says so explicitly.

**Stale annotation.** Fixed in 19052c9bac — the private-method cast in `SupabaseAuthProvider.vitest.ts` no longer declares the `notify` parameter this PR removes. Typecheck clean, suite green (9 tests in that file).

Thanks for the independent verification of the functional diff, particularly confirming that `finalizeApprovalResult` fills `lineChanges`/`userPatch` for early-resolved results too — that is the load-bearing assumption behind dropping the per-host fallbacks, and it is better to have it checked than asserted.

PR #9495, opened 2026-08-01
test(trace): restore strict minimal doubles

### LINKED ISSUE #9478: Trace test doubles: spiedTrace() is strictly more permissive than the minimal stubs it replaced
Deferred from the round-3 simplification sweep (#9477), where four independent reviewers flagged it.

## What changed

Several suites replaced deliberately minimal trace stubs with the shared `spiedTrace()` helper:

- `src/test-kernel/agent/runtime/ExecutionRegistry.vitest.ts` — was `{ emit: vi.fn() } as unknown as AgentTrace`
- `src/test-kernel/agent/output/outputOperations.vitest.ts` — was `{ warn } as unknown as AgentTrace`
- `src/test-kernel/agent/output/XmlOutputManager.vitest.ts` — was a local `createTraceStub()` exposing only `debug/info/warn/domain`

`spiedTrace()` spreads `noopTrace` and adds `vi.fn()` spies for every member.

## Why it is worth a look

The minimal stubs were an implicit tripwire. A stub with only `emit` asserted, by construction, that the code path under test touches nothing but `emit` — any other trace call threw a `TypeError` and failed the test loudly. With the full spy object those calls silently no-op.

Concretely: if a future change adds a `trace.warn(...)` or `trace.error(...)` to the kill/publish path, the old fixture failed and the new one passes in silence.

## Why it was not reverted in #9477

No asserted behavior regresses today, every existing assertion still holds, and the fixture consolidation is otherwise sound — reverting it inside a large refactor PR would have traded one problem for another. But the repo treats silent degradation as a defect, and this is that pattern in test scaffolding.

## Possible directions

- Give `spiedTrace()` an opt-in strict mode that throws on members the caller did not explicitly provide, and use it at these three sites.
- Or keep the permissive default and add explicit `expect(trace.warn).not.toHaveBeenCalled()` assertions where the tripwire mattered.

Not urgent; filing so it is not lost.

LionSR:
## Summary

- add an opt-in strict mode to the shared `spiedTrace` test helper
- make strict doubles throw when production code reads a trace member the test did not explicitly permit
- restore the former minimal-double contracts in the execution registry, output operation, and XML output suites
- add direct regression coverage for the strict-access tripwire

## Reasoning

The shared permissive trace double made every `AgentTrace` member available as a silent no-op. In tests that previously used deliberately minimal stubs, this removed a useful structural assertion: a new logging or event call could pass unnoticed even when the test intended to permit only a small trace surface.

Strict mode retains the convenient shared helper while requiring the caller to name each allowed member. The default remains permissive for suites that need a general-purpose trace. The three affected fixtures now state their exact contracts: `emit` for waiting-execution publication, `warn` for recoverable output failures, and `debug`/`info`/`warn`/`domain` for XML output management.

## Validation

- `npm run format`
- `npm run compile:fast`
- `npm run typecheck`
- `npm run lint`
- `npm test` (8,121 passed, 4 skipped)
- focused strict-trace and affected-suite run (114 passed)
- dead-code ratchet unchanged at 17

Closes #9478.

---

> [!NOTE]
> **Low Risk**
> Test-only helper and fixture changes with no production runtime impact.
> 
> **Overview**
> Adds **opt-in strict mode** to the shared `spiedTrace` helper so tests can enforce a minimal `AgentTrace` surface: when `{ strict: true }`, a `Proxy` throws if production code reads any member not listed in the `overrides` object (default behavior stays permissive via `noopTrace`).
> 
> **Execution registry**, **output operations**, and **XML output manager** fixtures now use strict doubles with explicitly named members (`emit`; `warn`; `debug`/`info`/`warn`/`domain` respectively), restoring the old “only these trace calls are allowed” contracts. A small **`spiedTrace.vitest.ts`** regression test covers the tripwire.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b5826c7b29bacce0da23a72b75bb064a2d6a4a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/agent/output/XmlOutputManager.vitest.ts
+++ b/src/test-kernel/agent/output/XmlOutputManager.vitest.ts
@@ -48,12 +48,23 @@ function createXmlManager(
   inputFiles: string[] = ['paper.tex'],
   options: XmlManagerOptions = {},
 ): XmlOutputManager {
+  const logger =
+    options.logger ??
+    spiedTrace(
+      {
+        debug: vi.fn(),
+        info: vi.fn(),
+        warn: vi.fn(),
+        domain: vi.fn(),
+      },
+      { strict: true },
+    );
   return new XmlOutputManager(
     {
       inputFiles,
       outputFiles: options.outputFiles ?? [],
     } as unknown as AgentConfig,
-    options.logger ?? spiedTrace(),
+    logger,
     new TaskRunFileService('xml-output-manager-test'),
   );
 }
--- a/src/test-kernel/agent/output/outputOperations.vitest.ts
+++ b/src/test-kernel/agent/output/outputOperations.vitest.ts
@@ -38,7 +38,7 @@ describe('tryOperation', () => {
       loggedType,
     }) => {
       const warn = vi.fn();
-      const logger = spiedTrace({ warn });
+      const logger = spiedTrace({ warn }, { strict: true });
 
       const result = await tryOperation(
         async () => {
--- a/src/test-kernel/agent/runtime/ExecutionRegistry.vitest.ts
+++ b/src/test-kernel/agent/runtime/ExecutionRegistry.vitest.ts
@@ -372,7 +372,7 @@ describe('executionRegistry', () => {
       'parent-waiting-kill-publish-result-test' as StreamTabId;
     const childStreamId =
       'child-waiting-kill-publish-result-test' as StreamTabId;
-    const trace = spiedTrace({ emit: vi.fn() });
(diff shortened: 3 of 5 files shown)


LionSR:
Review note on the XML fixture: `warn` is retained intentionally. Issue #9478 identifies the former local minimal double's contract as `debug`/`info`/`warn`/`domain`; this change restores that known contract while preventing access to every other `AgentTrace` member. Tightening the historical contract further would be a separate test-design change rather than restoration of the lost tripwire.

Issue #10, opened 2024-10-10
A better interface/management system for agents


LionSR:
Looking forward, I'd like to make the reflection-based agent/prompt chain pattern more extensible to support some exciting patterns from recent academic papers (MCTS or multi-agent debate).

- right now there are two kinds of agents: DirectWrite Agent and ThinkAndWrite Agents, both are child of the reflection agent.
- there should be some example agents provided and customizable by the users (and can be reset to default)
- in the main repo, some self-used agents (such as the agents for correcting the work-in-progress book) should not be there [better to use the global storage of vs code now I think, but it should be possible to be imported. How to?)
- it should be easy to add/delete agents. there should be an easy way to import/export agents. Right now it is managed by the visible arguments in vs code. VS code prefers JSON.
- in the future when the backend is remote, users should be able to customize agents without having needed to write any codes. xml pseudo codes are acceptable. Users should also be able to define what other files other than input/reference/sample/aux files need to be extracted in a code free way.
- maybe another github repo/market page where people can share agents 
- maybe this should be managed by a lightweight database such as sql-lite

LionSR:
yaml is the answer

LionSR:
<img width="941" alt="image" src="https://github.com/user-attachments/assets/0bce459b-e2b0-480b-8868-609cc3801d7c">

closed by #45

PR #9523, opened 2026-08-01
docs: open-source readiness audit, backend decoupling plan, directory survey

LionSR:
**Split as requested.** This PR is now documents only — 7 files. All code moved to three independently reviewable PRs:

| PR | Contents |
|---|---|
| #9535 | Workflow hardening — `claude.yml`, `allowed-tools.json` |
| #9536 | Package/alias fixes — trace-viewer deps, three dead aliases, build-map gate |
| #9537 | The five relocations plus every importer, test-literal and doc-link rewrite |

Each verifies standalone (`npm run typecheck` clean across all seven projects on each branch, plus the relevant gates).

## What's left here

- **`2026-08-01-open-source-readiness-audit.md`** — eleven parallel audits with adversarial verification; 90 findings survived, 33 refuted and recorded so they aren't re-proposed. Supersession headers now scope the three conclusions it disproves in `2026-07-29-open-source-readiness.md`.
- **`2026-08-01-backend-decoupling-plan.md`** — revised substantially in review (see below).
- **`2026-08-01-directory-organization.md`** — the tree is structurally sound; the problem is legibility and written rules that no longer match it.
- **`AGENTS.md`**, **`src/README.md`**, **`check-guidance-refs.mjs`** — two AGENTS.md rules that were measurably false against the tree, plus a `src/` index (there was none; the only four READMEs are under `src/agent/`).

## Backend plan changes from review

Per-facet availability gates instead of one proxy fact · Port A owning `apiAccessMode`/`quotaAutoSwitched` rather than Port C duplicating them · `AccountPlane` split into a shared schema and a controllers-side runtime registry, rather than letting the edge ratchet site a service locator in the browser-reachable zone · one `BACKEND_MODULES` array deriving both lint rules · the `@supabase/supabase-js` removal moved out of step 4 · step 10 rescoped against the **36** suites that actually import the modules it deletes.

The contract-ownership section now addresses module identity head-on and recommends factory exports over a shared singleton — a private package registering into a public singleton needs the same module *instance*, which SHA-pinned CI does not provide.

## Still open — and this PR should not merge until they're resolved

Your four substantive objections stand, and I have not tried to paper over them:

1. **The audit reproduces confidential correspondence and a working relay-accounting weakness.** I still don't know what "closing" these means beyond scheduling the history rewrite and fixing the client-attested spend cap server-side, both outside this PR. Tell me the shape you want and I'll do it.
2. **No importable public contract for `@texra/hosted`.** Documented now, with `packages/agent` assessed as the candidate — but the package decision is yours.
3. **`AccountPlane.signIn()` returning a boolean across three distinct host session protocols.** Not yet addressed.
4. **The dependency-license verdict is broader than the pinned npm and Deno/JSR versions actually examined.** Not yet narrowed.

Items 3 and 4 are straightforward once you confirm the direction; 1 and 2 are decisions rather than edits.

https://claude.ai/code/session_01K3sUAZ2ygoZ58tBPerPjS6

### DIFF
--- a/docs/proposals/2026-08-01-backend-decoupling-plan.md
+++ b/docs/proposals/2026-08-01-backend-decoupling-plan.md
@@ -1,0 +1,935 @@
+# Backend decoupling plan: finishing the seam that already exists
+
+**Status:** proposal. Supersedes nothing; complements
+`docs/proposals/2026-07-29-open-source-readiness.md`, which covers licensing,
+repo identity, and publication decisions this plan assumes are settled.
+
+**Goal.** Decouple the Supabase backend and the LLM relay from the UI and core
+so that (a) the hosted implementation can live in a private repo, (b) the
+open-source client works standalone on the user's own API keys with no dead
+UI, and (c) a different backend can be installed.
+
+**Method.** The repo already has the idiom and one finished instance of it.
+`src/model/includedModelAccess.ts:28-63` is a typed capability port with a loud
+null object (`BYOK_ONLY`, `:71-91`, which _throws_ at `getRelayBaseUrl` rather
+than fabricating a URL) installed once per process from
+`src/controllers/modelAccess/installTexraModelAccess.ts` next to
+`initPlatform()`. `packages/agent/src/index.ts:225,244` already ships as an
+embedder that never installs it and passes `loadAgents({ includeRemote: false })`.
+This plan adds **two more ports of the same shape**, **one lint block copied
+from an existing recipe**, and **no new machinery**. Every proposed layer that
+turned out to be a pass-through was cut; those cuts are named in §9.
+
+---
+
+## 1. The cut
+
+```mermaid
+graph TB
+  subgraph PUB["PUBLIC repo — texra (open source)"]
+    direction TB
+    CORE["core zones<br/>src/agent · src/model · src/tools · src/shared<br/>src/telemetry · src/common · src/controllers · src/latex"]
+    P1["<b>IncludedModelAccess</b><br/>src/model/includedModelAccess.ts<br/><i>BYOK_ONLY</i>"]
+    P2["<b>RemoteAgentSource</b><br/>src/agent/index/agentSource.ts<br/><i>NO_REMOTE_AGENTS</i>"]
+    P3["<b>AccountPlane</b><br/>src/shared/schemas/account.ts (schema)<br/>src/controllers/account/ (runtime)<br/><i>NO_ACCOUNT</i>"]
+    WIRE["Zod wire contracts<br/>src/shared/schemas/{relayErrorEnvelope,usageLog,spendingStatus}.ts"]
+    HOSTS["hosts — extension · desktop · cli<br/>UI gated on hasIncludedModelAccess()"]
+    INST["src/controllers/texra/installTexraBackend.ts<br/><i>the only file allowed to name the backend</i>"]
+    CORE --> P1 & P2 & P3
+    P1 & P2 & P3 -.-> WIRE
+    HOSTS --> INST
+  end
+  subgraph PRIV["PRIVATE repo — texra-hosted (@texra/hosted)"]
+    ADAPT["adapters: installTexraModelAccess · installTexraRemoteAgents<br/>installTexraAccount · host auth surfaces"]
+    AUTH["src/auth/{SupabaseClient,config,serverKeys,relayToken,SupabaseSession,…}"]
+    EDGE["supabase/functions/** · migrations · seed"]
+    PROMPT["deploy pipeline + storage sync"]
+    ADAPT --> AUTH --> EDGE
+  end
+  INST -.->|"optional dependency<br/>absent ⇒ BYOK, said out loud"| ADAPT
+  WIRE -.->|"imported by both runtimes"| EDGE
+```
+
+|                    | Public (`texra`)                                                                                                                                                                            | Private (`texra-hosted`)                                                                                                  | Interface between                                                                |
+| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------- |
+| **Model routing**  | `IncludedModelAccess` port + `BYOK_ONLY`; `ModelHandler.resolveClientCredential` (`src/agent/modelHandlers/ModelHandler.ts:494-565`); `ProxyConfig` union (`ProxyConfigResolver.ts:89-173`) | `installTexraModelAccess.ts` incl. `capIncludedReasoningEffort` (tier pricing policy, `:30-48`); `src/auth/serverKeys/**` | 12-member interface + `hasIncludedModelAccess()` + `setUseIncludedModelAccess()` |
+| **Agent catalog**  | `RemoteAgentSource` port; registry fan-out (`src/agent/index/agentRegistry.ts:206`); the 21 prompt YAMLs in `prompts/agents/remote/`                                                        | PostgREST query, `remote_agents` table + column list, `PGRST204`/`42703` sniffing, `get-agent-config` client              | 3-member interface + `RemoteAgentCatalogEntrySchema`                             |
+| **Account / plan** | `AccountPlane` port; `ProfileMessageBuilder` assembly; settings + CLI render models                                                                                                         | `SupabaseClient`, GoTrue session, tier/quota policy, `log-usage` POST, host OAuth flows                                   | 5-member interface + `AccountSnapshotSchema`                                     |
+| **Errors**         | `relayDetection.ts` parses a published Zod envelope; `isRelayError` on the trace wire                                                                                                       | relay `jsonError` builds its body from the same schema                                                                    | `RelayErrorEnvelopeSchema` (versioned)                                           |
+| **Abuse controls** | —                                                                                                                                                                                           | `_shared/emailPolicy.ts`, `relay/requestGate.ts`, `relay/enforcement.ts`, RLS migrations                                  | none — never client-visible                                                      |
+| **Enforcement**    | one `no-restricted-imports` + one `no-restricted-syntax` block in `eslint.config.mjs`                                                                                                       | private CI runs the public suite                                                                                          | lint fails on any new backend import in core                                     |
+
+**Read the boundary as one sentence:** core zones may _ask_ whether included
+access, a hosted catalog, or an account exists; only `src/controllers/texra/`
+may _answer_, and that directory is the unit that moves private.
+
+### Today's violations — the entire size of the problem
+
+`grep "from '@auth/" src/` minus `src/test-kernel/` and `src/auth/` yields
+**11 production files** (`@auth/codex` and `@auth/constants` excluded — see §2):
+
+| File                                                                    | Imports                                    | Fixed by                                               |
+| ----------------------------------------------------------------------- | ------------------------------------------ | ------------------------------------------------------ |
+| `src/agent/remote/remoteAgentConfigClient.ts:4`                         | `SUPABASE_CONFIG`                          | Port B (file moves)                                    |
+| `src/agent/remote/remoteAgentList.ts:17,18`                             | `SUPABASE_CONFIG`, `SupabaseClient`        | Port B (file moves)                                    |
+| `src/agent/remote/RemoteAgentLoader.ts:16`                              | `SupabaseClient`                           | Port B (file deleted)                                  |
+| `src/telemetry/UsageLogService.ts:6,7`                                  | `SupabaseClient`, `SUPABASE_CUSTOM_DOMAIN` | Port C                                                 |
+| `src/tools/setup/platform.ts:15,17`                                     | `relayToken`, `SupabaseClient`             | Port C                                                 |
+| `src/controllers/settingsView/ProfileMessageBuilder.ts:21,22,23`        | all three                                  | Port C                                                 |
+| `src/controllers/mainView/teamCatalogPorts.ts:3`                        | `SupabaseClient`                           | Port B                                                 |
+| `src/controllers/onboarding/setupLaunch.ts:1`                           | `serverKeys`                               | **Step 0 — deletion, no port**                         |
+| `src/controllers/settingsView/SettingsModelSelectionController.ts:8`    | `FREE_TIER`, `MAX_TIER`                    | **Step 0 — deletion, no port**                         |
+| `src/controllers/settingsView/SettingsRemoteAgentPromptController.ts:2` | `ULTRA_TIER`                               | Port B (controller deleted)                            |
+| `src/controllers/modelAccess/installTexraModelAccess.ts:13-16`          | all                                        | sanctioned adapter — moves to `src/controllers/texra/` |
+
+Plus **14 non-test importers** of `getServerSideKeyService()` across all three
+hosts and `src/controllers/`, which route around the one port that already
+exists. Member tally across those call sites:
+`setUseIncludedModelAccess` ×7, `getUseIncludedModelAccess` ×6,
+`clearAllCaches` ×6, `getUserTier` ×3, `canUseServerSideKeys` ×2, and one each
+of `getSpendingStatus`, `getRelayBaseUrl`, `isRelayQuotaExceeded`,
+`wasQuotaAutoSwitched`, `isProviderOnServer`, `shouldUseServerSideKeysSync`,
+`canUseModelSync`.
+
+---
+
+## 2. Why here and not elsewhere — the seams NOT cut
+
+Some coupling stays. Naming it is part of the design.
+
+| Left coupled                                                                                                                                            | Why                                                                                                                                                                                                                                                                                                   | Evidence                                                                         |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------- |
+| **`@auth/codex/**` and `@auth/constants`** stay importable everywhere                                                                                   | Codex/ChatGPT is a _third-party_ subscription, not TeXRA's backend; it already reaches storage only through `platform().secrets`. `AUTH_COMMANDS` are VS Code command IDs, not endpoints.                                                                                                             | `src/tools/setup/InvokeCommandTool.ts:5`, `src/agent/runtime/ModelFactory.ts:15` |
+| **`isRelayError` stays on the durable trace format**                                                                                                    | It is a field of `ProviderErrorObjectSchema` (`src/shared/schemas/errors.ts:123`) republished on the trace bus (`src/agent/trace/events.ts:281`). Renaming it to `viaProxy` is a persisted-data migration, not a refactor. Under BYOK it is never `true`, so the branches are unreachable, not wrong. |                                                                                  |
+| **`AGENT_SOURCE.REMOTE = 'remote'`** keeps its value                                                                                                    | It is a wire-contract enum member (`src/shared/schemas/agent.ts:22`) consumed by settings IPC, proposals, and the roster, and is persisted in registry keys. Its _meaning_ changes from "Supabase" to "the installed catalog source"; the string does not.                                            |                                                                                  |
+| **`AgentModePreset.texraHostedAgents`** keeps its field name                                                                                            | Persisted in user team presets (`src/shared/schemas/agentPresets.ts:56`). Renaming needs a `z.union()` legacy transform at the entry point — worth doing, but not in the boundary PR series.                                                                                                          |                                                                                  |
+| **The four hosted-vocabulary `ModelAvailabilityKind`s** (`'included-access'`, `'not-included'`, `'included-login-required'`, `'relay-quota-exhausted'`) | `computeModelOptions.ts:102-165` already normalizes them once into a client-owned view model that renderers consume verbatim, and `BYOK_ONLY` collapses all four. This is the seam working.                                                                                                           |                                                                                  |
+| **`RetryState`'s relay-401-refresh control flow** (`src/agent/core/flows/RetryState.ts:196-215`)                                                        | It goes _through_ the port (`getAccessToken(true)`), so it is backend-agnostic already. Generalizing it to `onAuthFailure()` is defensible but is retry-engine surgery unrelated to the cut.                                                                                                          |                                                                                  |
+| **`UsageMonitor`'s awaited flush** (`src/agent/utils/UsageMonitor.ts:296-310`)                                                                          | The `await` on relay rounds exists because the relay enforces its monthly cap from the DB aggregate; the comment says so. `usedRelay = usage.usageRoute === 'relay'` is permanently `false` with no relay installed, so leaving it untouched is both correct and free. **Do not "simplify" this.**    |                                                                                  |
+| **`packages/cli/src/runtime/relayUsage.ts`** (hand-built PostgREST keyset queries over `usage_logs`)                                                    | It is one CLI command in a host zone. It moves wholesale with the private package rather than getting a port nothing else would use.                                                                                                                                                                  | `:195-215`                                                                       |
+| **`supabase/functions/_shared/emailPolicy.ts`** never gets a public contract                                                                            | A 36-domain disposable-mailbox blocklist annotated "observed farming the Researcher Access Program", a privacy-relay blocklist, and `MIN_GITHUB_ACCOUNT_AGE_DAYS = 30`. This is enforcement logic whose value depends on non-publication. It is never client-visible and must not be.                 | `emailPolicy.ts:3-46`                                                            |
+
+### The placement constraint no proposal noticed
+
+`src/test-kernel/architecture/subsystemEdgeRatchet.vitest.ts` fails on any **new
+directed subsystem pair** against `config/ratchets/architecture-edges-baseline.json`.
+The baseline has **no `tools → controllers` and no `telemetry → controllers`
+edge**. It does have `tools → shared`, `telemetry → shared`, `agent → shared`,
+`model → shared`, and `controllers → shared`.
+
+Since `AccountPlane` is consumed by `src/tools/setup/platform.ts`,
+`src/telemetry/UsageLogService.ts`, and `src/controllers/settingsView/`, it
+**must** live in `src/shared/`. Placing it under `src/controllers/account/`
+would create two new edges and fail the ratchet — and would invert layering
+besides. This settles a question the source designs disagreed on.
+
+Conversely, the cut **removes** `agent → auth`, `tools → auth`, and
+`telemetry → auth` (all currently `value` edges in the baseline). Removing edges
+never fails the ratchet; regenerate the baseline in the same PR so the win is
+recorded.
+
+---
+
+## 3. The contract
+
+### Port A — `IncludedModelAccess` (existing, `src/model/includedModelAccess.ts`)
+
+Twelve members unchanged (`:30-63`). Two additions, both paid for by deletion:
+
+```ts
+export interface IncludedModelAccess {
+  /* …existing 12 members… */
+
+  /**
+   * Persist the user's included-vs-personal preference. The provider owns the
+   * write because flipping it must also invalidate entitlement caches and
+   * notify subscribers; a caller touching globalState cannot do that.
+   */
+  setUseIncludedModelAccess(enabled: boolean): Promise<void>;
+}
+
+/**
+ * Whether an included-access provider is installed at all — distinct from every
+ * interface member, which answer "may this user route through it *now*". UI asks
+ * this to decide whether an account surface should exist.
+ */
+export function hasIncludedModelAccess(): boolean; // `installed !== null`
+```
+
+`BYOK_ONLY.setUseIncludedModelAccess` **throws** with
+`INCLUDED_MODEL_ACCESS_REMEDY` (`:97`), matching `getRelayBaseUrl`'s posture
+(`:80-87`) — unreachable behind `hasIncludedModelAccess()`, and a silent no-op
+write is exactly the silent degradation CLAUDE.md bans.
+
+_Justification for `setUseIncludedModelAccess`:_ seven host call sites bind it
+to the singleton (`packages/cli/src/runtime/initPlatform.ts:405,407`,
+`packages/cli/src/runtime/apiAccessMode.ts:57`,
+`packages/desktop/src/main/index.ts:757-758`,
+`packages/desktop/src/main/desktopAgentExecution.ts:478-479`,
+`packages/extension/src/progressView/ProgressViewMessageHandler.ts:603-604`,
+`packages/extension/src/settingsView/SettingsViewMessageHandler.ts:182-183`),
+and `src/controllers/` _already declares it as an injected dependency_
+(`ProgressApiKeyRetryController.ts:44`, `SettingsProfileController.ts:69`).
+The controllers treat it as a port member today; only the hosts don't.
+
+**Deliberately NOT added** (each was in a source design; each is cut):
+
+- `canUseServerSideKeysForModel` — `src/auth/serverKeys/ServerSideKeyService.ts:376-380`
+  defines it as `(await canUseServerSideKeys()) && canUseModelSync(m)`, both
+  already on the port. The bypass at `setupLaunch.ts:49-50` is fixed by
+  composing, not widening.
+- `getSpendingStatus` — spend is an account fact. It lives on Port C, which the
+  CLI status bar can read instead of reaching `getServerSideKeyService()` from
+  inside an Ink render component (`packages/cli/src/chat/tui/panes/StatusBar.tsx:168-176`).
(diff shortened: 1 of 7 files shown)


LionSR:
All seven Codex findings verified against the tree and applied in `95e653ac0`. None were refuted.

One of them was confirmed empirically rather than by review: while executing the `texFormatter` move in `219b76272` I hit exactly the failure described in the line-67 comment — the file's own `./formatter/latexindentpt` and `./formatter/texfmt` imports resolving to a nonexistent `formatter/formatter/` from the new location. That is the same class of omission as the `DelegationTools` comment, and it justified re-checking every remaining blast-radius cell.

Corrections made:

| Finding | Verification |
|---|---|
| `DelegationTools` relative imports | Confirmed — `src/tools/DelegationTools.ts:50,58,59,61` |
| knip baseline keys the old desktop path | Confirmed — `knip-baseline.json` pins `packages/desktop/src/desktopCommandSurface.ts`; the row wrongly claimed all ratchets unaffected |
| `HOST_DIRS` scan roots on the test-kernel move | Confirmed — `hostAgentMockRatchet.vitest.ts:48-51`. The silent-green point is the important half: assertions are `≤` baseline, so zero found passes |
| Ban `@texra/hosted` in the lint block | Confirmed gap; added to both the static group and the dynamic-import selector |
| Keep Supabase resolvable until extraction | Confirmed, and the doc was factually wrong — it claimed the dep survives in "the desktop/CLI manifests", but neither declares it. Root `src/auth/` imports it until step 10, so the step-4 deletion would fail both host builds |
| Move private adapter tests with implementations | Confirmed and **understated**: 36 suites reference the modules step 10 deletes, not the four listed |

The last one changed the plan rather than just annotating it. Seven of the 36 are direct implementation tests that move private; the other ~29 are consumer tests that reach for `@auth/SupabaseClient` only to mock it — which is evidence their subjects are still coupled to the backend rather than to a port. That count is now step 10's readiness metric: while it is 29, extraction is blocked regardless of how the implementation files move.

Also note this PR is docs-plus-code now, not docs-only. `219b76272` and the two commits before it landed items 4-7, 10 and 11 from the organization plan plus the `claude.yml` hardening. Typecheck, lint, prettier, the architecture ratchets and the affected suites all pass locally.

---

.github/workflows/claude.yml
```diff
@@ -128,6 +132,13 @@ jobs:
...
-          issue-body: ${{ github.event.issue.body || '' }}
-          issue-title: ${{ github.event.issue.title || '' }}
+          # The issue body and title are authored by whoever opened the issue,
+          # who need not be trusted — a maintainer replying "@claude" on an
+          # outside contributor's issue is a normal and supported flow. Gate the
+          # untrusted *data* rather than the job: the run still happens, but an
+          # untrusted body is not forwarded as agent-visible instructions.
+          # github.event.issue.author_association is empty for non-issue events,
+          # which correctly yields ''.
+          issue-body: ${{ contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.issue.author_association) && github.event.issue.body || '' }}
```
LionSR: **Blocking:** this does not sanitize the input seen by Claude. These values are passed to the later `auto-create-issue-pr` action, which only scans them for the `auto-fix` label. The preceding `LionSR/agent-ci-actions@v1` step invokes `anthropics/claude-code-action@v1` with no explicit prompt, so it derives context from the original GitHub event before this expression is evaluated. Keep the author gate (or supply an explicitly sanitized prompt to the Claude step); gating data only in the post-run PR-creation step does not mitigate prompt injection.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,821 @@
...
+  ...HOST_LAYER_RESTRICTED_IMPORT_PATTERNS,
+];
+
+// `no-restricted-imports` cannot see `await import(...)`. One builtin selector
+// closes that hole without a custom rule. (The single dynamic backend import in
+// core, src/agent/index/remoteAgentMeta.ts:47, is deleted in step 2; this stops
+// it coming back.)
+const NO_DYNAMIC_BACKEND_IMPORT = {
+  selector:
+    'ImportExpression[source.value=/^(@supabase\\/supabase-js|@auth\\/(SupabaseClient|SupabaseSession|SupabaseAuthCoordinator|supabaseSessionTypes|TokenProvider|config|relayToken|serverKeys))/]',
```
LionSR: The static restriction above now includes `@texra/hosted`, but this separately maintained regex still omits it, so `await import('@texra/hosted')` remains legal in every backend-free zone. This is the exact SSOT drift the boundary is meant to prevent. Define one `BACKEND_MODULES` array and derive both the static patterns and dynamic selector from it.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,821 @@
...
...lers/texra/installTexraBackend.ts` calls the three installers; replace the one `installTexraModelAccess()` line at `extension.ts:252`, `desktop/platform/index.ts:300`, `cli/initPlatform.ts:295`. Add the lint blocks from §5. Delete `@supabase/supabase-js` from `package.json:117` and `packages/agent/package.json:59`. Regenerate `architecture-edges-baseline.json`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             | 1 new, 3 one-liners, 1 config, 1 baseline | `npm run lint` fails on a planted `import { SupabaseClient } from '@auth/SupabaseClient'` in `src/agent/` **and** on a planted `await import(...)` of the same | after 2 and 3            |
```
LionSR: This still instructs step 4 to delete `@supabase/supabase-js`, directly contradicting the bold correction at lines 521–535 that says doing so here breaks desktop/CLI and that the dependency must move in step 10. Remove this sentence from step 4 and put the three-manifest move in step 10 so the migration table remains the canonical executable sequence.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,821 @@
...
+
+---
+
+## 4. Two implementations
+
+| Port member                              | Hosted (`@texra/hosted`)                                                                                  | BYOK (nothing installed)                                  | UI under BYOK                                                                                                                                              |
+| ---------------------------------------- | --------------------------------------------------------------------------------------------------------- | --------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `hasIncludedModelAccess()`               | `true`                                                                                                    | `false`                                                   | **Single gate for every account surface.** Account tab omitted, LoginBanner absent, `/login` `/auth` `/api` unregistered, onboarding sign-in choice absent |
```
LionSR: This makes one facet (`IncludedModelAccess`) the proxy source of truth for a different facet (`AccountPlane`). Because the three ports install independently, an account-capable backend without included model access—or a partially installed backend—will have its entire account UI hidden. Gate account surfaces on canonical account capability, remote-agent surfaces on source capability, or install one atomic capability bundle and derive all gates from it.

src/README.md
```diff
@@ -0,0 +1,82 @@
...
+reachable from them today — `@utils/core`, `@utils/core/boundedIdSet`,
+`@utils/errors/errorMessage`, `@utils/files/pastedImageName`,
+`@utils/text/stringUtils`. Adding an import to any of those five, or to their
+transitive dependencies, can break a webview build in a way `tsc` will not catch.
+
+## Picking between the general-sounding names
+
+`shared/`, `common/`, and `utils/` are the three placements newcomers get wrong.
+
+- **`shared/`** — a type or schema that crosses a process or wire boundary
```
LionSR: This new canonical-looking rule says `shared/` is for cross-process/wire shapes. That supports `AccountSnapshotSchema`, but not the proposed `AccountPlane` mutable singleton (`signIn`, `signOut`, `submitUsage`, `onChange`, module-level setter/accessor). Split the shared schema from the process-side registry instead of using the current edge baseline to force runtime orchestration into a browser-reachable contract zone. Also add this README to the guidance scanner if it is intended to carry standing architecture rules.

LionSR (commented):
## Architecture / SSOT review — not merge-ready yet

The core direction is good: finish an existing seam, keep contracts public, move implementation policy private, and enforce the cut mechanically. The alias removals, `zod` declaration, formatter/text relocations, and most of the audit work are sound. But the current head still has one proven red build, broken moved documentation, an ineffective security refinement, and several places where the proposal creates parallel owners instead of one source of truth.

### Blocking correctness/security

1. **Head fails `typecheck:test-kernel`.** The helper moved from `@test/helpers/streamStatusTestUtils` to `@test/support/...`, but four `.mts` importers were missed:
   - `src/test-kernel/cli/TuiStateAndFocus.vitest.mts:102`
   - `src/test-kernel/cli/WorkflowScriptStreamTranscript.vitest.mts:44`
   - `src/test-kernel/desktop/DesktopAgentExecution.vitest.mts:54`
   - `src/test-kernel/desktop/DesktopWindowTitle.vitest.mts:18`

   Verified in an isolated checkout: `pnpm run typecheck:test-kernel` reports TS2307 at all four sites. The last CI run was on `e5f18937d2`; the code-moving commits after it have no validating CI run attached.

2. **The 21 checkpoint moves break 50 relative links.** Five proposal siblings stayed under `docs/proposals/`, but moved checkpoints still link as `](./<name>.md)`. Rewrite those links to `../../proposals/...`; the archival carve-out means current guidance checks will not catch them.

3. **The latest issue-body “gate” does not gate model input.** The new expressions are inputs to `auto-create-issue-pr`, whose action only scans them for an `auto-fix` label. The earlier `LionSR/agent-ci-actions@v1` step wraps `anthropics/claude-code-action@v1`, which derives its prompt from the GitHub event when no explicit prompt is supplied. Thus the untrusted issue context remains available to the agent; this commit only changes post-run PR labeling. The second checkout also still persists a write-capable token in `.trusted-actions/.git/config`.

4. **Existing confidentiality and abuse-control findings remain unresolved.** The readiness audit still republishes confidential correspondence that its own purge recipe does not remove, and it documents a live relay-accounting bypass before the server-side fix is a flip prerequisite. Those existing inline findings should be closed, not merely left for the later history rewrite.

### Architecture / single source of truth

5. **Backend availability has three registries but the UI uses a fourth, proxy fact.** `IncludedModelAccess`, `RemoteAgentSource`, and `AccountPlane` install independently, yet every account surface is gated by `hasIncludedModelAccess()`. An account-capable or catalog-only backend can therefore be installed and remain invisible. Make availability one atomic installed capability set, or give each facet its own canonical availability and transport that fact in the view model.

6. **Ports A and C duplicate routing facts.** `AccountSnapshot.apiAccessMode` and `quotaAutoSwitched` restate Port A's `getUseIncludedModelAccess()` and `wasQuotaAutoSwitched()`. Model routing reads A while profile/UI reads C, with no declared derivation direction. Port A should own these facts and C should derive from A, or they should be removed from C and composed once at the presentation boundary.

7. **`src/shared/` is the right home for the account schema, not clearly for a mutable process singleton.** The new `src/README.md` now says `shared/` is for shapes crossing a process/wire boundary; `AccountPlane` includes sign-in, sign-out, usage submission, listeners, and module-level registration. Split `AccountSnapshotSchema` into shared and put the runtime accessor in a process-side canonical owner. Do not let the current edge baseline force a service locator into the browser-reachable zone; the ratchet records current edges, not architectural intent.

8. **The public/private contract is not importable yet.** A private `@texra/hosted` cannot register against public singleton instances unless those contract modules are exported by one importable public package. Pinning a repository SHA in private CI does not create runtime module identity. Resolve the package/interface boundary before prescribing the split.

9. **Backend import policy is duplicated and already drifting.** The static restricted-import list now includes `@texra/hosted`; the separately hand-written dynamic-import regex does not. Derive both from one `BACKEND_MODULES` list, and apply it to the existing broad `src/**` lint block rather than maintaining a second closed subsystem enumeration that omits several current and all future subsystems.

10. **Alias ownership is still incomplete.** Root `tsconfig.json` correctly derives extension and desktop maps, but `tsconfig.build.json` remains an ungated hand-maintained subset—the same place stale `@logger` and now stale `@model` entries appeared. Add a build-map derivation target with an explicit exclusion list. Derive `SUBSYSTEM_ALIASES` from the root map as well instead of retaining dead `@types` metadata.

11. **Guidance now has multiple owners.** `src/README.md`, `AGENTS.md`, and `CLAUDE.md` repeat layout facts, while only the latter two are scanned by `check-guidance-refs.mjs`. The new README already describes `shared/` more narrowly than the directory proposal's own audited description. Declare `AGENTS.md` (or `src/README.md`) canonical, make the other pointer-only, and include whichever is canonical in the guidance gate.

12. **Open-source readiness has competing live plans.** The 2026-07-29 document still says no history rewrite and no vendored third-party source, while the new audit proves both false. Add explicit `Supersedes` / `Superseded by` headers and align titles with filenames. Also remove cross-document line-number citations; this PR's own formatting/fix commits have already shifted most of them.

### Simplification recommendation

Split the merge surface into four independently reviewable changes: (1) workflow hardening, (2) alias/dependency fixes, (3) relocations plus all importer/link rewrites, and (4) architecture/audit documents. For the backend plan itself, prefer one importable public contract, one atomic capability installation point, one canonical owner per fact, one derived backend-module list, and one canonical guidance document.

### Verified sound

- Generated extension/desktop tsconfig maps match root and `check:tsconfig-paths` passes.
- `@types/*`, bare `@logger`, and bare `@telemetry` removals are valid; `@types/*` was a real bundler-resolution hazard.
- Trace-viewer's `zod` declaration and lockfile entry are consistent.
- `appendTail`, `texFormatter`, and `executionFormState` relocations have complete source import rewrites.
- The `_relay` version-string correction, 11-file backend coupling census, 96-edge ratchet claims, and most load-bearing audit claims check out.
- The “do not do this” and “layers cut” sections show the right minimum-machinery discipline; the findings above are where that discipline needs to be applied one more time.

LionSR:
Thanks — the typecheck failure was a real break I shipped, and the diagnosis is exactly right.

**Root cause of #1:** the `streamStatusTestUtils` importer rewrite used a grep limited to `--include=*.ts --include=*.tsx`, which silently skipped `.mts`, and I verified with `tsc --noEmit` (the workspace project) rather than `npm run typecheck`, which composes seven projects including `typecheck:test-kernel`. Four `.mts` importers were left dangling. Fixed in `c56bff9b8` with an extension-agnostic rewrite; full `npm run typecheck` passes across all seven. I've stopped using per-project spot checks for anything involving a move.

## Fixed

| # | |
|---|---|
| 1 | Four `.mts` importers repaired; full seven-project typecheck green |
| 2 | 68 relative links rewritten (siblings keep `./`, proposal targets become `../../proposals/`, the `../dev/audits/` form that resolved to `docs/dev/dev/audits/` becomes `./`), plus one `../prds/` → `../../prds/`. Every relative target now resolves — verified by walking each one |
| 3 | Author gate restored; the data-gating reverted since it was ineffective for the reason you gave. Also added `persist-credentials: false` to the `.trusted-actions` checkout you spotted |
| 6 | Step 4's `@supabase/supabase-js` deletion removed; the dependency moves in step 10 |
| 9 | Both lint rules now derive from one `BACKEND_MODULES` array |
| 5 | Per-facet gating: `hasIncludedModelAccess()` / `hasAccountPlane()` / `hasRemoteAgentSource()`. The atomic-bundle alternative is written up as a decision — I recommend against it, since a backend offering accounts without a relay is the case this exercise exists to enable, but it's your call |
| 11 | `AGENTS.md` declared canonical, `src/README.md` scoped to orientation and added to `GUIDANCE_FILES` |
| 12 | Reciprocal supersession headers, scoped to the three disproved conclusions |

Your `src/README.md` findings were both right and both embarrassing: the auth row asserted port-mediated access while `src/tools/setup/platform.ts:15-17` imports `@auth/SupabaseClient` directly, and the `shared/` rule excluded the 26 Lit-importing UI modules. Writing target architecture in the present tense is precisely the defect this PR corrects in `AGENTS.md`, and I reproduced it in a new file.

## Not yet addressed

**#6** (Ports A/C duplicating `apiAccessMode`/`quotaAutoSwitched`), **#7** (splitting `AccountSnapshotSchema` from the runtime registry — I agree the ratchet records current edges, not intent, and shouldn't drive placement), **#8** (the contract needs one importable public package before the split is prescribable), **#10** (`tsconfig.build.json` derivation target), **#4** (the confidentiality and relay-accounting findings), and Codex's point that trace-viewer needs `lit`/`@lit/context`/`@awesome.me/webawesome` declared, not just `zod`.

**#8 is the one that changes the plan's shape** rather than annotating it, so I'd rather resolve it with you than guess.

## On the four-way split

Agreed the merge surface is too broad. Before I do it: the split is cheap for (1) workflow and (2) alias/dependency fixes, but (3) relocations and (4) documents are entangled — the documents describe the relocations and were corrected *by* executing them. Splitting those cleanly means the doc PR lands with stale status headers until the relocation PR merges. I'll take direction on ordering rather than pick.

---

LionSR:
Round 3 applied in `e4952e5cc`. All seven verified; none refuted.

**The P1 was a gap in my own fix.** I gated `issue_comment` and stopped there — `pull_request_review_comment` and `pull_request_review` authenticated the reviewer but not the PR author, so `@claude` on an outsider's PR still checked out attacker-controlled content and ran with write-capable tools. Both branches now require a trusted `pull_request.author_association`.

**The guidance-gate finding is the one I'd have never caught.** Adding `skills/` to `PATH_PREFIXES` made `src/README.md`'s table pass *for the wrong reason* — its `` `skills/` `` means `src/skills/`, but the checker resolved it against the root product directory. A green gate over a stale table is worse than no gate. Every row is now spelled `src/<name>/`, which also means the whole table is gated rather than just the one accidental match.

Also fixed: `LoginBanner` still on `hasIncludedModelAccess()` against my own facet table; and three more blast-radius omissions — `stateSettings.ts:192`'s `GIT_WORKTREE_RUNTIME_REACHABILITY` (existence-asserted at `stateSettings.vitest.ts:181-190`), three literal `read()` paths in `DesktopControlSystem.vitest.mts`, and `ElectronCompositionRoot.vitest.mts:219-222` where `exemptConsumers` hardcodes the suite's own path — after a test-kernel move it flags *itself*. Item 12 was 5 citations, not 4.

**Item 10** — I went with a validation gate rather than a derivation. The build map is a principled subset (core-only `.d.ts` emit, and its values differ in shape: `src/types/*.ts` not `src/types/*`), so deriving it would be wrong. What it lacked was any check at all — which is how the `@logger` entry pointing at a nonexistent file survived. `scripts/sync-tsconfig-paths.mjs` now rejects orphan keys, non-glob targets missing on disk, and any root alias neither included nor in `BUILD_MAP_EXCLUSIONS`, so a new root alias forces a decision.

**Items 6, 7, 8** are done in the plan. On 8: `packages/agent` is further along than either of us assumed — it already has a three-entry `exports` map, `publishConfig.access: public`, and `scripts/rewrite-declaration-aliases.mjs`, which is exactly the alias-free `.d.ts` machinery an importable contract needs. But I'd argue the identity problem shouldn't be solved with a shared singleton at all: have `@texra/hosted` export **factories** and let the public composition root call `installAccountPlane(createTexraAccountPlane(deps))`. One writer, in the public repo, so duplicate module instances can't cause a silent mis-registration. Whether `@texra-ai/agent` is that package is still yours to call — it doesn't block steps 0-9.

Full typecheck (7 projects), all three gates, and prettier are clean.

---

LionSR:
Current-head review (`a18fe9b78c`): this PR is not mergeable yet and should be split before further correction.

The safe separation is:

1. workflow hardening;
2. package/alias dependency corrections;
3. already executed, independently tested relocations;
4. readiness and backend-design documents.

The document part remains blocked for substantive reasons, not merely its merge conflict. The current audit still reproduces confidential correspondence and a working relay-accounting weakness. The backend plan still lacks an importable public contract for `@texra/hosted`, treats a boolean `AccountPlane.signIn()` as sufficient for three distinct host session protocols, does not distinguish an absent optional package from a broken transitive import, and does not await installation at all three composition roots before capability-derived construction. Its dependency-license verdict also remains broader than the exact pinned npm and Deno/JSR versions actually examined.

The narrow code changes should not be held inside a 65-file, 2,500-line audit PR while these design and disclosure questions remain open. No merge is recommended on this head.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,935 @@
...
+const BACKEND_RESTRICTED_IMPORT_PATTERNS = [
+  BACKEND_RESTRICTED_IMPORT_GROUP,
+  ...HOST_LAYER_RESTRICTED_IMPORT_PATTERNS,
+];
+
+// `no-restricted-imports` cannot see `await import(...)`. One builtin selector
+// closes that hole without a custom rule, built from the SAME array. (The
+// single dynamic backend import in core, src/agent/index/remoteAgentMeta.ts:47,
+// is deleted in step 2; this stops it coming back.)
+const escapeForRegex = (s) => s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
LionSR, line 503: **This selector cannot parse once `@supabase/supabase-js` is interpolated.** esquery uses `/.../` delimiters for attribute regexes, but this character class does not escape `/`. Reproduced with the repo's `esquery@1.7.0`: `Invalid regular expression: /^(@supabase/su: Unterminated group`. Include `/` in the escaped class (for example `/[.*+?^${}()|[\]\\/]/g`) and add a config-level test that parses/matches every `BACKEND_MODULES` entry. Also qualify coverage: template-literal/nonliteral imports remain invisible.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,935 @@
...
+the ratchet records the edges the tree has today, not the architecture it should
+have, and letting it site a service locator inside the browser-reachable
+contract zone is exactly the tail wagging the dog. `signIn`, `signOut`,
+`submitUsage`, `onChange` and a module-level setter are runtime orchestration,
+not a wire contract.
+
+| File                                      | Zone               | Contents                                                                                                                          |
+| ----------------------------------------- | ------------------ | --------------------------------------------------------------------------------------------------------------------------------- |
+| `src/shared/schemas/account.ts`           | `src/shared/`      | `AccountSnapshotSchema`, `AccountSnapshot`, `SIGNED_OUT_ACCOUNT`, `UsageSubmitResult` — data only, browser-safe, no mutable state |
+| `src/controllers/account/accountPlane.ts` | `src/controllers/` | The `AccountPlane` interface, `installAccountPlane()`, `getAccountPlane()`, `hasAccountPlane()`                                   |
```
LionSR, line 255: This placement ignores the declaration-build boundary. `tsconfig.build.json` deliberately has no `@controllers/*` path, while `packages/agent` reaches `src/telemetry/UsageLogService.ts` through `src/agent/utils/UsageMonitor.ts`; step 3 then makes `UsageLogService` consume this port. The `@texra-ai/agent` declaration build will therefore fail to resolve `@controllers/account/accountPlane`. Add the declaration program as a placement constraint and either keep the runtime port in an already-emitted subsystem (Port A's `src/model/` precedent is relevant) or explicitly make controllers part of the published type surface.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,935 @@
...
+  /** Operator's plan name, opaque to core. Rendered, never branched on. */
+  planName: z.string().nullable(),
+  sessionProblem: z.enum(['expired', 'unavailable']).nullable(),
+  // NOTE: apiAccessMode and quotaAutoSwitched deliberately absent. They are
+  // routing facts owned by Port A (getUseIncludedModelAccess() /
+  // wasQuotaAutoSwitched()). Restating them here created two owners with no
+  // declared derivation direction — model routing read A while the profile UI
+  // read C, so the two could disagree and nothing said which won. The
+  // presentation boundary composes A and C once, where the view model is built.
+  spendingStatus: SpendingStatusSchema.nullable(),
```
LionSR, line 282: The SSOT correction immediately above is incomplete: spend and its derived routing predicates still straddle two independently installed ports. Today `getSpendingStatus()` and `isRelayQuotaExceeded()` read the same `TierService`, and quota exhaustion triggers Port A's auto-switch. With A installed but C absent, routing silently flips to BYOK while `RelayQuotaMeter` has no status to explain why. Put spending status/error on Port A beside quota/auto-switch, then compose A+C once at the presentation boundary.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,935 @@
...
+```
+
+**Add this block _before_ the `src/agent/core/**` block at `:552`** (flat config:
+a later block replaces the same rule wholesale, so ordering matters):
+
+```js
+// Backend-free zones. Core may ask whether included access, a hosted catalog,
+// or an account exists; only src/controllers/texra/ may answer.
+{
+  files: [
```
LionSR, line 520: Prefer a default-closed boundary: `files: ['src/**/*.{ts,tsx,mts,cts}', 'packages/agent/src/**/*.{ts,tsx,mts,cts}']` with explicit ignores for `src/auth/**`, the composition root, and tests. This allowlist already omits `utils`, `platform`, `logger`, `transcript`, `skills`, `housekeeping`, and `types`, and every future subsystem starts unprotected. Also mirror the existing auth rule's relative-path regex; a direct `../auth/SupabaseClient` currently bypasses this alias-only group.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,935 @@
...
+| ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Account tab — a literal member of `SETTINGS_TAB_ORDER` (`src/shared/schemas/settingsView/data.ts:75-88`) and of `SETTINGS_TAB_GROUPS` (`:128-137`)                                                                | **new**: a capability predicate on tab entries, filtered once where the nav is built. The array is `as const` and doubles as the `SettingsTabName` union, so filter at build time, don't edit the tuple                                                                       |
```
LionSR, line 644: This predicate is not available where the nav is built. `SETTINGS_NAV_GROUPS` is a module-level constant in the webview/renderer bundle; `hasAccountPlane()` is a process-side singleton in the extension host/Electron main process. Transport an explicit capability snapshot on the existing settings wire message and make navigation a function of it. Filtering also needs to drop empty/degenerate groups and the host must reject `SET_TAB` for gated panels, otherwise ACCOUNT remains reachable over IPC even when hidden.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,935 @@
...
+| **8** | **Make the guarded install real.** `installTexraBackend.ts` becomes one `await import('@texra/hosted')` guarded on `ERR_MODULE_NOT_FOUND`, logging at `info` ("no hosted services package in this build — running bring-your-own-key") and rethrowing anything else. **Resolve the bundler question first** (§9, R2): all three hosts are esbuild/Vite-bundled, so an absent optional package must be marked external or the guarded import is a build-time failure, not a runtime one                                                                                                                                                                               | 1 file + 3 build configs           | public CI installs with `--no-optional`, builds all three hosts, boots each                                                                                                | after 4 and 7                            |
```
LionSR, line 678: Two prerequisites are missing here. First, this turns today's synchronous install calls into an async operation, but every `has*()` gate is a synchronous one-shot sample with no installation-change channel. Require all three roots to `await installTexraBackend()` before command/UI registration and test that ordering. Second, a literal private-package import in the public source must still resolve during TypeScript checking and frozen-lockfile install even when optional dependencies are omitted. Decide registry/workspace/build injection, ambient public types, and bundler externalization before prescribing this step; R2 only covers bundling, not typecheck/distribution.

docs/proposals/2026-08-01-backend-decoupling-plan.md
```diff
@@ -0,0 +1,935 @@
...
+| **10** | **Extract `texra-hosted`.** `git filter-repo` `supabase/**`, `src/auth/{SupabaseClient,config,SupabaseSession,SupabaseAuthCoordinator,supabaseSessionTypes,TokenProvider,relayToken}.ts`, `src/auth/serverKeys/**`, `src/controllers/texra/**` (except `installTexraBackend.ts`), `packages/cli/src/runtime/{relayUsage,relayTokensClient,supabaseAuthDeviceCode}.ts`, and the 6 cross-importing test files (§8)                                                                                                                                                                                                                                            | repo surgery                         | public `npm test` green; public build boots BYOK; private CI deploys | after 9  |
```
LionSR, line 685: Make the audit's confidential-history purge a literal precondition of this surgery. The same PR identifies `agents/prl/example_rebuttal_package.txt` as containing confidential correspondence, but neither this path list nor §8 Mechanics removes `agents/prl/**`; executing step 10 verbatim would publish it. Also ratchet production host-package `@auth/*` imports to zero, not only test mocks—the extension/desktop/CLI host trees are part of the extraction blast radius and otherwise leave step 10 uncompilable.

LionSR (commented):
## Supplemental architecture review — latest force-pushed head

I re-reviewed the current docs-only head (`4044d1ed645`) against current `main`, focusing on whether the migration can actually land under the repository's declaration-build, host/webview, lifecycle, and package constraints. The direction is still good, and the force-push removes the earlier mixed code/config merge surface. However, the plan is **not safe to execute as written**.

### Must correct in the plan

1. **The proposed dynamic-import selector is syntactically invalid.** `@supabase/supabase-js` contains `/`, but `escapeForRegex` does not escape `/`; esquery sees an unterminated regex and ESLint fails while loading config. I reproduced this with the repo's `esquery@1.7.0`.
2. **Port C's placement conflicts with the public declaration build.** Step 3 rewires `UsageLogService` and setup code to `@controllers/account/accountPlane`, but `@controllers/*` is absent from `tsconfig.build.json`. `packages/agent` reaches telemetry transitively through `UsageMonitor`, so the SDK declaration build cannot resolve the new import. Decide whether controllers become emitted SDK surface or place the runtime port in an already-emitted subsystem.
3. **Spend still has two owners.** Port A owns quota exhaustion and the auto-switch derived from `TierService.isQuotaExceeded()`, while Port C owns the underlying `spendingStatus`. Independent installation permits A-without-C: routing switches to BYOK while the quota UI has no fact to render. Keep spend with Port A and compose it with account identity at the presentation boundary.
4. **The settings gate crosses a process boundary the plan does not model.** `SETTINGS_NAV_GROUPS` is a browser/renderer module-level constant; `hasAccountPlane()` is a Node/extension-host singleton. Carry explicit capability facts over the existing settings message, derive navigation from them, drop empty groups, and reject host-side `SET_TAB` requests for disabled panels.
5. **The guarded install needs an explicit async lifecycle invariant.** Step 8 changes a synchronous composition-root call into `await import(...)`, while the proposed `has*()` gates are one-shot synchronous samples and have no install-change event. All three roots must await installation before registering commands/building UI, or early samples permanently hide `/login`, account navigation, etc.
6. **The private-package distribution/typecheck mechanism remains undecided.** A literal `import('@texra/hosted')` in the public tree reverses the claimed dependency direction for TypeScript resolution; `--no-optional` does not make an unresolved package disappear from typecheck or a frozen lockfile. Specify registry/workspace/build injection, a public ambient contract if needed, bundler externals, and the one lint exception before step 8.
7. **The boundary rule defaults new subsystems to unprotected.** The closed `files` list already omits `src/{utils,platform,logger,transcript,skills,housekeeping,types}`. Invert it to all `src/**/*` with explicit backend/composition/test exclusions; add a relative-import pattern too, otherwise `../auth/...` bypasses the alias ban.
8. **Step 10's extraction recipe omits known confidential history and most host coupling.** Make audit B1 (`agents/prl/**`) an explicit hard precondition/invert-path entry, and census/ratchet the host-package `@auth/*` consumers—not only the 29 test suites—before claiming extraction is mechanical.

### SSOT cleanup

The plan also needs one vocabulary and one installation model. The table says `installAccountPlane/getAccountPlane`, the code says `setAccountPlane/accountPlane`, and §8 correctly concludes that the private package should export factories while the public composition root is the sole writer. Apply that conclusion back through §3/§7/step 10; otherwise the plan preserves mutually incompatible designs.

### Current-head validation

- Confirmed the PR now contains only five documents plus `AGENTS.md` and `check-guidance-refs.mjs` changes.
- Current CI static checks, guidance references, builds, and one kernel shard are green; one kernel shard was still running when this review was submitted.
- Earlier stale-import and moved-link failures are not present in this force-pushed diff.

Because this is my own PR/account, GitHub only permits a commented review; architecturally I would treat the items above as documentation blockers before anyone starts the migration.

LionSR:
This oversized review surface has been split without discarding its work:

- #9535 — workflow hardening (merged)
- #9539 — alias ownership and trace-viewer dependencies
- #9540 — five relocations plus all importer/link rewrites
- #9541 — architecture/readiness documents and canonical guidance

Local recovery refs retain both the advanced original head and reconciled document edits. Closing #9523 so review and merge decisions happen only on the focused PRs.

PR #9528, opened 2026-08-01
refactor: replace hand-rolled terminal truncation with cli-truncate

LionSR:
## Summary

Follow-up from a codebase-wide scan for ad-hoc implementations that duplicate maintained packages:

1. `packages/cli/src/chat/tui/render/terminalText.ts` now uses `slice-ansi` for hard clipping and `cli-truncate` for ellipsis truncation. The existing exported helpers remain the call-site boundary.
2. Three filename-fragment call sites (`debugMessageSaver.ts`, `subagentDiffs.ts`, `compileCheck.ts`) now use the existing `sanitizePathSegment` helper with options equivalent to their former inline replacements.
3. Direct terminal-text tests now pin narrow-width, wide-character, grapheme, ANSI, and invalid-width behavior.

## Issue acceptance gates

No linked issue; this is proactive cleanup from a scheduled code-quality scan. Gate: remove the confirmed duplicate implementations without regressing valid terminal-width clipping or filename sanitization.

## Single source of truth

- `slice-ansi` owns ANSI/grapheme-safe hard clipping.
- `cli-truncate` owns ellipsis truncation.
- `sanitizePathSegment` owns filename-segment sanitization at all six matching call sites.

The local helpers remain because they define the project's width-normalization contract and are used by 17+ CLI call sites.

Removed the hand-rolled character loop and three inline sanitizing replacements. Added no project abstraction: both truncation helpers delegate directly to maintained packages, and the sanitizer already existed.

Deliberately left alone: timestamp formatting in `src/housekeeping/utils.ts` and HTML-attribute escaping in `src/shared/highlighting/highlightCode.ts`, which are different concerns.

## Defined width behavior

- Finite widths are floored to integers and clamped to zero.
- Non-finite widths throw `TypeError` instead of relying on accidental loop/library behavior.
- Hard clipping preserves complete graphemes and closes ANSI styles.
- Ellipsis truncation uses `cli-truncate`'s Unicode ellipsis; a zero-column budget returns an empty string.

## Net elements (R6)

- One focused test file added.
- Two direct CLI dependencies declared (`cli-truncate` and `slice-ansi`).
- No exported symbol added or removed; helper signatures are unchanged.

## Host impact

Extension: none.

Desktop: none.

CLI: terminal text clipping/truncation is now package-backed with an explicit normalized-width contract.

## Validation

- Focused terminal-text tests: 10/10 passed.
- Existing sanitizer/consumer suites passed.
- CLI typecheck passed.
- Test-kernel typecheck passed.
- Focused ESLint passed.
- Prettier and `git diff --check` passed.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -69,6 +69,7 @@
     "citty": "^0.2.2",
     "cli-highlight": "^2.1.11",
     "cli-table3": "^0.6.5",
+    "cli-truncate": "^6.1.1",
     "diff": "^9.0.0",
     "esbuild": "^0.28.1",
     "ink": "^7.1.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -80,6 +80,7 @@
     "picocolors": "^1.1.1",
     "react": "^19.2.8",
     "semver": "^7.8.5",
+    "slice-ansi": "^9.0.0",
     "string-width": "^8.2.2",
     "strip-ansi": "^7.2.0",
     "typescript": "^6.0.3",
--- a/packages/cli/src/chat/tui/render/terminalText.ts
+++ b/packages/cli/src/chat/tui/render/terminalText.ts
@@ -1,3 +1,4 @@
+import cliTruncate from 'cli-truncate';
 import stringWidth from 'string-width';
 import stripAnsi from 'strip-ansi';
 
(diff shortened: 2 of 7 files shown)


LionSR:
Pass through

LionSR:
@claude review
@codex review

Please re-review current head 40fd7cded1. The width-one hard-clipping regression is fixed with direct `slice-ansi`, and focused terminal-text tests now define the edge-case contract.

PR #9536, opened 2026-08-01
fix: declare trace-viewer's dependencies, delete dead aliases, gate the build map

LionSR:
Split (2 of 3) out of #9523. Seven files, no behavior change.

## trace-viewer declared no dependencies at all

Only `typescript`, `vite`, `vite-plugin-singlefile` as devDependencies — while `src/traceDataSchema.ts:7` imports `zod` and `src/main.ts` imports `@shared/wa` and `@progressView/frontend`, which pull `lit`, `@lit/context`, `@lit-labs/signals`, `@awesome.me/webawesome`, `signal-polyfill`, `@fortawesome/free-solid-svg-icons`, `katex` and `mutative`.

Five of those are already named in the package's own `vite.config.ts` `resolve.dedupe`, which is the tell. Under pnpm's isolated layout they resolved only because unrelated workspace importers hoisted them, so a filtered or isolated install could not build the package. All nine now declared at the versions the root manifest pins; `packages/trace-viewer/node_modules` resolves them locally.

## Three dead aliases

Zero usages across 9,279 internal specifiers.

**`@types/*` was worse than clutter.** `scripts/aliasUtils.mjs:18` filters only values ending `.d.ts`, and `"@types/*": ["src/types/*"]` doesn't — so it was injected as a **live esbuild and Vite alias in every host build**, shadowing the reserved DefinitelyTyped scope. A bare `@types/...` specifier from any bundled dependency resolved into `src/types/`. `src/types/ambient.d.ts` is unaffected: it resolves through the `include` glob and the explicit `turndown-plugin-gfm` mapping.

**Bare `@logger` and `@telemetry`** — neither directory has an `index.ts`, and `tsconfig.build.json` pointed `@logger` at `src/logger/index.ts`, a file that does not exist. All 164 `@logger/*` and 8 `@telemetry/*` deep imports are untouched. The bare entries implied a barrel exists, inviting the convenience barrel `CLAUDE.md` forbids.

Bare `@transcript`, `@model` and `@housekeeping` are deliberately kept — live usages, real `index.ts` files.

## A gate for `tsconfig.build.json`, not a derivation

The build map is a *deliberate* subset: the agent declaration build emits for core only, and its values differ in shape (`src/types/*.ts`, not `src/types/*`) because they resolve for `.d.ts` emit. Deriving it would be wrong. What it lacked was any check at all — which is how the dead `@logger` entry pointing at a nonexistent file survived.

`scripts/sync-tsconfig-paths.mjs` now rejects keys absent from the root map, non-glob targets missing on disk, and any root alias that is neither included nor listed in `BUILD_MAP_EXCLUSIONS` — so adding a root alias forces a conscious include/exclude decision. Runs in both sync and `--check` modes.

## Verification

`check:tsconfig-paths` passes and `npm run typecheck` is clean across all seven projects, on this branch alone.

https://claude.ai/code/session_01K3sUAZ2ygoZ58tBPerPjS6

---
_Generated by [Claude Code](https://claude.ai/code/session_01K3sUAZ2ygoZ58tBPerPjS6)_

---

> [!NOTE]
> **Low Risk**
> Tooling and dependency-manifest fixes with no described runtime behavior change; alias removals match zero usages.
> 
> **Overview**
> **`@texra/trace-viewer`** now lists its real runtime deps (`zod`, `lit`, Web Awesome, KaTeX, etc.) so pnpm isolated installs can build without relying on hoisting from other workspace packages.
> 
> **TypeScript path aliases** drop unused bare `@logger` / `@telemetry` (no `index.ts` barrels; build map had pointed `@logger` at a missing file), remove `@types/*` (was shadowing DefinitelyTyped in bundler aliases), and propagate the slimmer map through root, extension, desktop, and `tsconfig.build.json`.
> 
> **`scripts/sync-tsconfig-paths.mjs`** adds `validateBuildMap`: every root alias must appear in `tsconfig.build.json` or `BUILD_MAP_EXCLUSIONS`, build keys must exist in the root map, and non-glob targets must exist on disk—checked in sync and `--check` CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40791dd634c096d2a5aa62663da2146072c327b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Superseded by #9539, which preserves this alias/dependency work on the current split branch and is receiving the remaining dependency-census/validation fixes. Closing this duplicate to keep one review surface.

PR #9548, opened 2026-08-01
fix(tools): report truncated start-only file reads

### LINKED ISSUE #9543: Fix read_file silently truncating paginated reads with start-only ranges
## Context

Follow-up from #9530 (refactor(model,transcript,tools): retire token diagnostics and three duplicated shapes).

On a 5000-line file, `read_file(path)` correctly emits `...(truncated, 3000 more lines)` but `read_file(path, range:{start:1})` emits no marker at all — `ReadTool` pre-clamps the default window to exactly `start + 1999`, so `formatFileView`'s `truncated = rangeSize > maxLines` computes `2000 > 2000 → false`. A model paginating with `{start: N}` is silently told it reached EOF.

## Details

The clamping removal in #9530 makes the truncation logic cleaner but does not fix the underlying bug: when a range `{start: N}` is used with an omitted `end`, the default window size (`READ_FILE_MAX_LINES`) exactly equals the threshold for truncation-marking, making the truncation marker impossible to trigger for start-only ranges.

Fix: either offset the window-trim by 1 so `formatFileView` can tell the difference between "read 2000 lines because the file is 2000 lines" and "read 2000 lines but the file is longer," or reduce the default read size by 1 so the truncation check triggers.

## References

- PR: #9530

LionSR:
## Summary

The read_file tool now reports remaining lines when a start-only range exceeds the 2,000-line display limit. The tool passes the complete requested range to the shared formatter, which remains the single owner of line limiting and truncation reporting.

A regression test covers a 2,500-line file read from line 2 and verifies both the visible range and the 499-line truncation notice.

Closes #9543.

## Validation

- Focused ReadTool and formatting suites: 10 tests passed
- Changed-file Prettier and ESLint checks passed
- Test-kernel TypeScript check passed
- Diff whitespace check passed

---

> [!NOTE]
> **Low Risk**
> Localized read_file range/truncation behavior with a focused test; no auth, data, or API surface changes.
> 
> **Overview**
> **Start-only `read_file` ranges** that span more than 2,000 lines now show the `...(truncated, N more lines)` notice and a summary that reflects the visible window (e.g. lines 2–2001), instead of silently capping the range in `ReadTool` before formatting.
> 
> `ReadTool` no longer pre-limits omitted `range.end` to `READ_FILE_MAX_LINES`; it passes **EOF** into `formatFileView`, which remains the single place that applies the line cap and truncation messaging. `READ_FILE_MAX_LINES` is **module-private** in `formatting.ts` (no longer exported).
> 
> A regression test uses a 2,500-line fixture with `range: { start: 2 }` and asserts the summary and truncation footer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0644821af538dd978aef385d6c23209159b9e6e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test-kernel/tools/ReadToolRange.vitest.ts
+++ b/src/test-kernel/tools/ReadToolRange.vitest.ts
@@ -14,6 +14,9 @@ const EXECUTION_ID = 'read-range-exec';
 
 /** 10 lines: "line 1" … "line 10". */
 const SMALL = Array.from({ length: 10 }, (_, i) => `line ${i + 1}`).join('\n');
+const LARGE = Array.from({ length: 2500 }, (_, i) => `line ${i + 1}`).join(
+  '\n',
+);
 
 async function callRead(input: unknown) {
   const tool = new ReadFileTool();
@@ -30,7 +33,10 @@ describe('read_file line ranges', () => {
   beforeEach(async () => {
     await installFakePlatform({
       workspacePath: '/workspace',
-      files: { '/workspace/small.txt': SMALL },
+      files: {
+        '/workspace/large.txt': LARGE,
+        '/workspace/small.txt': SMALL,
+      },
     });
   });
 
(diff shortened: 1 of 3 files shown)


LionSR:
Addressed at current head `0644821af5`: `READ_FILE_MAX_LINES` is now module-private, and the focused 10-test run, changed-file lint/format checks, test-kernel typecheck, and pre-push checks pass after rebasing onto current main.

PR #9547, opened 2026-08-01
feat(runtime): execute the gold-standard core stages

LionSR:
## Summary

Executes stages 1–14 of the north-star plan — the wave the gap audit judged now-cheap after the campaign's eight rounds dissolved the blockers. The audit's verdict held: 68 design items were already done on main; this PR lands the remaining core. 43 of 44 stage items implemented; the one skip is itself a finding (below).

## Issue acceptance gates

No issue; the north-star round of the campaign. The audit's 9 rulings were all accepted as recommended and are recorded in a new living ledger (`docs/proposals/2026-08-01-architecture-rulings-ledger.md`).

## Single source of truth

- **TaskState retired.** The last parallel run-config vocabulary: six host sites and the store's own config→taskState→config round trip collapse onto `AgentConfig`. Exactly two legacy boundaries survive, both documented: the `meta.taskState` disk-read shim and the **frozen** CLI NDJSON `setTaskState` projection (byte-identical, gate-checked).
- **RunDescriptor edge-trim**: `kind` required with a one-line `readMeta` backfill at the existing canonicalization seam; the two name-sniffing fallback chains deleted.
- **One error-with-result carrier** through the flow exit — the lifecycle proposal's last structural item — plus placement last-mile and the `context-window` `AgentErrorKind` (producer-verified kinds only, per ruling Q8).
- **ModelCell**: the primitive now owns the handler instance; the persisted `shared.modelId` migration rides the #9434 boundary; the client/credential seam moved into the cell; `RunContext`'s live model getter deleted.
- **`runAbort.signal.reason`** is the single record of why a workflow run stops (D2 Half B; Half A landed in #9519).
- **ExecutionHandle interaction ownership** with the minimal ruled surface (owner-token + refcount + `ownedBy`; CLI-only writer).
- **The ledger sweep**: the error-pipeline proposal flipped to Implemented with per-tier LANDED/SUPERSEDED/OPEN tables re-derived from HEAD (not trusted from the plan); the stale CLAUDE.md/AGENTS.md boundary claim corrected — the SDK surface exists, is fenced by eslint + four ratchets, and the invariant is "never widen a baseline".

## Net elements (R6)

| | |
| --- | --- |
| Files changed | 169 |
| Net LoC | +1,985 (4,473 insertions, 2,488 deletions) — insertions dominated by pinned tests and the ledger docs |
| Exported symbols | −2 (14 added, 16 removed) |

## Consumer counts (R8)

[...]

PR #9553, opened 2026-08-01
fix(runtime): share waiting repair results

LionSR:
## Summary

- replace the per-stream waiting-repair `Set` with the in-flight `Promise<boolean>` for that stream
- make concurrent callers await the same repair verdict instead of treating the second submission as immediately unrepaired
- preserve probe cleanup after success, false results, and rejection so a later submission can retry
- add runtime and extension-command regressions proving two concurrent follow-ups trigger one probe and are admitted in submission order only after repair

## Compatibility boundary

This preserves the restart/follow-up admission boundary: a persisted resumable execution must be restored to `WAITING` before any concurrent `texra.sendFollowUp` submission reaches `submitFollowUp`. It does not change resumability detection, stream transition rules, or later retry behavior.

This is a focused follow-up to #9547; it corrects the concurrency behavior of the lazy waiting repair introduced there without redesigning session repair.

## Validation

- `env -u FORCE_COLOR npx vitest run --config vitest.config.mjs src/test-kernel/agent/runtime/SessionWaitingRepair.vitest.ts src/test-kernel/commands/agent/FollowUpCommandWaitingRepair.vitest.ts` — 2 files, 7 tests passed
- `npm run typecheck` — passed all workspace/package typechecks
- `npm run lint` — passed
- changed-file Prettier check — passed
- `git diff --check` — passed

---

> [!NOTE]
> **Medium Risk**
> Touches follow-up admission and stream phase transitions before submitFollowUp; behavior change is narrow but affects concurrent restart-repair paths.
> 
> **Overview**
> Fixes a concurrency bug in **`repairWaitingIfResumable`**: a second concurrent caller no longer gets an immediate `false` while the first probe is still running.
> 
> **`SessionHandle`** now tracks per-stream in-flight repair work as a **`Promise<boolean>`** (replacing a presence-only `Set`). Later callers **await the same probe**; resumability is still read once per stream. Probe logic moves into **`probeWaitingRepair`**, with map cleanup keyed to the specific promise so a later submission can retry after success, failure, or rejection.
> 
> Tests expect both concurrent repairs to resolve **`true`** when sharing one probe, and a new **`texra.sendFollowUp`** regression confirms two concurrent follow-ups trigger **one** `deriveResumability` call and reach **`submitFollowUp`** in submission order after the stream is restored to **`WAITING`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea3ed0ffae877ac9685ebb20c5ab79603b2cfec5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/runtime/SessionHandle.ts
+++ b/src/agent/runtime/SessionHandle.ts
@@ -169,8 +169,11 @@ export class SessionHandle {
    */
   readonly workflowControls: WorkflowControlRegistry;
   private restartRepairPromise: Promise<unknown> | undefined;
-  /** Streams with a {@link repairWaitingIfResumable} probe already in flight. */
-  private readonly waitingRepairProbes = new Set<StreamTabId>();
+  /** Per-stream {@link repairWaitingIfResumable} result currently in flight. */
+  private readonly waitingRepairProbes = new Map<
+    StreamTabId,
+    Promise<boolean>
+  >();
   private readonly restartRepairQueue = new PQueue({ concurrency: 1 });
   private readonly restartRepairRetry = new RestartRepairRetryScheduler();
   private readonly restartRepairAbort = new AbortController();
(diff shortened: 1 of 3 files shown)


LionSR:
Follow-up stale-probe race fix opened as #9556. It guards the pending resumability probe against intervening resume, clear, and execution replacement while preserving shared overlapping verdicts.

Issue #9532, opened 2026-08-01
Bound policy-approved model retry batches and background retry lifetime


LionSR:
## Severity

High: one root run and one delegated child produced overlapping multi-hour retry storms. The durable logs imply about 66 failed provider attempts under default batch sizing before cancellation, while queued work remained blocked.

## Verified production evidence

`error6.txt:6-29` contains 12 `Model request failed / Connection error` banners from root execution `3ee19b37dc0b` (stream-log sequence 974-985).

Delegated child execution `56239afe8a99` independently contains 10 equivalent banners (sequence 61-70), ending `interrupted/cancelled` after about 7h09m. The errors arrived roughly every 28-38 minutes, consistent with a default three-attempt batch where individual provider operations approach a ten-minute timeout.

These banners are emitted only after a bounded automatic batch is exhausted. Every banner after the first proves that the preceding retry request was accepted and another full batch ran. The CLI TUI formatting and automatic progression strongly indicate inherited `yolo` policy, although effective policy and retry decisions are not persisted, so this cannot be proven from execution metadata.

Under the default three attempts per batch, the two streams represent approximately 36 + 30 = 66 failed provider attempts. For OpenAI background mode many may be retrieves of one retained response rather than new generations; current telemetry cannot distinguish them.

## Root cause

The retry layers are individually bounded but compose into an unbounded loop:

1. `RetryableInvocationNode` configures a finite `pRetry` batch.
2. `Node._exec()` wraps that batch in `for (;;)`.
3. After exhaustion, `retryPrompt()` asks the host for a nominally manual decision.
4. CLI `yolo` auto-accepts an ordinary transient connection failure.
5. `Node._exec()` restarts the entire automatic batch with no cumulative attempt, batch, or wall-time cap.

Additional amplification/risk:

- `texra.model.retry.maxAttempts` has no finite maximum in the canonical schema or extension manifest.
- OpenAI background retries retain the response ID, but each new poller invocation resets the nominal three-hour polling clock; cumulative retrieval time is unbounded.
- Google background retrieval failure currently falls back to a fresh interaction submission, potentially orphaning the prior job and multiplying charged generations.
- `ModelRetryGate` coordinates route cooling/probes but intentionally does not own retry count.
- Delegated children inherit the parent CLI policy, so approving delegation is effectively followed by unbounded automatic retry approval under `yolo`.

## Expected behavior

- Every model operation has one finite cumulative retry budget across automatic batches and policy decisions.
[...]

PR #9566, opened 2026-08-01
fix(model): preserve OpenAI background deadline

LionSR:
## Summary

- preserve the original three-hour polling deadline when an OpenAI background response is resumed after a transient failure
- stop automatic and manual retry batches once that deadline has expired, before making another provider request
- let the shared background poller preserve provider-specific terminal timeout errors from an earlier polling lifetime

This is a bounded part of #9532. It prevents an approved retry from restarting the lifetime of the same OpenAI background response. Provider-lifetime handling for other background APIs and broader retry telemetry remain separate work.

## User-visible changes

An OpenAI background response now has one finite polling window. Reconnecting to the same response after a temporary failure no longer grants it another three hours, and an expired response is reported as terminal rather than being offered another ineffective retry.

## Design

`BackgroundRunLifecycle` owns the response identity and its absolute deadline together. The shared `BackgroundPoller` accepts that deadline without acquiring responsibility for persistence or provider-specific policy, and preserves a provider-specific error when the deadline expires. This keeps the provider lifecycle as the source of truth while allowing polling to continue across separate invocations.

## Validation

- `npm run format`
- `npm run compile:fast`
- `npm run lint`
- `npm test` (8,405 passed, 4 skipped; 8,409 total)
- `npm run typecheck`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,9 @@ All notable changes to this project will be documented in this file.
 
 #### Bug Fixes
 
+- **OpenAI background responses keep one polling deadline across retries** —
+  reconnecting to the same response no longer restarts its three-hour polling
+  window.
 - **Automatic model retry settings are bounded** — the retry setting now
   accepts zero to five additional attempts after the initial request. Invalid
   stored values fall back to the documented default instead of creating an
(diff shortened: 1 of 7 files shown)


LionSR:
Fact check this number

LionSR:
Confirmed from the final local `npm test` output on exact head `5c142b79382fc31a0f110e30bd8f1abe4d83233b`: **8,404 passed, 4 skipped, 8,408 total**, across 826 test files (825 passed, 1 skipped). The PR description reports the passing and skipped counts separately.

LionSR:
The review finding is corrected on exact head `8d314dc103236df3883cfb87da1ac34601a740b7`. Both deadline paths now throw a provider-tagged terminal error (`userRetryable: false`), so neither automatic nor manual retry offers can repeat the expired response. The message no longer advises retrying. The shared poller also preserves the typed error rather than reducing it to text. Regression tests cover classification and error identity. Final full suite: **8,405 passed, 4 skipped, 8,409 total**.

PR #9569, opened 2026-08-02
refactor(agent): make run signal cancellation SSOT

### LINKED ISSUE #9435: Cancellation SSOT: collapse InterruptManager and the pendingInterrupt compensators onto RunScope.signal (stages 2–3)
> ## ✅ Stage 1 shipped — the correctness bug is fixed
>
> **The dropped-Stop defect is resolved.** #9452 / 5162ed7 removed the `if (isInterrupted) return` early exit from `onInterrupt` and made `setAbortController` abort immediately when the latch is already set. Verified at HEAD in `InterruptManager.ts:18-37`. The commit says so itself: *"This is stage 1 of #9435. The RunScope.signal SSOT (stages 2 and 3) is not included."*
>
> **What remains is the SSOT cleanup, not a live bug.** Retitled accordingly — the original title claimed an unfixed dropped-Stop, which is no longer true. Stages 2 and 3 are unstarted: `InterruptManager.ts` still exists, `setAbortController` still appears at 14 non-test call sites, and `RunScope` has no `.signal` field.
>
> This issue was auto-closed by the PR's `Closes` keyword and manually reopened for exactly this reason.
>
> *Audited 2026-07-31 against HEAD 96dbb2c.*

---

## Cancellation is edge-triggered into a single mutable slot, and a Stop can be dropped entirely

Seam: flow-engine-cycles. Risk: high (staged). Corrected net LoC: **-80** (scout claimed -130).

### Current design

`createInterruptCallbacks()` (`src/agent/runtime/InterruptManager.ts:18-33`) builds a per-run closure over two independent pieces of state:

- a monotone boolean `isInterrupted`, polled by observers through `checkInterruption()` (36 non-test sites),
- a single mutable `abortController` slot, overwritten by nodes through `setAbortController(ctrl | null)`, declared as a port on `BaseFlowContextInit:34`.

`onInterrupt()` sets the latch and calls `abortController?.abort()`. It aborts only whatever happens to be registered at that instant. Nodes register a controller and null it in `finally` (`RetryState.ts:288-295`, `ToolUseDispatchNode.ts:178/225, 462/479`). Because the slot holds exactly one controller, `execCall` carries a hand-rolled chaining branch (`ToolUseDispatchNode.ts:455-470`: `addEventListener('abort')`, `detachBatchAbort`, `if (!batchSignal) options.setAbortController(controller)`) so a per-call controller does not clobber the batch controller.

One layer up, `ExecutionHandle` compensates for the same edge delivery with a second mechanism: `pendingInterrupt` / `acceptsPendingInterrupt` / `enablePendingInterrupt()` / `closePendingInterruptWindow()` / `hasPendingInterrupt` (`ExecutionHandle.ts:160-229`), replaying a stop that arrived before the flow attached.

### The defect, spelled out — ✅ fixed by stage 1

Delivery and observation are different mechanisms, so a stop can land in a gap where neither bites.

1. `ToolUseRoundPrepNode.prep` snapshots `const interrupted = this.services.checkInterruption()` at `ToolUseRoundPrepNode.ts:46`. It is a snapshot; `post` acts on it later at line 68.
2. The previous node's `finally` already ran `services.setAbortController(null)` (`RetryState.ts:293`). The slot is empty.
3. During the awaits that follow, the user presses Stop. In the tool-use round the window covers `post` running `appendFollowUpAsUserMessage` (media file IO). In the reflection flow it is wider: `ResponsePrepNode.prep` snapshots at `ResponseCycleFlow.ts:119`, then awaits `FlexibleFS.exists()` and `getSystemPromptWithRules()` (real rule-file reads), then `post` awaits `saveCycleDebug` (a file write).
4. `onInterrupt()` sets `isInterrupted = true` and calls `abortController?.abort()` on `null`. Nothing is aborted. Worse than the scout reported: `onInterrupt` opens with `if (isInterrupted) return`, so the abort edge is consumed permanently. A second Stop press cannot reach the controller registered afterwards. There is no user-side recovery.
5. `ModelInvocationNode.exec` gates only on `prepRes.shouldStop` (`ModelInvocationNode.ts:123-125`), the pre-interrupt snapshot. Its `InvocationServices` Pick (`:73-84`) does not even include `checkInterruption`.
6. `RetryableInvocationNode._exec` constructs a fresh, unconditionally non-aborted `AbortController` (`RetryState.ts:280-289`) and issues `createResponse` with that signal. `RetryableNodeServices` (`RetryState.ts:59-69`) exposes `setAbortController` but not `checkInterruption`, so the retry node structurally cannot see the latch.

Meanwhile `registry.terminate` sees `handle.interrupt()` return true (a handler and flow context are attached) and immediately runs `cancelStreamStatus`. The UI shows CANCELLED while a fully billed model turn keeps streaming to completion. Verification checked every other candidate guard: `flowContext.interrupt()` (`runToolUseFlow.ts:370-378`) only calls `onInterrupt`, `interactions.cancel`, and `sessionLifecycle.interrupt()`, which is just `followUps.dispose()` (`ToolUseSessionLifecycle.ts:63-66`). Nothing else can abort an in-flight HTTP request. No vitest pins the gap; `ResponseCycleCancellation.vitest.ts` pins post-hoc classification only.

### Smaller and safer redesign, single owner

Single owner: **one sticky `AbortSignal` per run, created in `AgentLaunchContext` and published on the frozen `RunScope`.** Not the `ExecutionHandle` constructor: `RunScope` is built before `new AgentExecutionHandle` (`AgentRunLifecycle.ts:324`), so the launch context must own the controller and hand it over.

Stage it.

- ~~**Stage 1, about 10 lines, low risk, ship first.** In `createInterruptCallbacks`, remove the `if (isInterrupted) return` early exit, and make `setAbortController(ctrl)` abort `ctrl` immediately when `isInterrupted` is already true. That alone makes "interrupt fired while the slot was empty" non-lossy and closes the second-press dead end. The entire correctness gain is here.~~ ✅ **shipped via #9452 / 5162ed7**
- **Stage 2, SSOT, about -80 net.** Add `readonly signal: AbortSignal` to `RunScope`. Derive `checkInterruption()` from `runScope.signal.aborted` and delete `InterruptManager.ts` (33 lines). Delete the `setAbortController` port and its 13 sites. `RetryableInvocationNode` and `ToolUseDispatchNode` derive their controller as `AbortSignal.any([runScope.signal, local.signal])`, which also deletes the hand-rolled `batchSignal` / `detachBatchAbort` chaining. Precedent for `AbortSignal.any` already exists at `src/tools/timeouts.ts:95` and in `runWorkflowScript.ts`'s run-scoped `runAbort`. Plumbing cost is low: one non-test `createRunScope` site (`AgentLaunchContext.ts:393`) plus about ten test fixtures.
- **Stage 3, attach window.** Delete `pendingInterrupt` / `acceptsPendingInterrupt` / `enablePendingInterrupt` / `closePendingInterruptWindow`; `hasPendingInterrupt` becomes `signal.aborted` at its one read site (`AgentRunLifecycle.ts:409`). `attachToolUseFlow` keeps a level-triggered `if (signal.aborted) context.interrupt()` plus an abort listener, because `flowContext.interrupt()` still owns the queue policy branch (`interruptPreservingQueue` vs `interrupt`, gated on `inResumeStartupWindow`) and `interactions.cancel`.

### Corrected numbers

| | scout | verified |
|---|---|---|
| net LoC | -130 | -80 |
| states or paths cut | 8 | 5 |
| risk | high | high for stages 2-3, low for stage 1 |

The scout's "43 non-test references to the compensators" is inflated; the grep catches the CLI's unrelated `pendingInterruptedFollowUps`. The real pending-interrupt mechanism is two fields plus three methods in `ExecutionHandle`, about 20 lines, with three read sites in `AgentRunLifecycle`. `checkInterruption` to `signal.aborted` is LoC-neutral across its 36 sites, and callback plumbing (`FollowUpQueue.waitForNext`, `IToolUseSession.waitForFollowUp`, `RoundPersistedFlow.callbacks.checkInterruption`) becomes signal plumbing, another wash.

### Risks and out-of-scope

- **`terminateWaitingHandle` is not a compensator for this and must not be touched.** `executionRegistry.ts:841-900` tears down a parked WAITING run: runs the registered waiting-cleanup, publishes the terminal `result`, settles `handle.result` (or a consumer awaiting it hangs forever), persists terminal status, and has a lease-loss recovery arm. A sticky signal aborts nobody there, because the run is suspended and nothing is awaiting. `interrupt()` must keep returning false for "nothing live" so `terminate` can fall through. That is roughly 90 of the scout's claimed lines that do not exist, and it brushes the settled two-arm terminal split guarded by `claimTerminalFinalize`.
- **`_userCancelled` cannot collapse to `signal.aborted`.** It has a second source: `retryPrompt` sets it from `handleManualRetryPrompt().userCancelled` (`RetryState.ts:301-305`) when the user declines a manual retry, which is not a run interrupt. It feeds `shared.userCancelledRetry` and the flow-record-preserved-for-resume policy. Collapsing it reclassifies a declined retry from cancelled to failed and drops the resume record. Keep it, or narrow it to `_retryPromptCancelled` and OR with `signal.aborted`.

  > **Update (2026-07-31):** this warning is now load-bearing in the other direction too. #9442 found that a declined retry derived `COMPLETED` while the `finally` preserved its record as resumable, and the fix in 0b18c30 makes `shared.userCancelledRetry` feed the `cancelled` fact in `deriveRunOutcome`. Stage 2 must preserve that: `userCancelledRetry` is a distinct terminal cause from a run interrupt, and both now reach the outcome.
- Per-handle stickiness is safe: a new handle is constructed per `runFlowWithLifecycle` call including resumes, so an aborted signal cannot poison a later turn.
- **Required test, does not exist today:** fire the interrupt between `ResponsePrepNode.prep` and the model call, assert `createResponse` is never issued or rejects immediately.

---

Filed from the 2026-07-30 status/flow/race design audit (3 deep scouts, adversarial verification of the top findings). This item survived verification; the numbers are the verifier's corrections, which cut the scout's estimates substantially.

LionSR:
## Summary

- Give every agent run one sticky `AbortSignal`, owned by its launch context and published through `RunScope`.
- Remove `InterruptManager` and the execution handle’s pending-interrupt state machine.
- Make reflection, tool-use, retries, follow-up waits, and child loops observe the same cancellation state.
- Attach the lifecycle interrupt target before execution tracking, while refusing stops once live runner work has ended.

## Issue acceptance gates

Closes #9435. Stages 2 and 3 are complete: `RunScope.signal` is the cancellation source of truth; `InterruptManager` and the pending-interrupt compensators are deleted; tool-use attachment is level-triggered; and the requested preparation-to-model-call regression test proves that an aborted run does not issue the provider request. The waiting-execution terminal path and manual-retry cancellation remain distinct, as required.

## Single source of truth

`AgentLaunchContext` owns the sole run cancellation controller. The frozen `RunScope` carries its signal to every flow and node, so cancellation state cannot disagree across callback snapshots or mutable controller slots.

The lifecycle installs the interrupt handler before the handle enters the execution registry. This preserves successor handoff without a second pending-stop mechanism. Tool-use flow attachment observes the sticky signal both immediately and through an abort listener, retaining the queue-policy behavior owned by the live flow. Waiting executions continue through their existing terminal-settlement path.

The change deletes the parallel `InterruptManager` owner and the `ExecutionHandle` pre-attachment state machine. It adds no wrapper or pass-through layer: `RunScope.signal` is read directly by the existing flows, nodes, queues, and lifecycle. Declined manual retries remain a separate terminal cause rather than being folded into run interruption.

## Net elements (R6)

- Files: 46 changed; 2 deleted and 0 added.
- Lines: 246 additions, 341 deletions; net -95.
- Exported symbols: 0 added, 1 deleted (`createInterruptCallbacks`); net -1.
- Classes/interfaces/enums: none added or deleted; `RunScope` gains one required field.
- State mechanisms: the three-state `PreAttachStop` type, its field, and its four compensating operations are deleted.

## Consumer counts (R8)

Repository grep over production `src/agent/**/*.ts` before the change found:

- `createInterruptCallbacks`: 2 call sites, now 0; both route through the lifecycle-owned launch controller.
- `checkInterruption()`: 13 executable observations, now 0; all read `runScope.signal.aborted`.
- separate `.abortSignal` accesses: 7 executable accesses, now 0; all use `runScope.signal`.
- `IToolUseSession.waitForFollowUp`: 2 call consumers plus its interface and implementation; both consumers now pass an `AbortSignal`.
- `FollowUpQueue.waitAndDrainAll`: 2 call consumers; both now pass a signal.
- `AgentExecutionHandle.attachToolUseFlow`: 1 call consumer; it now supplies the sticky run signal.

No event key or external subscriber surface changes.

## Host impact

Extension: Shared agent runs use the consolidated cancellation path; no host-specific API changes.

Desktop: Shared agent runs use the consolidated cancellation path; no host-specific API changes.

CLI: Shared agent runs use the consolidated cancellation path; no CLI command or display changes.

## Scheduled refactoring

None. This completes the remaining stages of #9435.

## Validation

- `npm run format`
- `npm run typecheck:workspace`
- `npm run typecheck:test-kernel`
- `npm run lint`
- `npm run compile:fast`
- `npm test` (824 files passed, 1 skipped; 8,412 tests passed, 4 skipped)

Regression coverage now interrupts between response preparation and model invocation and proves that the provider call is never issued. It also covers an active follow-up wait settling on cancellation and stop delivery during execution-registry handoff.


LionSR:
@codex review

LionSR:
@claude review

LionSR:
Addressed the review-process finding without changing the reviewed commit. The PR body now follows the repository template and includes measured `## Net elements (R6)` and `## Consumer counts (R8)` sections.

@claude review

PR #9582, opened 2026-08-02
fix(data-structures): tighten correlated-optional shapes flagged by audit

LionSR:
## Summary

Follow-up to a scheduled data-structure-complexity audit. Fixes the audit's High-severity finding and the contained Medium findings; two Mediums were investigated and deliberately left unapplied (see below).

- **`StreamLogStore` (High)**: `StreamLogSummarySchema` had `.catch(undefined)` on every field, including the crash-recovery flags (`hasRunningGroup`, `hasRunningStreamingText`, `hasNonterminalWorkflowTask`). A malformed field silently became `undefined` (treated as "not running"), which could skip the orphan-recovery sweep for a stream that actually needed it — exactly the "corruption → silent wrong default" pattern this repo's own review checklist calls a defect. Removed the per-field `.catch()` so a malformed field now fails the whole parse and falls back to the existing "rebuild from the authoritative log" path. Also unified `summaryOf()`/`summarizeEntries()` (previously two independently hand-written projections into the same summary shape) through one `toSummary()` helper.
- **`ToolEditApprovalResult` (Medium)**: was a flat object with 6 fields correlated by `accepted` (content fields only meaningful when `true`), forcing a runtime `appliedContent == null` throw and `??`-defaults at every consumer. Now a discriminated union on `accepted`, so `appliedContent` is a compile-time guarantee on acceptance.
- **`UpdateFollowUpTextMessageSchema` (Medium)**: `kind`/`error` were flat, unrelated fields — `error` was populated by the sender only on `polishError` but the frontend handler never read it, an invariant nothing enforced. Now a discriminated union on `kind`.
- **Desktop `WEBVIEW_READY` (Medium)**: `desktopProgressIpc.ts` validated the message via `ProgressViewInboundMessageSchema` but then read `message.view` off the raw, pre-validation object, because that schema's `WEBVIEW_READY` variant was `commandOnly(...)` (no `view` field) and Zod strips unknown keys on parse. Swapped in the already-existing, properly-typed `WebviewReadyMessageSchema` (used by the settings-view IPC path) so `result.data.view` is validated and typed.

### Investigated, not applied

- **`StreamLogEntry.data: unknown` classification duplicated across ~21 call sites** (the audit's other High finding) — left out of scope; it spans the CLI TUI, extension webview, and transcript recorder, and deserves a dedicated, carefully-tested PR rather than a rider on this cleanup.
- **`tool.end` event `result` typing** — the audit suggested typing it as `Partial<ToolUseLog>`, but `ToolUseLog` is a TeXRA/progress-view-specific shape, and `src/agent/trace/events.ts`/`TraceEmitter.ts` are explicitly documented as "agent-general, no TeXRA-specific sugar." Applying the suggested fix would have violated that boundary, so left as `unknown` (a genuine, already-documented escape hatch).
- **`FormProgress` "discriminated union on `status`"** — on inspection, `message`/`copyableMessage`/`archivedCopyableMessage`/`archiveCopyable` are genuinely independent of `status` in actual usage (`registerBuiltins.tsx`): the record is incrementally mutated via `{...current, field}` as an async action progresses, not reconstructed as one of several correlated variants. A discriminated union here would have broken that update pattern for no type-safety gain.

## Issue acceptance gates

Ad hoc fix from a scheduled audit, not tracked against a filed issue. Acceptance gate: each fix above compiles, is covered by (updated or existing) tests, and preserves current behavior except where the behavior itself was the bug (`StreamLogStore` corruption masking).

## Single source of truth

- `StreamLogStore.toSummary()` is now the one place a `StreamLogSummary` is projected from either a live `StreamLog` or a raw entries scan.
- `WebviewReadyMessageSchema` (`commonViewMessages.ts`) is now the one schema for the `WEBVIEW_READY` message shape reused across progress-view and settings-view IPC, instead of the progress-view path re-deriving `view` from the unvalidated raw message.

Removed: the hand-duplicated `summaryOf`/`summarizeEntries` object-literal construction (now both go through `toSummary`); the `result.accepted && result.appliedContent == null` runtime guard and the `??`-default fallbacks in `toolEditApproval.ts` (now compile-time guaranteed by the discriminated union). No new abstractions beyond the type-level discriminated unions and the one `toSummary` projection, all single-purpose and directly motivated by the audit findings.

## Net elements (R6)

15 files changed, 168 insertions(+), 94 deletions(-). No new files. New exported symbols: `AcceptedToolEditApprovalResult` (narrowed type alias, consumed by `fileEditFlow.ts`). No new classes/interfaces beyond the type narrowing already described.

## Consumer counts (R8)

n/a — no emit path or export was deleted/re-routed; `ToolEditApprovalResult`'s shape changed but its name/export site is unchanged, and every construction site across `src/`, `packages/cli/`, and tests was updated (grepped and fixed 13 call sites; see commit).

## Host impact

Extension: none functionally — `ToolEditApprovalController` (used by the VS Code webview host) behavior unchanged, only the redundant runtime check removed.

Desktop: `desktopProgressIpc.ts`'s `WEBVIEW_READY` routing now reads a validated `view` field instead of an unvalidated one; same runtime behavior for well-formed messages.

CLI: `toToolEditResult`/`subscribeApprovals.ts` already always supplied `appliedContent`; no behavior change, just now compile-time enforced.

## Scheduled refactoring

None. The `StreamLogEntry.data` 21-site duplication is flagged above as a candidate for a future dedicated PR but no follow-up issue was filed.

## Validation

- `npx tsc --noEmit` across the root workspace, `tsconfig.test-kernel.json`, `packages/extension`, `packages/cli`, `packages/desktop` (main/renderer/preload), and `packages/trace-viewer` — all clean.
- `corepack pnpm --filter @texra-ai/agent build` — clean.
- `npx vitest run` (full suite) — 8442 passed, 9 skipped (pre-existing), 0 failures. One test (`StreamLogStoreLoad.vitest.ts`) was rewritten because it was asserting the exact corruption-masking behavior this PR fixes.
- `npm run lint`, `npm run check:dead-code-ratchet`, `npx prettier --check` on all touched files — all clean.
- `npm run compile:fast` — builds successfully.

---
_Generated by [Claude Code](https://claude.ai/code/session_014PBqyD3eP5WsKvvgMpmFfD)_

---

> [!NOTE]
> **Medium Risk**
> `StreamLogStore` summary parsing change alters crash-recovery behavior (intentional fix); tool-edit and IPC changes are mostly type-level with behavior preserved for well-formed data.
> 
> **Overview**
> Follow-up to a data-structure audit: replaces **flat optional fields** with **discriminated unions** and fixes one **crash-recovery** bug in transcript summary caching.
> 
> **`StreamLogStore`**: Removes per-field `.catch(undefined)` on `StreamLogSummarySchema` so a malformed crash-recovery flag (e.g. `hasRunningGroup`) no longer silently reads as “not running” and skips orphan recovery; invalid summaries now fail parse and **rebuild from the authoritative log**. Adds shared `toSummary()` for live `StreamLog` vs entries scan.
> 
> **`ToolEditApprovalResult`**: Becomes a union on `accepted` with required `appliedContent` when accepted; drops runtime throws and `??` fallbacks in `toolEditApproval.ts` / `ToolEditApprovalController`. Bypass and disabled-approval paths always supply proposed content via `acceptProposedAsIs`.
> 
> **IPC/schemas**: Progress inbound uses `WebviewReadyMessageSchema` so desktop reads `result.data.view` after validation; `UPDATE_FOLLOW_UP_TEXT` is a union on `kind` so `error` only exists on `polishError`.
> 
> Tests updated for the new approval shape and for summary rebuild when summary fields have wrong types (replacing the old “salvage partial fields” expectation).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5510fcc9e7f9efe6758ad39bdad7962499f6ed54. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/desktopProgressIpc.ts
+++ b/packages/desktop/src/main/desktopProgressIpc.ts
@@ -108,16 +108,16 @@ export function createDesktopProgressIpc(
       const result = ProgressViewInboundMessageSchema.safeParse(message);
       if (!result.success) return false;
 
-      const command = result.data.command;
       // WEBVIEW_READY and pass-through commands return false so sibling
       // handlers in the chain still receive them.
-      if (command === PROGRESS_VIEW_COMMANDS.WEBVIEW_READY) {
-        if (message.view !== 'progress') return false;
+      if (result.data.command === PROGRESS_VIEW_COMMANDS.WEBVIEW_READY) {
+        if (result.data.view !== 'progress') return false;
         withProgress((progress) => {
           void progress.completeWebviewReady().catch(reportAsyncError);
         });
         return false;
       }
+      const command = result.data.command;
       if (passThroughCommands.has(command)) return false;
 
       withProgress((progress) => {
--- a/src/controllers/approval/ToolEditApprovalController.ts
+++ b/src/controllers/approval/ToolEditApprovalController.ts
@@ -190,17 +190,11 @@ export class ToolEditApprovalController {
       if (!entry.isSettled()) {
         void this.runAction(entry, () => this.publishPromptWhenVisible(entry));
       }
-      const result = await settlement;
-
       // `requestToolEditApproval` derives userPatch, lineChanges, and startLine
-      // from the content this returns, so an accepted result only owes it the
-      // content the user actually approved.
-      if (result.accepted && result.appliedContent == null) {
-        throw new Error(
-          'Tool edit approval settled without the current proposed content.',
-        );
-      }
-      return result;
+      // from the content this returns; `ToolEditApprovalResult` requires
+      // `appliedContent` on acceptance, so every settle() call below already
+      // supplies it.
+      return await settlement;
     } finally {
       this.requests.delete(requestId);
       await preview.dispose();
--- a/src/controllers/progressView/ProgressFollowUpPolishController.ts
+++ b/src/controllers/progressView/ProgressFollowUpPolishController.ts
@@ -73,7 +73,7 @@ export class ProgressFollowUpPolishController {
       if (result.success) {
         return {
           kind: 'updated',
-          update: this.createUpdate(input.stream, 'polished', result.text),
+          update: this.createPolishedUpdate(input.stream, result.text),
         };
       }
       if (!result.error) {
@@ -81,9 +81,7 @@ export class ProgressFollowUpPolishController {
       }
       return {
         kind: 'failed',
-        update: this.createUpdate(input.stream, 'polishError', null, {
-          error: result.error,
-        }),
+        update: this.createPolishErrorUpdate(input.stream, result.error),
         userMessage: result.error,
       };
     } catch (error) {
@@ -91,9 +89,7 @@ export class ProgressFollowUpPolishController {
       const logMessage = `Error polishing follow-up: ${errorMsg}`;
       return {
         kind: 'exception',
-        update: this.createUpdate(input.stream, 'polishError', null, {
-          error: errorMsg,
-        }),
+        update: this.createPolishErrorUpdate(input.stream, errorMsg),
         userMessage: logMessage,
         logMessage,
         ...(error instanceof Error && { logData: error }),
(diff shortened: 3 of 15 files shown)


LionSR:
@codex review

PR #9604, opened 2026-08-02
fix(cli): resolve custom roster names

LionSR:
## Summary

- resolve persisted bare agent identifiers against the current catalog before building CLI roster selections
- preserve canonical custom identifiers and unknown entries while deduplicating aliases
- keep the  selection path unchanged

## Regression coverage

- a bare built-in name such as  resolves to 
- canonical custom entries remain available in the roster

This follow-up was prepared during the final review of #9593 but was not included in its merge.

---

> [!NOTE]
> **Medium Risk**
> Changes workspace roster persistence and toggle semantics for mixed bare/qualified identifiers; wrong matching could drop or duplicate agents in custom rosters, but scope is limited to CLI agent configuration.
> 
> **Overview**
> Fixes how **persisted agent roster identifiers** are interpreted in the CLI when building selections and when toggling agents on/off.
> 
> **`selectedAgentKeys`** (used for chat-default and custom roster UI) now maps each stored id through `agentMatchesIdentifier` against the current catalog, emits canonical `source:name` keys, **deduplicates** bare vs qualified aliases, and **keeps unknown** ids unchanged.
> 
> **`setAgentEnabled`** in `AgentRosterController` matches stored entries by resolving candidates to the same canonical key as the toggled agent. Disabling removes **all** aliases for that agent (e.g. both `lead` and `builtInToolUse:lead`), while distinct same-name agents from different sources stay separate when only one source is turned off.
> 
> Regression tests cover alias deduplication, unknown ids, duplicate names across sources, and multi-alias removal on disable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 520adf5497f6a1d0cf66558769678cd7d48c755d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/tui/forms/AgentRosterForm.tsx
+++ b/packages/cli/src/chat/tui/forms/AgentRosterForm.tsx
@@ -17,7 +17,11 @@ import {
   type SelectWindowSize,
 } from '@cli/tui/selectWindow';
 import { platform } from '@platform/platform';
-import { agentKeyOf, type AgentCategory } from '@shared/schemas/agent';
+import {
+  agentKeyOf,
+  agentMatchesIdentifier,
+  type AgentCategory,
+} from '@shared/schemas/agent';
 import {
   AGENT_MODE_PRESETS,
   STARTER_AGENT_MODE_PRESET,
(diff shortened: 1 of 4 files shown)


LionSR:
Superseded by #9600, which fixes identifier canonicalization at the shared roster-controller boundary for CLI, extension, and desktop.

LionSR:
Closing without merge after final policy review. AGENTS.md lines 298-301 explicitly require CLI/TUI workflow-agent rosters to use only the current schema and prohibit rewriting old agent identifiers. Resolving persisted bare identifiers to whichever current catalog source wins would reintroduce a retired compatibility transform and can silently bind stale state to the wrong agent. The  pass-through in AgentRosterController.selectionKeys is therefore deliberate, not a missing canonicalization layer. No current-schema bug remains to fix here; the branch and tests will not be landed.

PR #9644, opened 2026-08-02
fix(cli): reject blocked Windows output paths

LionSR:
## Summary

Fix another bounded Windows compatibility cluster from #9409. Windows may report `ENOENT` rather than `ENOTDIR` when an output path descends through a file, causing TeXRA to accept an impossible destination and run the full workflow before writing fails.

- preflight with the same native recursive `mkdir` operation the output writer uses
- create `--output-dir` itself, or only the parent directory for `--output`
- reject blocked file ancestors before execution without custom path traversal
- preserve valid exact dangling `--output` symlinks that `copyFile` can follow
- report missing roots and dangling ancestors before the workflow starts
- cover Windows drive/UNC behavior through injected native-operation seams

## Contract decision

Preflight may create output directories before the workflow starts. This is intentional: exercising the real operation is simpler and more portable than reproducing filesystem traversal semantics, and the paths are destinations the successful workflow would create anyway. It does not create or overwrite the `--output` file.

## Scope

This does not claim to re-enable the full Windows matrix. Other path, environment, and fake-filesystem clusters remain independent.

## Verification

- Full local suite: 823 files, 8,400 tests passed; 1 file / 5 tests skipped
- `npm run format:check`
- `npm run lint`
- `npm run typecheck`
- `npm run compile:fast`
- `npm test`

Refs #9409

---

> [!NOTE]
> **Medium Risk**
> Changes CLI filesystem preflight and eagerly creates output directories before runs; behavior shifts on Windows and symlink edge cases, though scope is limited to `--output` / `--output-dir` validation.
> 
> **Overview**
> Fixes a Windows case where output paths descending through a **file** could look missing (`ENOENT`) during `stat`, so TeXRA ran the full workflow and only failed at write time.
> 
> **`probeOutputPath`** no longer relies on stat-only ancestor checks. After `ENOENT` on the target, it runs the same **recursive `mkdir`** the output writer uses: for `--output-dir` it creates the full directory; for `--output` only the parent. **`ENOTDIR` / `EEXIST`** from that mkdir are mapped to the existing “parent path component is a file” usage errors; persistent **`ENOENT`** gets clearer “cannot be created” messages. Dependencies are injectable for tests (`probeOutputPathForTests`).
> 
> Preflight **may create output directories** before the workflow starts (intentional—same destinations a successful run would create); it does **not** create the `--output` file. **Dangling `--output` symlinks** to absent files remain allowed; **dangling `--output-dir`** symlinks are rejected before execution.
> 
> Tests cover Windows drive/UNC paths with mocked native ops, mkdir failure modes, and integration updates (e.g. missing `--output-dir` is now created during assert).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5333e127e58d70ac48e591edf6a80ac1e7bcc6e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/runtime/workflowOutput.ts
+++ b/packages/cli/src/runtime/workflowOutput.ts
@@ -31,23 +31,26 @@ function joinCwdRelative(target: string, cwd: string): string {
 
 type OutputPathStats = Pick<Awaited<ReturnType<typeof fs.stat>>, 'isDirectory'>;
 
+type OutputFlag = '--output' | '--output-dir';
+
 interface OutputPathProbeDependencies {
   readonly stat: (target: string) => Promise<OutputPathStats>;
-  readonly lstat: (
+  readonly mkdir: (
     target: string,
-  ) => Promise<OutputPathStats & { isSymbolicLink(): boolean }>;
+    options: { readonly recursive: true },
+  ) => Promise<string | undefined>;
   readonly dirname: (target: string) => string;
 }
 
 const outputPathProbeDefaults: OutputPathProbeDependencies = {
   stat: fs.stat,
-  lstat: fs.lstat,
+  mkdir: fs.mkdir,
   dirname: path.dirname,
 };
 
 function parentFileUsageError(
   target: string,
-  flagLabel: '--output' | '--output-dir',
+  flagLabel: OutputFlag,
 ): CliUsageError {
   return new CliUsageError(
     flagLabel === '--output-dir'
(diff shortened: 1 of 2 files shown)


LionSR:
Looks ad hoc to me. Are we reinventing the wheel

LionSR:
Keeping this draft pending a concrete preflight-side-effect decision. Independent review confirmed the Windows ENOENT defect is unique, but the current walker has two blockers: it rejects a valid exact dangling `--output` symlink that `copyFile` can follow/create, and trailing separators can bypass dangling-link detection. It also duplicates filesystem traversal logic.

The simplest robust replacement is to preflight with the native operation the writer will use: recursive `mkdir(dirname(output))` for `--output`, or recursive `mkdir(outputDir)` for `--output-dir`, then retain the exact-target directory check. That handles file ancestors, missing depth, drive/UNC roots, and dangling ancestors cross-platform, but it intentionally creates directories before the workflow runs. If that side effect is unacceptable, the alternative is a much smaller Windows-on-ENOENT parent check. Please choose that contract before implementation. The focused Windows lane should then include `OutputGuards.vitest.mts`.

LionSR:
U decide

LionSR:
Decision implemented: preflight now uses the same native recursive mkdir operation as the eventual writer. This intentionally permits creation of destination directories before execution, removes the custom stat/lstat ancestor walker, preserves exact dangling `--output` symlinks, and turns blocked ancestors into early failures. Independent review found no correctness issue; its test-seam and deterministic ENOENT coverage findings are addressed in eb2962c239. Full local format, lint, typecheck, compile, and 8,400-test suite passed.

LionSR:
Addressed Cursor’s Windows-specific follow-up in 5333e127e5: recursive `mkdir` collisions reported as `EEXIST` now map to the same early `CliUsageError` as `ENOTDIR`. Deterministic drive and UNC coverage now exercises both native codes (22 focused tests pass). The broader eager-directory side effect remains the explicit contract decision documented in the PR body.

LionSR:
The latest CI failure is an unrelated existing timing flake in `SystemUtils.vitest.ts` (`startSleeper` read an incomplete PID file); the changed output-guard tests passed, every other lane passed, and the full local suite passed 8,402 tests. Rerunning the failed shard.

Issue #9590, opened 2026-08-02
Tracking: reduce StreamLogStore/StreamSnapshotStore surface and remove current-path resolvers


LionSR:
## Current status

Stages 1–6 are complete. The original design inventory below has been replaced by the current authority rules on `main`:

- exact stream identity is required for run-local mutations;
- current execution-to-stream association comes from registered execution metadata;
- legacy scans and suffix decoding are isolated from current paths;
- current metadata writes no longer maintain the retired mirrors removed in these stages.

Completed implementation:

- Stage 1: exact `clearMissingOutputs` identity — #9647
- Stage 2: current transcript identity — #9657
- Stage 3: explicit legacy boundary — #9661
- Stage 4: current execution lookup — #9663
- Stage 5: current lifecycle ownership — #9665
- Stage 6: current projection cleanup — #9667

## Remaining work

Stage 7 is the only remaining stage and must not begin before **2026-11-01**:

1. remove the legacy identity scan and stream-suffix decoding arms;
2. remove the read-only sidecar `description` mirror and fallback;
3. delete compatibility-only schemas, branches, comments, fixtures, and tests with those readers;
4. recount the remaining public store methods and production references;
5. verify that current trace, archive, output, restart, and deletion paths still use registered execution metadata directly.

The field-local invalid-descriptor behavior reported in #9675 is a separate correctness issue. Its fix must preserve valid sidecar metadata without restoring the retired pre-`kind` descriptor inference.

## Completion condition

Close this issue after Stage 7 lands and the following searches find no current-path identity reconstruction:

- sidecar directory scans used to associate a current execution;
- suffix-derived execution ownership;
- configuration-derived stream identifiers for run-local mutations;
- the snapshot `description` mirror or fallback.

Historical proposal text is available in the issue edit history. This body intentionally describes only the present implementation and the remaining work.

LionSR:
Answers to the eight review questions, verified against main @ `6f66839401` by two read-only research passes. File:line citations are to current main.

## Q1. Is `ExecutionMeta.streamId` viable as the sole authority?

Yes for the fresh-run path, with four lifecycles the authority rule must name explicitly.

[...]

PR #9647, opened 2026-08-02
refactor(transcript): require exact stream addressing for clearMissingOutputs

LionSR:
## Summary

Stage 1 of #9590: run-local mutations are now exactly addressed. The `clearMissingOutputs` mutation requires the exact `StreamTabId` selected by the initiator; the configuration-based fan-out that could clear missing-output markers on every look-alike workflow tab (same agent/model/inputFile/outputFiles) is deleted end to end — payload arm, both resolver methods on `StreamSnapshotStore`, and the CLI TUI's resolver port. Producers that have no stream context (command palette / main-view pack and clean buttons) now clear nothing instead of broadcasting by configuration; producers that know their tab address it directly.

## Issue acceptance gates (#9590 Stage 1, rule A3)

Rule A3: "Any mutation of run-local state, including `clearMissingOutputs`, must carry the exact `StreamTabId` selected by the initiator. Agent/model/input/output configuration is query/display data, not command authorization."

Proof obligations satisfied:

- **(3) rejects configuration-only addressing** — the payload type is now `{ streamId: StreamTabId }`, so configuration-only addressing is unconstructible: every in-repo producer fails to compile without an exact streamId, and the `streamConfig` resolution machinery no longer exists to receive one. Per the owner's review, this is enforced at the type contract (all producers are repository-controlled); no runtime rejection guards or removed-shape tests were kept.
- **(4) duplicate configurations affect only the explicitly addressed stream** — `src/test-kernel/transcript/StreamSnapshotStore.vitest.ts`, "clears missing outputs only on the exactly addressed stream, even with duplicate run configurations (#9590 A3)": two streams carry identical workflow configs and missing-output markers; clearing one via the session event leaves the other's markers intact.
- **(11) deleted resolver symbols have zero production references** — grep proof below.

## Single source of truth

The mutation's address is the initiator-selected `StreamTabId`, carried on the payload from emit to every consumer. There is no secondary resolution rule anywhere: no configuration matcher, no target resolver, no per-host reimplementation (the CLI TUI's `MissingOutputTargetResolver` port is gone with the store methods).

Schema/compat note: `ClearMissingOutputsPayload` is a TypeScript type on the in-memory `SessionEventHub` — it is not Zod-validated, not persisted, and never parsed from a wire. The only external surface is CLI `--output-format ndjson`, which is output-only; `{ streamId }` was already a valid arm of the old union, so emitted records only narrow. No legacy-arm union/transform is needed because no old-format events can be queued or replayed.

2 public methods deleted from `StreamSnapshotStore` (`resolveMissingOutputTargets`, `findWorkflowStreamsMatching`), 0 methods, ports, facades, or helpers added. Also deleted: the private `WorkflowStreamMatch` type and its two matching helpers (`normalizeOutputFiles`, `sameOutputFiles`), the CLI `MissingOutputTargetResolver` port, and the extension's `FileOpTarget` emit-descriptor interface (`emitClearMissingOutputs` now takes the streamId directly).

## Net elements

```
packages/cli/src/chat/chatSessionController.ts     |   4 +-
 .../cli/src/chat/tui/state/subscribeRuntimeHost.ts |  42 ++----
 .../src/commands/housekeeping/cleanCommands.ts     |   9 +-
 .../src/commands/housekeeping/packCommands.ts      |  15 +--
 .../src/commands/housekeeping/streamEventUtils.ts  |  32 ++---
 .../backend/events/ProgressFactApplier.ts          |  15 +--
 src/shared/schemas/progressEvents.ts               |  22 +--
 src/test-kernel/cli/TuiStateAndFocus.vitest.mts    | 148 ++++++++++-----------
 .../transcript/StreamSnapshotStore.vitest.ts       |  56 ++++++++
 src/transcript/StreamSnapshotStore.ts              |  83 +-----------
 10 files changed, 178 insertions(+), 248 deletions(-)
```

Net −70 lines overall, −85 lines in `StreamSnapshotStore.ts` alone. Public method count on `StreamSnapshotStore`: 42 → 40 by declaration count on this diff (the issue's AST baseline counted 43 → this change removes exactly 2, adds 0).

## Consumer counts

Deleted symbols, production references after this PR (grep over `src/` and `packages/`, tests excluded): `resolveMissingOutputTargets` — 0, `findWorkflowStreamsMatching` — 0, `WorkflowStreamMatch` — 0, `streamConfig` payload arm — 0.

Producer migration (every emitter of the `clearMissingOutputs` session event):

| Producer | Before | After |
| --- | --- | --- |
| `packages/extension/src/commands/housekeeping/streamEventUtils.ts:11` (`emitClearMissingOutputs`, sole emitter) | streamId or config fan-out descriptor | exact `streamId` only |
| `packages/extension/src/commands/housekeeping/cleanCommands.ts:109` (`handleClean`) | config broadcast when no streamId | emits only with `config.streamId` |
| `packages/extension/src/commands/housekeeping/packCommands.ts:82` (`handlePack`) | config broadcast when no streamId | emits only with `config.streamId` |
| `cleanCommands.ts:59` (`runCleanCommand`: single/multiple clean) | config broadcast | no emit (no stream context) |
| `packCommands.ts:91,101` (`handlePackSingle`/`handlePackMultiple`) | config broadcast | no emit (no stream context) |

The progress-view toolbar path (`ProgressWorkflowActionsController.runFileOperation`, `src/controllers/progressView/ProgressWorkflowActionsController.ts:85`) already passes its exact `streamId` plus `skipProgressViewClear: true` and is unchanged.

## Host impact

- **Extension**: `ProgressFactApplier.handleClearMissingOutputs` resets the addressed tab directly. Palette/main-view pack and clean no longer clear markers on config-matching tabs (deliberate: those invocations select no stream). Toolbar behavior unchanged.
- **CLI**: TUI subscriber (`subscribeRuntimeHost.ts`) drops the snapshot-store resolver parameter and applies the reset to the addressed stream; `chatSessionController` no longer passes the store into the subscription. Headless NDJSON output emits only the `{ streamId }` shape (already valid before). Renderer (`runProgressRenderer`) ignored this fact and still does.
- **Desktop**: no producers or bespoke consumers; it shares `StreamSnapshotStore.attachSessionEvents` and `ProgressFactApplier`, so it gets exact addressing for free. Verified no desktop-side references to the deleted symbols.

## Scheduled refactoring

Stages 2–7 of #9590 (direct current-record identity, the dated legacy boundary, reverse/deletion semantics, writer-only mutation, description authority, legacy retirement) remain tracked on the issue. This PR deliberately does not touch reload/root-replacement/flush machinery (#9603 atomic workspace-transition boundary) or the executionId dual-write sites.

## Validation

- `npm run typecheck` — clean (workspace, test-kernel, CLI, extension, trace-viewer, desktop)
- `env -u FORCE_COLOR npx vitest run src/test-kernel/transcript src/test-kernel/progressView src/test-kernel/cli` — 214 files, 2790 tests, all pass
- `env -u FORCE_COLOR npx vitest run src/test-kernel/architecture` — 10 files, 71 tests, all pass (no ratchet widened)
- `npm run check:dead-code-ratchet` — OK, 18 vs 18 baselined (no new dead exports; deleted methods were live surface, not baseline entries)
- `npx eslint` on all touched production files — clean; `npm run format` — no drift

### DIFF
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -196,9 +196,7 @@ export function createChatSessionController(
   // projecting status, output, and approval-related facts after the root
   // promise settles. Installing this once also avoids duplicate projections
   // when another root starts while an earlier detached child is still alive.
-  disposers.push(
-    attachTuiRunFactSubscription(runtimeSession.events, snapshotStore),
-  );
+  disposers.push(attachTuiRunFactSubscription(runtimeSession.events));
 
   // Shared prelude of the three run-starting paths (start, resume,
   // follow-up-wake resume): resolve the model-keyed session context and
--- a/packages/cli/src/chat/tui/state/subscribeRuntimeHost.ts
+++ b/packages/cli/src/chat/tui/state/subscribeRuntimeHost.ts
@@ -118,13 +118,6 @@ function applyStage(streamId: StreamTabId, stage: StreamStage): void {
 function applyClearMissingOutputs({
   streamId,
 }: ClearMissingOutputsPayload): void {
-  // Exact addressing only (#9590 rule A3). A payload without a streamId is a
-  // defect: throw so the hub logs the dropped mutation loudly.
-  if (!streamId) {
-    throw new Error(
-      'clearMissingOutputs requires the exact streamId selected by the initiator; configuration-based addressing is not accepted',
-    );
-  }
   // The empty map is a destructive reset, not a round patch. Record its
   // source revision even when the current map is already empty so a cold
   // read that started before this fact cannot restore older disk warnings.
--- a/packages/cli/src/chat/tui/state/subscribeRuntimeHost.ts
+++ b/packages/cli/src/chat/tui/state/subscribeRuntimeHost.ts
@@ -13,7 +13,6 @@ import {
   type UpdateQueuedFollowUpsPayload,
   type UpdateStreamUsagePayload,
 } from '@shared/schemas';
-import type { StreamSnapshotStore } from '@transcript';
 import { assertNever } from '@utils/core';
 
 import {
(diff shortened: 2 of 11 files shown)


LionSR:
The current-contract core is sound, but the PR adds compatibility-rejection boilerplate at an in-memory typed boundary. `ClearMissingOutputsPayload` now requires `streamId`; all producers are repository-controlled, so the new `if (!streamId) throw` guards in both consumers and the `as unknown as ClearMissingOutputsPayload` rejection test conflict with the repository rule to trust controlled inputs and the owner direction not to add tests whose purpose is proving a retired format fails. Please destructure/use the required `streamId` directly, delete `ProgressBackendFactProjection`’s removed-shape test, and keep the consequential duplicate-configuration test only for the positive assertion that exact addressing clears one selected stream and leaves the other unchanged. That preserves the real A3 contract with less code and no synthetic legacy payload.

LionSR:
Addressed in 7bcffa55d7: removed the `if (!streamId) throw` compatibility-rejection guards from all three consumers (the typed in-repo contract is the enforcement — producers are repository-controlled), deleted the removed-shape rejection test in ProgressBackendFactProjection, and reduced the StreamSnapshotStore test to the positive assertion only (exact addressing clears the selected stream; the duplicate-configuration look-alike keeps its markers). Also fixed the CI kernel-shard failure: the new test now awaits `store.flush()` before the temp-dir cleanup (the ENOTEMPTY race came from unsettled async sidecar writes, not the contract change). PR body updated to match.

Issue #9531, opened 2026-08-01
Resumed native subagent turns can reuse stale results and lack idempotent delivery


LionSR:
## Severity

High: stale or duplicate child results can flood the parent model context, while an accepted follow-up may produce no visible new work.

## Relationship to prior fix

Related to #9345 / #9347, but narrower. #9347 correctly established generation-scoped continuation ownership. This issue covers the remaining result-boundary gap: resumed native subagent cycles have no stable turn/delivery identity, no idempotent parent admission, and can select or expose prior-cycle state.

`AGENT_INCIDENT_REPORT.md` records a 2026-07-31 incident in which a resumed tool-use child reportedly re-emitted its prior result about 60 times, never landed the requested run-2 edit, finished interrupted, and exposed no run-2 conversation rows. The historical execution artifact is no longer available and the running product version/reload state is unknown, so this report alone does not prove #9347's lease race recurred. The current code nevertheless has independently verifiable gaps below.

### Resumed response selection is not turn-scoped

`runToolUseFlow` derives the returned response by searching the entire persisted message history (`findLastAssistantText(shared.messages, ...)`) and may fall back to `shared.lastResponse`. If a resumed cycle reaches its result boundary without adding a new assistant message, it can select the prior cycle's response.

### Report/result storage is one mutable latest-value slot

`childRunLoop` overwrites one execution-level report/result-meta only after a turn reaches `deliverTurn`. Accepting a new follow-up does not create a pending turn record. If the resumed cycle is interrupted first, run 1 remains exposed as the latest result with no way to tell that run 2 was accepted and failed to settle.

### Result delivery is not idempotent

#9347 prevents competing continuation owners, but `PendingChildDelivery` / `FollowUpQueueInput` carry no logical turn or delivery ID. The parent queue blindly appends each admitted child result. If any producer repeats a result callback, the admission boundary cannot identify and suppress the replay.

### Existing tests bypass the production failure boundary

The closest tests mock resume and/or delivery persistence. There is no deterministic test that traverses:

`DelegateAgentTool(execution_id)` → real child queue/recovery → persisted WAITING cursor → resumed model turn → report/result persist → parent queue admission → reopened conversation archive.

## Expected behavior

- Every accepted initial instruction or explicit follow-up belongs to a distinct logical turn.
- A resumed result is derived only from work produced in that turn.
- One logical turn result is admitted to the parent at most once, independent of transport retries or repeated producer callbacks.
[...]

PR #9664, opened 2026-08-03
fix(agent): give child turns stable identity and idempotent delivery

LionSR:
## Summary

Stage 3 of #9531 (stages 1–2 landed as #9576 and #9658). The parent admission boundary had no logical delivery identity: `PendingChildDelivery` / `FollowUpQueueInput` carried nothing a replay could be recognized by, so the parent queue blindly appended every admitted child result, and the single mutable latest-value report/result slot let an interrupted new turn hide behind the previous one.

This PR introduces a narrow, execution-owned turn identity at the existing boundary — the #9347 queue lease remains the sole turn driver, exactly per the issue's design tournament:

- **Identity**: the child-run loop mints a deterministic `{ token, deliveryId }` per accepted turn (`<executionId>:turn:<n>`), stamps the token on the subagent `ResultMeta`, and records accepted vs completed turns in a new `turn-state` key in the execution KV store (optional, additive; pre-identity records load unchanged — no invented default token).
- **Idempotent admission**: `ToolUseFollowUpQueue.submit` does a synchronous check-and-add against a per-stream bounded admitted-id set (`BoundedIdSet`, cap 1000). The same delivery admitted 100× yields at most one parent message and at most one wake; identical text under distinct ids stays distinct (no content dedup). In-memory, transport-level suppression only — documented as **not** crash-safe exactly-once; no outbox, per the issue's durability threshold.
- **Interrupted-turn attribution**: `/report` and `/result` prepend a one-line note when the accepted turn never persisted a result ("still running" while live, "was interrupted before producing a result" once terminal), so turn 1 is never silently presented as turn 2.

Refs #9531 (remaining: crash/restart durability only if reproduction warrants the outbox; the `appendFollowUp` bypass note is in the code).

## Issue acceptance gates

- [x] Each child turn has a stable logical identity carried through result formatting and parent delivery (token on `ResultMeta`, `deliveryId` on `FollowUpQueueInput`).
- [x] Replaying the same logical delivery 100 times appends at most one parent message and triggers at most one parent wake (regression tests through the real admission path + the production-shaped delegate/resume suite).
- [x] Distinct turns with identical text remain distinct (identity-keyed, never content-keyed).
- [x] Interrupting turn 2 before result persistence does not expose turn 1 as turn 2; `/report` and `/result` identify latest completed vs active/interrupted turn.
- [x] Concurrent same-id submissions admit at most once (synchronous check-and-add).
- [x] Existing #9347 ownership/race tests, the #9576 production suite, and the #9658 answerless-turn tests pass unchanged in behavior.
- [x] In-memory suppression is documented as not crash-safe; no outbox built.

## Single source of truth

Turn identity is minted in exactly one place (`mintChildTurnRef` in `childRunLoop.ts`); replay suppression lives in exactly one place (`ToolUseFollowUpQueue.submit`); turn attribution is read from one record (`turn-state` in the execution KV store).

No new registry, facade, or event system. One KV key, one optional field on two existing schemas, one bounded set per queue entry.

## Net elements (R6)

- Files: 18 changed (+645/−9, most of it tests)
- Exported symbols: +2 types via the existing `src/agent/storage` surface (`ChildTurnRef`, `ChildTurnState`), both with consumers in this PR; +1 `persistChildRunTurnState`; +1 `turnAttributionNote` (file-local)
- Classes/interfaces/enums: 0 added; one schema key (`turn-state`), two optional schema fields
- Dead-code ratchet: 18 vs 18 baselined — OK

## Consumer counts (R8)

- `FollowUpQueueInput.deliveryId`: 1 producer (child-run loop delivery), 1 consumer (queue admission); callers without ids are never suppressed.
- `readTurnState`/`writeTurnState`: 1 writer (child-run loop), 1 reader (`ExecutionsTool` attribution), plus test fakes.
- `turnToken`: 1 stamper, read by `/result` attribution.

## Host impact

Extension/Desktop/CLI: subagent result delivery becomes replay-safe; `/report` and `/result` gain the attribution note. No wire-protocol changes beyond the additive optional fields.

## Scheduled refactoring

Per the issue's durability threshold: an outbox/idempotent parent inbox is warranted only if crash/restart tests reproduce duplicate or lost insertion across a restart (the in-memory set and the 1000-id LRU cap are documented as transport-level). Possible small follow-up: migrate the two pre-existing production-suite tests to the new `launchWaitingChild` fixture helper.

## Validation

- Reproduction-first: all six new regression tests failed for the right reason before the fix (blind append, no turn identity), pass after
- `npx vitest run src/test-kernel/tools/ src/test-kernel/agent/` — 295 files / 2883 tests pass
- `npm run typecheck:workspace`, `typecheck:test-kernel`, `npm run lint`, `npm run format`, `check:dead-code-ratchet` — clean

---

> [!NOTE]
> **Medium Risk**
> Touches parent continuation admission, child-run delivery choreography, and execution KV writes—high-impact orchestration paths, but changes are additive with broad regression coverage and no durable outbox yet.
> 
> **Overview**
> Child runs now get **stable per-turn identity** and **replay-safe parent admission**, so duplicate child-result callbacks cannot spam the parent queue or re-wake the parent, and `/report` / `/result` no longer mis-label an interrupted newer turn as the current result.
> 
> The child-run loop mints deterministic `{ token, deliveryId }` per accepted turn, stamps `turnToken` on subagent `ResultMeta`, attaches `deliveryId` to parent follow-ups, and persists **`turn-state`** (`activeTurn` vs `lastCompletedTurn`) with serialized writes via `PQueue`. **`ToolUseFollowUpQueue.submit`** does a synchronous check-and-add on a per-stream bounded admitted-id set (cap 1000); repeats return **`duplicate`** and skip enqueue/wake. **`submitFollowUp`** and delegation tools surface that status; **`ExecutionsTool`** adds turn-attribution notes (and embeds them in `/result` JSON when needed).
> 
> Suppression is **in-memory only** (documented—not crash-safe exactly-once). Follow-ups without `deliveryId` behave as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 908723138528f8dadc590f100ff46bc81261a89a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
**Codex P1 (preserve valid JSON for attributed execution results)** — fixed in 3c15360. The attribution note now rides inside the JSON as a `turnAttribution` field instead of prefixed prose; the production regression now asserts `JSON.parse` succeeds and both the note and the prior turn's result payload are present. `/report` (prose endpoint) keeps the prefixed note.

LionSR:
**Cursor Bugbot (turn-state write ordering race)** — confirmed and fixed in 5466a38. `KVStore.write` indeed has no per-key serialization (the earlier comment's claim was wrong). Both turn-state writes now go through a per-loop `PQueue({ concurrency: 1 })`: the acceptance write stays fire-and-forget (synchronous first-turn launch timing preserved) but is guaranteed to land before `deliverTurn`'s completion record, which is awaited through the same queue.

PR #9669, opened 2026-08-03
Consolidate duplicate OpenAI client setup and reimplemented error helpers

LionSR:
## Summary

Scheduled code-consolidation sweep. Three independent Explore agents scanned `src/utils`/`src/shared`/`src/tools`, `src/agent`/`src/model`, and the three host packages for duplicate logic and hand-rolled reimplementations of native/existing helpers. The codebase turned out to already be heavily consolidated — most categories (debounce, sleep, retry/backoff, dedup, UUID gen, promise-queue serialization) came back clean. This PR lands the handful of genuine, low-risk findings that survived review.

## Issue acceptance gates

No linked issue — this is a scheduled maintenance sweep, not a fix for a filed bug. Acceptance gate: each change is a behavior-preserving consolidation, verified by full typecheck, lint, and the relevant test suites.

## Single source of truth

- `OpenAIClientModelHandler` (new, `src/agent/modelHandlers/openai/OpenAIClientModelHandler.ts`) now owns OpenAI-family client construction (`createOpenAIClient`/`getClient`/`getRetryEndpoint`) for both `ModelHandlerOpenAI` and `ModelHandlerOpenAIResponse`, kept off the generic `ModelHandler` base so non-OpenAI providers don't carry the OpenAI SDK dependency (mirrors the existing rationale on `logOpenAICompatibleClientConfig`).
- `ensureError` / `extractErrorMessage` / `isObject` (all pre-existing in `@utils/errors/errorMessage` and `@utils/core`) are now the single source for error-to-string/Error coercion in `extension.ts`, `createLanguageModelPort.ts`, and the desktop renderer bootstrap path, replacing three independent reimplementations. Note (per review discussion below): this swap is byte-identical for `ensureError`, but `extractErrorMessage` trims and drops empty/whitespace-only messages where the two replaced inline checks did not — a deliberate, safer edge-case behavior (falls back to a meaningful default instead of a blank message), not a strictly zero-diff substitution. Similarly `isObject` excludes arrays where one replaced inline check didn't, on a path (VS Code Language Model errors) that never surfaces arrays.
- The Execute-button shortcut label in `InstructionPanel.ts` is now derived from the `texra.execute` entry in the shared command catalog (`commandCatalogById` + `formatDesktopAccelerator`) instead of a second hardcoded/independent platform-sniffing string.

Removed: two byte-identical ~30-line method clusters (`createOpenAIClient`/`getClient`/`getRetryEndpoint`) duplicated across `modelHandlerOpenAI.ts` and `modelHandlerOpenAIResponse.ts`; three independent reimplementations of `ensureError`/`extractErrorMessage`-shaped logic; one independently-hardcoded keyboard-shortcut string that could silently drift from the real keybinding.

Added: one new abstract class, `OpenAIClientModelHandler`. It owns exactly one invariant — "OpenAI-family handlers share one client-construction/credential/logging path" — and sits between `ModelHandler` (generic, SDK-agnostic) and the two concrete OpenAI handlers, matching the existing `openAIChatHelpers.ts` pattern of keeping OpenAI SDK specifics off the generic base. `ModelHandlerCodex`'s existing `super.createOpenAIClient()` override chain is unaffected — it now resolves through the new intermediate class instead of `ModelHandlerOpenAIResponse` directly.

No other new abstractions; the error-helper and shortcut-label changes are pure call-site substitutions with no new exports.

## Net elements (R6)

Diffstat: 7 files changed, 93 insertions(+), 107 deletions(-) (net −14 lines).

Exported symbols (`^[+-]export` over the diff):
- `+export abstract class OpenAIClientModelHandler<...>` (new)
- `-export class ModelHandlerOpenAIResponse extends ModelHandler<...>` / `+export class ModelHandlerOpenAIResponse extends OpenAIClientModelHandler<...>` (same export, new base)

Net: +1 class, 0 net new public API surface beyond the new shared base itself.

## Consumer counts (R8)

No emit paths or subscriptions touched. `createOpenAIClient`/`getClient`/`getRetryEndpoint` had exactly 2 duplicate call-site definitions before this change (grepped and confirmed by an independent audit agent); both now delegate to the single base implementation. `ModelHandlerCodex` (1 additional override site) still calls `super.createOpenAIClient()`, which now resolves one level higher in the prototype chain — verified via full typecheck and the Codex-specific test suites (`CodexExperimentalTransports`, `CodexSubscriptionFallback`, `CodexEncryptedReasoning`).

## Host impact

Extension: `extension.ts` (Supabase init error handling), `createLanguageModelPort.ts` (VS Code Language Model API error translation), `InstructionPanel.ts` (Execute-button shortcut label) — all VS Code-allowed zones, no `vscode`-coupling changes.

Desktop: `packages/desktop/src/renderer/main.ts` bootstrap-fallback error message — browser-bundled renderer code, using the already browser-safe `@utils/errors/errorMessage`.

CLI: No changes.

## Scheduled refactoring

None. A larger, higher-risk finding (~18 call sites of `vscode.env.openExternal`/`revealFileInOS` across extension command handlers that could route through the existing `ExternalOpener` port) was identified by the audit but deliberately not implemented here — it needs light dependency-injection wiring into several plain command handlers rather than a pure substitution, so it's left as a follow-up rather than bundled into this low-risk sweep.

## Validation

- `npm run typecheck` — full workspace (root, test-kernel, agent, extension, cli, trace-viewer, desktop) — all pass.
- `npm run lint` — clean on the full touched-file set (two import-order nits were caught and fixed).
- `npx prettier --check` on all touched files — clean.
- `npm run check:dead-code-ratchet` — 18 findings, matches baseline exactly, no new dead code/exports.
- `npx vitest run` on the full `src/test-kernel/agent/modelHandlers/` suite (63 files, 706 tests, 4 pre-existing skips) plus `src/test-kernel/frontend/createLanguageModelPort.vitest.ts` — all pass.

### DIFF
--- a/packages/desktop/src/renderer/main.ts
+++ b/packages/desktop/src/renderer/main.ts
@@ -76,6 +76,7 @@ import {
 import { renderEmptyState } from '@shared/wa/emptyState';
 import type { TeXRAIconName } from '@shared/wa/iconNames';
 import { waIcon } from '@shared/wa/webAwesomeIcons';
+import { extractErrorMessage } from '@utils/errors/errorMessage';
 
 import {
   DesktopOpenWorkbenchMessageSchema,
@@ -1187,9 +1188,7 @@ function rerenderShell(): void {
 }
 
 function toBootstrapErrorMessage(error: unknown): string {
-  if (error instanceof Error) return error.message;
-  if (typeof error === 'string') return error;
-  return 'TeXRA could not finish starting up.';
+  return extractErrorMessage(error) ?? 'TeXRA could not finish starting up.';
 }
 
 function renderBootstrapFallback(error: unknown): void {
--- a/packages/extension/src/extension.ts
+++ b/packages/extension/src/extension.ts
@@ -100,7 +100,7 @@ import { setLeanLanguageServices } from '@tools/lean/leanLanguageServices';
 import { setInlineCommentProvider } from '@tools/comment/InlineCommentTool';
 import { ephemeralTranscriptWarning, StreamLogStore } from '@transcript';
 import { StorageFS } from '@utils/files';
-import { toErrorMessage } from '@utils/errors/errorMessage';
+import { ensureError, toErrorMessage } from '@utils/errors/errorMessage';
 
 // Local file imports
 import { ProgressViewProvider } from './progressView/ProgressViewProvider';
@@ -478,9 +478,7 @@ export async function activate(context: vscode.ExtensionContext) {
       }
     }
   } catch (error) {
-    const initError =
-      error instanceof Error ? error : new Error(toErrorMessage(error));
-    SupabaseClient.setInitError(initError);
+    SupabaseClient.setInitError(ensureError(error));
     logger.error(
       'extension',
       `Failed to initialize Supabase authentication: ${toErrorMessage(error)}`,
--- a/packages/extension/src/frontend/lm/createLanguageModelPort.ts
+++ b/packages/extension/src/frontend/lm/createLanguageModelPort.ts
@@ -14,6 +14,8 @@ import {
   type LanguageModelReference,
   type LanguageModelResponsePart,
 } from '@platform/languageModel';
+import { isObject } from '@utils/core';
+import { extractErrorMessage } from '@utils/errors/errorMessage';
 
 function translateLanguageModelError(
   error: unknown,
@@ -31,10 +33,7 @@ function translateLanguageModelError(
     );
   }
 
-  const code =
-    typeof error === 'object' && error !== null && 'code' in error
-      ? (error as { code?: unknown }).code
-      : undefined;
+  const code = isObject(error) && 'code' in error ? error.code : undefined;
   switch (code) {
     case 'NoPermissions':
       return new LanguageModelPortError(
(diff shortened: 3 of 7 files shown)


LionSR:
Thanks for the careful read — the two behavioral deltas you flagged are real and I agree with your assessment on both:

- **`extractErrorMessage` trims and drops empty/whitespace-only messages.** This is intentional and, on reflection, an improvement rather than a neutral swap: a bootstrap or language-model error with a blank/whitespace `message` now falls through to a meaningful default (`'TeXRA could not finish starting up.'` / `` `Language model${model} request failed.` ``) instead of rendering nothing useful to the user. I should have described this as "behavior-preserving for the realistic error shapes these paths see, with a couple of edge-case improvements" rather than flatly "behavior-preserving" — that claim in the PR description was overstated. No code change intended here; leaving as-is.
- **`isObject` excludes arrays** where the old inline check didn't. As you noted, the VS Code Language Model API never throws an array as an error, so this is a dead-path narrowing with no practical effect. No action needed.

Not changing the code for either finding, since both are the correct/safer behavior and neither has a real call path where the old behavior mattered. Will amend the PR description language for future reference.


---

LionSR:
The OpenAI client-construction work was superseded by the smaller focused PR #9672, which is now merged. The remaining error-helper and shortcut-label edits are unrelated to one another and should be proposed separately only if they still justify their maintenance cost.

PR #9684, opened 2026-08-03
refactor: retire legacy flow checkpoints

LionSR:
## Summary

TeXRA now reads and writes one current persisted-flow checkpoint envelope. This removes support for version-1 or unstamped records, absent replay cursors, node histories without node identifiers, the removed per-flow params channel, and unknown envelope fields.

The current writer has emitted schema version 2 with an explicit graph cursor since July 2026. CLI and TUI workflow checkpoints are intentionally current-format-only, so retaining a second replay algorithm and silently restamping old records only increases maintenance cost.

## What changed

- require schema version 2, the `texra` flow name, an explicit cursor, and node identifiers
- reject unknown envelope, cursor, and node fields at the persistence boundary
- remove node-history replay and schema-restamping paths
- validate an existing record before `PersistedFlow` caches it
- delete compatibility-only tests and convert current-contract fixtures to the current envelope

## Scope and compatibility

This is an intentional breaking change for historical internal CLI/TUI flow checkpoints. It does not change the current writer or any public wire protocol. Old checkpoints fail as unsupported instead of being inferred or rewritten.

## Validation

- `corepack pnpm run typecheck` (all workspace targets)
- `corepack pnpm run lint`
- focused Vitest run: 9 files, 145 tests passed
- `git diff --check`

## Maintenance effect

11 files, 72 insertions, 128 deletions. The deletion removes one compatibility parser branch, one replay implementation, one mutation helper, and two compatibility-specific tests.

Related to #6981.

---

> [!NOTE]
> **Medium Risk**
> Intentionally breaks resume for historical checkpoints and changes persistence validation at the resumability boundary; current writers are unchanged but any stored legacy record will no longer resume.
> 
> **Overview**
> **Retires legacy persisted-flow checkpoint compatibility** so CLI/TUI resume only accepts the current envelope instead of inferring or rewriting older shapes.
> 
> The **`FlowRecord` contract is tightened**: `schemaVersion` must be `2`, `flowName` must be `texra`, `cursor` is required, and node audit entries must include `nodeId`. Zod validation switches to **`strictObject`**, so unknown fields (including removed `params`) fail at read time. **`stampFlowRecordSchemaVersion` is removed**—writes persist the record as-is without bumping/restamping on every KV write.
> 
> **Resume semantics drop the legacy path**: `replayLegacyNodePath` and cursor-less replay from `nodes[]` are gone; **`resolveCursor` only follows `cursor.nextNodeId`**. `PersistedFlow.ensureRecord` loads existing records via **`readPersistedFlowRecord`** so invalid envelopes are rejected before caching.
> 
> **Tool-use recovery** still self-heals **shared** state but no longer restamps the envelope when patching `flowRecord.shared`.
> 
> Tests drop legacy migration cases and update fixtures to the current envelope; old internal checkpoints surface as **unsupported** rather than being migrated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 550e8003f3cb25e18341ad88b8d9f5fabc1dcf38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/implementations/flows/tooluse/runToolUseFlow.ts
+++ b/src/agent/implementations/flows/tooluse/runToolUseFlow.ts
@@ -21,7 +21,6 @@ import {
   PersistedFlowStateError,
   flowKey,
   readPersistedFlowRecord,
-  stampFlowRecordSchemaVersion,
 } from '@agent/node/persistedFlow';
 import {
   AgentCategory,
@@ -479,10 +478,7 @@ export async function runToolUseFlow<C = unknown>(
       };
       if (!isDeepStrictEqual(flowRecord.shared, resumedShared)) {
         flowRecord.shared = resumedShared;
-        await kv.write(
-          flowKey(executionId),
-          stampFlowRecordSchemaVersion(flowRecord),
-        );
+        await kv.write(flowKey(executionId), flowRecord);
       }
     } else if (flowRecord) {
       const migrationResult = migrateSharedState(flowRecord.shared);
@@ -525,10 +521,7 @@ export async function runToolUseFlow<C = unknown>(
           logger.debug('Normalized persisted tool-use shared state');
         }
         flowRecord.shared = migratedData;
-        await kv.write(
-          flowKey(executionId),
-          stampFlowRecordSchemaVersion(flowRecord),
-        );
+        await kv.write(flowKey(executionId), flowRecord);
       }
     }
     // Cleanup may delete a terminal flow record only after absence was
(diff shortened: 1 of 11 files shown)


LionSR:
Closing after the required surface audit. PersistedFlow is a shared released-session boundary used by extension tool-use and reflection resumes, not merely the current-only CLI/TUI workflow-script journal. Its version-2 cursor writer dates from July 2026, so it is still inside the three-month compatibility window. The deletion is mechanically clean but not yet justified under the retirement policy.

Issue #9627, opened 2026-08-02
Tracking: dated retirement queue for small compatibility arms (Aug–Nov 2026)


LionSR:
Rollup tracker for small, temporary compatibility readers. Larger persisted-data cohorts remain in #9422 and the final transcript-identity retirement was completed in #9590.

Every row is removed as one unit with its schemas, transforms, branches, comments, fixtures, and compatibility-only tests. Before removal, identify the former writer, verify that current writers cannot reproduce the old shape, and search both `src/` and `packages/` for migration or merge code that can carry it forward.

## Open rows

| Retire after | Temporary reader | Removal condition |
|---|---|---|
| 2026-09-14 | One-shot onboarding `firstRunDone` backfill | Introduced in v0.38.8. Delete the backfill and migration-only tests after the direct-upgrade window. |
| 2026-09-28 | `codicon-` prefix stripping in persisted custom-team icons | Current team writers have emitted canonical icon names since v0.39.0. Unknown-icon fallback remains an input-boundary safeguard and is not part of this retirement. |
| 2026-09-30 | Pre-refactor `.tex` resume probe in `runReflectionFlow.ts` | The legacy location is still a live carrier (every release through 0.40.0 wrote it; the switch landed in 0.40.1 on 2026-08-08). Remove only after the pre-0.40.1 resume window closes. |
| 2026-09-30 | Text-only `FollowUpQueue.drain()` | No production consumer remains. Delete the method and its compatibility-only tests after the support window. |
| 2026-10-13 | Team roster fallback from missing `texraHostedAgents` to all unresolved names | Current team writers always persist explicit hosted-agent provenance. |
| 2026-10-26 | `rN` task-group stage inference in `taskGroupProjection.ts` | Current group writers persist typed `kind` and `index` metadata. |
| 2026-10-31 | Legacy `EndGroupStatus` projection in `taskGroupProjection.ts` | Current group-end rows use `TaskGroupStatus`. |
| 2026-11-01 | Reads of former VS Code `ui.showOrchestratorBanner` and `ui.showLoginBanner` settings | Global-state dismissal keys have been the only current writer for the full support window. |
| 2026-11-03 | Copilot `copilot:` model-id cohort | Remove `normalizePersistedCopilotModelId`, transcript/runtime inference, `migrateCopilotModelRouteSelections`, `COPILOT_ROUTE_MIGRATION`, and `VscodeMainViewPersistedStateSchema` together. |
[...]

PR #9697, opened 2026-08-03
docs(proposals): consolidate run classification to one discriminant

LionSR:
## What

Adds `docs/proposals/2026-08-03-run-classification-consolidation.md`. Docs-only, no code change.

TeXRA has six overlapping vocabularies for "what kind of thing is running":

| Vocabulary | Verdict |
|---|---|
| `AgentCategory` `{workflow, toolUse}` | Real, irreducible. It is `AgentConfigSchema`'s persisted discriminant. Stays. |
| `RunKind` `{agent, process, workflowScript}` | The correct run axis. Already exists. Nothing treats it as authoritative. |
| `RunDescriptor.agent` + `.category` | Required beside `kind`, so process and script runs fabricate them. |
| `CreateChildStreamOptions.streamCategory` | False at 2 of 3 writers. |
| `ExecutionMeta.category` | Untyped persisted string mixing both axes. |
| `isProcessAgent` / `resolveExecutionDisplayCategory` / host-local kinds | The same decision respelled. |

A stream can carry a native agent, an external agent CLI, a background bash process, or a deterministic multi-agent script. Only the first two are agents at all, so the run has been borrowing a fact that belongs to agents.

## The proposal

One `RunIdentity` discriminated union where each arm carries only the identity that arm has, `RunKind` derived as `RunIdentity['kind']` rather than a parallel enum, and a `RunSurface` chrome table both hosts index instead of each computing labels and icons. Structure comes from `kind`, chrome from `surface`.

Includes an ownership table (one writer per field) and a declarative-UI section showing the exact render-time reconstructions that die.

**Net -16 elements**, plus 22 render-time reconstructions and 6 silent `?? Workflow` defaults removed. The counting method is stated inline: renames and retypes count zero, relocations onto a union arm count zero, branch sites counted separately.

## Two live defects found while mapping

Worth fixing regardless of whether the refactor proceeds:

1. `kind` was added at `runDescriptor.ts:24` without bumping the version at `:6`, so descriptors written before that change fail `z.strictObject` and are warn-and-dropped today. A sampled developer bucket showed 142 of 478 affected.
2. Four of six `registerExecution` call sites omit `category`, which is why no execution on disk is ever tagged `workflowScript`.

## Staging

Five PRs. **PR 1 is self-contained, touches no persisted shape, and unblocks `delegate_multi_agents`** on the `software-engineer` and `lean-project` teams, which declare `workflowAgents: []` and therefore cannot launch the tool at all today.

PR 1 must ship a checkpoint-digest fixture test. `deriveWorkflowScriptCheckpointId` salts on the resolved agent name and feeds `deriveExecutionId`, hence the executions directory, stream id and lease. Making the input nullish makes it tempting to drop from the hash, which would orphan every in-flight journal and defeat the double-launch guard.

## Provenance and honesty notes

Produced by a 14-agent audit with an adversarial verification pass. Two claims from earlier analysis were disproved and are recorded in the doc rather than quietly dropped:

- `texra.enabledAgents` / `texra.enabledToolUseAgents` have no production reader; the live key is `texra.agentRosterSelection`. Why the workflow roster resolves empty on those teams is **not** established, only the symptom is.
- `packages/trace-viewer/src/traceDataSchema.ts` does not hand-mirror `ExecutionMeta`; the `_IsExact` hazard is stale for that file.

Disk-scale numbers come from one developer bucket in a prior measurement pass and were not re-run. The Risks section keeps everything the critics flagged that could not be resolved from the files.

## Non-goals

The roster `workflowAgents`/`toolUseAgents` pair collapse (~25 field pairs, a genuine reduction) is a different axis needing a three-host wire change. Deferred.

https://claude.ai/code/session_01GE7LWMagZ9NyNg2fLZjjwV

---

> [!NOTE]
> **Low Risk**
> Documentation-only PR with no code, schema, or migration execution; risk is limited to planning accuracy for a future large refactor.
> 
> **Overview**
> Adds **`docs/proposals/2026-08-03-run-classification-consolidation.md`** (~1.3k lines). **Docs only** — no runtime changes.
> 
> The proposal replaces six overlapping “what is running?” vocabularies with **`RunIdentity`**, durable **`ExecutionMeta.identity`**, and the same struct on the wire; **`AgentCategory` stays** on agent config. It also folds in **26 lineage representations**, **Part II** (lifecycle `outcome`, retire `legacyExecutionIdentity`, single artifact/usage accumulators, `ByCategory<T>` roster collapse), and **Part III** (one session view-model + command layer for extension/desktop vs CLI).
> 
> **Delivery model shifts** from staged PRs to **one atomic change** with eight **holistic phases** (revision 10) so intermediate layers like upgraded `StreamSlice` are not built then deleted. **Migration** is **idempotent store-entrance stamping** (no generation marker); **`identity` required at `registerExecution`**, **optional on read/export** schemas after red-team review.
> 
> Documents **three live defects** (descriptor schema version, missing `workflowScript` tags, progress Resume borrowing script identity), compatibility classes across shared `~/.texra`, removal inventories with grep gates, red-team amendments, and an open-problems register. **Rejects** revision 1’s shared `RUN_SURFACE` table; workflow-roster root cause stays a **separate track**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05d79083c29f085bbaaf748be5191661d5226386. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Blocking merge pending a narrower revision. The authoritative `RunIdentity` discriminated union is a sound direction and the diagnosed classification defects are real, but the current five-PR blueprint exceeds that result:

- Keep shared code to one validated run identity. `RUN_SURFACE` currently combines identity with host-specific pane, resume, file-action, model-display, and log policy; extension and CLI should project their own UI unless the behavior is demonstrably identical.
- Do not persist or transmit a weakly validated `surface`. Derive it after parsing `RunIdentity`, or define a real runtime schema if a boundary truly needs it.
- Choose one durable authority for run identity. Copying related facts into descriptors, execution metadata, and optional trace fields weakens the single-source claim unless each copy has a necessary contract and an enforced derivation.
- State the result honestly as one authoritative identity with typed projections, not “six vocabularies to one”; several projections and `AgentCategory` remain.
- Every temporary compatibility reader must record its introduction date and exact retirement condition/date. The recent descriptor break may justify a narrow reader inside the three-month window, but it must be explicitly scheduled for deletion with its schema, fixture, and test.
- Separate the workflow-roster failure from this architecture proposal. Reproduce its actual cause at the true boundary before changing nullability or adding a checkpoint test; the proposal itself says the empty-roster mechanism is not established.

Recommended scope: one validated current run identity, one explicitly dated minimal reader for the recent descriptor break, and local host projections. The workflow-roster bug should be handled independently after causal reproduction.

LionSR:
Thanks both. All four citation nits were real; each was re-verified against the tree before editing, and all four are fixed in 5d63e00fc9.

- **`RunKind` "module-private"** was wrong, and the correction is sharper than the original claim. The `RunKind` *type* is exported at `runDescriptor.ts:9` and already consumed at `childStream.ts:17,28`. Only `RunKindSchema` stays unexported, which is the actual problem: nothing validates a kind at a boundary. The row now says that instead.
- **`streamState.ts`** now cites `:195`, the `kind: z.literal(AgentCategory.Workflow)` field that substantiates the claim, with the union at `:206-209` kept as a secondary reference.
- **`enabledToolUseAgents`** has zero hits anywhere; only `texra.enabledAgents` has the single `WorktreeMemento.vitest.ts:34` fixture hit. The old wording implied they shared it. Split explicitly.
- **`SubagentList.tsx`** duplicate is `panes/SubagentList.tsx:728-732`. Fixed, and the `panes/` path segment added.

**One reviewer claim did not survive checking, so the original count stands.** `texra-review` reports five `registerExecution` call sites and suggests "4 of 6" is a miscount. There are six:

| Site | `category` |
|---|---|
| `agentCliShared.ts:197` | omitted |
| `subagentExecution.ts:165` | omitted |
| `WorkflowScriptTool.ts:371` | omitted |
| `inBandSubagentExecution.ts:446` | omitted |
| `bash.ts:415` | `'process'` |
| `runAgent.ts:102` | `config.agentCategory` |

`runAgent.ts:102` is the missed one. The `executionLifecycle.ts:392` hit nearby is a doc comment, not a call, which is the likely source of the confusion. The ownership row now lists all six with their arguments so the count is auditable rather than asserted.

**On the PR 2 migration question**, which was the most useful comment: the posture was implicit and is now stated. It is read-time healing, forward-only, no rewrite pass over `streamData/*/meta.json`. The reasoning is that the descriptor is a derived, rebuildable projection of `config.json` rather than authored data, so a bad row costs a degraded tab label, not lost work. The consequence being accepted, now written down, is that the legacy arm is permanent for as long as those buckets exist and must be covered by a fixture test rather than treated as temporary scaffolding.

The checkpoint-hash and `normalizeWriterCategory` concerns need no doc change; both were already in Risks and both reviewers agree on the framing. The `normalizeWriterCategory` consumer grep is a PR 3 precondition, not a doc gap.

One caveat I want on the record rather than buried: the disk-scale numbers (478 descriptors, 142 kind-less) come from a prior measurement pass and were not re-run for this proposal. They set migration expectations, not correctness, since the legacy arms are total either way. If anyone wants them load-bearing, they should be re-measured first.

Adding `re-review` since pushes do not re-trigger the bots on this repo.

LionSR:
Revision 2 pushed as 52d4301969. I owe you an apology on process: my previous comment landed 27 seconds after your blocking review and did not respond to it. I had only read the bot summaries. That was my error, not a disagreement.

Taking the six points in order.

**1. Keep shared code to one validated run identity.** `RUN_SURFACE` is withdrawn. You are right that it smuggled host policy into the shared layer: pane, resume, file actions, model display and child-log behavior are host decisions, and bundling them into an identity table was the same category error the proposal exists to fix, one level up. The shared layer is now `RunIdentity` and nothing else. Each host projects its own UI. A shared table returns only for behavior demonstrably identical across hosts, proven case by case.

**2. Do not persist or transmit a weakly validated `surface`.** Removed from every wire and persisted shape. It is derived locally after parsing `RunIdentity`. The `z.custom<RunSurface>()` that revision 1 flagged as a known trade is gone rather than defended.

**3. One durable authority.** Named: `ExecutionMeta`, written once at birth by `registerExecution` inside the fresh-lease scope. Each remaining copy now has to pass both of your tests, and the doc states them in a table:

| Copy | Necessary contract | Enforced derivation |
|---|---|---|
| `ExecutionMeta.identity` | the authority | sole writer, inside the lease |
| stream descriptor | display renders before the config snapshot resolves, the window `streamTabInfo.ts:36-42` documents and today fills by parsing the stream ID | built by `buildRunDescriptor` from the same value passed to `registerExecution`, one call site (`childStream.ts:128`) |
| `trace.json` identity | an exported trace is standalone, with no access to `~/.texra` | written by the exporter from `ExecutionMeta` at export time |

This also resolves the divergence a bot reviewer raised inline: `identity.category` versus `config.json`'s copy. The enforcement is sequencing. `identity.category` is introduced only in the stage that deletes `normalizeWriterCategory`, so a frozen copy and a correctable copy never coexist. If that sequencing cannot hold, the copy gets dropped and display loads the config instead. That is written down as a condition, not an assumption.

**4. State the result honestly.** Retitled to "One authoritative run identity, with typed projections". The "six vocabularies to one" framing is gone from the title, the opening, and the results section, which now says plainly what remains: `AgentCategory` unchanged, `StreamTabInfo.kind` and `ExecutionListingEntry.kind` as typed projections, host presentation types local. The net -16 element claim is **withdrawn rather than restated**, because it credited host-local branch collapses to a shared table this revision no longer builds. Only measured figures are kept, and per-stage deltas get counted against the tree at implementation time.

**5. Dated compatibility readers.** Every reader now carries an introduction date, an exact retirement condition, and a retirement date, with schema, fixture and test deleted together. Descriptor and `ExecutionMeta` legacy arms: introduced 2026-08-03, retire by 2026-11-03. This reverses what I added earlier today, where I argued the descriptor reader should be permanent because the descriptor is derived. The one reader marked permanent is the `trace.json` fallback, and only because an exported trace is an immutable artifact; it is confined to that read boundary.

**6. Separate the workflow-roster failure.** Removed from the proposal entirely. You are right that it was incoherent to open with a nullability change while the same document says the mechanism is not established. It moves to its own track with the sequence stated: reproduce deterministically and identify the boundary, distinguishing roster-selection resolution from the admission gate at `WorkflowScriptTool.ts:287`; fix where the reproduction points; only then decide whether nullability changes and whether the checkpoint fixture is needed. The checkpoint salt hazard is recorded there so it is not lost.

Two smaller things also fixed from inline bot comments: `buildRunDescriptor` is called at `childStream.ts:128`, and `normalizeWriterCategory` lives under `src/agent/storage/`.

One item I have not resolved and am not going to paper over: the 142-of-478 disk figures are withdrawn pending measurement rather than re-measured. Nothing in the design depends on them, since the legacy arms are total either way. I can measure them against real buckets if you want them load-bearing for the retirement window.

LionSR:
Not yet happy with the TUI workflow agent displays and the workflow script.

They are related problems

LionSR:
Weighing in on revision 2 with one constraint applied that the doc never states, and that changes several of its decisions once stated: **the two hosts have asymmetric persistence obligations.** The CLI/TUI owns no durable format — its state model, the NDJSON progress vocabulary, and its interaction policy are in-memory or wire shapes whose two ends ship in the same binary. Everything there can break freely and atomically; the only external contract is headless output parity. The extension's histories (`executions/*/meta.json`, exported traces) can never break. Revision 2 classifies its compatibility machinery by *reader*; it should classify by *data class* — authored durable history, derived rebuildable projection, host-local ephemeral — and the three classes get three different regimes, not one dated table.

Applied concretely:

**1. The `ExecutionMeta` v1 legacy arm must not retire on 2026-11-03.** This is the one place revision 2 over-applied the "date every reader" review point. Mechanism, verified in the tree: `ExecutionKVStore.readValidated` (`src/agent/storage/ExecutionKVStore.ts:215-233`) warn-and-nulls a meta.json that fails validation, and `resumability.ts:116-128` returns `resumable: false` / `INVALID_META` on the same failure. Retire the legacy arm and every execution recorded before 2026-08-03 loses its timestamp, outcome, and description in listings and becomes non-resumable — which is precisely "breaking histories," delivered on a schedule. The descriptor arm is genuinely fine to date-retire: it is a rebuildable projection, worst case is a degraded tab label, and dropping is already today's behavior for those rows. But `ExecutionMeta` is the doc's own *authority* — authored once at birth, never regenerable. The doc's rationale for the permanent trace reader ("an exported trace is an immutable artifact") applies to it verbatim: a past execution is exactly as immutable as an exported trace. Either the v1 arm is permanent, or Stage 2 ships a one-time read-heal-rewrite migration at the version bump. A dated deletion of the authority's reader is the one option the extension's constraint rules out.

**2. The trace embedding contradicts the doc's own permanent reader.** `TraceDocumentSchema.meta` embeds `ExecutionMetaSchema.nullable()` directly (`src/transcript/traceDocumentSchema.ts:10,20`), and `packages/trace-viewer` consumes `TraceDocumentSchema` via `@transcript`. So the 2026-11-03 retirement of the `ExecutionMeta` v1 arm would also break `parseTraceData` (which throws) on previously exported traces — through the embedded meta, not through the `identity` field the doc's permanent `legacyRunIdentity` reader covers. Stage 2's "retype `category` to the identity" flows into the trace document schema whether or not that was intended. Making the v1 arm permanent (point 1) resolves this automatically; otherwise the trace boundary needs its own frozen copy of the v1 meta shape, which is worse.

**3. Invert the build order: CLI first, persistence last.** The staging currently puts the only zero-risk, fully breakable work — the CLI model replacement — at the tail of Stage 3, behind two stages that touch persisted shapes. Nothing forces that ordering. `RunIdentity` can be threaded **in-memory** from the three launch sites (which already know the truth: `bash.ts` knows it is a process, `WorkflowScriptTool.ts` knows it is a script) through the `run.start` progress fact into the CLI before any persisted byte changes. That one PR kills the five CLI reconstructions and the two-kind `StreamSlice` model atomically — no compatibility readers, no "add before remove, strictly" ceremony, since both ends of that wire ship together. It also delivers the thing actually complained about — the TUI workflow-agent displays and workflow-script rendering — first, and lets the persisted-authority stages (`ExecutionMeta`, descriptor) follow under the stricter regime with the schema already proven in use. The current ordering spends the caution where none is owed and defers the payoff.

**4. The last mixed concern: `category` on the agent arm.** The Problem section states, correctly, that `AgentCategory` "is a fact about an *agent*... not a fact about a *run*" — and the target model then persists it on the run identity. The sequencing defense (introduce `identity.category` only in the PR that deletes `normalizeWriterCategory`) answers the *staleness* objection; it does not answer the *concern-mixing* one. The descriptor copy earns its place through an access path: keyed by stream id, readable before the config snapshot resolves. The question to settle before Stage 2 is whether `category` has the same justification or only rides along: does anything in the early-render window actually branch on category, or could that window live on `{kind, name}` and pick up category when the config resolves? If the latter, the agent arm shrinks to `{kind, agent, tool?}`, the identity is purely the run axis, and the one remaining cross-axis field disappears — that is the genuinely clean version. If the former, keep it, but the doc should record the access-path reason, because sequencing alone defends a copy that maybe shouldn't exist.

Direction verdict: the rev-2 core — one validated authority, host-local projections, no shared chrome table, roster bug separated — is right and worth keeping. Points 1 and 2 are blocking (they violate the never-break-histories constraint the design must serve); point 3 is a reordering that costs nothing and front-loads the visible fix; point 4 is a question the doc should answer either way before Stage 2 lands.

## Verified

- Read the full revision-2 document at 52d4301 and all prior review rounds and author responses.
- `src/shared/schemas/stream.ts:77-119` — `ExecutionMetaBaseSchema` with `category: z.string().optional()`, `schemaVersion` prefault, and the `ExecutionMetaSchema` transform.
- `src/agent/storage/ExecutionKVStore.ts:215-241` — `readValidated` warn-and-null on parse failure; `readMeta`/`readMetaStrict` policies.
- `src/agent/storage/resumability.ts:100-130` — `safeParse` failure → `resumable: false`, `INVALID_META`.
- `src/transcript/traceDocumentSchema.ts:10,20` — `meta: ExecutionMetaSchema.nullable()` embedded in `TraceDocumentSchema`; `packages/trace-viewer/src/traceDataSchema.ts` imports it via `@transcript`.

---

LionSR:
Follow-up to my previous comment, after a full surface audit. The compatibility constraint has been refined: bash and workflow-script session history may break even in the extension, and everything desktop-side may break. Only native agent-run history in the VS Code extension is inviolable. Mapping every surface against that rule changes the proposal more than expected — mostly by **deleting compatibility machinery** — and it settles the `identity.category` question with evidence rather than argument.

## The constraint, stated precisely

One structural fact forces a reformulation: **all three hosts share the same `~/.texra` store.** Desktop resolves `resolveDesktopDataRoot()` → `DEFAULT_NODE_STORAGE_ROOT` (`packages/desktop/src/main/platform/paths.ts:45-52`, `src/platform/defaults/nodeStorage.ts:15-18`), the extension migrates its legacy `storageUri` data *into* that same root (`packages/extension/src/frontend/vscode/sharedStorageRoot.ts`), and the CLI uses it too (`packages/agent/src/node.ts:109`). So "desktop can break" cannot mean "desktop's persisted formats can break" — the same bytes are the extension's history. Breakability is a property of **row cohorts and surface classes**, not hosts:

| Class | Members | Regime |
|---|---|---|
| **Protected** | `executions/*/meta.json` + `config.json` rows whose run is a native agent; exported traces of agent runs (`TraceDocumentSchema.meta` embeds `ExecutionMetaSchema`, `src/transcript/traceDocumentSchema.ts:20`, and replay stamps `StreamState.kind` from `trace.config.agentCategory`, `replayTrace.ts:169`, where the discriminator **fails hard** on an unknown value) | Must parse forever, or be migrated once. Never date-dropped. |
| **Breakable durable** | The same files' `process` and `workflowScript` rows; `streamData/*/meta.json` descriptors (rebuildable — `descriptorFromConfig` fallback already exists, `StreamSnapshotStore.ts:232-241`); main-view `sessionType` webview state | Version-bump freely; degradation acceptable. |
| **Ephemeral** | Every UI wire shape on every host. Verified: `StreamTabInfo`, `StreamMetadata`, `SYNC_STREAM_CONTENT`, proposals, history messages, `StreamState`, `ProgressStreamRunDetails` (documented "not persisted", `ProgressViewState.ts:89`), CLI `StreamSlice` and NDJSON facts, desktop IPC. No memento or localStorage anywhere stores a kind/category from these. | Recut atomically. **Zero compat machinery, ever.** |

A second simplification: there are only **two UI stacks**, not three. Desktop reuses the extension's Lit frontends and the identical `ProgressBackend`/`WebviewUpdater` wire (`packages/desktop/src/renderer/main.ts:22-53`, `desktopAgentExecution.ts:256-266`); only the transport differs (postMessage vs Electron IPC). Recutting the extension webview shapes recuts desktop for free. The other stack is the Ink TUI.

## Consequence 1: the compat-reader table collapses to one row

Rev 2 carries three readers (descriptor arm dated, `ExecutionMeta` v1 arm dated, trace fallback permanent). Under the real constraint:

- **Descriptor legacy arm: delete it entirely, no date needed.** The descriptor is rebuildable from `config.json` via the existing `descriptorFromConfig` fallback, its failure mode is a degraded tab label, and the cohorts that most depend on `kind` (process/script) are breakable anyway. Bump `RUN_DESCRIPTOR_SCHEMA_VERSION` and move on.
- **`ExecutionMeta` v1: the schema-transform legacy arm is the wrong mechanism anyway.** `ExecutionMeta` does not carry an agent name, so a pure `category → identity` transform cannot synthesize the agent arm — it doesn't have `agent`. The join already happens at the established normalization boundary, `executionListing.ts:142-158`, which builds `ExecutionListingEntry.kind` from `meta.category` + config presence. So: make `identity` **required at write, optional at read**; readers treat `identity == null` as "pre-identity agent-era row, resolve from config" — which is literally today's code path, kept **permanently for agent rows** (that's the one protected obligation, and it satisfies the trace-embedding problem for free) and allowed to degrade for process/script rows. No `workflow-script#` stream-id sniffing needs building at all — that recovery only served a breakable cohort.

Net: **zero dated readers, one permanent absence-tolerant read path that already exists.** That is strictly cleaner than rev 2's table.

## Consequence 2: the responsibility boundaries, established from the behavior audit

I mapped every site where classification changes *behavior* rather than chrome. The pattern is unambiguous:

**`AgentCategory` is an execution-mode fact owned by the agent definition, and the runtime already treats it that way.** Every substantive behavior branch reads `config.agentCategory` or `setting.agentCategory`: flow-engine selection (`executeAgent.ts:468`), resume-type routing (`SessionResumeRetrieval.ts:111-124`, `resolveAndResumeStream.ts:123`), structured-output tool injection (`runToolUseFlow.ts:228-231`), model-handler mode and token safety buffers (`ModelHandler.ts:325-335,1873`), GPT background-mode eligibility (`modelHandlerOpenAIResponse.ts:434`), skill-catalog loading (`userVars.ts:192`), usage accounting (`UsageMonitor.ts:200`), helper-model refusal (`helperModelPreference.ts:37-42`), workflow-output opening (`runAgent.ts:139-141`). And `AgentLaunchContext.ts:276-293` already enforces that the YAML **wins and overwrites** the config at launch. `deriveResumability` itself has **no classification branch at all**. So the runtime's authority chain for category is settled and correct: setting → config. `RunIdentity` must not become a second behavioral source for it.

**`RunIdentity.kind` is a launch-site fact** — the three launch sites already know it and today either fabricate configs around it (`bash.ts:409-413`, `WorkflowScriptTool.ts:353-356`) or leak it through names. Everything currently recovering kind does so by name-parsing: `isProcessAgent` (5 sites), the CLI's four-prefix stream-id regex (`subscribeStreamLog.ts:138-142`), `meta.category === 'process'` string sniffing (`executionListing.ts:149`), `replayTrace.ts:159`. All of these are the sites `identity.kind` replaces.

**Three responsibility violations the audit surfaced, worth adding to the doc:**

1. **Runtime semantics sourced from a display projection.** `AgentExecutionHandle.category` reads the *descriptor* (`ExecutionHandle.ts:159-161`), and real behavior keys on it: `executions action=wait` returns immediately for a WAITING tool-use child (`waitCoordination.ts:43-50`), report suppression (`summaryFormat.ts:46`), and the `attachToolFlow` guard (`ExecutionHandle.ts:211`). The descriptor exists for the early-render window; runtime behavior reading it inverts the ownership. Re-source the handle's category from config/the authority when the descriptor changes shape.
2. **Write-time mutation serving a UI default.** `normalizeWriterCategory` rewrites persisted `config.json` bytes solely so `chatDefaults.loadHistoryDefaults` (`chatDefaults.ts:108-114`) won't adopt a background bash row's model. Once listing filters on `identity.kind !== 'agent'`, the mutation's purpose evaporates — and `runtimeCategory`, which exists purely to hide that demotion from the category column, goes with it.
3. **Mis-classification is destructive in one place.** `runExecution.ts:115-120`: a category mismatch marks the run ERROR **and deletes the flow record**, permanently killing resumability. This is the strongest argument in the tree for validating identity at birth rather than reconciling it later.

## Consequence 3: `identity.category` — answered, not argued

My previous comment posed this as a question. The audit answers it: **no runtime behavior reads a category copy outside `config`/`setting`** (the one exception is the descriptor-sourced handle above, which is a bug to fix, not a contract to preserve). The only legitimate consumers of a category copy outside config are display-timing ones — `streamTabInfo.ts:32`'s early-render window and listing normalization. So `category` on the agent arm of `RunIdentity` would be a **display snapshot living inside the shared identity** — exactly the mixed concern this proposal exists to remove.

The clean cut: `RunIdentity` carries the run axis only — `{kind:'agent', agent, tool?}`, `{kind:'process', tool}`, `{kind:'workflowScript', workflowName}`. If the early-render window needs category before the config snapshot resolves, the snapshot belongs on the **descriptor wrapper** (which *is* the display-timing projection, ephemeral in consequence and rebuildable) — explicitly named as a snapshot — never inside the identity or the authority. This also dissolves rev 2's most fragile dependency: the "introduce `identity.category` only in the PR that deletes `normalizeWriterCategory`" sequencing becomes unnecessary because there is no `identity.category`.

## Revised staging that follows from the classes

- **Stage A — ephemeral only, everything breakable.** Thread `RunIdentity` in-memory from the three launch sites through the progress facts; recut both UI stacks atomically: the Lit shapes (which fixes desktop simultaneously) and the CLI slice. This is also where the naming trap dies — `StreamMetadata.kind`, `StreamState.kind`, and `SYNC_STREAM_CONTENT.kind` all carry *category* under the name `kind` today; rename freely, it's all ephemeral. Delete the UI-side name-parsers.
- **Stage B — durable, under the one protected obligation.** `registerExecution` requires identity at all six sites; readers treat absent `identity` as legacy-agent via the existing listing/config join, permanently. Delete `normalizeWriterCategory`, `runtimeCategory`, and the `meta.category` string. Re-source `ExecutionHandle.category`. Filter `loadHistoryDefaults`/`isUserVisibleExecution` on `kind`.
- **Stage C — descriptor + trace.** Descriptor version bump with no legacy arm; exporter writes identity; replay prefers it with the permanent config-based fallback for old agent traces.

This ordering front-loads the TUI workflow-display and workflow-script rendering fixes (all Stage A), and every irreversible decision sits in Stage B under exactly one stated obligation instead of three dated ones.

## Verified

- Storage sharing: `packages/desktop/src/main/platform/paths.ts:45-52`, `src/platform/defaults/nodeStorage.ts:15-18`, `packages/extension/src/extension.ts:200-218`, `src/common/storage/storageLayout.ts:8-17`.
- Desktop reuses extension frontends/backend: `packages/desktop/src/renderer/main.ts:22-53`, `desktopAgentExecution.ts:256-266`, `hostBridge.ts:39-52`.
- Ephemerality of all wire shapes: `ProgressViewState.ts:89,153-154` (documented not-persisted), `streamState.ts` frontend-only call sites, webview-state audit (only `mainViewState.sessionType` persists, `mainView/state.ts:151-171`).
- Behavior audit: read every predicate cited above in `resumability.ts`, `SessionResumeRetrieval.ts`, `resolveAndResumeStream.ts`, `executeAgent.ts`, `executionListing.ts`, `chatDefaults.ts`, `executionLifecycle.ts:50-63`, `waitCoordination.ts`, `summaryFormat.ts`, `ExecutionHandle.ts`, `AgentLaunchContext.ts:276-293`, `runExecution.ts:115-120`, `replayTrace.ts:145-196`, `subscribeStreamLog.ts:138-142`, plus the model-handler and delegation-gating sites.
- Descriptor rebuildability: `StreamSnapshotStore.ts:232-241` (`descriptorFromConfig`); `ExecutionMeta` carries no agent name: `src/shared/schemas/stream.ts:77-96`.

---

LionSR:
Revision 3 pushed as 06667ec, applying the two follow-up comments directly to the document rather than leaving them as review commentary. What changed, and why each change is in the doc itself:

**A removal inventory is now a first-class section.** The previous revisions listed additions precisely and deletions as prose ("what is deleted outright"). Revision 3 inventories every removal in four tables — shared schemas/types, classification-recovery helpers, CLI reconstructions, silent defaults — each row carrying the element, its `file:line`, its replacement, and the stage that deletes it, with the rule stated that deletions ship in the same PR as their replacement's adoption and leave no re-export shims. A fifth table records **deferred removals** so they are parked rather than lost: the `DELEGATION_AVAILABILITY_CATEGORY` fabricated row (its consumer `annotateDelegationTool` writes the model-facing tool description, so removal is roster-track work), the fabricated per-run `config.json` records, and `ActiveChildInfo.kind`.

**The compatibility-class matrix replaces the dated-reader table.** Protected durable (agent-run rows + exported agent traces, including the `TraceDocumentSchema.meta` embedding), breakable durable (process/script rows, descriptors, `sessionType`), ephemeral (every UI wire shape on both stacks — audited, none persisted). Consequence written into the doc: zero dated readers; `ExecutionMeta.identity` required at write, optional at read, absence resolved from config at the existing listing/resume boundaries permanently for agent rows; descriptor gets **no** legacy arm because `descriptorFromConfig` already rebuilds v1 rows — which also subsumes live defect 1 instead of patching it.

**`identity.category` is gone from the model.** The Responsibilities section now records the audit result as a ruling: category's authority chain is `setting → config`, enforced at `AgentLaunchContext.ts:276-293`, with every behavioral consumer cited; the identity carries the run axis only. The pre-config render window is served by a provisional display hint on the descriptor wrapper, explicitly non-behavioral, with the self-correction path (`streamStateMerge.ts:40-45`) cited and a stated preference for dropping the hint if the window proves it can render without it. The rev-2 `normalizeWriterCategory` sequencing dependency disappears with the field.

**Three responsibility violations are recorded with their fixes staged**: `ExecutionHandle.category` sourced from the descriptor while driving wait/report/attach behavior (re-sourced from config, Stage B); `normalizeWriterCategory` mutating persisted bytes for a CLI UI default (deleted when listing filters on `identity.kind`); `runExecution.ts:115-120` deleting the flow record on category mismatch (pinned by a fixture test).

**Stages reordered ephemeral-first.** Stage A recuts both UI stacks atomically with zero persistence risk and delivers the TUI workflow-display and workflow-script rendering fixes first; Stage B touches the durable store under the single protected obligation; Stage C does descriptor and trace.

Also fixed the one open citation nit (`streamTabInfo.ts:34-38` for the render-window comment) and resolved that thread.

One thing revision 3 deliberately does *not* do: it keeps the workflow-roster failure out, unchanged from revision 2 — the empty-roster mechanism is still not established, and the checkpoint-salt hazard remains recorded on that track.

---

LionSR:
Deeper review, taking the maintainer's chair properly this time. Revision 4 is pushed (c7c4e04) with the verdicts applied; here is the review that produced it. The standard is the old kernel rule pair, and both halves matter: **never break the user's data, and never let compatibility leak past a single boundary into code structure.** Rev 3 honored the first and fumbled the second.

**1. "Readers treat `identity == null` as legacy" was the wrong design, and I wrote it.** That sentence means every reader in the tree grows a branch, forever. That's not compatibility, that's a tumor with a nice name — every future contributor has to know that `identity` might be null and what the folklore fallback is. Bad. The correct shape: the *disk* schema keeps `identity` optional (old rows must parse, non-negotiable), the *in-memory* type has it **required**, and exactly one function at the store boundary — `resolveRunIdentity(meta, config)`, five lines, sitting where the meta/config join already happens (`executionListing.ts:142-158`) — derives it for pre-identity rows. No consumer above the store ever sees an identity-less row. Compatibility is a data problem; solve it at the edge and the rest of the codebase has ONE shape. That directly answers the "compatibility makes code harder to maintain" concern: it only does when you let it shape the code instead of confining it to the boundary.

**2. The "provisional category hint" was cowardice and it's gone.** A field whose documented contract is "don't trust me for anything real" is not a design, it's an apology in the type system. Look at what's actually true: the descriptor is written at birth, from the same launch context that writes `config.json`, and once `normalizeWriterCategory` is dead (PR 4, *before* the descriptor recut in PR 5) that config field has exactly one writer. A copy written once from a single-writer source is **correct by construction** — a plain denormalization, the most boring thing in the world. So the recut descriptor is `{schemaVersion: 2, streamId, executionId, identity, agentCategory?}` with `agentCategory` present iff the run is an agent, display-only by ruling, and the PR ordering is what makes that claim true rather than hopeful. Name the thing what it is or don't have it.

**3. The data structure was hiding in plain sight at the launch seam.** `CreateChildStreamOptions` currently takes `streamCategory`, `runKind`, `agentName`, AND `toolName` — four fields describing one fact, two of them false at two of three call sites, plus a `toolName` "for icon selection" that is identity leaking out a third hole (`childStream.ts:27-34`). Get the data structure right and the code fixes itself: one `run: RunIdentity` field, three call sites passing the truth they already know. Also found while reading: **`run.start` already emits the full descriptor on the trace wire** (`childStream.ts:135`). The CLI's entire two-kind blindness is because `projectCliRunFact` throws that event away. The TUI fix everyone wants is projection wiring, not new event vocabulary. That's the difference between "we need a new protocol" and "read the event you're already sent."

**4. The plan is now six bisectable PRs, each with an explicit delete list and a grep-able acceptance gate** — because "what to delete" is the deliverable, and a deletion without a grep proving it's gone doesn't count:

- **PR 1** — `RunIdentity` exists; launch sites speak it; `CreateChildStreamOptions` loses its four identity fields. Gate: `rg -c streamCategory src/` → 0.
- **PR 2** — Lit stack (extension **and** desktop, same frontends) reads identity from birth via `run.start`; delete `ProgressStreamRunDetails`, the `streamId.split('@')` name-parse, and the `?? Workflow` defaults. Gate: no `isProcessAgent`, no `?? AgentCategory.Workflow` under `src/controllers/`.
- **PR 3** — CLI recut, atomic (both wire ends ship in one binary): `StreamSlice.identity` in, all nine reconstructions out, `SetActiveStreamPayload.agentCategory` out. This PR is the TUI workflow-display and workflow-script fix. Gate: headless parity + zero `isFullLogChildStream`.
- **PR 4** — the durable authority: `ExecutionMeta.identity` written at all six `registerExecution` sites inside the lease; the choke-point resolver; delete `normalizeWriterCategory`, `runtimeCategory`, the `meta.category` string sniff, `resolveExecutionDisplayCategory`, the `handle.toolName` slot; re-source `ExecutionHandle.category` from config (its consumers `waitCoordination`/`summaryFormat`/`attachToolFlow` are *behavior* reading a display projection today — that's the real bug in the tree). Fixtures pin the pre-identity agent row and the `runExecution.ts:115-120` flow-record-deleting path. Fixes live defect 2.
- **PR 5** — descriptor v2 (no legacy arm — v1 rows rebuild via the `descriptorFromConfig` path that already exists) + `TraceDocument.identity`; `agentKind.ts` deleted entirely with its last consumer. Gate: `rg -c isProcessAgent` → 0 in the production tree. Subsumes live defect 1.
- **PR 6** — shrink the knip baseline (never widen), confirm no new `@agent/*` deep-import specifier, close out.

Every PR type-checks (`npm run typecheck` — the builds don't), tests, and is independently revertable.

**What I did not change:** the protected cohort. Old native-agent rows and old exported traces parse forever — that's the "don't break the user's data" half, and it costs five lines at one boundary, which is the correct price. Process and workflow-script history degrade exactly as they read today, and nobody builds `workflow-script#` stream-id sniffing to heal a cohort that's licensed to break.

## Verified
- Read `childStream.ts:1-145` in full (options shape, `buildRunDescriptor` call, `run.start`/`run.config` emits), `streamTabInfo.ts:1-120` in full (the `?? Workflow` default, the name-parse fallback, per-arm construction), `progressEvents.ts:1-60` (`SetActiveStreamPayload.agentCategory`), `cliState.ts:175-205` (`StreamSlice.category` and its doc comment).
- All other `file:line` citations carried over from the two audit passes posted earlier in this thread, each of which was read directly.

---

LionSR:
Revision 5 pushed as 08e92bc. Title changed to match what it now is: **one struct, one home, one wire.** Three inputs drove it: the directive that projection layers and resolver functions are not acceptable, the decision to pull every deferred removal into scope, and a deep audit of the multi-agent orchestration layer — which found the spaghetti is worse than the classification problem that started this PR.

## The audit numbers

**26 distinct representations of the parent–child edge.** Two durable authored (`ExecutionMeta.parentExecutionId` and the `child-{id}` KV records — written non-atomically in the same `Promise.allSettled`), one in-memory authority (the handle), and the rest a mix of derived wire copies (the parent stream id travels on **three** separate webview messages), hand-threaded parameters (the same orchestrator stream id has **three names** in one call chain: `orchestratorStreamId` / `parentStreamId` / `deliveryTargetStreamId`), duplicate predicates (`isChildExecution` open-coded **5×**, caller-ownership authorization **4×**), string-encoded ids parsed back at three sites, and two stream-id formulas whose source comment literally says they "must stay in lockstep." One genuine hazard found: `agentCliSessionStores` keeps its own `parentStreamId`, unsynced with `handle.detach()`, so the external-CLI follow-up authorization can contradict the live handle.

**A dead union arm.** `ActiveChildInfo.kind: 'process'` is never constructed in production — `getActiveChildren` hard-codes `'subagent'` and bash children register as full handles. The CLI roster filter that branches on it filters nothing; the webview icon logic sniffs `toolName === 'codex'` instead. The union, the filter, and the sniffing all go; roster rows carry `identity`.

**A third live defect.** The progress-view Resume button on a workflow-script stream relaunches the *borrowed default agent's config* as a plain workflow run (`ProgressViewHost.ts:62-79` branches only on the fabricated `Workflow` category), and `RESTORE_STATE` pushes a synthetic bash config into the launcher form. Fixed in PR 2 by gating resume/rerun/restore on `identity.kind === 'agent'`.

**The fabricated-row removal is cheaper than deferred-status implied.** `requireVisibleAgent` and `selectAvailableDelegationModel` already take category *literals* — the `DELEGATION_AVAILABILITY_CATEGORY` map's only production effect is description annotation, and `delegate_multi_agents`' description has no anchors, so both blocks are appended as trailing paragraphs today. The row is also *wrong*, not merely fabricated: the script runner is bi-categorical (`ToolUse` iff `schema`, `Workflow` otherwise). One `Workflow` entry cannot describe that tool.

## What revision 5 changes structurally

**No resolver.** The rev-4 `resolveRunIdentity()` choke point is gone. A **one-shot migration** stamps `identity` into every pre-existing agent row (small terminal JSON files, atomic per-file, idempotent, agent name from the adjacent config, dry-run count logged first); after it, the persisted schema **requires** `identity`. One shape on disk and in memory, zero optional fields, zero read-path branches. The migration module is the only legacy artifact and lives in no hot path.

**No projections.** Three copies rev 2–4 defended as "projections with contracts" are deleted from the model: the descriptor's identity+category copy (the stream sidecar keeps a **foreign key** — at rehydration the descriptor read is a disk read and so is reading the authority by executionId, same timing; live launches get identity in-memory from `run.start`, which `childStream.ts:135` already emits), `TraceDocument.identity` (the trace embeds `ExecutionMeta` wholesale — new exports carry identity for free, one confined agent-fallback for old immutable exports), and the per-arm wire re-declarations (`StreamTabInfo.identity: RunIdentity` verbatim — the same struct schema→disk→wire→component; hosts add display fields *beside* it). **`RunDescriptor` dies as a concept** — `buildRunDescriptor`, `PersistedRunDescriptorSchema`, the schema version, and `descriptorFromConfig` all reduce to one FK field.

**Orchestration ruling.** Lineage gets the same discipline: one durable authority, one in-memory authority (sole mutator `detach()`), **one** wire copy (`StreamTabInfo.parentStreamId`; `UPDATE_PARENT_STREAM` and the `activeState` copy are deleted — all ephemeral), the `child-{id}` records kept but labeled as what they are (an index, not a second truth), checkpoint hashes kept by design (the double-launch guard depends on them), and the in-band/async split kept as a durability contract — only the birth preamble (id mint + register + stream-id) is unified.

**The plan is now nine PRs**, 1–5 as before (sharpened: PR 4 gains the migration, PR 5 deletes the descriptor concept), plus PR 6 (orchestration lineage dedup — the full delete table is in the doc), PR 7 (per-tool roster declaration on `ToolDefinition`, a documented forward-compatible `looseObject`; the fabricated map row dies), PR 8 (honest run records — `RunRecord` union; the fabricated `agentCategory`/`model` writes and the `PROCESS_HIDDEN_FIELDS` filter that exists to hide them both go), PR 9 (ratchet close-out). Each with delete lists and grep-able acceptance gates.

Kept out, still: the workflow-roster failure (mechanism unestablished; checkpoint-salt hazard recorded).

## Verified
- Orchestration audit: read every cited site in `executionRegistry.ts`, `ExecutionHandle.ts`, `childRunLoop.ts`, `childStream.ts`, `subagentExecution.ts`, `inBandSubagentExecution.ts`, `nativeSubagentStrategy.ts`, `agentCliShared.ts`, `agentCliSessionStores.ts`, `WorkflowScriptTool.ts`, `workflowScriptAgentRunner.ts`, `runWorkflowScript.ts`, `workflowScriptRun.ts`, `streamApprovalQueue.ts`, `executionInteractionOwnership.ts`, `waitCoordination.ts`, `summaryFormat.ts`, `DelegationTools.ts`, `ExecutionsTool.ts`, plus the CLI/webview roster consumers.
- Feasibility audit: full reader inventory for the four fabricated-config cohorts (listing, history, `/executions`, trace assembler, replay, workspace files, resume, progress backend), both delegation maps and their consumers, and the `ActiveChildInfo` writer/reader census (including the grep showing `kind:'process'` constructed only in test fixtures).

---

LionSR:
Revision 6 pushed as abf6140, answering "are there further surface areas" with three more audits. Yes — four more axes have the same disease, and each now has a ruling, a removal list, and a PR in **Part II (PRs 10–15)**. The unifying principle you named — migrate at the entrance, never in middle layers — turns out to be the precise diagnosis for the worst two findings.

## Axis S — lifecycle state (PRs 10–11)

**19 distinct state vocabularies. 18 sites re-derive status independently of the status machine. 20 middle-layer healing sites.** The centerpiece is the persisted terminal fact being *two* fields — `terminalStatus` (untyped string) + `outcome` — kept consistent by a `superRefine` and healed by a `.transform` that runs on **every read and every write** (`stream.ts:98-119` via `ExecutionKVStore.ts:290`). That is a write-side problem being fixed in a middle layer for eternity. Worst individual sites: `WorkflowScriptTool.ts:505` decides a script call's success by comparing the raw string; `AgentRunLifecycle.ts:183-189` reads the phase back out of the machine and silently **overrides the caller's outcome**; the two hosts have parallel history-status projections reading *different* source fields into *different* vocabularies; and the TUI detects child errors by lowercasing arbitrary strings and regexing for exit codes (`childExecutionStatus.ts:6-15`). Ruling: `StreamPhase` (machine-owned) for live, `RunOutcome` (meta-owned) for terminal, one-shot stamp of `outcome` on old rows, `terminalStatus` never written or read again, `ExecutionStatus` surviving only as the trace-export projection.

## Axis T — opaque stream ids + retiring the legacy resolver (PR 12)

Two discoveries. First: **the `@model` segment of `agent@model#executionId` has zero production consumers** — no parser anywhere reads it, uniqueness comes entirely from the `#executionId` suffix — yet `AgentLaunchContext.ts:564` hard-fails launches without a model *purely to mint the id*, and the fabricated display models exist substantially to feed the dead segment. Second: **`legacyExecutionIdentity.ts` (374 lines) is the largest permanent middle-layer in the tree and it's on hot paths** — every completed `/executions/{id}` summary and every trace assembly pays a bounded O(all-streams) sidecar scan unless the row carries registration provenance, and its backfill writes a cache that its own contract (`registeredStreamId`) forbids from ever becoming authoritative. Textbook entrance-migration target: the migration walker runs the resolver **once** per unresolved row, stamps unique matches as authoritative, and the module is deleted — same results, computed once instead of per read. Also folded in: `parseLegacyTaskState`/`TaskState.ts`, the sidecar `description` mirrors with their per-stream meta read on every load (#9627), and the `taskRuns` directory probe. The one legacy grammar that stays is the filename-era workflow-output parser — it reads **user files on the user's disk**, which no migration of our store can rewrite.

## Axis U — one accumulator per fact (PR 13)

Round artifacts are accumulated in **four** places (store sidecar, the CLI's live+durable merge with generation counters, the CLI's *second* independent live spread, the extension frontend's copy), usage deltas in **five** (including per-render summing at two webview components and the status bar's private map). And `missingOutputs` exists in two representations that **disagree**: three `XmlOutputManager` paths write the transcript row without emitting the run fact, so the sidecar and the transcript diverge. Plus one gem: `UserMessage.ts:217` re-parses structured data *out of rendered message text*. Ruling: facts carry deltas, the `StreamSnapshotStore` is the one accumulator, consumers read through it; one `reportMissingOutputs` helper emits row+fact together; and the workspace-files conversation-scrape — which one history reader uses and the other doesn't — moves to **finalize-time**, persisted once, so the read-time scraper dies.

## Axis R — the roster pair (PR 14)

The count is now exact: **39 `workflow*`/`toolUse*` field-pair declarations across four packages, and exactly one of them is durable** (`AGENT_ROSTER_SELECTION`). Every other pair is a re-shaping of that one value for a different consumer. Ruling: `ByCategory<T> = Record<AgentCategory, T>` everywhere, memento migrated at its entrance, prompt variables keep their names as the external contract. All 38 non-durable pairs and their branch-per-category sites go.

## Program shape

Fifteen PRs total: Part I (1–9) is run identity, orchestration lineage, delegation gating, honest run records. Part II (10–15) is terminal state, projections, opaque ids + legacy retirement, single-accumulator derived data, the roster collapse, and close-out. Two entrance migrations in total (executions store; roster memento), both idempotent, one-shot, quarantined; the only permanent legacy readers left standing are the trace-viewer's (immutable exported artifacts) and the user-disk filename grammar (files we don't own). Everything else that says "legacy" in a middle layer is scheduled deletion with a grep gate.

## Verified
- Part II claims sourced from three fresh audits: the vocabulary/conversion/writer census across `stream.ts`, `streamStatus.ts`, `StreamStatusService.ts`, `resumability.ts`, `restartRepair.ts`, `executionLifecycle.ts`, both history-status projections, and the CLI status mirrors; the full parser census of the stream-id format (every `split('@')`, `lastIndexOf('#')`, prefix regex in the tree) plus `legacyExecutionIdentity.ts:1-374` and every `.transform`/`.catch`/`.prefault` on persisted schemas; and the writer/merger/reader chains for round artifacts, usage, todos/plan, workspace files, and all 39 roster pair sites with file:line.

---

LionSR:
Revision 7 pushed as cdeaea3: **single-shot implementation, no staged merges.**

The staging wasn't just overhead — it was actively working against the proposal's own goal, and the doc now says so. Staged delivery is a compatibility generator: every intermediate PR that must stay green forces bridge code that exists for no other reason. Concretely, the fifteen-PR plan contained three pieces of pure staging scaffolding, all now deleted from the plan:

- **Step 1's `buildRunDescriptor` bridge** — deriving v1 persisted fields from the new identity solely so the on-disk shape could stay untouched until a later merge. In one shot, the descriptor never exists in a transitional form; it just dies.
- **The writer-ordering arguments** — "delete `normalizeWriterCategory` before the descriptor recut so the copy is correct by construction" was reasoning about which intermediate tree states are safe. When everything lands at once there are no intermediate tree states to reason about.
- **Three migration generations** — the staged plan walked the executions store in PR 4 (identity), again in PR 10 (outcome), again in PR 12 (streamId resolution). Now it is **one walker, one pass**: `identity`, `outcome`, and the resolved `streamId` stamped together, idempotent, atomic per file, dry-run count logged first.

So the final accounting gets cleaner: the tree goes from the current state directly to the final state, and the only compatibility artifacts that *ever exist* — at any point in history, not just at the end — are the two entrance migrations (executions store, roster memento) plus the two permanent readers justified by data we don't own (exported traces, user-disk filenames).

The fifteen former PRs remain in the doc as the **build-and-review checklist inside the single change** — implementation order and review grouping, nothing more. The whole change ships when the full acceptance battery passes: `npm run typecheck` **and** the trace-viewer tsc (which the root typecheck doesn't cover), `npm test`, `check:dead-code-ratchet` with shrunk baselines, every per-step grep gate (zero hits for `isProcessAgent`, `streamCategory`, `normalizeWriterCategory`, `runtimeCategory`, `terminalStatus` outside the export projection and migration, `legacyExecutionIdentity`, `DELEGATION_AVAILABILITY_CATEGORY`, the 39 roster pairs, …), and headless output parity per the `texra-cli` skill.

One cost recorded honestly in Risks rather than argued: a single change of this size loses per-stage bisectability and concentrates the review surface. The mitigations are structural — the removal inventory doubles as the review checklist, every deletion has a grep gate that proves it rather than asserts it, and the checklist ordering keeps the review groupable even though the merge is atomic.

---

LionSR:
Revision 8 pushed as 2da0ab1 — the red-team pass. Three independent adversarial subagents attacked revision 7 (migration & persistence; deletion claims; runtime semantics). They refuted **five load-bearing claims**, found **three landing blockers**, and forced one genuine design retreat. Everything is folded into the doc; the highlights:

## The two blockers, found independently by all three reviewers

**1. "Required `identity`, zero optional fields" was impossible — twice over.** The shared `ExecutionMetaSchema` is transitively the *export* schema (`TraceDocumentSchema.meta` embeds it, and `parseTraceData` throws on any mismatch) — so a required field bricks **every pre-migration exported trace** at the parse boundary, before the planned "one line of fallback" can ever run. Revision 2 knew this hazard; revisions 5–7 reintroduced it. And separately: the schema is `z.object`, so a not-yet-updated binary's read-modify-write (`enqueueMetaUpdate` → `writeMeta`) **silently strips** the stamped field from disk — after which the row vanishes from listings, becomes unresumable, *and undeletable* (deletion admission treats the ZodError as corruption). The redesign: **required at the write boundary only** (`RegisteredExecutionMetaSchema`, used by `registerExecution` and the finalize writers); the shared read schema keeps `identity`/`outcome` optional forever. One deliberate seam instead of an absolutism the ecosystem cannot support.

**2. The store-generation marker guaranteed permanent damage.** Mixed-version hosts on the shared `~/.texra` (three independent release channels), the extension's own legacy-bucket merge injecting v1 rows *after* the mark, execution-lease fences that **throw** on live or freshly-crashed rows, and restored backups — each one leaves rows the marker swears are migrated but aren't, forever. The marker is abandoned: **stamping is idempotent at the store entrance, every time** — skip rows that carry `identity`, re-stamp on absence, lease-aware, inverted to one O(N+M) walk. Healing becomes a property of the store rather than an event that old binaries can undo.

## The refuted claims, corrected inline

- **`@model` is a reproduction contract, not a display string.** The CLI resume paths *re-derive existing stream ids* from `config.agent + config.model` — the "zero consumers" audit looked for parsers and missed re-derivers. FK-first resolution (`registeredStreamId(meta) ?? getStreamTabId(…)`) lands before the mint format changes, with a pre-migration resume fixture.
- **The phase-label join is scoped, not deleted.** Planned/cached/skipped workflow calls have no `childStreamId` (`workflowCallProgress.ts:31-36` says so explicitly) — the phase join is what makes the `1/5` progress headers correct. `childStreamId` joins row→task-card; `workflowPhase` keeps the phase-group headers.
- **Provisional kinds are forbidden.** The `StreamState` kind-flip reset preserves exactly one field and drops the user's typed-but-unsent follow-up text, todos, plan, and accumulated usage. Identity must be known before the first `createStreamState`; the CLI populates `StreamSlice.identity` on cold read from the store (a rehydrated terminal child never re-emits `run.start` — the regex deletion was unsound without this).
- **`CUSTOM_AGENT_PRESETS` is a second durable roster pair** — user-authored teams would have been silently dropped. It gets its own migration, and the roster selection moves to a **versioned key** with `safeParse`, because the shared `state.json` is read by desktop *and* CLI at possibly different versions.
- **The `runExecution` flow-record guard was misdiagnosed** — it checks the user's *command* demand against the resolved run, not classification drift. Corrected and moved out of scope.

## The runtime regressions caught before they shipped

- **`settleRun` would have destroyed stopped runs' resume state.** The current phase-override code implements user-stop-wins; a naive `settleRun(callerOutcome)` writes `COMPLETED` over a `CANCELLED` phase — and the fail-closed rule then **deletes the flow record** of a run the user merely stopped. The resolution now happens inside the machine, with a fixture pinning stop → report-completed → everything reads CANCELLED and the flow record survives.
- **The birth-preamble unification would have poisoned the in-band attempt ledger** (keys-without-marker is a hard reconciliation error) and could have broken the workflow-script double-launch guard (an `instanceof ExecutionLeaseActiveError` check that the agent-CLI path's bare `catch` would have swallowed). The preamble is now a post-reservation callback that rethrows the lease error untouched.
- **NDJSON is a declared frozen public contract** (`cliNdjsonProgressEvents.ts:26-34`; the texra-cli skill says byte-identical) — not "ephemeral" as the class table claimed. Two planned deletions would have changed it and one path would have *thrown* (`agentConfigToTaskState` has a throwing default arm, on the event path). All projected at the boundary now; the parity test lands before the deletions.
- Plus: `handle.category` sourced from launch-time in-memory config (three missed consumers found, including the required `ResultEvent.category`); codex/claude stay on the agent `RunRecord` arm with resume gated on `identity.tool === undefined` (otherwise defect 3 changes shape instead of dying); `readConfig()` deleted outright with all ~19 sites moved to `readRunRecord()` so the typechecker enumerates them.

**What survived unrefuted:** the dead `ActiveChildInfo` `process` arm, and the descriptor early-render timing argument — both held up under deliberate attack.

The suffix helper (`legacyExecutionIdFromStreamSuffix`, ~30 of the 374 lines) survives as the deletion-admission boundary — deleting it would orphan pre-descriptor executions forever — and `streamIdSource` survives as stamp provenance so a wrong stamp stays demotable. The acceptance gates are rewritten to match reality rather than aspiration.

---

LionSR:
Revision 9 pushed — **Part III: one language across hosts.** Two more audits mapped the last dual system, the CLI versus the shared extension+desktop stack, and the evidence is conclusive:

**The pipeline duality.** Four subscribers attach to the *same* `SessionEventHub` with the same dispatch idiom — `ProgressBackend`, the TUI's `subscribeRuntimeHost`, the headless `runProgressRenderer`, and the NDJSON projector — each with its own handler table, and the round/phase stage `index`/`total` guard logic is **cloned byte-for-byte in all four**. The two full state engines maintain ~20 field-for-field paired facts under different names and shapes (`outputFilesByRound` ↔ `files`; one discriminated `stage` slot ↔ two separate fields; nested `bypass` ↔ three flat booleans; category stored in *three* places on the Lit side). The child-roster tombstone cap is a value-copy whose own comment admits it mirrors the other engine's constant "as a value, not an import."

**The asymmetry that proves the thesis:** the display *formatters* (`streamStatusDisplay`, `formatWorkflowCallLine`, `taskGroupProjection`) are already consumed by all three hosts — but every *derivation* module (`childActivityReducer`, `buildStreamTabInfo`, `streamMetadata`, `streamOrdering`) is shared only inside the Lit stack. The CLI re-derives status mirrors, roster reconciliation, parent topology, tombstoning, artifact merges, phase grouping, transcript role classification, and a `subagentExecutionLabels` map that exists under the same name with independent implementations on both sides.

**The command layer diverges behaviorally, eleven ways** — including a second CLI resume entry (`resolveCliResume`) that bypasses `resolveAndResumeStream` with its own five-way outcome vocabulary; workflow resume supported by ext/desktop but rejected by the CLI; two history-status projections over different source fields; the workspace-file fallback on one side only; and **`AgentRosterForm` bypassing `AgentRosterController.allPresets()`** — custom presets visible in settings but invisible in the TUI, which is a live lead for the separated workflow-roster bug. Silent UX divergence also exists at the fact level: `followUpSent` (Lit no-op, TUI refresh), `goalPaused` (Lit no-op, TUI transcript row), `goalStateChanged`/`inquiryThreadUpdated` (Lit handled, TUI ignored).

**The ruling (steps 16–18):** one host-neutral **session view-model** — `ProgressViewState` + `ProgressFactApplier` generalized into `src/controllers/session/` (they already comply with the no-`vscode` rule) — is the single hub consumer and single owner of derived session state. Hosts are renderers over it: the `WebviewUpdater` projects it over postMessage as today, the Ink TUI reads it through a signals adapter, and the headless renderer plus the frozen NDJSON vocabulary are boundary projections of the same store. One command layer: one history service with host ports for surface verbs only, one resume funnel (the CLI gains workflow resume for free by deleting its bypass), unified launcher validation, the TUI roster form on the controller, the CLI on `selectAutoOpenFinalOutput`. The fact-coverage deltas get one decided behavior each, same on every host. Delete lists and grep gates per step; NDJSON byte-parity guards the whole move.

That completes the program: Part I (one struct, one home, one wire), Part II (one terminal fact, opaque ids, one accumulator, `ByCategory`), Part III (one view-model, one subscription, one command layer) — all red-teamed, all landing as the single change, with hosts reduced to what is genuinely theirs: rendering and input.

---

LionSR:
Revision 10 pushed — the holistic build order, plus a consolidated open-problems register.

**The ordering was checked adversarially against one criterion: is anything built that a later phase deletes?** The step numbering (1–18) failed it — it reflects *discovery* order, and followed literally it would have manufactured the exact waste this plan forbids: steps 2–3 recut the Lit wire shapes and build `StreamSlice.identity`, both of which step 16 then deletes when the unified view-model replaces `StreamSlice` wholesale. Intermediate layers by another name.

**The fix inverts Part III into the middle of the build.** Eight dependency-ordered phases (`0 → 1 → 2 → {3 ∥ 4} → 5 → 6 → 7`), nothing built twice, nothing built to be deleted:

- **Phase 0** — schemas + the idempotent stampers, pure additions, every red-team fixture running against old-format buckets before anything consumes the new shapes.
- **Phase 1** — runtime authorities: launch sites speak identity, `settleRun`, outcome-only writes, `normalizeWriterCategory` dies.
- **Phase 2** — storage reads: `readRunRecord()` everywhere, **resume goes FK-first before the id format changes** (the red-team's hard precondition), the forward resolver dies, the suffix helper survives as deletion admission.
- **Phase 3 ∥ 4** — orchestration lineage dedup, and the **session view-model built before any UI is touched**, with the final canonical shapes, so both stacks land directly on it.
- **Phase 5** — hosts become renderers: Lit projects the view-model (desktop free), the TUI's parallel engine is deleted in one motion, NDJSON byte-parity asserted, the fact-coverage deltas get their decided behaviors.
- **Phase 6** — one command layer (history service, one resume funnel — `resolveCliResume` deleted, CLI gains workflow resume — roster form on the controller).
- **Phase 7** — gating, honest records, the `ByCategory` recut, and the full gate battery.

The steps survive as per-area Add/Delete/Test/Gate specifications the phases reference; review proceeds phase-by-phase inside the single change.

**The open-problems register** pins the eight genuinely undecided items so they're decided deliberately, not discovered mid-implementation: the wait-predicate target (`category`-from-launch vs an explicit `deliversResultPerTurn` flag — decide at phase 1 review); the NDJSON non-agent `setTaskState` shape (today's default arm *throws*; fixed by the parity suite); the four fact-coverage delta decisions (`followUpSent`, `goalPaused`, `goalStateChanged`, `inquiryThreadUpdated` — one behavior each, all hosts); **old-host display degradation accepted, dual-write rejected** as a compatibility layer (cosmetic, bounded by update lag, heals on update); the `tool`-less migrated codex cohort (tolerated, unhealable); the roster-bug lead (the `AgentRosterForm` bypass — phase 6 fixes the bypass, the deterministic repro still owns the verdict); disk-scale numbers (the stamper's dry-run count *is* the measurement); and the confirmed non-bump of `EXECUTION_META_SCHEMA_VERSION` (a bump would itself warn-and-null every old row).

The doc is now complete as a program: three parts, red-teamed twice, one change, eight phases, two entrance stampers as the only compatibility artifacts that ever exist, and an explicit register of every remaining decision.

---

PR #9705, opened 2026-08-04
refactor: consolidate run classification onto one RunIdentity authority

LionSR:
## Summary

Implements docs/proposals/2026-08-03-run-classification-consolidation.md — Part I (steps 1–9), Axis S (steps 10–11), Axis R (step 14), step 8 (honest RunRecord), the rest of Axis T (step 12), Axis U (step 13), **and Part III step 17 (one command layer)**: **one struct, one durable home, one wire** for "what kind of thing is running", one terminal fact, one stream-id mint, one accumulator per fact, one resume funnel.

- **One struct** — `RunIdentitySchema` (`{kind:'agent',agent,tool?}` | `{kind:'process',tool}` | `{kind:'multiAgentWorkflow',workflowName}`). The six classification vocabularies, five name-parsers, and nine CLI reconstructions are deleted; every consumer keys on `identity.kind`.
- **One home** — `ExecutionMeta.identity`, required at the write boundary, optional on the shared read schema. Legacy rows heal by idempotent entrance stamping in `listExecutions` (identity only, from quarantined stream-id-prefix evidence plus the persisted config); cold hydrate derives from the same rule, using the sidecar's own id as evidence when the execution row predates registration-time streamId stamping (the association is proven by the registration-written `meta.executionId` edge).
- **One wire** — `run.start` carries `{streamId, executionId, identity}`; the struct travels verbatim on `StreamTabInfo`, roster rows, and CLI `StreamSlice`. The `RunDescriptor` concept is deleted; the stream sidecar keeps a foreign key.
- **Honest run records (step 8)** — `executions/{id}/config.json` is the `RunRecord` union: a real `AgentConfig` for agent runs, `{name, instruction, workingDirectory?, model?}` for bash and workflow-script containers. No fabricated `agentCategory`, default model, or tool-config on non-agent runs; `readRunRecord()` is the store seam with a write-side parse gate; `PROCESS_HIDDEN_FIELDS` deleted; the trace document embeds the union. codex/claude stay on the agent arm (they stamp `kind:'agent'`).
- **Opaque stream ids + resolver retired (Axis T)** — ids mint as `${name}#${executionId}` (no model segment; the model-required launch guard died). `legacyExecutionIdentity.ts` (378 lines) is deleted whole with **no replacement walker**: `meta.streamId` stamped at registration is the one execution→stream mapping, suffix resemblance is never ownership, and deletion admission / restart repair / trace assembly / `/output` reads use only the persisted FKs. `streamIdSource`/`registeredStreamId` and the trace-export ambiguity status died with it. CLI resume reads the stamped FK; a row without one is not resumable.
- **One accumulator per fact (Axis U)** — the `StreamSnapshotStore` is the only accumulator for round artifacts and per-run usage. The CLI's live+durable merge layer, revision counters, second live spread, and dual usage sum are deleted (TUI projects store state, exactly like the extension's fact applier); TUI taskGroups go incremental; `reportMissingOutputs` emits the transcript row and the run fact together at every producer; the edited-files conversation scraper is deleted in favor of the persisted list; `StatusBarUsageTracker` and both webview stream components project from the store; workflow delivery summaries travel typed beside row text (render-time re-parse deleted); `reasoningTokens` joined the one accumulated usage vocabulary.
- **One command layer (Part III step 17)** — `texra resume` funnels through the shared `resolveAndResumeStream` orchestrator: the tool-use arm reopens the chat TUI as before, and the workflow arm gains **headless workflow resume** under the persisted execution id, honoring the run's original `--output` via `resumeWorkflowOutputFile` (previously caller-less). The CLI-only five-way resume vocabulary (`sessionResume.ts`) is deleted; chat resume resolves inline over the shared retrieval. The workflow execute+outputs+emit block is one shared function (`executeCliWorkflowConfig`, used by run and resume). The duplicated user-visibility re-filter (`userStartedCliHistoryEntries`) is deleted — the rule's one home is `isUserVisibleExecution` at the listing; menu builders trust their input. `AgentRosterForm` consumes `AgentRosterController.allPresets()` instead of composing its own preset list (`resolveCliHistoryStatus` was already unified onto shared `resolveHistoryRunStatus` in an earlier phase).

Fixes the proposal's three live defects, including progress-view Resume/Rerun/Restore relaunching borrowed identity (now gated on `identity.kind === 'agent' && tool === undefined`).

## Deliberate clean breaks (no legacy readers)

- Legacy `terminalStatus` bytes are never converted; pre-consolidation rows have no recorded terminal state.
- Roster selection: one unversioned key, canonical record shape only; the v1 pair vocabulary and `.v2` key are deleted.
- Pre-FK sidecar `runDescriptor` is never lifted.
- Pre-registration executions have no persisted stream (no sidecar scans, no suffix derivation, no ambiguity candidates); old-format stream ids remain opaque strings that nothing parses.
- Old-format workflow/tool-use resume re-derivation is gone; resume is FK-first only.

## Reasoned deviations from step 17

- `selectAutoOpenFinalOutput` is a GUI editor-opening policy; the CLI has no auto-open action — its `--output` copy and stdout result line keep their explicit-flag/stdout semantics rather than adopting a setting-gated open policy.
- The chat-defaults history tier already reads the shared listing primitives (`listExecutions` + `isUserVisibleExecution`); its extra filters are tier-specific projection, not a duplicated rule.

## Acceptance gates

`streamCategory`, `isProcessAgent`, `buildRunDescriptor`, `?? AgentCategory.Workflow`, `normalizeWriterCategory|runtimeCategory|resolveExecutionDisplayCategory`, `DELEGATION_AVAILABILITY_CATEGORY`, `UPDATE_PARENT_STREAM`, `runDescriptor`, `legacyExecutionIdentity|streamIdSource|LEGACY_RESOLUTION`, `mergeArtifactSnapshot|streamArtifactRevision`, `listExecutionEditedFiles`, `resolveCliResume|resolveCliHistoryStatus|userStartedCliHistoryEntries` → 0 production hits. `terminalStatus` confined to the frozen trace-export projection; `workflowAgentKeys|toolUseAgentKeys` confined to the config-entrance migration for persisted team-run rows. Frozen NDJSON vocabularies unchanged (boundary projections asserted by tests).

Still proposed (not in this PR): the `taskRuns` legacy directory probe (#6981, its own dated retention policy), and Part III steps 16/18 (the session view-model generalization and TUI-as-renderer recut — a PR-sized change of its own). The proposal's status header says exactly this.

## Single source of truth

- Run kind: `ExecutionMeta.identity` (durable) / `AgentExecutionHandle.run` (live) / the `run.start` fact (wire).
- Run record: `RunRecord` union in `executions/{id}/config.json`, written only by `registerExecution` through the parse gate.
- Terminal state: `ExecutionMeta.outcome`, written only by `finalizeExecution`.
- Execution→stream: `meta.streamId`, stamped once at registration. Stream→execution: the sidecar FK.
- Artifacts/usage: the `StreamSnapshotStore` accumulators; every consumer projects.
- Roster shape: `ByCategory` with `AGENT_ROSTER_SELECTION` as the one durable key.
- Resume: `resolveAndResumeStream` on every host; the CLI's retrieval feed is `readCliToolUseResumeData` (FK-first) over the shared `retrieveSessionResumeData`.

## Validation

- `npm run typecheck` (all 7 configs) — clean
- `npm run lint`, `npm run check:dead-code-ratchet` (18=18, several baselines shrunk), `npm run format`, CLI TUI smoke scenarios (incl. the recut orchestrate-history pair) — clean
- Fixtures: legacy-row identity healing, honest non-agent records listing without fabricated fields, FK-only deletion admission / restart repair / resume, headless workflow resume via the shared funnel, store-projected TUI artifacts/usage incl. reasoning tokens, pre-migration trace parse/replay

https://claude.ai/code/session_013JK8BzoyKHjFg2jEam9KXv

---

> [!NOTE]
> **High Risk**
> Cross-cutting changes to execution identity, resume routing, and persisted metadata affect CLI, desktop, and extension together; mistakes could mis-resume workflows, drop artifacts, or gate resume incorrectly despite extensive test/ratchet updates.
> 
> **Overview**
> This PR finishes a large slice of the run-classification consolidation: **one `RunIdentity` struct** on durable metadata, `run.start`, and UI state replaces descriptor/name parsing; child rows, workflow dashboards, and resume gating key off `identity.kind` instead of category, tool names, or stream-id regexes.
> 
> **`texra resume`** now goes through shared **`resolveAndResumeStream`**: tool-use sessions still open the chat TUI (FK-stamped `meta.streamId`, not re-derived ids), while **workflow runs can resume headless** under the persisted execution id via shared **`executeCliWorkflowConfig`** (including stored `--output`). The CLI-only **`sessionResume.ts`** five-way resolver is removed; chat resume resolves inline with **`readCliToolUseResumeData`**.
> 
> The **CLI TUI** stops maintaining a second artifact/usage accumulator: **`attachTuiRunFactSubscription`** and focus hydration **re-read `StreamSnapshotStore`**, `run.start` seeds **`StreamSlice.identity`**, and task groups project incrementally. **History/listing** uses shared **`resolveHistoryRunStatus`** and listing entry shapes; frozen **NDJSON** still projects legacy child-row and status vocabularies at the boundary.
> 
> **Agent roster** durable state moves to **category-keyed `agentKeys` / `ByCategory`** (harness, config CLI, team plans). Extension/desktop **status bar usage** and progress **resume/rerun/restore** use **`getRunIdentity`** so non-native-agent streams cannot relaunch borrowed configs. Ratchets, docs, and TUI smoke scenarios (`orchestrate-history`) are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1609f5301993b36a48649abd730bb2df3b3f9b45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Re Bugbot's "Legacy outcome lost on stamp" (fd989ae): intentional, not a regression — the legacy process/terminal-tab cohort was explicitly declared breakable, and `fd989ae` deleted the `meta.category` residue that was the only durable evidence for it. Classifying legacy bash rows by config name would re-introduce exactly the `isProcessAgent` name-sniffing this PR deletes.

Scope of the accepted breakage:

- **New rows are unaffected** — identity is stamped at registration; defect 3's gate holds by construction.
- **Legacy workflow-script rows never had persisted evidence** (they were registered as their borrowed agent config, category-less), so their misclassification predates this commit and was never fixable — same as the migrated codex/claude cohort in the proposal's open problem 5.
- **Legacy bash rows** now heal as `{kind:'agent', agent:'bash'}`. Practical blast radius of the re-opened gate: Resume goes through the tool-use resume port, which finds no session flow record for a bash row and reports nothing to resume; Rerun fails at agent resolution ('bash' is not a registered agent). An error toast on a years-old terminal tab, not a silent wrong relaunch.

---

src/shared/schemas/agentRoster.ts
```diff
@@ -23,21 +25,49 @@ export const AgentRosterSelectionSchema = z.discriminatedUnion('kind', [
...
 });
 
 /** Exact delegation catalog attached to a run, independent of durable UI state. */
-export const AgentDelegationScopeSchema = z.strictObject({
-  workflowAgentKeys: AgentKeyListSchema,
-  toolUseAgentKeys: AgentKeyListSchema,
-});
+export const AgentDelegationScopeSchema = z.record(
+  AgentCategorySchema,
+  AgentKeyListSchema,
```
LionSR, line 44: **Blocker** — this recut (`z.strictObject({workflowAgentKeys, toolUseAgentKeys})` → `z.record(AgentCategorySchema, ...)`) has no entrance migration, but the pair shape was persisted into team-run `config.json` since #8403 (2026-07-14, inside the 3-month compatibility window). Those configs now fail `AgentConfigSchema.safeParse` → `readConfig()` returns null → rows list as `incomplete`, vanish from history, and can't be resumed. The roster selection got a `.v2` key and presets got a legacy union; this third consumer of the same retired pair needs the same legacy-member `z.union`+`transform` treatment, plus a fixture reading a real pre-PR team-run config.

LionSR (commented):
Deep review of the full diff (238 files) against base `5fc8caea`, cross-checked against the proposal doc and the base commit's actual write paths. Every load-bearing claim below was independently verified against PR-head sources; grep gates re-run.

## Blockers

**B1 — Persisted team-run `config.json` files break: `AgentDelegationScopeSchema` recut with no entrance migration.**
`src/shared/schemas/agentRoster.ts:68` changed the scope from `z.strictObject({workflowAgentKeys, toolUseAgentKeys})` to `z.record(AgentCategorySchema, AgentKeyListSchema)`. The pair shape was persisted into team-run `config.json` via `teamExecutionFields` → `registerExecution` → `writeConfig` since #8403 (2026-07-14 — three weeks into the repo's three-month compatibility window). Those configs now fail `AgentConfigSchema.safeParse` → `readConfig()` warns and returns null → `listExecutions` marks the rows `incomplete`: old team runs vanish from history, can't be resumed, and `/executions/{id}` detail degrades. This PR applied the mandated `z.union`+`transform` entrance migration to the two *other* durable values carrying the same retired pair (roster selection `.v2` key, custom presets) but missed this third consumer — the grep gate "`workflowAgentKeys|toolUseAgentKeys` confined to the v1 entrance migration" is precisely why it slipped through. Fix is the same legacy-member union on `AgentDelegationScopeSchema`, plus a fixture that reads a real pre-PR team-run `config.json`.

**B2 — The `npm test` validation claim does not hold at HEAD.**
`src/test-kernel/transcript/StreamSnapshotStore.vitest.ts:901` ("preserves legacy meta taskState and description when unrelated meta patches are written") and `:929` ("falls back to legacy taskState when an execution config is unreadable") still assert the retired shims: `StreamTabMetaSchema` strips both fields and `writeMeta` (`StreamSnapshotStore.ts:1466-1481`) rewrites the file with only `schemaVersion`/`executionId`/`parentStreamId`, so both must fail. The head commit's own message says "Known: src/test-kernel/transcript suites still assert the retired machinery; updated in the follow-up commit" — but no follow-up commit exists in this PR. The "8448/8450, only 2 pre-existing failures" claim is therefore not true of the head commit. Land the test recut before merge.

**B3 — The frozen NDJSON public contract was recut, twice, despite the "byte-identical" claim.**
(a) `updateActiveSubagents` is declared in `cliNdjsonProgressEvents.ts:27-34` as a payload "retained only for CLI NDJSON public-output compatibility", yet its row shape was recut at the schema level (`ActiveChildInfo`, `streamState.ts:25-61`: `kind` discriminant gone, `toolName` gone, `identity` added, `childStreamId` now required where `process` rows had none) and the projection spreads `event.items` verbatim (`sessionProgressSubscription.ts:153-159`). The only test asserts pass-through of the *new* shape with `expect.objectContaining` — zero bytes of the old shape pinned.
(b) `history-entry`/`history-detail` status vocabulary changed from raw `terminalStatus` (`interrupted`/`error`) to `RunOutcome` (`cancelled`/`failed`) at `history.ts:214,500`, while `docs/guide/texra-cli.md:283` declares the NDJSON form "stable for scripts". The smoking gun: the proposal-mandated boundary projection exists and is imported — `runOutcomeToExecutionStatus` at `history.ts:29` — but never called. Either project the old vocabularies at the boundary (proposal gate G), or accept the new ones deliberately: update the doc, drop the "byte-identical" claim, and add real byte-parity fixtures.

## Majors

**M1 — The entrance stamper durably misclassifies the legacy workflowScript and external-CLI cohorts as native agents — defect 3 is not fixed for any pre-migration row.**
`deriveLegacyIdentity` (`executionListing.ts:122-130`) special-cases only `meta.category === 'process'`. Verified against the base: legacy workflowScript rows registered with **no** category and `config.agent` = the borrowed default agent (base `WorkflowScriptTool.ts:354,371`), and legacy codex/claude rows also registered with no category (base `agentCliShared.ts:197`). Both stamp `{kind:'agent', agent, tool: undefined}` — which passes `isNativeAgentRun` (`ProgressViewHost.ts:59-61`) and the `RESTORE_STATE` gate (`ProgressViewCommandHandlers.ts:460-461`), so Resume/Rerun/Restore relaunches the borrowed/synthetic config: the exact defect the PR body lists as fixed. Because stamping is idempotent with no generation marker, the wrong identity is persisted forever. The evidence to distinguish these cohorts exists and is unused: `meta.streamId` prefixes (`workflow-script#`, `codex@`, `claude@`) and the legacy sidecar `runDescriptor.kind` lifted at the same parse boundary. Compounding it, `ExecutionListing.vitest.ts:131-159` enshrines a no-category `config('bash')` row — a fixture production never wrote, since bash always wrote `category:'process'` — becoming a user-visible native agent. Either teach the stamper the prefix/descriptor evidence (with fixtures for both real cohorts), or deliberately leave those rows unstamped (`incomplete`) rather than durably persisting a wrong identity — and if the stamping rule ever changes, add a stamp-version field.

**M2 — Main-view persisted state changed shape with no migration; upgrading users silently lose agent selection and unsent instruction drafts.**
`src/shared/schemas/mainView/state.ts:159-172` replaces durable flat `workflowAgent`/`toolUseAgent`/`workflowInstruction`/`toolUseInstruction` with `agent`/`instruction` records that `.prefault({})`. An old blob parses cleanly (unknown keys stripped, prefaults fire), resetting to `'correct'`/`'orchestrator'` + empty instructions, and `applyState` then persists the reset back. This is the *third* durable value carrying the retired pair — the roster selection got a versioned `.v2` key and custom presets got a legacy union, but this one got neither, and no test pins old-blob behavior (`MainAppPersistenceRestore.vitest.mts` seeds were rewritten to the new shape). Launching the wrong agent next session costs a real paid run. Add a legacy union member or dated preprocess reader, mirroring the pattern this PR already established.

**M3 — Sidecar `description`/`taskState` retired ~3 months before the dated retirement gate, and `writeMeta` now destroys them on any rewrite.**
The base schema documented `description` as "scheduled for deletion in #9590 Stage 7 / #9627 with the ≥ 2026-11-01 legacy-retirement cohort"; `legacyExecutionIdentity.ts:21-22` (still in the tree) repeats the gate and conditions it on accumulated usage evidence. Today is 2026-08-04; no evidence is cited. `writeMeta` (`StreamSnapshotStore.ts:1466-1481`) now rewrites the file with only three fields, so any unrelated meta patch deletes on-disk `description`/`taskState`, and `getDescription` lost the legacy-mirror fallback — pre-registration and desktop-imported streams permanently lose descriptions. Neither retirement appears in the PR body. Either restore the parse + round-trip until the cohort date, or update #9627, `TaskState.ts:5-7`, and the contradicting comments (`StreamSnapshotStore.ts:1468-1472` still claims it "cannot destroy legacy display data before its Stage 7 retirement"), and call the early retirement out explicitly.

**M4 — Resume/Rerun/Restore still render as enabled, clickable buttons on non-native streams; the gate makes them silent no-ops.**
`TERMINAL_STATE_BUTTONS` (`progressView/frontend/constants.ts:79-87`) enables RESUME/RUN_NEW/RESTORE_STATE for any terminal stream, and no frontend code consults `stream.identity` for toolbar gating. The new backend gates then silently `return` — no toast, no disabled state. Pre-PR: click relaunched the wrong agent. Post-PR: click does nothing, with no feedback. Hide the buttons per `stream.identity` (the `isKnownUnsupported` precedent exists in the same component), and surface a message if a click still reaches the backend.

## Minors

- `replayTrace.ts:156-160` — the fallback comment "every pre-migration export is an agent run" is false: pre-migration bash process traces existed (the deleted `isProcessAgent` branch handled them) and now replay as agent runs, losing the terminal renderer. `:147` also sets `agent: trace.config.agent` in `streamTabBase`, a field no longer on `StreamTabInfoSchema`.
- `traceAssembler.ts:89-91` reads `meta?.outcome` directly; unlike `ExecutionsTool.ts:652`, `history.ts:216`, and `summaryFormat.ts:80` it never applies `deriveLegacyOutcome`, so a direct by-id export of an un-healed row regresses to `terminalStatus: null` where the base projected a real status.
- `restartRepair.ts:142-162` — the unmappable-terminal protection is defeated by the stamper: the first `listExecutions` pass stamps `outcome:'failed'` on those rows, after which repair treats them as settled-failed. If intended, the "protected" comment is now stale; if not, gate the stamp.
- Roster v1→v2 shadowing is one-way (`AgentRosterController.ts:96-115`): v1 is read only when v2 is absent, so an old binary sharing `state.json` that edits the roster writes v1 only and the new binary never sees it — silent, permanent divergence. Correctly protects the old binary's value, but the divergence deserves a documented decision.
- Proposal doc vs reality: the status header still says "Parts II–III remain proposed" while the body (correctly) says Axis S (steps 10–11) and Axis R (step 14) landed — the PR body's "the proposal's status header says exactly this" is false. Also the "Demotion repair" migration bullet in the mechanics section was never implemented (`stampExecutionRow` never writes config; the recut `loadHistoryDefaults` made it unnecessary) — delete the bullet.
- `config/ratchets/store-public-surface-baseline.json:44` still lists `getRunDescriptor` and omits `getRunIdentity`. The ratchet is count-only so CI stays green, but the baseline's own semantics say every listed method is caller-verified and should be updated in the same PR.
- `ExecutionsTool.ts:988-990` — a hand-rolled legacy branch (`meta?.identity ? … : meta?.category === 'process'`) is a fourth derivation site, and it diverges from the stamper on config-less `'process'` rows (stamper fabricates `tool:'bash'`; this branch doesn't). `:609-613` also still displays the synthetic `handle.category` for running executions while the completed branch shows the identity-derived value.
- CLI `resume()` (`chatSessionController.ts:479-489`) never seeds `slice.identity`; only `tryResumeStream` (`:616-619`) does — the two cold entrances are now inconsistent, and "cold-read resume seeds identity" is half-true.
- `appInteractionPolicy.ts:285-287` — the child gate widened from `category === Workflow` to `identity?.kind === 'agent'`, so external-CLI agent children now get skip/retry controls that silently no-op per the module's own rationale; zero tests reference this function.
- `BackgroundTasksPanel.ts:515-527` — icon inversion: ordinary delegated native agents flip robot→server while external-CLI rows keep the robot — an unannounced user-visible change. Same hunk dropped the `kind === 'subagent'` guard in `taskStatusBadge`, so a lagging process row now renders in-flight despite the schema doc warning that `status` "must never be used" for membership.
- Pending identity still fabricates the workflow icon: `getAgentCategoryDecorator(undefined)` falls back to the workflow decorator (`icons.ts:105-113`), the tooltip reads "Category: Workflow" (`StreamTabs.ts:271-276`), and `StreamHeader.ts:351-354` falls back to `TOOLBAR_BUTTONS.workflow` — exactly the render-time fabricated defaults the PR claims to delete.
- `desktopCommandPalette.ts:377-389` — the `process` arm is gone; bash streams fall through to the literal `'Stream'`.
- `summaryFormat.ts:40-50` — adding `identity.kind === 'agent'` to `isCallerParentOfToolUseSubagent` re-surfaces the auto-delivered background-bash report in the parent's `/executions` summary (a duplicate result surface), unannounced.
- `childRunLoop.ts:894` — a detached subagent finalizing after detach now reports `isSubagent: false` on its terminal event; and `nativeSubagentStrategy.ts:226` still passes `isSubagent: true` as an `executeAgent` option, so the "3 spellings → `handle.isChildExecution`" claim overclaims.
- Legacy bash streams whose FK can't resolve lose the process UI (`ProgressViewState.ts:296-307` takes identity only from `getRunIdentity`; the old patch synthesized `kind:'process'` from `isProcessAgent`): tool-use chrome instead of `<process-stream-content>`, `command` withheld, `terminalMode` lost. Acceptable cohort degradation if deliberate — worth stating.
- `streamSnapshotRead.ts:87-98` — a malformed legacy `runDescriptor`/`executionId` now discards the whole meta where the old code dropped only the bad pointer and kept `parentStreamId`/`description`.
- `legacyExecutionIdentity.ts:189-197` — the per-read rescan is retired, so a wrongly stamped `streamId` no longer self-corrects. Combined with M1, wrong classifications are now permanent on both artifacts.
- `streamData.ts:45-46` — the legacy `runDescriptor` reader lacks the introduction-date/retirement-condition annotation AGENTS.md requires beside every temporary reader.
- Leftover hygiene from the recut: six unused imports invisible to lint (`no-unused-vars` off): `history.ts:25,29`, `executionFinalization.ts:3` (plus the vestigial `const status = outcome` alias), `App.tsx:18`, `appInteractionPolicy.ts:4`, `StaticConversationTranscript.tsx:15`; stale descriptor/terminalStatus-era comments at `StreamSnapshotStore.ts:283-284,2118`, `WebviewUpdater.ts:352` (recommends the deleted `updateParentStream`), `ProgressFactApplier.ts:448-449,631`, `ProgressViewState.ts:281-290` (dangling `{@link ProgressStreamRunDetails}`), `legacyExecutionIdentity.ts:95`, `executeAgent.ts:394`, `SessionStores.ts:281`, `sessionProgressSubscription.ts:75`, `workflowPlainOutput.ts:155`, `streamStatus.ts:38,46`, `restartRepair.ts:69,209,435`, `executionLifecycle.ts:228`, `streamStateMerge.ts:29-30`, `childExecutions.ts:39`.

## Test gaps (the riskiest new behavior is the least covered)

- The headline gate has **zero negative tests**: `ProgressViewHost.vitest.ts:45` and `ProgressViewCommandHandlers.vitest.ts:143` hardcode `getRunIdentity: () => ({kind:'agent',…})`. No test asserts RESUME / RUN_NEW / RESTORE_STATE are refused for `workflowScript`, `process`, or tool-carrying identities — the defect fixed is precisely the negative case.
- No stamper fixtures for the real legacy cohorts (no-category workflowScript row, no-category external-CLI row); the only no-category fixture enshrines the wrong answer.
- No byte-parity fixtures for `updateActiveSubagents` children or `history-entry`/`history-detail`.
- `executionStatusToRunOutcome` (all three mappings + unmappable→`failed`) and outcome-wins precedence on contradictory legacy meta lost their direct tests.
- Roster: both-keys-present precedence (v2 wins even when v1 is newer) and malformed-v2-with-valid-v1 untested.
- Cold CLI resume seeding from the store FK untested for missing/corrupt meta.
- Wire→frontend parent-edge clearing (the deleted `UPDATE_PARENT_STREAM` test's behavior) not recut.

## Verified clean

All six `registerExecution` sites pass correct identities; both `run.start` emitters carry identity and subscribers store the struct verbatim; `DELEGATION_AVAILABILITY_CATEGORY` has zero production hits and the per-tool `availabilityCategory` declarations are correct (bi-categorical caveat documented at `WorkflowScriptTool.ts:181-184`); `UPDATE_PARENT_STREAM` fully deleted with the parent edge riding `StreamTabInfo.parentStreamId`; `kind`→`category` rename fully applied with no `?? Workflow` left in controllers; `terminalStatus` writes confined to the sanctioned set; shared-schemas deep-import baseline shrunk only, knip baseline unchanged; custom-preset legacy union normalizes once at the boundary; the stamper's lease handling, idempotency, and 120 s half-born guard are sound; resume-hint predicate correctly excludes process/workflowScript/tool-carrying identities; completion scripts keep stable function names; no render-time anti-patterns introduced in the Lit components. All listed grep gates re-run and pass as phrased (modulo `isFullLogChildStream` surviving by name, now identity-keyed).

## What else should still be done

1. **B1**: legacy-pair union member on `AgentDelegationScopeSchema` + real pre-PR team-run config fixture.
2. **B2**: land the missing transcript-test recut (or the tests fail at HEAD).
3. **B3**: decide the NDJSON answer deliberately — boundary projections or accepted vocabulary change with doc updates and byte-parity fixtures.
4. **M1**: teach the stamper the stream-prefix/legacy-descriptor evidence (or leave the cohort unstamped) + fixtures + a stamp-version field if the rule can change.
5. **M2**: legacy reader for the old flat main-view persisted fields.
6. **M3**: reconcile the sidecar retirement with #9627's dated gate, or document the early retirement and fix the contradicting comments.
7. **M4**: gate the toolbar itself on `stream.identity`; toast on any click that still reaches the backend.
8. Add the negative gate tests, stamper cohort fixtures, and NDJSON parity fixtures listed above.
9. Shrink `store-public-surface-baseline.json` (`getRunDescriptor`→`getRunIdentity`), fix the proposal status header + demotion-repair bullet, and do the stale-comment sweep.
10. Deferred proposal work already acknowledged (step 8 `RunRecord` union, Axes T/U, Part III) is fine to leave out — but `replayTrace`'s process-trace fallback, `traceAssembler`'s un-healed-row regression, and the restart-repair/stamper interaction are Part-I-correctness items, not deferred-scope items.

src/agent/storage/executionListing.ts
```diff
@@ -75,8 +75,72 @@ export type ExecutionListingEntry =
...
+ */
+async function stampExecutionRow(
+  id: ExecutionId,
+  meta: ExecutionMeta,
+): Promise<void> {
+  if (Date.now() - Date.parse(meta.timestamp) < STAMP_MIN_AGE_MS) return;
+  try {
+    await runWithInactiveExecutionLease(id, async () => {
+      const store = getExecutionStore(id);
+      const [current, cfg] = await Promise.all([
```
LionSR, line 173: **Major** — verified against the base: legacy workflowScript rows registered with no `meta.category` and `config.agent` = the borrowed default agent (base `WorkflowScriptTool.ts:354,371`), and legacy codex/claude rows also had no category (base `agentCliShared.ts:197`). Both stamp here as `{kind:'agent', tool: undefined}`, which passes `isNativeAgentRun` and the RESTORE_STATE gate — so Resume/Rerun/Restore relaunches the borrowed config for exactly the cohort defect 3 was about, and the wrong stamp is durable with no generation marker. The distinguishing evidence exists and is unused: `meta.streamId` prefixes (`workflow-script#`, `codex@`, `claude@`) and the legacy sidecar `runDescriptor.kind`. Either use it, or leave these rows unstamped (`incomplete`) rather than persisting a wrong identity.

src/shared/schemas/mainView/state.ts
```diff
@@ -155,12 +156,20 @@ const MainViewPersistedStateBaseSchema = UIFileFieldsSchema.merge(
   launchTarget: LaunchTargetSchema.prefault('agent'),
   selectedTeamId: z.string().prefault(''),
   workingDirectory: z.string().prefault(''),
-  workflowAgent: z.string().prefault('correct'),
-  toolUseAgent: z.string().prefault('orchestrator'),
+  agent: z
```
LionSR, line 159: **Major** — the old flat `workflowAgent`/`toolUseAgent`/`workflowInstruction`/`toolUseInstruction` fields parse cleanly against this new shape (unknown keys stripped, `.prefault({})` fires), so upgrading users silently reset to `'correct'`/`'orchestrator'` with empty instructions, and `applyState` persists the reset back. This is the third durable value carrying the retired pair — the other two got a versioned key / legacy union in this PR; this one needs the same treatment (legacy union member or dated preprocess reader).

packages/cli/src/runtime/history.ts
```diff
@@ -19,7 +19,16 @@ import type { AgentConfig } from '@agent/core/definition/AgentConfig';
...
 import type { CliNdjsonRecord } from '@cli/schemas/cliOutput';
-import { ExecutionIdSchema, type ExecutionId } from '@shared/schemas';
+import {
+  ExecutionIdSchema,
+  RunOutcomeSchema,
+  type ExecutionId,
+} from '@shared/schemas';
+import {
+  resolveHistoryRunStatus,
+  type HistoryRunStatus,
```
LionSR, line 29: **Blocker (per the PR's own frozen-NDJSON gate)** — `runOutcomeToExecutionStatus` is imported but never called, and `:214,500` now emit the `RunOutcome` vocabulary (`cancelled`/`failed`) where the deleted `resolveCliHistoryStatus` emitted raw `terminalStatus` (`interrupted`/`error`) — while `docs/guide/texra-cli.md:283` declares the NDJSON form "stable for scripts". The proposal's boundary-projection machinery exists; it was wired in but never applied. Same story for `updateActiveSubagents` (`sessionProgressSubscription.ts:153-159` spreads the recut `ActiveChildInfo` verbatim). Either project the old vocabularies at the boundary or drop the "byte-identical" claim and add byte-parity fixtures.

src/shared/schemas/streamData.ts
```diff
@@ -28,30 +25,33 @@ import {
...
+export const StreamTabMetaSchema = z
+  .object({
+    schemaVersion: z
+      .literal(STREAM_TAB_META_SCHEMA_VERSION)
+      .prefault(STREAM_TAB_META_SCHEMA_VERSION),
+    parentStreamId: z.string().optional(),
+    /** FK into `executions/{executionId}/` — the durable run authority. */
+    executionId: ExecutionIdSchema.optional(),
+    /** Legacy pre-FK sidecar shape; only the FK inside it is read. */
+    runDescriptor: z.looseObject({ executionId: ExecutionIdSchema }).optional(),
```
LionSR: Two issues on this schema: (1) dropping `description`/`taskState` retires them ~3 months before the documented ≥ 2026-11-01 gate (#9590 Stage 7 / #9627; `legacyExecutionIdentity.ts:21-22` still conditions it on usage evidence), and `writeMeta` now destroys both on any unrelated meta patch — this early retirement is not mentioned in the PR body. (2) This legacy `runDescriptor` reader lacks the introduction-date/retirement-condition annotation AGENTS.md requires beside every temporary reader.

LionSR:
All review items addressed in `9c72223` (plus `ff30c51`, which the review predates). Point by point:

**B1** — fixed. `AgentDelegationScopeSchema` is now a `z.union` whose legacy member accepts the persisted `workflowAgentKeys`/`toolUseAgentKeys` pair and transforms into the record (same pattern as the presets migration). Fixture parses a realistic pre-PR team-run `config.json` through `AgentConfigSchema` and asserts it lists as `kind:'run'`.

**B2** — was already resolved at `ff30c51` (pushed while the review was being cut against the merge commit): the two transcript tests asserting the retired shims were recut there, and the full suite is green at HEAD minus the two root-container `ExtensionTexraConfig` artifacts.

**B3** — took the boundary-projection answer (gate G), not the vocabulary change: `history-entry`/`history-detail` project through the previously-unwired `runOutcomeToExecutionStatus` (`cancelled→interrupted`, `failed→error`), and `updateActiveSubagents` projects each row back to the public `kind`/`toolName`/`childStreamId` shape with `identity` never emitted. Byte-parity tests (`toStrictEqual`, all four cohorts) pin both.

**M1** — fixed with the evidence you pointed at: `deriveLegacyIdentity` reads `meta.streamId` prefixes (`workflow-script#` → `multiAgentWorkflow`, `codex@`/`claude@` → agent+tool, `bash@` → process), confined to the one quarantined migration function and documented as the only prefix-reader. Legacy non-native cohorts no longer stamp as native agents, so the resume gate holds for pre-migration rows; cohort fixtures cover all four prefixes, and the same rule feeds snapshot hydration and the trace-viewer fallback (which also lost its false "every pre-migration export is an agent run" comment).

**M2** — fixed: a parse-entrance preprocess lifts the legacy flat `workflowAgent`/`toolUseAgent`/`workflowInstruction`/`toolUseInstruction` fields into the records; old-blob fixture asserts selection + drafts survive.

**M3** — the early retirement of the sidecar `description`/`taskState` mirrors is deliberate (per review direction that the legacy sidecar cohorts may break now) and is now stated as such: `legacyExecutionIdentity.ts`'s header and the contradicting `writeMeta`/`TaskState.ts` comments are reconciled, and the legacy `runDescriptor` reader carries the required introduction/retirement annotation.

**M4** — fixed at both layers: the toolbar hides RESUME/RUN_NEW/RESTORE_STATE for non-native (or pending) identities via the existing hidden-button mechanism, and any click that still reaches the backend surfaces through the host info port instead of a silent return. Negative tests cover all three commands × `multiAgentWorkflow`/`process`/tool-carrying identities.

**Minors** — all applied where still live at HEAD (several were mooted by `fd989ae`'s outcome-only tightening, e.g. the `deriveLegacyOutcome` divergences and the stamper/repair interaction — the stamper no longer stamps outcome at all): stale-comment sweep across ~15 files, unused imports, `store-public-surface-baseline` `getRunDescriptor`→`getRunIdentity`, proposal status header + demotion-repair bullet, desktop palette process description, roster icon inversion + `taskStatusBadge` membership guard, background-bash report suppression restored in parent summaries, skip/retry gate narrowed to native agents, CLI `resume()` identity seeding, `readMeta` field-level tolerance (bad FK dropped loudly, `parentStreamId` kept).

**Deliberate divergences to flag**: the v1→v2 roster shadowing stays one-way (protects the old binary's value; divergence accepted and now documented), and legacy bash streams whose FK can't resolve keep the degraded non-process chrome — accepted cohort breakage per the terminal-tab ruling.

---

LionSR (commented):
Second-pass review, re-graded against the proposal's own compatibility ruling (protected durable / breakable durable / ephemeral, with the single NDJSON exception) rather than the repo's generic compatibility policy. Key claims this round are empirically verified at `f850797c` (ran the suites and the schemas, not just read them).

## Withdrawn or downgraded after reading the proposal — my earlier review over-called these

- **Main-view persisted state reset** — `mainView/state.ts:151-171` is named verbatim as **Breakable durable** ("version-bump freely; degradation acceptable"). Sanctioned. (previous M2 → withdrawn)
- **Sidecar `description`/`taskState` retirement** — Axis T explicitly deletes both, "#9627, folded in," as a breakable sidecar cohort; the proposal supersedes the 2026-11-01 schedule. What remains is bookkeeping (see below). (previous M3 → minor)
- **Legacy workflowScript/codex cohort stamping as `kind:'agent'`** — the PR implements the identity rule exactly as specified (proposal "One migration" identity rule), amendment D and open problem 5 accept the un-healable codex cohort, and the ruling says bash/workflow-script history may break. Spec-faithful, not a bug. Residual: the consequence (defect 3 stays alive for the legacy cohort, stamped permanently, no marker) deserves one line in the PR body, and `ExecutionListing.vitest.ts:131-159` enshrines a no-category `config('bash')` fixture production never wrote — bash always wrote `category:'process'`. (previous M1 → minor)
- Also withdrawn/downgraded per the proposal's text: roster v1→v2 one-way shadowing (the specified Axis R design), the `summaryFormat` wait/report predicate change (amendment C verbatim), the `legacyExecutionIdentity` rescan retirement (accepted under "Risks"), desktop palette/process-UI degradation ("desktop may break"), the trace-viewer agent fallback (the proposal's own confined fallback).

## What stands — each is the proposal's own requirement, unmet

**1. The NDJSON exception is violated twice, and the mandated parity test never landed.**
The compatibility table's one frozen class: "the headless NDJSON vocabulary is a declared frozen public contract … projected at its boundary, **never recut**." Amendment G is specific: "`history-entry.status` keeps the `ExecutionStatus` vocabulary via `runOutcomeToExecutionStatus` at that one boundary. **The byte-parity test lands *before* the deletions.**"
- `history-entry`/`history-detail` now emit `RunOutcome` (`cancelled`/`failed`) where `resolveCliHistoryStatus` emitted `ExecutionStatus` (`interrupted`/`error`); `runOutcomeToExecutionStatus` is imported at `history.ts:29` and never called.
- `updateActiveSubagents` children rows changed shape (`kind`/`toolName` gone, `identity` added, `childStreamId` now required) because the unchanged projection (`sessionProgressSubscription.ts:158`) spreads the recut `ActiveChildInfo` verbatim. Projection back is trivial — the proposal's own audit found the `process` arm was never constructed, so every old row was `kind:'subagent'` with `toolName` from what is now `identity.tool`.
- No byte-parity test exists for either surface. This is the proposal's explicit acceptance gate, not reviewer conservatism.

**2. A third durable pair-carrier the Axis R audit missed: team-run `config.json` — a Protected durable cohort.**
Axis R's findings say "exactly two of them are durable" (roster selection + custom presets), and the PR migrated both correctly. But `delegationAgentScope: AgentDelegationScopeSchema.nullish()` sits in `AgentConfigSchema` (`AgentConfig.ts:49`), and team runs persist the old pair shape into `config.json`. Verified empirically at HEAD with the repo's zod: `{workflowAgentKeys, toolUseAgentKeys}` now **fails** `AgentConfigSchema.safeParse` (the v4 record is exhaustive — missing `workflow`/`toolUse` keys invalid, pair keys unrecognized) → `readConfig()` warns and returns null → rows list as `incomplete`: old team runs vanish from history and can't be resumed. These are native agent runs — the inviolable class. Fix is the same legacy-member `z.union`+`transform` the PR already used twice, plus a fixture reading a real pre-PR team-run config.

**3. The suite is red at HEAD — 15 failures, not the claimed 2.**
Ran `src/test-kernel/transcript/` at `f850797c`: **15 failed / 214 passed across 2 files**, all asserting the retired machinery (taskState/description round-trips, descriptor handling, the rescan). The head commit message says "updated in the follow-up commit" — that commit is not in this PR. Two of the 15 also burn 10-second timeouts each. The body's "8448/8450, only 2 pre-existing failures" does not describe the head commit.

**4. Non-agent streams must show *no* resume/rerun/restore affordance — the step-2 test spec, not my invention.**
Proposal step 2, Tests: "a non-agent stream shows no resume/rerun/restore affordance." The PR gates only the backend, which silently returns; the frontend enables RESUME/RUN_NEW/RESTORE_STATE for every terminal stream (`TERMINAL_STATE_BUTTONS`) and never consults `stream.identity`. Clicking now does nothing with no feedback — better than relaunching the wrong agent, but not the specified behavior, and the negative gate tests the spec implies are absent (both host suites hardcode `kind:'agent'`).

## Cleanup debt — directly against this PR's stated purpose

- **Bookkeeping**: the status header says "Parts II–III remain proposed" while Axis S (steps 10–11) and Axis R (step 14) landed — and parts of Axis T landed too (the sidecar shims) while the header names Axis T as remaining. The "Demotion repair" bullet in "One migration" was never implemented (`stampExecutionRow` never writes config; the recut `loadHistoryDefaults` made it unnecessary) — delete the bullet.
- **Ratchet baseline**: `store-public-surface-baseline.json:44` still names `getRunDescriptor`, omits `getRunIdentity`. Count-only ratchet stays green, but the baseline's own semantics require a same-PR swap. (Dead-code ratchet verified: 18=18, passes.)
- **One invariant, two spellings**: `isNativeAgentRun` (`ProgressViewHost.ts:59-61`) vs the inline `identity?.kind !== 'agent' || identity.tool !== undefined` at `ProgressViewCommandHandlers.ts:460-461` — the PR's central gate duplicated in sibling files.
- **A fourth legacy-derivation site**: `ExecutionsTool.ts:988-990` re-implements the category sniff inline, diverging from the stamper on config-less `'process'` rows (the stamper fabricates `tool:'bash'`; this branch doesn't). "Healing lives in exactly one place" — except here.
- **"Absent ≠ Workflow" violated in the icon layer** (proposal's silent-defaults ruling): `getAgentCategoryDecorator(undefined)` still returns the workflow decorator (`icons.ts:105-113`), the tooltip reads "Category: Workflow" (`StreamTabs.ts:271-276`), `StreamHeader.ts:351-354` falls back to `TOOLBAR_BUTTONS.workflow`.
- **Lint-invisible leftovers**: six unused imports from the recut (`history.ts:25,29`, `executionFinalization.ts:3` + vestigial `const status = outcome`, `App.tsx:18`, `appInteractionPolicy.ts:4`, `StaticConversationTranscript.tsx:15`) — `no-unused-vars` is off, so this class of half-applied cleanup is invisible; worth a one-time sweep or enabling the rule for the touched packages. Plus ~15 stale comments still describing deleted machinery (top offenders: `WebviewUpdater.ts:352` recommends the deleted `updateParentStream`; `StreamSnapshotStore.ts:1468-1472` claims legacy description round-trips "before its Stage 7 retirement"; `ProgressViewState.ts:281-290` has a dangling `{@link ProgressStreamRunDetails}`; `streamStatus.ts:38,46` still describes `terminalStatus` as the persisted projection target).

## Empirically verified clean this round

Dead-code ratchet 18=18 passes; all six `registerExecution` sites emit correct identities; the stamper's lease handling, idempotency, and half-born guard are sound; the transcript suite is green apart from the 15 retired-machinery assertions; grep gates pass as phrased; `UPDATE_PARENT_STREAM` fully deleted; `kind`→`category` rename fully applied; `terminalStatus` writes confined to the sanctioned set.

## What else should still be done (ordered)

1. Land the NDJSON byte-parity suite + the two boundary projections (`runOutcomeToExecutionStatus` for history status; `{kind:'subagent', toolName}` row projection for `updateActiveSubagents`) — amendment G.
2. Legacy-pair union member on `AgentDelegationScopeSchema` + a real pre-PR team-run config fixture — protected-cohort data loss.
3. Land the transcript test recut — 15 red at HEAD.
4. Hide resume/rerun/restore per `stream.identity` (step-2 spec) + negative tests for the gate (workflowScript / process / tool-carrying / absent identity).
5. Bookkeeping sweep: status header, demotion-repair bullet, `getRunDescriptor`→`getRunIdentity` baseline swap, unused imports, stale comments.
6. One line in the PR body stating the accepted consequence for the legacy workflowScript/codex cohort (defect 3 persists there, permanently stamped), and a decision note on the roster-icon inversion (robot→server for native delegated agents).
7. Deferred scope (step 8 `RunRecord`, Axes T/U beyond what landed, Part III) is correctly out of this PR — no action.

PR #9711, opened 2026-08-04
fix(progressView): restore the shared focus ring, add non-color status cues, fix measured contrast

LionSR:
An interface review of the session view (`packages/extension/src/progressView/frontend/`) turned up eleven issues across accessibility, color, typography, and layout. This fixes ten. Every color claim below is a measured ratio, not an impression.

## Focus rings — the repo already made this call

`focusRingStyles` (`src/shared/styles/controlStyles.ts:34`) is documented as "one focus ring for every interactive element in the shadow root", set to `--focus-ring-width: 2px` with the rationale written next to it:

> A 1px ring is the kind of thing that reads as tidy and fails a low-vision user.

It reaches every affected shadow root already, via `commonViewStyles` / `logStyles`. Eight selectors were beating it on specificity with a 1px outline, and `.workflow-task--linked` replaced it with `outline: none` plus a border-color flip **identical to its hover state** — so a keyboard user had no way to tell focused from pointed-at.

All eight now defer to the shared ring, keeping only the `border-radius` the outline needs to follow the shape. One exception: `.code-block-copy` keeps a local rule, because a `:focus-visible` in the adopting sheet cannot reach inside `wa-button`'s shadow root — it is that control's only ring, so it was widened to the shared tokens rather than removed.

The eighth site (`src/shared/styles/markdownStyles.ts`) was found while implementing and is shared with the desktop host; the change there moves it in the same direction.

## Status was carried by hue alone

A run's state — running, failed, waiting, starting, and **blocked on your approval** — was encoded only in the `border-left-color` of a 2px edge. A red-green colorblind reader could not distinguish a failed run from a running one, and the one state that actively wants you had no shape at all. The label was already in the `aria-label` and the tooltip, so pointer-free visual scanning was the only path with nothing.

Rows now carry a glyph too, rendered in `currentColor` so it adds no new foreground to verify. It appears only for states that light the rail — completed, cancelled, and ready leave it transparent and get no glyph, so a settled rail stays as quiet as it is today.

## Contrast, measured

Against the four default VS Code themes:

| Fix | Before | After |
| --- | --- | --- |
| Approve label (`ApproveSplitButton`) | **3.90:1** Light+, **4.07:1** Light Modern | 5.63 / 5.88 / 7.65:1 |
| Usage strip text (`UsagePanel`) | **4.32:1** Dark+, **3.34:1** Light Modern | token value, no stacked alpha |
| Usage strip icons | **2.87 / 2.96 / 2.95 / 2.21:1** — under 3:1 in *all four* | as above |

The Approve button was using a `-fill-` token as a foreground; `controlStyles.ts:558` applies `--split-accent` as `color`. `--wa-color-success-on-quiet` is the role-correct token.

The usage strip stacked element `opacity` on top of `--color-text-secondary`, which is itself `color-mix(… 70%, transparent)` — compounding to 0.595 effective alpha on text and 0.4165 on icons. The token already carries the de-emphasis.

## Also

- **Five unlabeled textareas** — including the main follow-up composer — had only a placeholder for a name. `WorkflowToolUseFollowupSection` already showed the intended pattern.
- **`aria-label="Delete"`** repeated identically down the rail; now names its stream.
- **20px hit targets** in the rail raised to 24px (WCAG 2.5.8). Deliberately a literal, not `--control-size-s`: that step is bridged to 20px in the extension for editor-chrome density. A pseudo-element hit area was rejected — it would overlap the select target, so near-misses would *delete* a session.
- **`tabular-nums`** on the live token/cost counters, which shifted the right-aligned footer on every tick.
- **Conversation measure** 800px → 680px (~127 characters at the 13px body size, against a 60–75 target). Code blocks and diffs still overflow the column.
- **`word-break: break-all`** → `overflow-wrap: anywhere` on log lines; the file's own comment already made this trade for terminal streams.
- Two `...` → proper ellipses.

## Deliberately not fixed

The rail's `.tab-meta` is hidden until hover and revealed with `display: flex`, which grows the hovered row and pushes every row below it. That is a real reflow, but the fix is a design call about the recent declutter work (#9702) rather than a defect fix, so it is left for a decision.

## Verification

- `npm run lint` — clean.
- `npm run check:dead-code-ratchet` — 18 findings vs 18 baselined, no new exports.
- `npx vitest run` — **820 files / 8,421 tests passing**, 1 file and 5 tests skipped.
- Contrast computed in Node from the stock theme values (WCAG 2.x relative luminance), compositing `color-mix` alpha and element `opacity` before measuring.

**Not verified:** no rendered inspection — these webviews only mount inside a VS Code extension host — and no screen-reader pass. The status glyph, the 680px measure, and the 24px rail targets are the three changes most worth a look on screen before merge.

---

> [!NOTE]
> **Low Risk**
> Progress-view UI, styling, and ARIA only—no auth, data, or API changes; risk is mainly visual regression in the extension webview.
> 
> **Overview**
> Addresses accessibility and readability issues in the progress view session UI: focus visibility, stream status scanning, measured contrast, labels, and a few layout tweaks.
> 
> **Focus** — Removes local 1px `:focus-visible` outlines that overrode the shared **2px** `focusRingStyles` ring; linked workflow tasks no longer use hover-identical border styling on focus. `UserMessage` composes `focusRingStyles` for focusable LaTeX refs. `wa-button` copy controls keep `::part(base)` rings aligned to shared focus tokens.
> 
> **Stream tabs** — Adds `currentColor` status glyphs (including pending approval) beside hue-only rail colors; enlarges delete/expand controls to **24×24**; delete `aria-label` includes the stream title.
> 
> **Contrast & typography** — Approve split button uses `--wa-color-success-on-quiet`; usage footer drops stacked opacity on secondary text/icons; token/cost values use `tabular-nums`; conversation column **680px**; log lines use `overflow-wrap: anywhere` instead of `break-all`.
> 
> **Forms** — Adds `aria-label`s on several textareas (follow-up composer, rejection, external inquiry, user questions).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2f886f60869af3bf004a1761abff1cba6f79f8f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/src/progressView/frontend/components/ApproveSplitButton.ts
+++ b/packages/extension/src/progressView/frontend/components/ApproveSplitButton.ts
@@ -68,8 +68,13 @@ export class ApproveSplitButton extends LitElement {
         flex: 0 1 auto;
         min-width: auto;
         max-width: min(14rem, 100%);
-        /* Tints both halves of the shared split-button skin. */
-        --split-accent: var(--wa-color-success-fill-loud);
+        /* Tints both halves of the shared split-button skin. The on-quiet
+           token, not fill-loud: the skin applies this as the label color, and
+           a fill token used as a foreground measured 3.90:1 (Light+) and
+           4.07:1 (Light Modern) against the panel surface — under AA for the
+           primary action of the approval flow. The on-quiet token is the
+           role-correct one and measures 5.63 / 5.88 / 7.65:1. */
+        --split-accent: var(--wa-color-success-on-quiet);
       }
     `,
   ];
--- a/packages/extension/src/progressView/frontend/components/BaseFeedbackPanel.ts
+++ b/packages/extension/src/progressView/frontend/components/BaseFeedbackPanel.ts
@@ -93,6 +93,7 @@ export abstract class BaseFeedbackPanel<
       <div class=${containerClass}>
         <wa-textarea
           class=${inputClass}
+          aria-label=${placeholder}
           placeholder=${placeholder}
           rows="3"
           data-feedback-input
--- a/packages/extension/src/progressView/frontend/components/ExternalInquiryPanel.ts
+++ b/packages/extension/src/progressView/frontend/components/ExternalInquiryPanel.ts
@@ -400,7 +400,8 @@ export class ExternalInquiryPanel extends BaseFeedbackPanel<'externalInquiry'> {
         </div>
         <wa-textarea
           class="external-inquiry-request__answer-input"
-          placeholder="Paste the answer here..."
+          aria-label="Answer from the external model"
+          placeholder="Paste the answer here…"
           rows="4"
           resize="vertical"
           .value=${live(this.answerText)}
@@ -458,7 +459,8 @@ export class ExternalInquiryPanel extends BaseFeedbackPanel<'externalInquiry'> {
           </div>
           <wa-textarea
             class="external-inquiry-request__session-links-input"
-            placeholder="Paste one external session link per line..."
+            aria-label="External session links, one per line"
+            placeholder="Paste one external session link per line…"
             rows="2"
             resize="vertical"
             .value=${live(this.sessionLinksText)}
--- a/packages/extension/src/progressView/frontend/components/FollowUpInput.ts
+++ b/packages/extension/src/progressView/frontend/components/FollowUpInput.ts
@@ -404,6 +404,7 @@ export class FollowUpInput extends LitElement {
         <div class="composer-surface">
           <wa-textarea
             id=${ELEMENT_IDS.FOLLOW_UP_INPUT}
+            aria-label="Follow-up message"
             placeholder="Message TeXRA…"
             rows="6"
             resize="vertical"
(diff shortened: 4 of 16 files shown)


LionSR:
Both review findings confirmed and fixed in cec171bb0a — they were the same gap seen from two directions: a control whose shadow root the shared sheet does not reach.

**`UserMessage`** — verified it is the *only* `markdownStyles` consumer that does not also compose `commonViewStyles` (`MemoryItem`, `HistoryItemElement`, `ProposalRequestPanel`, and `logStyles` all do), so it was the only one that stopped getting a ring. It now composes `focusRingStyles` directly — the pattern `src/shared/styles/index.ts:12` already documents for a component that wants the control layer without the full common view.

**`.banner-content-copy`** — confirmed it is a `wa-button` (`htmlBuilders.ts:126`, `messageFormatters.ts:148`), and that WebAwesome ships no focus styling for it at all: `button.styles.js` is 233 bytes and contains zero occurrences of `focus`. So it was falling back to the UA default ring while its sibling `.code-block-copy` showed the branded one. It now gets the same `::part(base)` exception at the shared token width.

Also checked, since it is the same class of bug: nothing else in the diff removed an outline from a control whose ring came from a sheet it does not compose.

typecheck, eslint, and prettier clean.

PR #9714, opened 2026-08-04
fix(ui): make every custom property actually resolve, and take colour from the right role

LionSR:
Stacked on #9711 — **base is `fix/session-view-a11y-contrast`**, retarget to `main` once that merges. Only `src/shared/styles/` overlaps between the two.

This is PR 1 of a cross-surface interface review covering the launcher webview, settings view, desktop renderer, trace viewer, shared style layer, and CLI. It takes the items that are **not** design choices: declarations that quietly evaluate to nothing, and colour taken from the wrong role. Everything here is one-line-per-site.

## The desktop `:root` mirror had drifted

`themeTokens.css` keeps its own copy of the shared `:host` tokens because light DOM cannot adopt a Lit `CSSResult`. Its own comment states the risk:

> a token added there and missing here silently resolves to nothing in the desktop's light DOM

Six names had drifted out of sync. `styles.css` and `reviewPane.css` consume `--border-thin` **fourteen times**; with it undefined, a `border: var(--border-thin) solid X` shorthand is invalid at computed-value time and is dropped entirely — so those seams never rendered as authored. Same for `--border-medium`, `--letter-spacing-caps`, and the `--color-success` / `--color-error` / `--color-text-link` trio.

## Six names declared nowhere in the repo

No fallback on any of them, so each was a dead declaration:

| Was | Now | Site |
| --- | --- | --- |
| `--icon-size-m` | `--font-size-icon` | `styles.css:648` |
| `--letter-spacing-wide` | `--letter-spacing-caps` | `styles.css:718` |
| `--icon-size-s` | `--font-size-icon-sm` | `styles.css:789` |
| `--height-control-s` | `--height-control-compact` | `styles.css:860` |
| `--row-height-compact` | `--height-control-compact` | `styles.css:1598` |
| `--icon-size-l` | `--font-size-h1` | `reviewPane.css:181` |
| `--height-control-l` | `--height-control` | `settingsView/frontend/styles.ts:52` |

## Raw `--vscode-*` in the launcher

Seven declarations across three onboarding/banner components reached past the token bridge straight to the VS Code host. Those same components mount in the desktop app, which defines **no** `--vscode-*` at all — so the description text colour and the card border resolved to nothing there.

## Colour from the wrong role

- `--wa-color-danger-60` / `--wa-color-success-60` are raw WebAwesome palette steps that bypass the host bridge, so the forced-contrast block cannot reach them. Measured on the desktop light canvas: `#f3676c` = **3.01:1**, `#00ac49` = **3.00:1**. Now `--color-error` / `--color-success`.
- `--desktop-color-success` was `#0e9f6e` = **3.39:1** on white, and it feeds `--wa-color-success-on-quiet` — a *text* role. Darkened to `#0c875e` (**4.52:1**). Dark value untouched at 9.63:1.

## Two new tokens, on both bridges in the same commit

```css
--color-text-muted: color-mix(in srgb, var(--wa-color-text-normal) 55%, transparent);
--border-control:   color-mix(in srgb, var(--wa-color-text-normal) 55%, transparent);
```

They replace the `--color-text-secondary` + element-`opacity` pattern that multiplies two alphas into a value nobody wrote down. **55%, not 45%** — at 45% this measures 2.41:1 against Light Modern and misses the 3:1 non-text minimum; at 55% the worst of four stock VS Code themes plus both desktop appearances is 3.07:1. Two names for one expression is deliberate: a border reading a text token is the same role violation this PR exists to fix.

They land on **both** bridges here because adding one to `litStyles.ts` alone would reproduce the mirror gap above. PR 2 consumes them.

## Deliberately not included — these are design calls

| Item | Why it needs a decision |
| --- | --- |
| Raising `--surface-selected` (dark selection is 1.46:1) | It also drives `.btn-secondary:active`, so it changes the shared active state of every secondary button |
| Markdown `h1`–`h4` collapsed to one size | Fixing it makes h3/h4 the same size as body copy, distinguished by weight only |
| Inline `code` / `.latex-ref` rendering in UI sans | Changes the reading density of every agent response |
| `--wa-control-size-s: 20px` hit areas | The 20px carries an explicit "reads as editor chrome" rationale; raising it resizes every compact toolbar |
| Desktop terminal ignoring its 16 ANSI tokens (light-mode white is 1.16:1) | Needs a token-format change first — the values are `light-dark()`, which xterm cannot parse |

## Verification

- `npm run lint` — clean.
- `npm run check:dead-code-ratchet` — 18 vs 18 baselined.
- `npx vitest run` — **820 files / 8,421 tests passing**, 1 file and 5 tests skipped.
- Grep sweep confirming no phantom token name remains in the desktop light DOM, and that `--vscode-*` count in `webview/frontend/` is now 0.
- Token parity check: all 8 names present in both `litStyles.ts` and `themeTokens.css`.
- Contrast computed in Node from stock theme values, compositing `color-mix` alpha before measuring.

**Not verified:** nothing was rendered. These are token-resolution and colour-role fixes derived from the cascade; the desktop seams that were previously dropped (`--border-thin`) will now actually draw, which is a visible change worth a look.

`ShortcutsTab.ts` also picks up a whitespace-only reformat — it was already unformatted on `main` and the pre-commit prettier hook normalizes any file a commit touches.

---

> [!NOTE]
> **Low Risk**
> CSS token and colour-role fixes with no auth, data, or API changes; main visible risk is borders now rendering where shorthand was previously invalid.
> 
> **Overview**
> **Token resolution:** Replaces six undeclared custom properties (`--icon-size-*`, `--letter-spacing-wide`, `--height-control-s`, `--row-height-compact`, `--height-control-l`) with existing shared names so icon sizing, caps tracking, and compact control heights compute correctly.
> 
> **Desktop light DOM:** Adds missing mirror tokens in `themeTokens.css` (`--border-thin`, `--border-medium`, `--letter-spacing-caps`, `--color-success` / `--color-error` / `--color-text-link`) so `border: var(--border-thin) solid …` shorthands stop being dropped entirely. Introduces `--color-text-muted` and `--border-control` on both `litStyles.ts` and the desktop mirror, with high-contrast re-declarations on `<body>` so aliases pick up system colours.
> 
> **Colour roles:** Swaps raw WebAwesome palette steps and `--vscode-*` usage for bridged tokens (`--color-success`, `--color-error`, `--wa-color-text-quiet`, `--wa-color-surface-*`) in task shell, shortcuts, and onboarding components. Darkens light-mode `--desktop-color-success` for text-role contrast.
> 
> **Follow-up:** PR 2 will consume the new muted/border tokens; this change may visibly draw borders that were previously invalid.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 521c8538954d325333cbded8f4d7d92272b8bb41. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/renderer/reviewPane.css
+++ b/packages/desktop/src/renderer/reviewPane.css
@@ -178,7 +178,7 @@
 }
 
 .desktop-review-empty .empty-state-icon {
-  font-size: var(--icon-size-l);
+  font-size: var(--font-size-h1);
 }
 
 .desktop-review-empty .empty-state-title {
--- a/packages/desktop/src/renderer/styles.css
+++ b/packages/desktop/src/renderer/styles.css
@@ -645,7 +645,7 @@ wa-dialog.desktop-command-palette::part(body) {
   height: 1em;
   place-items: center;
   color: var(--wa-color-text-quiet);
-  font-size: var(--icon-size-m);
+  font-size: var(--font-size-icon);
 }
 
 wa-input.desktop-command-palette-input {
@@ -715,7 +715,7 @@ wa-input.desktop-command-palette-input::part(input) {
   color: var(--wa-color-text-quiet);
   font-size: var(--font-size-xs);
   font-weight: var(--font-weight-semibold);
-  letter-spacing: var(--letter-spacing-wide);
+  letter-spacing: var(--letter-spacing-caps);
   text-transform: uppercase;
 }
 
@@ -786,7 +786,7 @@ wa-input.desktop-command-palette-input::part(input) {
   border-radius: var(--wa-border-radius-s);
   background: var(--wa-color-surface-lowered);
   color: var(--wa-color-text-quiet);
-  font-size: var(--icon-size-s);
+  font-size: var(--font-size-icon-sm);
 }
 
 .desktop-command-palette-main {
@@ -857,7 +857,7 @@ wa-input.desktop-command-palette-input::part(input) {
 }
 
 .desktop-command-palette-customize::part(base) {
-  min-height: var(--height-control-s);
+  min-height: var(--height-control-compact);
   padding-inline: var(--wa-space-2xs);
   color: var(--wa-color-text-quiet);
   font-size: var(--font-size-xs);
@@ -1595,7 +1595,7 @@ wa-card.desktop-onboarding::part(footer) {
 }
 
 .desktop-editor-tree-status::part(item) {
-  min-height: var(--row-height-compact);
+  min-height: var(--height-control-compact);
   padding: var(--row-padding);
 }
 
--- a/packages/desktop/src/renderer/taskShell.css
+++ b/packages/desktop/src/renderer/taskShell.css
@@ -626,7 +626,7 @@ body[data-desktop-platform='darwin'] .task-sidebar-brand {
 }
 
 .task-environment-trailing.is-success {
-  color: var(--wa-color-success-60);
+  color: var(--color-success);
 }
 
 .task-environment-diff {
@@ -634,11 +634,11 @@ body[data-desktop-platform='darwin'] .task-sidebar-brand {
 }
 
 .task-environment-diff .is-added {
-  color: var(--wa-color-success-60);
+  color: var(--color-success);
 }
 
 .task-environment-diff .is-deleted {
-  color: var(--wa-color-danger-60);
+  color: var(--color-error);
 }
 
 .task-environment-more {
--- a/packages/desktop/src/renderer/themeTokens.css
+++ b/packages/desktop/src/renderer/themeTokens.css
@@ -121,7 +121,7 @@
   );
   --desktop-color-info: light-dark(#087ea4, #74c7ec);
   --desktop-color-info-background: light-dark(#e7f5fa, rgb(116 199 236 / 14%));
-  --desktop-color-success: light-dark(#0e9f6e, #56e0a0);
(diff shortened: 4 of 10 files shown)


LionSR:
Addressed the Bugbot finding in d98cb0f1c0.

The catch is correct and it is a subtle one: a custom property substitutes its `var()` source **at the point of declaration**, not the point of use. `--color-success` / `--color-error` / `--color-text-link` were declared on `:root`, and the high-contrast overrides live on `body.*-high-contrast` — a descendant — so overriding the `--wa-*` names could never reach the aliases. Task statuses would have kept desktop green and red instead of the system `Highlight` and `Mark`.

The aliases are now re-declared in that block. Same reasoning applies to `--color-text-muted` and `--border-control` (they read `--wa-color-text-normal`), so those are declared there too — `GrayText` and `CanvasText`.

On the other note: the `transparent` fallback dropped from `OnboardingWelcomeCard.ts:96` is deliberate. `--wa-color-surface-border` resolves on both bridges, and this PR's whole point is that a token which resolves nowhere should be found and fixed rather than papered over with a fallback that hides it.

LionSR:
Addressed the success-contrast thread: light `--desktop-color-success` is now `#0a7a54` (~4.87:1 on the raised `#f4f4f4` surface), clearing the 4.5:1 requirement used by review-pane success text.

PR #9715, opened 2026-08-04
fix(ui): stop stacking element opacity on tokens that already carry alpha

LionSR:
Stacked on #9714 (which is stacked on #9711). Retarget as those merge.

PR 2 of the cross-surface interface review. **One root cause, fourteen rules.** `--wa-color-text-quiet` is `color-mix(… 70%, transparent)`, so an element `opacity` on top multiplies into an effective alpha nobody wrote down — and every measured composite landed under its threshold.

| Site | effective α | measured | needs |
| --- | --- | --- | --- |
| Status dots, quiet states | 0.49 | 2.65:1 | 3:1 |
| Status dot, `is-ready` | 0.35 | **1.94:1** | 3:1 |
| Disabled `wa-option` | 0.49 | 2.65:1 | 3:1 |
| Unavailable model row | 0.35 | **1.91:1** | 3:1 |
| Session hint time | 0.595 | 3.18:1 | 4.5:1 |
| Login banner fineprint | 0.60 | 3.22:1 | 4.5:1 |
| Onboarding choice copy | 0.70 | 4.11:1 | 4.5:1 |
| Desktop split-panel grips | 0.45 | **1.99:1** | 3:1 |

The fix is the same everywhere: delete the `opacity`, express the step once in colour via `--color-text-muted` from #9714.

## Three sites needed judgement, not a mechanical swap

**Login banner fineprint** is the programme terms a user is agreeing to. It was `--font-size-xs` (10.4px) *and* faded. `--font-size-sm` would still be 11.7px, so it now inherits the 13px body size and takes its de-emphasis from the callout treatment rather than a second size step plus an alpha.

**The unavailable model row's dim** also covered `.model-metadata` and `.reasoning-level-select` — both rendered *outside* the `<wa-switch>`, so they never qualified for the disabled-control contrast exemption in the first place. I checked before removing it: the row already renders a `key` or `triangle-exclamation` icon with a title (`ModelSelectionList.ts:196-210`), so unavailability keeps a cue that survives a contrast check. This was flagged as a possible design decision during review; that cue is what resolves it.

**The saturated status dots** carried `opacity: 1` only to undo the base rule's fade. With the fade gone, they don't need it.

## Also here

- `.file-folder` moves off the 10.4px step to 11.7px.
- The two 2px gutters between adjacent icon-only buttons widen to 4px so their hit areas stop touching.

The buttons themselves stay 20px. Reaching 24×24 means raising `--wa-control-size-s`, which carries an explicit "reads as editor chrome" rationale and resizes every compact toolbar in the extension — left as a design decision, not shipped silently.

## Verification

- `npm run lint` — clean. `prettier --check` on every changed file — clean.
- `npx vitest run` — **820 files / 8,421 tests passing**.
- Every ratio above computed in Node against the surface each element actually sits on, compositing `color-mix` alpha and element `opacity` before measuring.

**Not verified:** nothing rendered. These are contrast and token changes derived from the cascade; the status dots and the model list are the most visible and worth a look.

---

> [!NOTE]
> **Low Risk**
> CSS-only visual and contrast adjustments with no logic, auth, or data-path changes; main risk is subtle appearance shifts worth a quick visual pass on status dots and the model list.
> 
> **Overview**
> Fixes **accessibility contrast failures** caused by layering element `opacity` on top of tokens like `--wa-color-text-quiet` that already use `color-mix` alpha. The effective contrast dropped below 3:1 (non-text) or 4.5:1 (small text) across ~14 rules.
> 
> **Core pattern:** remove redundant `opacity`, and express de-emphasis once via **`--color-text-muted`** (from the stacked token PR) or existing quiet text colors. **Status dots** no longer fade the base dot; quiet states use `--color-text-muted`, **`is-starting`** uses **`--color-pending`**, and redundant `opacity: 1` on saturated states is gone. **Disabled / requires-key `wa-option`** rows drop stacked opacity for `--color-text-muted` + italic.
> 
> **Targeted UX tweaks:** unavailable **model rows** lose row-level dim (it also washed metadata/reasoning controls outside the switch); key/warning icons still signal unavailability. **Login banner** fineprint drops xs size + fade and inherits body size. **Onboarding** choice copy and skip button use quiet color instead of opacity fade. **Desktop split handles** lose extra opacity. **File select** bumps folder label to `font-size-sm` and widens icon-button gaps from 3xs to 2xs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b780d1c80d3be3cead485e8f2a5856b0f0948db0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/renderer/taskShell.css
+++ b/packages/desktop/src/renderer/taskShell.css
@@ -99,7 +99,6 @@
   color: var(--wa-color-text-quiet);
   font-size: var(--font-size-xs);
   line-height: 1;
-  opacity: 0.45;
   pointer-events: none;
   transform: rotate(90deg);
 }
@@ -112,7 +111,6 @@
   color: var(--wa-color-text-quiet);
   font-size: var(--font-size-xs);
   line-height: 1;
-  opacity: 0.45;
   pointer-events: none;
 }
 
@@ -282,7 +280,6 @@ body[data-desktop-platform='darwin'] .task-sidebar-brand {
   color: var(--wa-color-text-quiet);
   font-size: var(--font-size-xs);
   line-height: 1;
-  opacity: 0.45;
   pointer-events: none;
 }
 
--- a/packages/extension/src/settingsView/frontend/components/profile/ModelSelectionList.styles.ts
+++ b/packages/extension/src/settingsView/frontend/components/profile/ModelSelectionList.styles.ts
@@ -151,9 +151,12 @@ export const modelSelectionListStyles: CSSResult = css`
   }
 
   /* Unavailable model rows (not in relay allowlist) */
-  .model-row--unavailable {
-    opacity: var(--opacity-disabled);
-  }
+  /* No row-level dim. At 0.5 on top of an already-translucent secondary text
+     token this composited to 1.91:1, and it also dimmed .model-metadata and
+     .reasoning-level-select, which render OUTSIDE the wa-switch and so never
+     qualified for the disabled-control contrast exemption in the first place.
+     Unavailability is already carried by the key / warning icon and its title
+     on the row, which is the cue that survives a contrast check. */
 
   /**
    * Inline icon following a model row's name. Variants set --_icon-color;
--- a/packages/extension/src/settingsView/frontend/components/profile/ModelSelectionList.ts
+++ b/packages/extension/src/settingsView/frontend/components/profile/ModelSelectionList.ts
@@ -212,10 +212,9 @@ export class ModelSelectionList extends LitElement {
 
   private renderModelRow(model: ModelSelectionItem): TemplateResult {
     const available = !model.disabled;
-    const unavailableClass = !available ? ' model-row--unavailable' : '';
 
     return html`
-      <div class="model-row${unavailableClass}">
+      <div class="model-row">
         <wa-switch
           ?checked=${model.enabled}
           ?disabled=${!available && !model.enabled}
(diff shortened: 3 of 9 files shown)


LionSR:
All three addressed in 1d07db18bf.

**`.is-starting`** — correct, and I had missed it. It was the one quiet state at full opacity, and collapsing it onto `--color-text-muted` made five semantically different states render as one dot. It now uses `--color-pending`, the accent already carrying in-progress elsewhere, so initializing reads apart from settled.

**`.skip-row` button labels** — also correct, and this one was a regression I introduced. `--color-text-muted` is calibrated to the 3:1 non-text floor; these are interactive labels needing 4.5:1, so the swap lowered their contrast. Removing the stacked opacity was the fix — recolouring was a step too far. The colour override is gone and the buttons use their own foreground.

**`.model-row--unavailable`** — right, the rule was deleted but the class was still applied. The class computation is removed too; there is no inert hook left.

**On `.banner-lead`**: fair point that it wasn't in the violation table. It is deliberate but under-stated — the rule was `opacity` on inherited text with no colour of its own, i.e. the same stacking pattern, just one that happened to still clear threshold. I have left it removed for consistency with the rest of the pass rather than keeping one instance of the pattern the PR exists to remove.

LionSR:
Both open review threads are addressed in `f2979ee` (`fix(ui): address review on the opacity pass`):

- `.is-starting` uses `--color-pending` so it stays distinct from the muted completed/cancelled/ready fill.
- The dead `model-row--unavailable` class computation is gone from `ModelSelectionList.ts`.

LionSR:
Addressed the onboarding choice-description thread: descriptions now use `--wa-color-text-quiet` like `.card-copy` / `.path-step__copy`.

PR #9718, opened 2026-08-04
fix(ui): name every control, expose its state, and restore the shared focus ring

LionSR:
Final PR of the cross-surface interface review. Stacked on #9715 → #9714 → #9711; retarget as they merge.

## Names that never reached the control

WebAwesome puts the role-bearing element **inside its shadow root**:

```
wa-switch → <label part="base"><input role="switch"> … </label>
wa-input  → <label for="input">…</label>  +  <input id="input">
```

So an `aria-label` on the host names the *custom element*, not the control that carries the role — and a sibling `<label>` with no `for` names nothing at all.

The fix hinges on a fact worth checking rather than assuming: **`WebAwesomeFormAssociatedElement.formAssociated = true`**, so these elements are labelable and a plain light-DOM `for`/`id` pair associates properly — with zero layout change.

- **`renderSettingsToggleRow`** builds **21 toggle rows across 8 settings files**. Its `<span class="settings-row-label">` is now a real `<label for>`. Those switches were reaching the accessibility tree **unnamed**. The label text also becomes part of the switch's hit target, which removes the dead zone between a toggle and the words describing it.
- Git author name/email and the provider custom-endpoint field get `for`/`id` pairs against their existing visible labels.
- The history search field and the desktop credential prompt were named only by a placeholder or a sibling subtitle. Both now use WebAwesome's own `label` API. In the desktop prompt the text **moves** into the label rather than being duplicated, so it isn't announced twice.
- The launcher composer, named only by a rotating placeholder, gets a stable name.

## State that existed only visually

The record toggle flipped its icon, tooltip and colour but not its name or pressed state — nothing about the change reached a screen reader. It now reports `aria-pressed` and renames to "Stop recording".

## One static name per list

"Delete" repeated down a list says nothing about which row it destroys. History, memory and provider-key actions now carry their row's subject, appended at the map site so the shared descriptor tables stay untouched.

## Focus rings, continuing #9711

`commonViewStyles` composes `focusRingStyles` — and then overrode it on `.list-item`, using `:focus-within`, which also fires on a **mouse click**. `AgentSelectionPanel` killed the ring outright with `outline: none` and substituted a 1px one; it now uses the shared `.focus-ring-inset` class built for exactly that case (a ring inside an `overflow: hidden` scroller). `MultiAgentTab`'s preset card defers to the shared ring; its delete button keeps a part-piercing rule at the shared token width.

**Two sites that looked like the same defect are deliberately untouched** — `mark[data-current]` in `historyStyles` and `wa-button.has-options` in `fileSelectStyles` are *state cues*, not focus indicators. Changing them would have been a false positive.

## Also

One canonical `.visually-hidden` in `commonViewStyles`, replacing a local copy in `InstructionPanel`. A differently-named `.sr-only` remains in `WorktreeChip`; it's in progressView, so it's left for that surface's follow-up rather than widened here.

## Tests

Two files pinned the old behaviour and are updated rather than worked around. `ToolsTab` selected switches by `wa-switch[aria-label="…"]` — the very attribute that never named anything — and now asserts the `<label for>` association that does. `MemoryItemActionGroup` expects the per-row subject.

## Verification

- `npm run lint` — clean. `prettier --check` on every changed file — clean.
- `npm run check:dead-code-ratchet` — 18 vs 18 baselined.
- `npx vitest run` — **820 files / 8,421 tests passing**.
- WebAwesome internals read from `dist/chunks/` to confirm where the role and the `for` actually live, and that `formAssociated` is set.

**Not verified:** nothing rendered. The settings toggle rows and the desktop credential prompt are the two places where the DOM changed shape, so they're worth a look.

---

> [!NOTE]
> **Low Risk**
> UI-only accessibility and focus-style changes with no auth, data, or backend logic; tests lock in the new labeling behavior.
> 
> **Overview**
> Improves **accessibility naming and keyboard focus** across settings, history, memory, provider keys, Git, the launcher composer, and the desktop credential prompt.
> 
> **Controls that were unnamed or misnamed** now use real `label`/`for` (or WebAwesome `label`) so screen readers hit the actual input inside shadow DOM—not a host `aria-label`. Shared **`renderSettingsToggleRow`** switches from span + `aria-label` to `<label for>` + stable ids for all settings toggles. History search, Git author fields, provider custom endpoint, instruction textarea, and the desktop prompt get proper labels; the prompt moves prompt text into `wa-input.label` and hides the empty subtitle to avoid double announcement.
> 
> **List actions** append row context to `aria-label` (history item title, memory filename, provider name) so repeated “Delete”/“Set” buttons are distinguishable. The **record** button toggles label (“Record” / “Stop recording”) and **`aria-pressed`** while recording.
> 
> **Focus styling** drops ad-hoc outlines and a `.list-item:focus-within` ring that fired on click; agent list rows use shared **`.focus-ring-inset`**, and multi-agent preset delete uses shared focus tokens. **`.visually-hidden`** is centralized in `commonViewStyles` (removed duplicate from InstructionPanel).
> 
> Tests updated to assert label association and per-row action labels instead of `wa-switch[aria-label]`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ae2cb531cee5962d43eb5883c080261ffde75ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/renderer/promptOverlay.ts
+++ b/packages/desktop/src/renderer/promptOverlay.ts
@@ -96,7 +96,12 @@ export function createDesktopPromptOverlay(
     // label. As a subtitle it named nothing — wa-input puts its control inside
     // a shadow root, so only WebAwesome's `label` produces a real <label for>.
     // Kept in one place rather than two so the text is not read out twice.
-    if (shell.subtitleEl) shell.subtitleEl.textContent = '';
+    // Hide the empty subtitle so its always-on top margin does not leave a gap
+    // between the title and the input.
+    if (shell.subtitleEl) {
+      shell.subtitleEl.textContent = '';
+      shell.subtitleEl.hidden = true;
+    }
     input.type = message.password ? 'password' : 'text';
     input.label = message.prompt;
     input.value = '';
--- a/packages/desktop/src/renderer/promptOverlay.ts
+++ b/packages/desktop/src/renderer/promptOverlay.ts
@@ -92,8 +92,13 @@ export function createDesktopPromptOverlay(
     if (current) sendSettlement(current, null);
     current = message;
     if (shell.titleEl) shell.titleEl.textContent = message.title;
-    if (shell.subtitleEl) shell.subtitleEl.textContent = message.prompt;
+    // The prompt text moves from the sibling subtitle into the field's own
+    // label. As a subtitle it named nothing — wa-input puts its control inside
+    // a shadow root, so only WebAwesome's `label` produces a real <label for>.
+    // Kept in one place rather than two so the text is not read out twice.
+    if (shell.subtitleEl) shell.subtitleEl.textContent = '';
     input.type = message.password ? 'password' : 'text';
+    input.label = message.prompt;
     input.value = '';
     dialog.setAttribute('aria-label', message.title);
     dialog.open = true;
--- a/packages/extension/src/settingsView/frontend/components/history/HistoryItemElement.ts
+++ b/packages/extension/src/settingsView/frontend/components/history/HistoryItemElement.ts
@@ -420,7 +420,12 @@ export class HistoryItemElement extends LitElement {
                 (!entry.toolUseOnly || presentation.isToolUse) &&
                 (!entry.hideWhenUnsupported ||
                   !isKnownUnsupported(this.unsupportedCommands, entry.command)),
-            ).map((entry) => renderIconActionButton(entry.button))}
+            ).map((entry) =>
+              renderIconActionButton({
+                ...entry.button,
+                label: `${entry.button.label}: ${presentation.title}`,
+              }),
+            )}
           </div>
         </div>
         ${this.renderInstructionBlock(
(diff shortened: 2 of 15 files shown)


LionSR:
Both addressed in 8504f9cb23.

**The composer finding is the sharper of the two, and it caught me contradicting this PR's own premise.** I set `aria-label` on the `wa-textarea` host — exactly the pattern the rest of the PR removes. WebAwesome keeps the role-bearing control in its shadow root, so that names the custom element and nothing else; the composer would have stayed unnamed.

**The search field** used WebAwesome's label slot, which does name the control — but filling that slot activates the component's has-label chrome, and hiding only the slotted span does not collapse the shadow `form-control-label` part, so the row could gain vertical space and fall out of alignment.

Both now use a visually-hidden light-DOM `<label for>`. `wa-textarea` and `wa-input` share the form-associated base (`WebAwesomeFormAssociatedElement.formAssociated = true`), so the association is real, and nothing is added inside either shadow root — no label chrome, no layout risk. That also makes them consistent with the 21 settings toggle rows, which use the same mechanism.

LionSR:
Addressed both review threads: memory action labels now use `displayPath`, and the desktop prompt subtitle is `hidden` when empty so its margin does not create a gap.

LionSR:
Addressed the history-action naming thread: Delete/Setup/Rerun/Export labels now truncate `presentation.title` with `truncateWithEllipsis(..., LONG_INSTRUCTION_CHARS)` so screen readers are not announced multi-thousand-character instructions.

PR #9723, opened 2026-08-04
fix(cli): warn once when an approval gate is denied

### LINKED ISSUE #9721: Log a warning on policy denial so nonzero exit has a visible cause
## Context

Follow-up from #9720 (fix(cli): a denied approval gate is not a failed run).

When a run exits with code 4 (`CliExitCode.ApprovalDenied`), stderr is empty — the denial writes nothing to the terminal. The nonzero exit arrives with no stated cause, which makes diagnosing "why did the run report failure?" unnecessarily time-consuming.

## Details

A one-line `warn` log at the point where a policy denial is marked (in `hasCliApprovalDenied` or at the decision site in the approval adapter) would have made the root cause of the regression obvious immediately.

The fix should:
1. Emit a diagnostic message (to stderr or via the CLI logging channel) when a gate is denied under any policy.
2. Keep it at `warn` level — it's diagnostic, not an error.
3. Ensure the message identifies which gate was denied and under which policy.

## References

- PR: #9720
- Relevant section in PR body: "Not covered here" — "The denial is invisible. It writes nothing to stderr, so the nonzero exit arrived with no stated cause — which is most of why this took a while to pin down."
- Files: `packages/cli/src/runtime/terminalStatus.ts`, `packages/cli/src/runtime/exitCodes.ts`

LionSR:
## Summary
- Closes #9721 (follow-up from #9720): first `markApprovalDenied` now emits a `warn` on the `cli-approval` channel naming the gate and policy.
- Repeat marks on the same context stay silent.
- Focused unit coverage in `ApprovalDeniedWarn.vitest.mts`.

## Test plan
- [x] `vitest` `ApprovalDeniedWarn.vitest.mts`
- [ ] Manual: `texra run … --approval-policy never` against a tool that needs approval and fails the run → stderr shows `[warn] [cli-approval] … denied under policy "never".`

### DIFF
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -373,7 +373,8 @@ export function createChatSessionController(
             registerExecution: true,
             enforceCategory: true,
             approvalPromptsUnavailable: approvalsUnavailable,
-            onApprovalPolicyDenial: () => markApprovalDenied(sessionContext),
+            onApprovalPolicyDenial: () =>
+              markApprovalDenied(sessionContext, 'Tool or edit approval'),
             runtimeUnavailableTools: CLI_UNAVAILABLE_TOOLS,
             onStreamResolved: (resolvedStreamId) => {
               // Each chat round mints a fresh root StreamTabId (new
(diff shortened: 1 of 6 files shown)


LionSR:
Approval fail should not fail CLi

Just tell the LLM say it is rejected

LionSR:
Addressed the log-sink thread: `markApprovalDenied` now writes `[warn] [cli-approval] …` via `writeTextStderr`, matching other operator-facing CLI warnings.

PR #9717, opened 2026-08-04
refactor(approval): unify TeXRA policy across hosts (#9597)

LionSR:
## Summary
- Shared `decideTexraApproval` distinguishes `deny-policy` / `deny-unpresentable`; retry and human-input decisions live in `@shared/approvalPolicy`.
- Delete the CLI decision mirror; survivors are prompt-only in `approvalPrompts.ts`. Adapters call shared `decide*` via `settleApprovals.ts`.
- Extension and desktop seed (and re-seed on settings write / workspace transition) from `texra.approvalPolicy`; Tools tab control and status-bar policy line (refreshed on re-seed).
- Drop dead `SUPER_YOLO_ENABLED`; shared `approvalBypassKind`; rename internal reliability handlers (wire literals unchanged). Authority ratchet in `approvalPolicyAuthorityRatchet.vitest.ts`.
- No old-format persistence migration — missing/invalid config uses the default.

Closes #9597 (finishes PRD stages A–E; prior stages already on main).
#9721 closed via #9723 (merged); this branch carries the same fix forward through the approvalPrompts.ts rename — rebase reconciles the two.

## Issue acceptance gates
- #9597 PRD stages A–E: honest deny, CLI mirror deletion, host seed/settings, vocabulary hygiene, authority ratchet.

## Single source of truth
- Policy vocabulary + `decide*` evaluators: `src/shared/approvalPolicy.ts`
- Live session value: `SessionHandle.approvalPolicy` / `setApprovalPolicy`
- Persisted spelling: `texra.approvalPolicy` in `.texra/config.json`

## Duplication and abstraction budget
- Removed CLI `approvalPolicy.ts` decision mirror and duplicated headless/TUI settle helpers (consolidated into `settleApprovals.ts`).
- No policy service / observer registry; host seeds remain `read → setApprovalPolicy`.

## Net elements (R6)
- Files (vs `main`): +4 added (`settleApprovals.ts`, `approvalPolicyTooltipRefresh.ts`, `approvalBypassKind.ts`, authority ratchet test), −1 deleted (`humanInputHandlers.ts`), 1 rename (`approvalPolicy.ts` → `approvalPrompts.ts`); ~54 paths touched.
- Exported symbols (`^[+-]export` over the diff): roughly +27 / −13 — growth is the shared decide/message/read API and settle helpers; deleted exports are the old CLI mirror (`immediateDecision*`, denial helpers, `ApprovalInstructionContext`, etc.).
[...]

PR #9737, opened 2026-08-04
fix(cli): retire the approval-denied exit code

LionSR:
Follow-up to the review comment on #9723: *"Approval fail should not fail CLI — just tell the LLM it is rejected."*

## Why

A denied approval gate is not a run outcome. The gate already hands the denial back to the model as tool feedback (`rejectedResult` → `userInstruction`), the model routes around it, and the turn continues. Giving denial its own process exit code meant any caller that reads a nonzero exit as "produced no result" threw away runs that had in fact succeeded — TeXRA's own PR review workflow among them.

#9720 narrowed the damage by nesting the denial check inside `FAILED`. This removes the concept instead of narrowing it.

## What changed

- Deleted `CliExitCode.ApprovalDenied` (4), the `WeakSet<CliContext>` denial marker, and `hasCliApprovalDenied`.
- `runOutcomeExitCode` no longer takes a `CliContext`; a run's exit code now depends on its outcome alone.
- `markApprovalDenied` becomes `warnApprovalDenied`: same one-line `[warn] [cli-approval] …` on stderr, warn-once per run, but with no effect on exit status. Operators keep the diagnostic that #9723 added.
- Removed the call sites that only ever existed to set exit-code state — interactive user rejections, cleared/replaced TUI entries, and prompt failures — rather than converting them to warns. Those are user decisions or transport failures, not policy denials.
- Updated the CLI exit-code contract table and the approval-policy PRD to record the retirement and its rationale.

## Verification

- `npm run typecheck` — green
- `npm run lint` — green
- `npm run check:dead-code-ratchet` — 18 findings, same as baseline
- `ApprovalDeniedWarn`, `ApprovalAdapter`, `TuiApprovalRetry`, `RunExecution` — 111 tests pass

Tests that pinned the retired exit code were deleted rather than rewritten around the new behavior; the surviving assertions cover the denial *reason* delivered to the model, which is the part that still matters.

Net: 127 insertions, 170 deletions across 15 files.

### DIFF
--- a/docs/prds/2026-05-04-prd-cli-app.md
+++ b/docs/prds/2026-05-04-prd-cli-app.md
@@ -952,7 +952,7 @@ Each phase is independently reviewable. The extension and the desktop port never
 - Plan / proposal / retry / external-inquiry handlers via `ClackPromptHost` (C3).
 - `--allowed-tools` / `--disallowed-tools` flag wiring through `resolveTools()`.
 - `texra resume [<run-id>]` for tool-use snapshots.
-- **Exit criteria:** `texra run orchestrator --instruction "..." --approval-policy yolo` runs an end-to-end multi-tool flow against a real LaTeX project. `--approval-policy never` correctly fails when an approval is needed (exit code 4). Approvals on TTY render diffs and respect user input.
+- **Exit criteria:** `texra run orchestrator --instruction "..." --approval-policy yolo` runs an end-to-end multi-tool flow against a real LaTeX project. `--approval-policy never` denies the gate and returns the denial to the model as feedback (exit code 4 was retired in 2026-08). Approvals on TTY render diffs and respect user input.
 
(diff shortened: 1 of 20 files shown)


LionSR:
Rebased onto current `main` (post-#9717) and re-applied the retirement against the unified approval shape:

- Deleted `CliExitCode.ApprovalDenied`, `CliContext.approvalDenied`, and `hasCliApprovalDenied`
- `markApprovalDenied` → `warnApprovalDenied` (stderr only; live `SessionHandle` policy)
- Interactive rejection sites removed rather than converted to warns
- Tests updated; 112 pass in the four affected suites

The earlier CI failure on the host-agent import ratchet was a stale-base artifact (main already carries `@agent/runtime/helperModelName`). The TeXRA `review` job exiting 4 was the exact bug this PR retires.

PR #9742, opened 2026-08-05
refactor(tools): share SDK-export and binary-cache helpers across claude/codex

LionSR:
## Summary

The Claude Code and Codex CLI import modules (`src/tools/claudeAgentImport.ts`, `src/tools/codexImport.ts`) each hand-rolled two identical shapes:

1. **SDK export resolution** — walking the three ESM/CJS interop shapes esbuild can produce (`mod.X` → `mod.default.X` → `mod.default`), the `typeof … !== 'function'` guard, and the "not found … NOT in esbuild externals" error.
2. **The per-session binary cache** — the `if (cached !== undefined) return cached; … if (result) cached = result` wrapper around `resolveBinary()`.

Both are now extracted into `src/tools/support/externalBinaryUtils.ts` as `resolveSdkExport()` and `createCachedBinaryResolver()`. The two tool files shrink to their SDK-specific config plus the inline `await import()`.

One deliberate non-change: the `await import('<specifier>')` that loads each SDK **stays inline** in each tool file. esbuild only rewrites dynamic `import()` to `require()` when the specifier is a static string literal, so hoisting the import itself into a shared helper would silently break the packaged Electron/VSIX build. Only the post-import logic is shared; this constraint is documented in the module header.

## Issue acceptance gates

No issue — this is a self-contained slice of an internal tech-debt audit. Acceptance: no behavioral change, all existing tests pass, no new lint/typecheck/dead-code findings.

## Single source of truth

`src/tools/support/externalBinaryUtils.ts` now owns both the SDK export-interop resolution (`resolveSdkExport`) and the binary-resolution cache lifecycle (`createCachedBinaryResolver`) for every CLI vendor tool. The error text and cache semantics are defined once.

Removed: two copies of the ESM/CJS export-resolution + validation block, and two copies of the resolve-and-cache wrapper. No pass-through layer added — both new helpers carry real logic (interop resolution/validation; cache state + short-circuit). `resolveBinary()` was de-exported because its only remaining caller is `createCachedBinaryResolver()` in the same module.

## Net elements (R6)

- Files: 3 changed, 0 added, 0 deleted.
- Exported symbols: **+1 net** — added `resolveSdkExport`, `createCachedBinaryResolver`; removed the `export` on `resolveBinary` (symbol kept, now internal). `findClaudeBinaryPath`/`findCodexBinaryPath` stay exported (converted from `async function` to `const`, same call signature).
- Net LoC: `+116 / −69` (+47). The raw count rises because the two shared helpers carry doc comments explaining the esbuild-literal-import and cache-retry invariants; the two tool files are net-negative. The win here is drift elimination, not line count.

## Consumer counts (R8)

- `resolveBinary` re-routed from exported to internal: previously imported by 2 modules (`claudeAgentImport`, `codexImport`); now **0 external importers, 1 internal caller** (`createCachedBinaryResolver`). Verified by grep.
- New exports `resolveSdkExport` and `createCachedBinaryResolver`: 2 consumers each (both tool files).

## Host impact

These are host-agnostic modules under `src/tools/`, consumed identically by all hosts through the tool layer.

Extension: no behavioral change.

Desktop: no behavioral change.

CLI: no behavioral change.

## Scheduled refactoring

None. Larger audit items (cross-host progress-view wiring dedup, the two parallel transcript persistence engines, model-handler template-method extraction) are tracked separately and intentionally out of scope for this low-risk slice.

## Validation

- `npm run typecheck:workspace` — clean.
- `npx vitest run` on the four affected suites (`codexImport`, `CodexResumeFallback`, `ClaudeAgentResumeFallback`, `ClaudeAgentProgressEvents`) — 17/17 pass.
- `npx eslint` on the three changed files — clean.
- `npx prettier --check` on the three changed files — clean.
- `npm run check:dead-code-ratchet` — OK, back to baseline (18=18) after de-exporting `resolveBinary`.

https://claude.ai/code/session_011BF7vgmqKoEGgQZJs2SZUk

---
_Generated by [Claude Code](https://claude.ai/code/session_011BF7vgmqKoEGgQZJs2SZUk)_

### DIFF
--- a/config/ratchets/host-agent-import-baseline.json
+++ b/config/ratchets/host-agent-import-baseline.json
@@ -18,6 +18,7 @@
       "@agent/runtime/detachSubagentsOnStop",
       "@agent/runtime/executeAgent",
       "@agent/runtime/executionRegistry",
+      "@agent/runtime/helperModelName",
       "@agent/runtime/HostInteractions",
       "@agent/runtime/resolveAndResumeStream",
       "@agent/runtime/resumeQueuedToolUse",
--- a/src/tools/claudeAgentImport.ts
+++ b/src/tools/claudeAgentImport.ts
@@ -20,7 +20,10 @@ import * as path from 'node:path';
 import { isModuleNotFoundError } from '@common/errors';
 import { AbsoluteFS } from '@utils/files';
 import { IS_WINDOWS } from '@utils/system/platformPaths';
-import { resolveBinary } from './support/externalBinaryUtils';
+import {
+  createCachedBinaryResolver,
+  resolveSdkExport,
+} from './support/externalBinaryUtils';
 // Mirror the native `query` signature exactly (no hand-rolled structural copy).
 type QueryFn = typeof import('@anthropic-ai/claude-agent-sdk').query;
 
@@ -52,20 +55,11 @@ export async function importClaudeAgentSdk(): Promise<QueryFn> {
     throw err;
   }
 
-  const queryFn =
-    mod.query ??
-    (mod.default as Record<string, unknown> | undefined)?.query ??
-    mod.default;
-
-  if (typeof queryFn !== 'function') {
-    const keys = Object.keys(mod).join(', ');
-    throw new Error(
-      `query() not found in @anthropic-ai/claude-agent-sdk. Module keys: [${keys}]. ` +
-        'Ensure @anthropic-ai/claude-agent-sdk is NOT in esbuild externals.',
-    );
-  }
-
-  cachedQuery = queryFn as QueryFn;
+  cachedQuery = resolveSdkExport<QueryFn>(mod, {
+    exportName: 'query',
+    specifier: '@anthropic-ai/claude-agent-sdk',
+    errorLabel: 'query()',
+  });
   return cachedQuery;
 }
 
@@ -96,8 +90,13 @@ const PLATFORM_PACKAGES: Record<string, readonly string[]> = {
 /** Native CLI binary filename for the current platform. */
 const CLAUDE_BINARY_NAME = IS_WINDOWS ? 'claude.exe' : 'claude';
 
-/** Cached result — found paths are cached; misses are retried. */
-let cachedBinaryPath: string | undefined;
+/** The platform binary sits directly in the platform-package directory. */
+async function claudeBinaryInPlatformPackage(
+  platformPkgDir: string,
+): Promise<string | undefined> {
+  const binary = path.join(platformPkgDir, CLAUDE_BINARY_NAME);
+  return (await AbsoluteFS.exists(binary)) ? binary : undefined;
+}
 
 /**
  * Locate the native Claude Code CLI binary. Results are cached for the session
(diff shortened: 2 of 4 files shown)


LionSR:
The `review` check (run [30988999044](https://github.com/LionSR/TeXRA/actions/runs/30988999044/job/92250390175)) is red because the DeepSeek-backed review agent (`deepseekproT`) exited 4 **after** the agent ran to completion — the "Normalize review" and "Post review" steps were skipped, so it failed to emit a postable review. That's a reviewer-tooling/provider failure, not a finding against this diff, and there's no code change that resolves it.

Corroboration that it's independent of the change:
- The two other LLM reviewers on this same commit both passed — `claude-review-anthropic` and `claude-review-deepseek` — and the Anthropic reviewer posted a clean, no-findings review.
- All build/test gating checks on this commit are green (`build artifacts (linux)`, `webview smoke (macOS)`, `CLI workflow input paths (Windows)`, `guidance references`, `detect changes`); `kernel tests` and `static checks` were still running at time of writing (typecheck, the four affected vitest suites, eslint, prettier, and the dead-code ratchet all pass locally).

Recommend re-running the `review` job; it should clear on retry. Flagging rather than "fixing" since the failure is in the reviewer runner, not the code.

---

LionSR:
`kernel tests (linux, shard 1/2)` and `validate` are red, but **on the base branch too** — not from this diff.

Root cause: the host-agent deep-import ratchet (`hostAgentDeepImportRatchet.vitest.ts`) fails with

```
cli @agent/* deep-import specifiers grew from 31 to 32:
  + @agent/runtime/helperModelName
```

That import lives in `packages/cli/src/runtime/enabledModels.ts`, added by **#9724** ("enable model catalog from TUI", already merged to `main`), and the `cli` entry in `config/ratchets/host-agent-import-baseline.json` was never updated to match. This PR changes only three files under `src/tools/` and never touches the CLI or that module, so the CLI import set — and therefore this ratchet result — is identical to `main`'s. `main` is currently red on this check, and every open PR inherits it. `validate` is just the aggregation gate; its log shows `test result: failure` and no other failure, so it's red purely because `kernel tests` is.

I'm not folding a `cli`-baseline change into this `src/tools` refactor: widening the deep-import baseline is the exact thing the ratchet is meant to gate ("never widen a baseline"), so whether to baseline `@agent/runtime/helperModelName` or drop the import belongs in a CLI-scoped fix a reviewer signs off on, not buried here. This PR's own gating checks are otherwise green (build, webview smoke, CLI workflow paths on Windows, guidance references) and will go green on this ratchet once `main`'s baseline is fixed and merged in.

(The separate `review` check is an unrelated LLM-reviewer tooling failure — see the earlier comment.)

---

LionSR:
Pushed `901642d` to fix the base-branch ratchet failure (maintainer asked me to take it here rather than a separate PR).

The fix records `@agent/runtime/helperModelName` in the `cli` list of `config/ratchets/host-agent-import-baseline.json` — the import #9724 added to `packages/cli/src/runtime/enabledModels.ts`. That specifier is already a sanctioned deep-import for the `desktop` and `extension` hosts, so this restores the "never widen a baseline" invariant without expanding the overall frozen `@agent` surface. The ratchet test (`hostAgentDeepImportRatchet.vitest.ts`, 4/4) passes locally with the change.

This unblocks `kernel tests` + `validate` on this PR and clears the same failure for every other open PR once merged. The `review` check remains a separate LLM-reviewer tooling failure (re-run should clear it).

---

PR #9766, opened 2026-08-05
fix(approval): make the approval flow readable and correctly targeted

LionSR:
An interface pass over the progress-view approval flow — the surface every agent run blocks on. Four fixes, all of them things a user sees.

## The armed panel was invisible

`y`/`n` act on the **newest** request, while sections render in a **fixed kind order**. With mixed kinds pending, the top panel on screen is not the one a keypress hits — and the action it triggers can be executing a shell command. The dispatch code already acknowledged the divergence in a comment; nothing surfaced it to the user. The armed panel now carries `data-armed` and a visible outline.

## Rejecting wore the approval glyph

The confirm step of a **rejection** rendered as a checkmark labelled "Submit". Wrong icon, and a label that never named the consequence. It keeps the reject icon and reads "Send rejection".

## The working directory was the least readable thing in the prompt

The field that decides *where a command runs* used `--font-size-xs`, which resolves to **10.4px** (`13px × 0.8`), at a colour measuring **4.40:1 on Light Modern** — under AA. It moves to `--font-size-sm` and the path takes the normal text colour, so the value outranks its label. The `span` previously resolved to *exactly* its parent's colour, so the markup's label/value split did nothing at all.

## The rejection field lost its description on the first keystroke

It had only a placeholder. Now it has a visible label, and the placeholder carries an example instead.

## Scope

The section title is a heading element rather than a styled span — the right element for a section title, one tag plus a three-line margin reset.

**This branch adds no ARIA.** An earlier revision included a live region and `aria-labelledby` wiring; those were removed as out of scope for this product.

## Considered but rejected

| Location | Candidate | Rejected because |
| --- | --- | --- |
| `requestPanelSharedStyles.ts` details/actions | Cap details height, pin the actions row | Already correct: `max-height: min(34vh, 26rem)` with `overflow-y: auto`, actions `position: sticky` |
| `StreamTabs.ts`, `FileSelectGroup.ts` | `@click` on a `<div>` | Event delegation over real `<button>`s inside `stream-tab` |
| `BaseApprovalPanel.ts` | Require a modifier for single-key `y` | Intentional design, advertised in every button title; the defect was mis-targeting, not the accelerator |
| Panel motion | Add reduced-motion handling | Already covered in `litStyles`, `controlStyles`, and `common.css` |

## Verification

- `npm run typecheck` and `npm run lint` clean.
- 88 suites / 698 tests pass, re-run after each revision.
- Contrast measured against each stock theme's own `editorWidget.background`.
- The armed-panel selector was corrected during review: `:is(${ITEMS})[data-armed]` would never have matched, because `data-armed` lands on the panel host in `RequestPanels`' shadow root while `ITEMS` is a class inside each panel's own root. It targets `[data-request-panel][data-armed]`.

**Not verified:** not rendered. The armed outline and the new visible label have not been eyeballed. The outline is worth checking with two pending requests of different kinds, since that is the exact case it exists for.

### DIFF
--- a/packages/extension/src/progressView/frontend/components/BaseFeedbackPanel.ts
+++ b/packages/extension/src/progressView/frontend/components/BaseFeedbackPanel.ts
@@ -72,10 +72,13 @@ export abstract class BaseFeedbackPanel<
   }
 
   protected renderRejectButton(rejectTitle: string): TemplateResult {
+    // Both states keep the reject icon and name the consequence. Switching to
+    // a checkmark labelled "Submit" made the confirm step of a rejection wear
+    // the approval glyph, which is the one misreading this flow cannot afford.
     return renderLabeledActionButton({
-      icon: this.showFeedback ? 'check' : 'xmark',
-      text: this.showFeedback ? 'Submit' : 'Reject',
-      title: this.showFeedback ? 'Submit rejection (n)' : rejectTitle,
+      icon: 'xmark',
+      text: this.showFeedback ? 'Send rejection' : 'Reject',
+      title: this.showFeedback ? 'Send rejection (n)' : rejectTitle,
       action: 'reject',
       disabled: this.archived,
       onClick: () => this.handleRejectAction(),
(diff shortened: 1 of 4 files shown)


LionSR:
Cross-reference: this is complementary to #9771, not competing.

#9771 fixes the **layout** cause — `request-panels` was mounted inside `.conversation-prelude`, height-capped and shared with todos/plans, so the whole card could sit below the fold where a sticky action row cannot recover it. That is the root cause of the original report.

This PR fixes what remains once the panel is on screen: which panel the `y`/`n` accelerators actually target, the reject step wearing the approval checkmark, and the working directory's size and contrast.

Verified the two branches **auto-merge cleanly** (`git merge --no-commit` on `requestPanelSharedStyles.ts`, no conflicts). One interaction was found and handled here: the armed-panel ring is now drawn inset, because #9771's approval dock sets `overflow-x: hidden` and would have shaved the left and right edges off an outward ring. The inset version is correct under either layout, so merge order does not matter.

PR #9775, opened 2026-08-05
refactor(cli): drop unreachable transcript render paths, fix /config enum budget

LionSR:
Three unreachable paths in the CLI transcript, plus one budget constant that was wrong because it was written twice.

## `ToolUseRow`: three props no caller passes

`omitHeader`, `neutralStatus`, and `showOutput` are declared, threaded into `toolUseStyledLines`, and branched on — by nobody. All three `<ToolUseRow>` call sites (`ConversationPane.tsx`, `TranscriptEntry.tsx` ×2) pass only `maxRows`/`toolUse`/`width`/`subagentExecutionLabels`, no test constructs the component, and no caller of `toolUseStyledLines` or `toolUseDisplayLines` sets `showOutput` or `neutralStatus`. So `allLines.slice(1)`, the neutral status color, the `showOutput` override, and two constant segments of the memo cache key were all dead.

Removing them also narrows that cache key from five segments to three, so entries no longer vary on values that never differ.

One subtlety worth flagging for review: `toolRenderers.tsx` has **two** `showOutput`s. `options.showOutput` is the dead prop; `opts.showOutput` is live per-tool config from `toolRowOptions` that decides whether e.g. a bash row shows output at all. Only the former is gone — `const showOutput = options.showOutput === true || opts.showOutput` becomes `const showOutput = opts.showOutput`.

## `TodosPlanPanel`: an unreachable branch that said so

The panel had a bounded branch and an uncapped one. The uncapped branch's own comment read *"Reached only when maxRows is undefined (the compact branch above owns every bounded case)"* — and the only production caller is `ConversationRegion.tsx:289`, `<TodosPlanPanel maxRows={todosPlanRows} />`, where `allocateConversationBottomPanelRows` declares `todosPlanRows: number`. It was never undefined. Tests import only `compactTodosPlanRows` and `todosPlanPanelRowCount`, never the component.

`maxRows` is now required and the branch is gone, which takes the `Todos`/`Plan` heading rows and their margins with it. `TodoRow`'s `compact` prop follows: the one surviving caller always passed `compact`, so its three conditionals collapse to their compact values.

## `/config`: 4 where the identical shape says 5

`ENUM_CHROME_ROWS = 4` sat beside `LIST_CHROME_ROWS = 5`, but both layouts wrap one `Select` in the same chrome — two `FormFrame` border rows, the title, the `KeyHints` margin, and the hints row. Both even pass `showCloseHint={false}`. The enum picker could therefore size its `Select` one row past its budget.

## Net elements (R6)

- Files: 4 changed (+42 / −95 LoC)
- Exports: 0 exported symbols added or removed
- Constructs removed: `omitHeader`, `neutralStatus`, `showOutput` (ToolUseRow props); uncapped `TodosPlanPanel` branch; `TodoRow.compact` prop; `LIST_CHROME_ROWS`, `ENUM_CHROME_ROWS` constants
- Constructs added: `SELECT_CHROME_ROWS` (unifies the two removed)

## Consumer counts (R8)

n/a — no emit paths or external-consumed exports changed; all removed symbols were internal to their modules.

### DIFF
--- a/packages/cli/src/chat/tui/forms/ConfigForm.tsx
+++ b/packages/cli/src/chat/tui/forms/ConfigForm.tsx
@@ -187,9 +187,9 @@ type ConfigFormMode =
       readonly category: string;
     };
 
-	/* Every /config layout wraps one Select in the same chrome: two FormFrame
+/* Every /config layout wraps one Select in the same chrome: two FormFrame
 	   border rows, the title, the KeyHints margin, and the hints row. */
-	const SELECT_CHROME_ROWS = 5;
+const SELECT_CHROME_ROWS = 5;
 
 /** Inline text editor for a string/number setting (its own input buffer). */
 function ConfigTextEditor(props: {
--- a/packages/cli/src/chat/tui/forms/ConfigForm.tsx
+++ b/packages/cli/src/chat/tui/forms/ConfigForm.tsx
@@ -187,8 +187,9 @@ type ConfigFormMode =
       readonly category: string;
     };
 
-const LIST_CHROME_ROWS = 5;
-const ENUM_CHROME_ROWS = 5;
+	/* Every /config layout wraps one Select in the same chrome: two FormFrame
+	   border rows, the title, the KeyHints margin, and the hints row. */
+	const SELECT_CHROME_ROWS = 5;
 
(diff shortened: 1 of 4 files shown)


LionSR:
Rebase

PR #9776, opened 2026-08-05
fix(cli): collapse doubled blank rows between transcript entries

LionSR:
Every boundary between two transcript entries that both declared a separator was spending two blank rows on one gap.

`ROLE_GEOMETRY` gives a user turn both `marginTopRows: 1` and `marginBottomRows: 1`, and a phase divider `marginTopRows: 1`. Yoga does not collapse adjacent margins the way CSS does, so `user`→`user`, `user`→`phase`, and a multi-line tool row→either each cost two rows instead of one.

Measured on a rendered frame at 60 columns, `user`/`user`/`assistant`/`user`/`phase`:

```
before (12 rows)              after (10 rows)
 0|                            0|
 1| › first user turn          1| › first user turn
 2|                            2|
 3|                    <-- ×2  3| › second user turn
 4| › second user turn         4|
 5|                            5|assistant reply
 6|assistant reply             6|
 7|                            7| › third user turn
 8| › third user turn          8|
 9|                           9|◆ Scan
10|                    <-- ×2
11|◆ Scan
```

In a TUI that appends to native scrollback and keeps a capped live region, rows are the scarcest resource — every wasted one pushes real content out of the visible window, and this fired on every turn boundary.

## Approach

`transcriptEntryLayout` now accepts the entry printed directly above and returns `max(0, ownTop - previousBottom)` — the CSS margin-collapsing rule, applied at the geometry SSOT that every render path and every row budget already reads. Nothing computes spacing at render time.

A caller that cannot identify the entry above passes nothing and keeps the full declared margin, so no path can under-reserve. That keeps the blank above the very first entry, and leaves the live pane's leading row alone.

Threaded through the scrollback path on **both** sides of the render/budget pair, which is the invariant that matters: `StaticConversationTranscript` passes the preceding item to `TranscriptEntry` and to `staticTranscriptItemRowCount`, so painted rows and counted rows stay equal. A session header or printed-output block above an entry counts as no neighbor, since neither declares a bottom margin.

## Verification

- New `TranscriptMarginCollapse.vitest.ts` pins the rule at the SSOT: collapse when both sides declare, keep the separator when the entry above declares none (assistant→user, assistant→phase), full margin when no previous entry is known, and the bottom margin unchanged either way.
- `npx vitest run src/test-kernel/cli` — 2163 passed, 1 skipped, 148 files
- `npm run typecheck` — clean across all four projects
- `npm run lint` — clean
- `npm run check:dead-code-ratchet` — 18 current vs 18 baselined, unchanged
- The before/after frame above came from an Ink render through `inkTestHarness`, not from reading the geometry table.

**Not verified:** I did not drive a live TTY at multiple widths. The regression test covers the geometry, and the render/budget pair is threaded from one source, but a real session is the honest check for scrollback appearance.

## Related

Fourth from the same audit, after #9773, #9774, and #9775. No shared files with any of them.

---

> [!NOTE]
> **Low Risk**
> Display-only spacing logic in the CLI transcript layout SSOT; no auth, data, or API changes. Risk is limited to scrollback/live transcript appearance and row-budget alignment.
> 
> **Overview**
> **Fixes doubled blank rows** at transcript entry boundaries where both neighbors declared a separator (e.g. user→user, user→phase). Ink/Yoga does not collapse adjacent `marginTop`/`marginBottom` like CSS, so each boundary was costing two rows instead of one in scrollback and compact row budgets.
> 
> **`transcriptEntryLayout`** now accepts optional `previousEntry` and computes `marginTopRows` as `max(0, ownTop − previousBottom)` via a small `entryMarginBottomRows` helper. Callers without a known neighbor keep the full declared top margin so the first row and budget estimators do not under-reserve.
> 
> **Static scrollback** threads the preceding static item through both paint and count: `StaticConversationTranscript` passes `previousItem` into `TranscriptEntry` and into `staticTranscriptItemRowCount`; `entryAbove` only treats a prior `entry` item as a neighbor (headers and printed transcript blocks do not participate in collapse).
> 
> **Tests:** `TranscriptMarginCollapse.vitest.ts` pins collapse rules at the geometry SSOT.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34b37c5c32e4225564fdc525e543a17086d8da3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
@claude fix

PR #9792, opened 2026-08-06
ci(review): set texra-review approval-policy to always

LionSR:
Test: change approval-policy from never to always so the review action exits 0 instead of 4 when it finds issues.


LionSR:
Test complete. Fix landed via separate branch.

PR #9796, opened 2026-08-06
Document invariants behind non-null assertions found by data-structure audit

LionSR:
## Summary

A scheduled audit flagged data structures across the codebase for unnecessary complexity: deeply nested/generic types, duplicated data transformations, and code requiring frequent type assertions or null checks to use safely (full report in the originating session). This PR follows up on that audit by verifying each finding against its actual call sites and landing the ones that survived scrutiny.

Three non-null assertions were flagged as unsafe with no visible justification. Each is backed by a real control-flow invariant that just wasn't documented at the call site — exactly the kind of hidden invariant CLAUDE.md calls out as deserving a comment:

- `ZoteroCollectionsTool.buildTree` (`src/tools/zotero/ZoteroCollectionsTool.ts`): the parent-linking loop walks the same `collections` array that seeded every map key in the prior loop, so the lookup can't miss.
- `ModelRetryGate.acquire` (`src/agent/runtime/ModelRetryGate.ts`): `acquireHealthy` only returns `undefined` when it read an *existing* non-healthy route state (the create-on-miss branch always returns a permit), so the route is guaranteed to still be keyed in the map.
- `externalInquiryStorage` (`src/tools/inquiry/externalInquiryStorage.ts`): the preceding `turns.length === 0` guard makes `turns.at(-1)` provably defined.

## Issue acceptance gates

Originating scheduled audit findings: "structures that require frequent type assertions, null checks, or conditional logic to use safely." Gate: each of the three flagged non-null-assertion sites now documents the invariant that makes it safe. No behavior change.

## Single source of truth

n/a — this PR adds documentation only, no new module/schema/coordinator.

No duplication removed or added; no new abstraction introduced. Three one-line comments added, zero constructs added.

## Net elements (R6)

3 files changed, +7/-1 lines. No new exports, classes, interfaces, or enums.

## Consumer counts (R8)

n/a — no emit path, export, or public symbol changed or re-routed.

## Host impact

Extension: none (host-agnostic files under `src/agent/`, `src/tools/`)
Desktop: none
CLI: none

## Scheduled refactoring

None filed as a follow-up issue. Note for reviewers: the majority of the audit's other findings (per-provider reasoning-effort clamp "duplication", separate Codex/Claude CLI effort Zod schemas, the remote-agent `agentCategory` field name, the `agentYamlScanner` cast, the `TextEditorTool` nested `Map<Map>`) were investigated and found to be intentional external-contract fidelity (DB column names, SDK-specific enums, provider-specific accepted value ranges) or already-justified control flow — not debt. They were deliberately left untouched rather than churned to inflate this diff; if any are worth revisiting, that should be a separate, individually scoped design discussion given they touch provider-integration and DB-contract code.

## Validation

- `npm run typecheck` — passed (workspace, test-kernel, agent, extension, cli, trace-viewer, desktop)
- `npm run lint` — passed, no findings
- `npx vitest run src/test-kernel/agent/runtime/ModelRetryGate.vitest.ts` — 20/20 passed
- No test files exist for `ZoteroCollectionsTool` or `externalInquiryStorage`; changes there are comment-only so no behavior to test.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RXvoUt8RfTCixCV5LNzk1c)_

---

> [!NOTE]
> **Low Risk**
> Comment-only changes with no logic, exports, or contracts altered; lowest possible deployment risk.
> 
> **Overview**
> Follow-up to a data-structure audit: **three `!` non-null assertions** now document the control-flow invariant that makes each safe—no runtime or type behavior change.
> 
> In **`ModelRetryGate.acquire`**, a comment explains that when `acquireHealthy` returns `undefined`, an existing non-healthy route state was already in the map (create-on-miss always returns a permit), so `this.routes.get(route)!` cannot miss.
> 
> In **`withOpenTurnUpdate`** (`externalInquiryStorage`), a comment notes that the earlier `turns.length === 0` guard guarantees `turns.at(-1)` is defined.
> 
> In **`ZoteroCollectionsTool.buildTree`**, the parent-linking loop comment states it walks the same `collections` array used to seed `nodeMap` by `c.key`, so `nodeMap.get(c.key)!` cannot miss.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 831f8303f7578a04a156c2816abb2b77d5407091. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/texra-code-review.yml
+++ b/.github/workflows/texra-code-review.yml
@@ -112,7 +112,7 @@ jobs:
         uses: texra-ai/texra-action/review@7800c7098bcb4d4bf956aa5dc72437eb5d0af6ac
         with:
           prompt-file: .trusted-review-prompt/.github/prompts/texra-code-review-prompt.md
-          approval-policy: never
+          approval-policy: yolo
           require-write-access: 'true'
           allow-bots: dependabot[bot],github-actions[bot]
           model: ${{ vars.TEXRA_REVIEW_MODEL }}
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,5 @@ out/
 
 
 errorLogs/
+# Playwright MCP debug artifacts
+.playwright-mcp/
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "delay": "^7.0.0",
     "diff-match-patch": "^1.0.5",
     "dotenv": "^17.4.2",
+    "es-toolkit": "^1.50.0",
     "escape-string-regexp": "^5.0.0",
     "execa": "^10.0.1",
     "fast-json-stable-stringify": "^2.1.0",
--- a/packages/agent/package.json
+++ b/packages/agent/package.json
@@ -68,6 +68,7 @@
     "deepmerge": "^4.3.1",
     "delay": "^7.0.0",
     "diff-match-patch": "^1.0.5",
+    "es-toolkit": "^1.50.0",
     "escape-string-regexp": "^5.0.0",
     "execa": "^10.0.1",
     "fast-json-stable-stringify": "^2.1.0",
--- a/packages/cli/scripts/validate-run.mjs
+++ b/packages/cli/scripts/validate-run.mjs
@@ -337,7 +337,7 @@ function validateMultiAgentListAvailability() {
     );
     const leanProjectAvailability = leanProjectJson?.availability;
     assert(
-      leanProjectAvailability?.toolUse?.label != null,
+      leanProjectAvailability?.agents?.toolUse?.label != null,
       `multi-agent list JSON should include planned availability\nstdout:\n${json.stdout}`,
     );
     assert(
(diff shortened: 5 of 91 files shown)


LionSR:
CI on this PR (head `cd19cd5`) is red on `webview smoke (macOS)` and `kernel tests (linux, shard 1/2)`. Confirmed these are not caused by this diff:

- Both failures reproduce identically on `main`'s own tip commit (`2be3de7`, CI run [31081228141](https://github.com/LionSR/TeXRA/actions/runs/31081228141)), independent of this branch: `webview smoke (macOS)` fails the same "Main view fixture rendered too few file rows: 0" error, and `kernel tests (linux, shard 1/2)` fails the same `ToolEditRequestPanel.vitest.ts:69:43` assertion.
- `main` was force-pushed with a batch of `wip(ui)`/`wip(desktop)` checkpoint commits (`af5c268`, `0e79d74`, `146b0bc`, `2be3de7`) after this branch was originally cut, apparently without full CI validation.
- I merged `main` into this branch to pick up those commits, which also surfaced (and I fixed, `cd19cd5`) a `prettier --check` failure in `ProgressApp.ts`/`SettingsApp.ts` that was already present on `main`'s tip in isolation.
- The two remaining failures (webview smoke, kernel tests shard 1) are genuine regressions on `main` itself, unrelated to this PR's three-file, comment-only diff. Fixing them is out of scope here.

I'll re-run once `main` recovers.

---

PR #9802, opened 2026-08-06
fix: batch triage of small follow-up issues (#9794, #9779, #9780)

LionSR:
## Summary

Batch fix for three small, well-scoped follow-up issues filed against recent PRs:

- **#9794** — `FileSelectGroup.ts` bound two `@click` handlers on the same element (`handleRemoveClick`, `handleMoveClick`). The HTML parser drops duplicate attribute names, so only the first survived and the move-up/move-down buttons never dispatched `filesReordered`. Merged both into a single delegating `handleFileListClick`.
- **#9779** (priority:p0, risk:high) — `finalizeChildStream`'s failure prologue (`logger.error`, `toErrorMessage`, `classifyAgentError`) ran before `finalizeRunTerminal`. A throw there would leave the handle stuck in `ExecutionRegistry` forever (`claimTerminalFinalize()` makes a retry a silent no-op). Wrapped the prologue in try/catch with a safe fallback outcome/error so `finalizeRunTerminal` — and the untrack it performs — always runs.
- **#9780** (risk:medium) — `BashBackgroundSession.interrupt()` sent a bare SIGTERM with no SIGKILL escalation or stream cleanup, unlike `executeCommand`'s own kill ladder. Replaced the duplicated pid/kill bookkeeping with an `AbortController` wired into `executeCommand`'s existing `signal` option, so background bash interrupts get the same SIGTERM→SIGKILL escalation as foreground runs.

**#9795** (remove committed `.playwright-mcp/` artifacts) was already fixed on `main` prior to this branch — verified the files are gone and `.gitignore` already excludes the directory. No change needed here.

## Test plan

- [x] `npx tsc --noEmit --checkers 8` (root workspace) — clean
- [x] `corepack pnpm --filter texra typecheck` (extension package, covers `FileSelectGroup.ts`) — clean
- [x] `npx eslint` on all three changed files — clean
- [x] `npx vitest run` on `BashTool`, `DelegationTools`, `DelegationHeadless`, `AgentRunLifecycle` suites — 100/100 passing

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fbeitka9Ru48fbvqhogUiF)_

---

> [!NOTE]
> **Medium Risk**
> Changes background process termination and child-stream terminal cleanup in the execution registry; both are correctness fixes with bounded scope but touch run lifecycle and subprocess teardown.
> 
> **Overview**
> Background bash interrupts now route through **`executeCommand`'s `AbortSignal`** instead of a separate **`onPid` + bare SIGTERM** path in `BashBackgroundSession`, so stops get the same **SIGTERM → SIGKILL** escalation and stream teardown as foreground runs. **`signalProcessGroup`** is no longer exported from `execUtils` because bash no longer calls it directly.
> 
> **`finalizeChildStream`** wraps its failure prologue (logging, error formatting, outcome derivation) in **try/catch** so a throw there cannot skip **`finalizeRunTerminal`**—avoiding execution handles stuck in the registry after the terminal-finalize guard has already been claimed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a94616ca4e091cacde89925913751eb395e73dfb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/scripts/validate-run.mjs
+++ b/packages/cli/scripts/validate-run.mjs
@@ -337,7 +337,7 @@ function validateMultiAgentListAvailability() {
     );
     const leanProjectAvailability = leanProjectJson?.availability;
     assert(
-      leanProjectAvailability?.toolUse?.label != null,
+      leanProjectAvailability?.agents?.toolUse?.label != null,
       `multi-agent list JSON should include planned availability\nstdout:\n${json.stdout}`,
     );
     assert(
@@ -346,7 +346,7 @@ function validateMultiAgentListAvailability() {
       `lean-project JSON should report degraded or unavailable status\nrecord:\n${JSON.stringify(leanProjectJson, null, 2)}`,
     );
     assert(
-      leanProjectAvailability?.toolUse?.label !== '7',
+      leanProjectAvailability?.agents?.toolUse?.label !== '7',
       `lean-project JSON should not claim full tool-use availability\nrecord:\n${JSON.stringify(leanProjectJson, null, 2)}`,
     );
 
@@ -357,7 +357,7 @@ function validateMultiAgentListAvailability() {
       'multi-agent list NDJSON',
     ).find((record) => record.preset?.id === 'lean-project');
     assert(
-      leanProjectNdjson?.preset?.availability?.toolUse?.label != null,
+      leanProjectNdjson?.preset?.availability?.agents?.toolUse?.label != null,
       `multi-agent list NDJSON should include planned availability\nstdout:\n${ndjson.stdout}`,
     );
   } finally {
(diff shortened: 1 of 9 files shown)


LionSR:
CI red on `kernel tests (linux, shard 1/2)` (and the downstream `validate` gate), but it's failing on `main` too — verified by running the same two tests in a clean worktree of `origin/main` (`fca529b`):

```
FAIL src/test-kernel/progressView/ProposalRequestPanel.vitest.ts > uses a shared tooltip for the direct Setup action
FAIL src/test-kernel/progressView/ToolEditRequestPanel.vitest.ts > uses a shared tooltip for the direct diff action
```

Both assert `hasAttribute('title')` is `false` and get `true` — pre-existing, unrelated to this PR's diff (`FileSelectGroup.ts`, `childStream.ts`, `bash.ts`/`execUtils.ts`). Not fixing here; will re-run when the base branch recovers.


---

PR #9803, opened 2026-08-06
refactor: consolidate the five largest ad hoc implementations

LionSR:
## Summary

A codebase-wide audit identified the largest ad hoc / duplicated implementations; this PR refactors the five highest-impact areas, one commit each. It also fixes two pre-existing failures inherited from the base branch, so **CI is green here while `main` is red**.

The honest headline on the refactor itself: **it removes duplication but does not reduce production line count** — net **+67 LoC** across the five refactor commits (see R6). Three of the audit's proposed refactors did not survive contact with the code and were deliberately not implemented rather than forced; each is documented below and in its commit message. The wins are single-ownership of facts, removal of drift hazards, and testability — not line savings.

| Commit | Area | Prod LoC |
|---|---|---|
| `67ea9f1` | tools: `executed()` adoption, Lean command SSOT | −44 |
| `5a61a98` | transcript: `StagedDeletionCoordinator` extraction | +162 |
| `b7ef88d` | agent: shared server-chain lifecycle | −9 |
| `a3b2254` | extension: state-setting rows + subscription providers | −107 |
| `1d01b52` | cli: one workflow-dashboard model, TUI scaffolds | +65 |
| `6421abd` | style: repo-wide prettier gate | — |
| `2c16c0d` | fix: restore webview file-list remove/reorder | — |
| `5d8ef28` | fix: scope native `title` fallback to icon-only buttons | — |

## Issue acceptance gates

No tracked issue — this originated from a scheduled codebase audit. No acceptance gates to satisfy.

## Single source of truth

Each area names a new owner for a fact previously derived in two or more places:

- **`packages/cli/src/chat/tui/state/workflowDashboardModel.ts`** — owns the workflow row model (child→task dedup, unique-child resolution, row-value lists). Previously rebuilt independently in `App.tsx` and `SubagentList.tsx`, which had to agree on row identity/order or Alt+1..9 focus desynced from the screen. `workflowDashboardPanelItemCount` now reads it too, so the rows the panel *reserves* and the rows it *renders* come from one active-group rule.
- **`src/agent/modelHandlers/support/ServerChainState.ts`** — owns server-side response-chain state for both OpenAI and Google. Google previously re-implemented the concept as three loose handler fields plus a private `invalidateChain()`.
- **`src/transcript/StagedDeletionCoordinator.ts`** — owns `deletionStates` and the crash-safe staged-deletion/rollback machine, previously entangled in a 2250-line store.
- **`src/tools/lean/leanTypes.ts`** — one frozen `Record<name, spec>` table per tool is now the single source for the `z.enum`, the tool prose, the schema `.describe()` strings, and the result summaries. Previously declared three times.
- **`packages/extension/src/settingsView/handlers/subscriptionHandlers.ts`** (`SubscriptionProvider` descriptor) — owns per-provider subscription wiring for ChatGPT/Codex and Grok/xAI, previously duplicated end to end across two handlers and two Lit sections.

**Removed:** 94 hand-rolled `status:'executed'` literals; the triplicated Lean command lists; Google's inline chain-state fields; two duplicated subscription handler classes and two duplicated Lit sections; six tabs' worth of `UPDATE_STATE_SETTING` plumbing plus `catalogEnumOptions`/`catalogSelectOptions` (verbatim copies of each other) and seven now-dead toggle handlers; a duplicated workflow model across two files; three structurally identical rollback blocks; five near-identical orchestration step scaffolds; three copies of a widest-first "first row that fits" cascade.

**Added, with the invariant each owns:** `ServerChainState` (chain anchor + sent-count + cumulative tokens); `StagedDeletionCoordinator` (crash-safe deletion ownership, behind a documented 6-member host port); `workflowDashboardModel` (row identity and order shared by the reducer, the renderer, and the row budget); `SubscriptionProvider`/`SubscriptionSectionProvider` (per-provider wiring); `stateSettingRows` (state-setting post + catalog option derivation); `firstRowThatFits` / `attemptRollback` / `StepShell`.

**Deliberately NOT abstracted** — four proposals rejected on this budget:

1. **Shared debounced-persist scheduler for the two transcript stores.** The premise is false: `StreamSnapshotStore` has no debounce, no coalesced save and no awaiter promises, and its dirty map is keyed per `(stream, category)` *with values* rather than by stream id; `StreamLogStore` has a 300 ms debounce, a stream-id `Set`, and one coalesced write. A shared scheduler would have had one real caller and would have had to change flush/cancel timing inside crash-recovery code.
2. **`createResponseImpl` template driver.** `applyTokenCountLimit` and `maybeCompactByInputTokens` are *already* the shared envelopes; the three remaining error-recovery paths are different algorithms, not copies. Unifying them needs 4+ injected predicates and moves the termination proof away from the code that depends on it.
3. **`OpenAIResponseCompactionController`.** The compaction cluster reaches ~20 handler members, so the controller would need a 20-callback deps bag or a back-reference to the handler — leakage, not encapsulation.
4. **Unifying the two status-bar segment fitters.** One is a `compactPriority` sweep; the other never reads `compactPriority` and is positional with mid-segment truncation. Unifying needs a mode flag and ~8 options to produce a larger function than the two it replaces. Likewise the `ExternalInquiry`/`DiffView` sites would only gain a wrapper around `Array.prototype.find` — the shallow indirection AGENTS.md bans.

No pass-through layers were added; all deleted files were removed outright with no re-export shims.

## Net elements (R6)

Measured over the five refactor commits; the two fix commits and the formatting commit are excluded.

- **Files:** 82 changed — 10 added, 6 deleted, 66 modified (`+3208 / −2519`).
- **Exported symbols** (`^[+-]export` over the diff): **+33 / −13**. Net element count therefore *grows*; the growth is the six named owners above replacing implicit contracts, and `check:dead-code-ratchet` confirms every new export has a consumer.
- **Classes / interfaces / enums:** classes `+4 / −5`; interfaces `+17 / −6`; enums `0 / 0`; type aliases `+13 / −9`.
- **Net LoC:** production **+67** (`+2480 / −2413`); tests **+622** (`+723 / −101`); docs **0** (`+5 / −5`).

The two areas that grow production LoC are extractions, stated as such rather than presented as savings: `StagedDeletionCoordinator` (+162, of which the module docstring and documented port are a substantial share) and the CLI workflow model (+65, of which ~51 lines are comments/blank).

## Consumer counts (R8)

No `bus.emit` path, `AgentEvent`, or `SessionFact` channel was added, removed, or re-routed — **n/a for emit paths**.

Exports deleted or renamed, with remaining references verified at 0 by full-repo typecheck plus grep:

| Deleted export | Disposition | Remaining refs |
|---|---|---|
| `ResponseChainState` | renamed → `ServerChainState` (`support/`) | 0 (incl. docs; PRD updated) |
| `ChatGptSubscriptionHandlers`, `GrokSubscriptionHandlers` | merged → `SubscriptionHandlers` + descriptors | 0 |
| `CodexSubscriptionSection`, `GrokSubscriptionSection` | merged → `<subscription-section>` | 0 |
| `LEAN_FILE_COMMANDS`, `LEAN_PROJECT_COMMANDS` | re-typed: frozen name arrays → frozen spec `Record` | all call sites migrated |
| `LeanFileCommand`, `LeanProjectCommand` | now derived via `keyof typeof` | unchanged shape |
| `FILE_/PROJECT_COMMAND_DESCRIPTIONS` | derived from the spec table | 0 |
| `codexSandboxMode`, `codexReasoningEffort`, `codexApprovalPolicy` | re-typed `string` → concrete union | consumers restored in-area |
| `orchestrationKeyHints` | folded into `orchestrationStepKeyHints`; export retained | 1 (a test) |

`check:dead-code-ratchet`: **18 findings vs 18 baselined** — no new unused exports, baseline not widened.

## Host impact

**Extension:** Settings-view internals — six tabs now post `UPDATE_STATE_SETTING` through shared helpers, and the two subscription providers render from one component. No user-visible change there: labels/copy are still supplied by each caller precisely because the catalog's `title`/`description` differ materially from the shipped UI copy.

Two user-visible *fixes* land here (both repairing regressions from `af5c268`):
- The webview file list rendered zero rows, so **remove and reorder were both dead**.
- Labelled action buttons carrying a `wa-tooltip` also set a native `title`, stacking a browser tooltip on top of the tooltip showing the same hint.

**Desktop:** No files changed. The `src/shared/settingsView` `chatGpt`/`grok` message builders were **not** folded in, because their consumer `packages/desktop/src/main/desktopCredentialSettingsController.ts` is outside this change's scope; they are now one indirection from merging.

**CLI:** TUI internals only. Rendering verified unchanged by diffing byte-for-byte Ink output before/after for the launcher, teams, resume and model-access steps; all six rollback error strings byte-preserved.

## Scheduled refactoring

1. **Narrow-list ordering divergence (CLI):** narrow `listValues` is transcript order while the narrow list renders grouped order. These coincide only while same-phase tasks stay contiguous, which is what workflows emit today; an interleaved-phase workflow would seed the reducer's default row differently from the highlighted row. Predates this PR; documented in `workflowDashboardModel.ts` rather than silently changed.
2. **Missing `await` (tools):** `LeanInspectTool.execute` dispatches with `return this.executeGoal(...)` and no `await`, so its `catch` was unreachable. The provably dead `catch` was removed; adding the `await` would change behaviour and is left as a separate fix.
3. **Minor coverage loss:** the `markOpenRouterSkipLogged` idempotence unit test was removed with the setter it covered and not replaced; the "log OpenRouter compaction-skip once" path now has no behavioural test.
4. `src/shared/settingsView` `chatGpt`/`grok` builders — merge once desktop can be edited in the same change.
5. **`ExtensionTexraConfig` is environment-sensitive:** two cases `chmod 0o500` a directory to force a write failure, which silently passes when the suite runs as root (they fail locally in a uid-0 container, pass in CI). Worth a `process.getuid()` guard or skip so the result doesn't depend on the runner.

## Validation

**CI is green on `5d8ef28`** — `validate`, both kernel shards, static checks, webview smoke (macOS), build artifacts, CLI workflow input paths (Windows), guidance references, and Cursor Bugbot (no findings). Branch is merged up to `main` @ `fca529b`.

Local, on the integrated branch:

- `npm run typecheck` (whole repo) — clean.
- `npm run lint` — clean.
- `npm run check:dead-code-ratchet` — OK, 18 vs 18 baselined.
- `npm test` — **828 files / 8501 tests passed**; the only 2 failures are the `ExtensionTexraConfig` uid-0 cases described above, which pass in CI. For comparison, the same suite before these fixes: 3 failed files / 4 failed tests.

Evidence for the two fixes:
- **File-list actions:** the new `FileSelectGroupActions` tests were run against the duplicate-binding version and **fail 3/3**, and pass against the fix — confirming they actually catch the regression rather than merely passing.
- **Tooltip:** traced to `af5c268` changing `title=${ifDefined(useWebAwesomeTooltip ? undefined : nativeTitle)}` → `title=${ifDefined(nativeTitle)}`. The two assertions predate that commit, so they encode the original contract. Both failing buttons carry visible text and are therefore already named by their content, so scoping the fallback to icon-only buttons satisfies the accessibility intent *and* the tests without weakening either.

Area-level evidence for the refactors: `GoogleInteractionsChaining` (12 tests incl. stale-id one-shot retry bound) passes unchanged; moved transcript method bodies machine-diffed against the prior revision and identical modulo `this.` → `this.host.`; generated Lean tool definitions diffed against originals (byte-identical except where the hand-written prose had already drifted from the description map — the drift this removes).

New tests: `WorkflowDashboardModel.vitest.ts` (5 cases, drives the real mounted `SubagentList` to pin that arrow navigation reaches exactly `listValues` in order, narrow and wide), `StagedDeletionCoordinator.vitest.ts` (9 cases on port-contract ordering the store's public surface cannot observe), `SubscriptionSection.vitest.ts` (per-provider command wiring/copy), `FileSelectGroupActions.vitest.ts` (3 cases pinning row render + remove + reorder), plus frozen-spec rows in `SharedLiteralImmutability`.

---

> [!NOTE]
> **Medium Risk**
> Changes shared agent chaining state and CLI selection logic, but behavior is intended to stay equivalent with added tests; risk is moderate refactors in multi-turn API and TUI focus paths rather than new product surface.
> 
> **Overview**
> This PR **centralizes facts that were previously derived in two places**, so keyboard focus, API chaining, and settings writes no longer depend on parallel copies staying in sync.
> 
> **CLI TUI:** **`workflowDashboardModel`** is now the single source for workflow dashboard rows (phase groups, task order, `listValues`, child-stream ownership). **`App`** and **`SubagentList`** both consume it instead of rebuilding grouping/dedup independently. Orchestration launcher steps share **`StepShell`** and **`orchestrationStepKeyHints`**; status-bar binding rows use **`firstRowThatFits`**. Approval retry rollback in **`subscribeApprovals`** uses **`attemptRollback`** instead of three duplicated restore blocks.
> 
> **Agent handlers:** OpenAI Responses and Google Interactions both use **`ServerChainState`** in `support/` (replacing **`ResponseChainState`** and Google’s inline chain fields) for anchor id, sent-count delta, compaction tokens, and “send all next turn” bookkeeping.
> 
> **Extension settings:** ChatGPT and Grok subscription **handlers** and **Lit sections** collapse into descriptor-driven **`SubscriptionHandlers`** and **`subscription-section`**, with **`stateSettingRows`** (`postStateSetting`, `renderStateSettingToggleRow`, `catalogEnumChoices`) replacing per-tab **`UPDATE_STATE_SETTING`** and enum-option duplication across several tabs.
> 
> Docs reference **`ServerChainState`** for the xAI Responses PRD. New/updated tests cover the workflow dashboard model, server chain state, and related wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc8676b74d7025f095561a267e165f7868a65cb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
## CI status

### `static checks (linux)` — fixed in `6421abd`

`prettier --check .` failed on three files. One was a genuine regression from this branch and two predate it:

- **`docs/prds/2026-08-04-prd-xai-responses-previous-response-id.md` — mine.** Renaming `ResponseChainState` → `ServerChainState` shortened a cell in a markdown table and broke the column alignment prettier enforces. Verified clean on `main` and failing at this branch's HEAD.
- **`ProgressApp.ts`, `SettingsApp.ts` — pre-existing**, already failing `prettier --check` on `main`.

I'd originally kept the two pre-existing files out of the diff to avoid unrelated churn, but the gate is repo-wide, so it cannot go green while they're unformatted. All three are now formatted. Both `.ts` changes are formatting-only — the reflowed regions are brace/indent placement inside `${...}` expression delimiters and a multi-line attribute split, neither significant to the rendered template. Confirmed with `git diff -w` and `typecheck:extension`.

### `webview smoke (macOS)` — pre-existing on the base branch, not from this diff

Fails with `Main view fixture rendered too few file rows: 0`. The same job fails identically on `main` at `2be3de7`, this PR's base commit (run [31081228141](https://github.com/LionSR/TeXRA/actions/runs/31081228141)), and no file it touches is in this diff.

Root cause, for whoever picks it up: `packages/extension/src/webview/frontend/components/FileSelectGroup.ts:270-271` binds `@click` twice on the same element:

```ts
return html`<div
  role="list"
  @click=${this.handleRemoveClick}
  @click=${this.handleMoveClick}
>
```

lit rejects duplicate attribute bindings, so `renderFileList()` throws and renders zero rows — which is exactly what the smoke fixture asserts against. The same error surfaces locally as an unhandled rejection in `MainAppPersistenceRestore.vitest.ts`. It means the remove and move actions in the selected-files list are both dead in the shipped webview, so it's worth fixing on its own merits, not just for CI.

I've left it alone rather than fold an unrelated behavioural fix into a refactor PR; it's listed as follow-up #1 in the description. Happy to send it as a separate PR if wanted.

For completeness, `main` is currently red on several other jobs too (`kernel tests shard 1/2`, `runtime validation`, `static checks`), so this branch inherits that baseline.

---

LionSR:
## CI final status (`6421abd`)

**Green:** `static checks (linux)`, `build artifacts (linux)`, `CLI workflow input paths (Windows)`, `detect changes`, `guidance references`, `label`.

**Red — both pre-existing on the base branch, neither caused by this PR:**

1. **`kernel tests (linux, shard 1/2)`** — fails on exactly two assertions:
   - `src/test-kernel/progressView/ProposalRequestPanel.vitest.ts:84`
   - `src/test-kernel/progressView/ToolEditRequestPanel.vitest.ts:69`

   Both reproduce identically on a clean `origin/main` checkout, and no progressView file appears in this diff. The shard also logs the `FileSelectGroup` duplicate-binding error via `MainAppPersistenceRestore.vitest.ts` — same root cause as #2.

2. **`webview smoke (macOS)`** — `Main view fixture rendered too few file rows: 0`, caused by the duplicate `@click` binding detailed in the previous comment. Same job fails on `main` at `2be3de7`.

`validate` is an aggregate gate, so it stays red while either of the above does.

One data point worth recording: the two `ExtensionTexraConfig` failures I saw locally did **not** occur in CI. That confirms they were purely an artifact of my container running as uid 0 — the suite `chmod 0o500`s a directory to force a write failure, and root ignores permission bits. CI runs unprivileged, so they pass there. No action needed.

The single CI failure genuinely attributable to this branch was the PRD markdown table alignment, fixed in `6421abd`; `static checks` is green as a result.

---

LionSR:
## CI green ✅ (`5d8ef28`)

Every check passes: `validate`, both kernel shards, `static checks`, `webview smoke (macOS)`, `build artifacts`, `CLI workflow input paths (Windows)`, `guidance references`, and Cursor Bugbot (no findings). Branch is merged up to `main` @ `fca529b`.

This supersedes my two earlier comments, which reported the base-branch failures as out of scope. Both turned out to be **regressions from the same commit**, `af5c268` ("wip(ui): checkpoint design-review fixes … before main merge"), and both are fixed here:

1. **`webview smoke (macOS)`** — `FileSelectGroup` bound `@click` twice on the list container; lit rejects duplicate attribute bindings, so `renderFileList()` threw and rendered zero rows. Remove and reorder were both dead in the shipped webview. Fixed in `2c16c0d` with regression tests that were verified to fail 3/3 against the buggy version.

2. **`kernel tests shard 1/2`** — `af5c268` also changed
   `title=${ifDefined(useWebAwesomeTooltip ? undefined : nativeTitle)}` → `title=${ifDefined(nativeTitle)}`,
   to stop icon-only buttons being unnamed (wa-button forwards `title`, but not the host's `aria-label`, into its shadow `<button>`). That regressed two older assertions pinning the original contract. Both concerns are satisfiable at once: the failing buttons carry **visible text** and are already named by it, so the `title` fallback is only needed for icon-only buttons — where it now stays. Fixed in `5d8ef28`; no test was weakened.

I was wrong in my earlier comment to characterise the tooltip failure as a stale test needing a judgement call. The tests predate the change that broke them.

One follow-up worth filing separately: the two `ExtensionTexraConfig` cases that `chmod 0o500` a directory to force a write failure silently pass when the suite runs as root, so they fail in a uid-0 container and pass in CI. A `process.getuid()` guard would stop the result depending on the runner.

---

LionSR:
## Green again on `bc8676b` ✅

`validate`, both kernel shards, `static checks`, `webview smoke (macOS)`, `build artifacts`, `CLI workflow input paths (Windows)`, `guidance references`, and Cursor Bugbot (no findings). This supersedes the "green on `5d8ef28`" note above — the branch has since taken #9805 plus a merge of `main`.

Two things from this round worth flagging:

### The merge with #9805 produced a green-but-wrong `actionButtons.ts`

#9805 and this branch independently fixed the same duplicate hover hint from `af5c268`, and they chose different resolutions:

- **#9805:** `title` is suppressed for *every* button that has a `wa-tooltip` sibling, on the basis that `aria-label` already names the button unconditionally.
- **This branch:** `title` was kept for *icon-only* buttons, on the (weaker) assumption that `aria-label` is not forwarded into wa-button's inner `<button>`.

The merge kept this branch's `title=${ifDefined(nativeTitle)}` binding while taking #9805's comment, so icon-only buttons were still receiving the native `title` that #9805 had deliberately removed — with a comment above them saying the opposite. **The test suite did not catch this**, because the two tooltip assertions only cover labelled buttons, where both resolutions agree.

`bc8676b` restores `src/shared/wa/actionButtons.ts` to #9805's version verbatim. Deferring to it on both counts: it is the maintainer's call on this component, and its `aria-label` reasoning is the stronger one. The `ProposalRequestPanel` and `ToolEditRequestPanel` assertions pass under it, since suppressing the native title for every `wa-tooltip` button is a superset of suppressing it only for labelled ones.

The `FileSelectGroup` side of the merge came out correctly — #9805's cleaner `handleFileListClick` (dispatching by `data-*` marker) with a single `@click` binding.

### The earlier `kernel tests shard 1/2` failure was a flake, not a regression

It was `Test timed out in 10000ms` at `src/test-kernel/architecture/subsystemEdgeRatchet.vitest.ts:351`, not an assertion. That test does a full-repo import scan; it uses ~1.9s of its 10s budget locally, passed on the preceding head, and has now passed again on `bc8676b`. No timeout bump was applied — worth remembering as a candidate if it reappears under shard contention.

---

Issue #9798, opened 2026-08-06
Terminology: unify 'background tasks' / 'child streams' / 'subagent' user-facing naming


LionSR:
Deferred from the #9791 design sweep (multi-agent UI review, nit #7).

The same concept family appears under three user-facing terms:
- "Background tasks" — extension `BackgroundTasksPanel`
- "child stream(s)" — extension `StreamTabs` aria-labels
- "subagent" — CLI `SubagentList` pane

These are not perfectly overlapping scopes (background tasks include non-agent bash work; child streams are a stream-tree concept), so unification is a product terminology decision rather than a sweep edit. Pick one canonical term per concept and apply it consistently across extension, desktop, and CLI.

PR #9813, opened 2026-08-06
fix(cli): keep approval actions reachable and name the way out

LionSR:
## Summary

A harsher second pass over the TUI turned up one card that can become
unanswerable, one control that does the opposite of its own label, and a
wrap-budget bug that lets the terminal re-wrap our own rows. This fixes those
plus three copy defects where the interface described something other than
what happens.

**Retry approval could lose its own actions.** `RetryRequest` was the only
approval card that never received `availableRows`, so it rendered the provider
error unbounded. A stack trace pushed `y retry` / `n dismiss` past the live
region's clip, leaving the run blocked on a prompt with no visible way to
answer it. It now budgets the body through `scrollableModalTextRowsBudget`
like `BashApproval` and `PlanApproval` do, so the error scrolls and the
actions stay pinned.

**Empty Enter rejected the inquiry it promised to submit.** The external
inquiry footer advertises `Enter submit answer`, but an empty Enter resolved
the decision as `{ accepted: false, userMessage: 'cancelled' }` — silently
rejecting the agent's question. Esc already skips, with a clear message, so
an empty Enter is now a no-op.

**Wrapped rows overflowed their column budget.** `wrapWithPrefix` sized the
wrap width from `firstPrefix` but prepended `continuationPrefix` to every
later row. Where the continuation gutter is wider — `phase` entries indent
2 columns under a flush first line — each wrapped row ran 2 columns past
`columns` and the terminal re-wrapped it, producing stray one- and two-character
rows. It now budgets against the wider of the two, measured with
`textDisplayWidth` rather than `.length` so a wide status marker counts
correctly.

Three copy fixes, each replacing a statement of internal state with something
the reader can act on:

| Before | After |
| --- | --- |
| `Session selection active.` | `Choosing a session; press Esc to return to the prompt.` |
| `Use foreground panel shortcuts` | `Keys go to the panel above` |
| `Signed in with TeXRA as X (included models enabled).` | `Signed in with Researcher Access as X. Model calls now use included access.` |

The sign-in line matters beyond phrasing: `src/shared/copy/modelAccess.ts`
declares the canonical vocabulary and states outright that "Researcher Access
is the free program you sign in to, not a name for included access." The old
string contradicted both halves of that. It now imports `INCLUDED_ACCESS`
instead of paraphrasing.

Two findings from the review were dropped after checking the source. The
`/memories` in the memory-path error is the memory root, not a mistyped
`/memory` slash command, so it was already correct. And `Tab children` is
genuinely poor vocabulary, but renaming it collides with the existing
`Alt-s subagents` binding — that needs a design decision, not a mechanical
sweep, so it is left alone.

## Test plan

- [x] `npm test` — 8488 passed, 5 skipped
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run check:dead-code-ratchet`
- [x] `npm run validate:tui` — 4 failures, all pre-existing. Three
      (`escape-stops-focused-main-only`, `escape-stops-focused-child-only`,
      `escape-captures-focused-child`) reproduce identically on the unmodified
      baseline; the fourth (`narrow-slash-palette-command-names`) expects an
      `/api` description that has since gained "Grok".
- [x] New regression test: a 40-line error at `availableRows: 12` keeps both
      actions visible, shows the overflow marker, and stays within budget.

The old retry test called `RetryRequest({...})` as a plain function to inspect
`ConfirmCard` props, which the new `useWindowSize()` call breaks. It is now a
behavioral test that drives a real `n` keypress through the approval queue —
a better check of "dismiss is immediate" than asserting the prop that
implements it.

Note: this branch also carries in-flight work from concurrent sessions in the
same worktree (progress-view stream tabs, delegation tooling, login command
extraction). Two build breaks in that work are fixed here so the branch is
green: a `StreamTabs.getStatus` return type too loose for `isInFlightPhase`,
and a duplicate `AgentEntry` import in `WorkflowScriptTool`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "texra-workspace",
   "version": "0.40.1",
   "private": true,
-  "packageManager": "pnpm@11.5.2+sha512.71c631e382066efc25625d5cf029075de07b61b37f6e27350fbd84b1bda5864c8c1967adc280776b45c30a715c0359a3be08fef42d5bb09e2b99029979692916",
+  "packageManager": "pnpm@11.20.0+sha512.9a6f330a95b66446ea088faf1521405a8a01f07fde7124cc9958dfed52d4bb436737e65b08f85f37b46fcba375092558ac51262b816844b22f63406ed166bfee",
   "workspaces": [
     "packages/*"
   ],
--- a/packages/cli/scripts/validate-tui.mjs
+++ b/packages/cli/scripts/validate-tui.mjs
@@ -58,8 +58,12 @@ const PAGE_DOWN = ESC + '[6~';
 const ANSI_SGR_PATTERN = new RegExp(`${ESC}\\[[0-?]*[ -/]*m`, 'g');
 const RUNNING_STATUS_PATTERN = /◆ [-|\/\\] running/;
 const STOPPED_SUBAGENT_INPUT_MESSAGE_START =
-  'Subagent is no longer accepting follow-ups; press Tab to select a session';
+  // Keep in sync with FOCUSED_BACKGROUND_TASK in src/shared/copy/nestedRuns.ts.
+  'This background task is no longer accepting follow-ups; press Tab to select a session';
 const STOPPED_SUBAGENT_INPUT_MESSAGE = `${STOPPED_SUBAGENT_INPUT_MESSAGE_START}.`;
+const STOPPED_SELECTED_BACKGROUND_TASK_MESSAGE =
+  // Keep in sync with FOCUSED_BACKGROUND_TASK.selectedNoLongerAccepting.
+  'The selected background task is no longer accepting follow-ups.';
 const LONG_BASH_APPROVAL_COMMAND = [
   "python3 << 'EOF'",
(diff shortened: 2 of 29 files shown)


LionSR:
Follow-up commit for the jargon consistency ask: instead of more one-off string swaps, this adds \`src/shared/copy/nestedRuns.ts\` (same pattern as \`modelAccess.ts\`) and migrates hosts to import from it.

**Ruling (#9798)**
| Concept | Term |
| --- | --- |
| Anything nested under a run | background task |
| Delegated TeXRA agent | subagent |
| Stream-tree edge | child stream — code only, never UI |
| Tab-opened list | session list (\`Tab sessions\`) |

Deliberate exception (documented in the canon): the status-bar count stays \`N subagents\` / \`N sub\` because those rows are overwhelmingly agents and \`N bg\` is less legible in the footer.

LionSR:
Superseded by a clean PR against `main` that carries the same TUI UX + vocabulary work without the `ui/better-skills-sweep` history.

PR #9881, opened 2026-08-08
fix(build): stop shipping 12MB of unreachable Monaco workers in the VSIX

LionSR:
## The regression

The packaged extension jumped from **8.59 MB** (`v0.40.0`) to **13.39 MB** on `main`. All 4.8 MB of it is Monaco language-worker bundles that nothing in the extension can load.

| path | before | after this PR |
|---|---|---|
| `dist/assets/*.worker-*.js` | 11.77 MB (ts.worker alone 8.49 MB) | gone |
| `resources/traceViewer/` | 6 files, 14.14 MB | 1 file, 5.12 MB |
| VSIX total | 12.77 MB / 184 files | **8.27 MB / 172 files** |

## Why it happened

The extension host has no Monaco — it diffs through VS Code's own `vscode.diff` (`frontend/ui/diffView.ts:28`, `approval/VscodeDiffViewHost.ts:24`, `commands/latex/compareCommands.ts:140`). Monaco belongs to the desktop renderer, which loads it for the editor pane and registers `<texra-diff-view>`.

But **Vite emits a worker bundle for every `import('...?worker')` expression it finds in the module graph, at transform time, before tree-shaking.** When `ToolEditRequestPanel` started importing `monacoLanguageForPath` — a pure path → language-id `switch` that happened to live in `monacoLoader.ts` — the five `monaco-editor/**?worker` imports inside `loadMonaco` became visible to that plugin and all five were written out. Rolldown then dropped `loadMonaco` itself as unreachable, so Monaco's *core* never entered the bundle at all: `bundle.js` grew by 1 KB while 11.8 MB of orphan workers landed next to it.

The trace viewer doubled the cost — it imports the whole progressView frontend (`trace-viewer/src/main.ts:11`), and `vite-plugin-singlefile` cannot inline `?worker` chunks, so they sat beside `index.html` inside the packaged dir.

Neither shipped artifact references any of those files. I grepped `bundle.js` and `traceViewer/index.html` for every emitted worker filename: zero hits.

## The fix

Move the table into `@shared/monaco/monacoLanguage` — a leaf module with no `monaco-editor` import — and point both callers at it. It remains the single owner of the mapping; the module comment explains why it must stay Monaco-free. Desktop is untouched: it still imports `loadMonaco` from `monacoLoader` and ships Monaco exactly as before.

The language-table tests move to `MonacoLanguage.vitest.ts` with a plain static import, which is itself the assertion — if that file ever needs the mocked-Monaco harness again, the split has regressed.

## Regression guard

`check:vsix-contents` already runs in CI (`ci.yml:301`) and release (`release.yml:73`) and did not notice 12 MB appearing. It now fails on any packaged Monaco worker, with a message pointing at the right import. Verified the check passes on the fixed build and that its matcher hits all four entry-name shapes the regressed build produced (`ts.worker-Cp7iE73a.js`, `editorWebWorkerMain.js`, …) and none of the legitimate ones.

## Verified

- `npm run typecheck` — clean
- `npm run check:dead-code-ratchet` — 17 findings vs 17 baselined, no new
- `npm run check:vsix-contents` — passes, including the new guard
- `npm run build:fast` — VSIX 8.27 MB / 172 files; `dist/progressView/bundle.js` byte-identical in size (2,526.55 kB); `resources/traceViewer/` back to a single `index.html`
- Affected vitest files (`MonacoLanguage`, `MonacoLoader`, `TexraDiffView`) — 8 passed

## Not in this PR

`<texra-diff-view>` is registered only by the desktop (`desktop/src/renderer/main.ts:24`). `ToolEditRequestPanel` renders that tag in the shared component, so in the **extension webview and trace viewer** expanding the inline diff yields an unregistered, empty element. Latent since the component landed in May. Fixing it is a design call — register Monaco in the extension webview (megabytes, and `vscode.diff` already covers it) or hide the toggle in hosts that cannot render it — so it wants its own issue rather than a drive-by here.

---

> [!NOTE]
> **Low Risk**
> Packaging and import-path refactor with tests and CI guard; no runtime behavior change for desktop Monaco loading.
> 
> **Overview**
> **Moves the path → Monaco language-id mapping into `@shared/monaco/monacoLanguage`** so callers that only need a label (e.g. `ToolEditRequestPanel`, desktop diff IPC) no longer import `monacoLoader`, which pulls in `monaco-editor` and Vite’s `?worker` graph. That accidental import had emitted ~12MB of language-worker bundles into the extension VSIX even though shipped code never loads them.
> 
> Desktop and editor code now use the shared `monacoLanguageForPath`; the duplicate `monacoLanguageForFilePath` table in `desktopDiffMessages` is removed. Language-mapping tests live in `MonacoLanguage.vitest.ts` with a plain static import as a regression guard.
> 
> **`check:vsix-contents` fails if any packaged Monaco worker stems appear**, with guidance to import `monacoLanguage` instead of `monacoLoader` when only a language id is needed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdc73555e3b7017dd3feb873711be5fe159f3a0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/desktopDiffHost.ts
+++ b/packages/desktop/src/main/desktopDiffHost.ts
@@ -8,6 +8,7 @@ import {
   type DiffSource,
   type DiffViewHost,
 } from '@hosts/uiHosts';
+import { monacoLanguageForPath } from '@shared/monaco/monacoLanguage';
 import {
   computeLineChangeSummary,
   computeUserPatch,
@@ -17,7 +18,6 @@ import { createTexraTempDir } from '@utils/files/tempDir';
 import {
   DESKTOP_DIFF_COMMANDS,
   type DesktopShowDiffMessage,
-  monacoLanguageForFilePath,
 } from '../shared/desktopDiffMessages.js';
 import {
   tryShowInRenderer,
@@ -63,7 +63,7 @@ export function createDesktopDiffHost(
         proposedText: proposedContent,
         additions: lineChanges.added,
         deletions: lineChanges.removed,
-        language: monacoLanguageForFilePath(proposed.filePath),
+        language: monacoLanguageForPath(proposed.filePath ?? ''),
       } satisfies DesktopShowDiffMessage,
     );
     if (shownInRenderer) return { original, proposed, title };
--- a/packages/desktop/src/renderer/editorPane.ts
+++ b/packages/desktop/src/renderer/editorPane.ts
@@ -20,9 +20,9 @@ import { html, nothing, render, type TemplateResult } from 'lit';
 import type { DesktopThemeKind } from '@shared/schemas';
 import { renderIconActionButton } from '@shared/wa/actionButtons';
 import { waIcon } from '@shared/wa/webAwesomeIcons';
+import { monacoLanguageForPath } from '@shared/monaco/monacoLanguage';
 import {
   loadMonaco,
-  monacoLanguageForPath,
   monacoThemeForHostTheme,
   type MonacoModule,
 } from '@shared/monaco/monacoLoader';
--- a/packages/desktop/src/shared/desktopDiffMessages.ts
+++ b/packages/desktop/src/shared/desktopDiffMessages.ts
@@ -1,7 +1,5 @@
 import { z } from 'zod';
 
-import { getBasename } from '@utils/core';
-
 // IPC plumbing for the in-app Review workbench.
 //
 // The desktop main process used to satisfy `DiffViewHost.openDiff` by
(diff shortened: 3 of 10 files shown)


LionSR:
Both review findings addressed in cdc73555e3.

**1. "Single owner" claim was false** — confirmed: `desktopDiffMessages.ts:59-95` did carry a second `EXTENSION_TO_MONACO_LANGUAGE` + `monacoLanguageForFilePath`, consumed by `desktopDiffHost.ts:66`. Rather than weaken the comment, I deleted the duplicate and routed the desktop diff host through the shared owner, so the claim is now true.

Behavior delta, deliberate: the desktop diff view now recognises the languages the desktop *editor pane* already did — `.lean`, `.toml`, `.go`, `.rs`, `.java`, `.c`/`.cpp`, plus the Dockerfile/Makefile basenames — where it previously fell back to `plaintext`. Same host, same Monaco instance, so this only makes the two desktop surfaces agree. The call site passes `''` for a missing path, which the shared table already maps to `plaintext` (covered by a new test case). The mapping tests moved from `DesktopDiffMessages.vitest.ts` onto the surviving owner, including the basename/separator and case-insensitivity cases the desktop suite held.

**2. Regex too broad** — correct on both counts, and Codex was right that `requiredMonacoWorkers` already exists in `extension-package-utils.mjs:10`. The guard now matches those stems (plus `editorWebWorkerMain`) rather than every `*.worker-*` artifact, with `[A-Za-z0-9_-]+` for the hash so a base64url hyphen cannot slip through. Re-verified: all six regressed entry names still trip it, and `search.worker-abc123.js` / `pdf.worker-Xy1.js` correctly do not.

Re-ran after the change: `typecheck` clean, dead-code ratchet 17 vs 17 baselined, the four affected vitest files pass (13 tests), and a fresh `build:fast` is still 8.27 MB / 172 files with `check:vsix-contents` green.

PR #9886, opened 2026-08-08
fix(sessions): sweep leftover background shells, summarize workflow runs

LionSR:
Two defects in the Sessions rail, found from the user's own persisted data.

## 1. Finished background shells stay listed forever

A `bash@tool#…` child is ephemeral: `autoClose` drops its tab the moment the command finalizes, and the command's output is already delivered into the parent run's transcript as a tool result. Its transcript, sidecar and execution row nevertheless stay on disk, and `SessionState.load` hydrates *every* persisted stream — so a host that exited mid-command (or a deletion the execution lease refused) leaves the row listed for good. It accumulates: **987 of the 4610 persisted streams** in my local workspaces are background shells, ~21%.

The existing child filter in `StreamTabs.renderStreamNode` (#9818) cannot catch them. Stream phases are in-memory only, so a hydrated shell arrives with *no* status — `isTerminalOutcomePhase(undefined)` is false, and the row is kept. That filter can only ever work while a live status exists.

`SessionState.load` now sweeps them through the existing `clearStream` path, deleting the durable state instead of hiding it at render time (per the "never compensate for a data-model problem in the renderer" rule). A shell this host is still running is skipped via `isActivePhase`, and a refused deletion is warned about rather than silently leaving a discrepancy.

**Legacy rows.** Streams written before identity stamping (#9705) hydrate with no `RunIdentity` at all — and that is the *entire* population of a workspace last used before Aug 4 (0 of 4610 sidecars here carry one). So the sweep falls back to the minted `bash@tool` prefix for rows that never got an identity. This is exact rather than heuristic: `BASH_CHILD_STREAM_PREFIX` is the prefix's only minter, no agent name can contain `@`, and across all 987 such rows not one has an authored agent other than `bash`. It is confined to this single legacy-data question in the sweep — no run is addressed, resolved or classified through it, so #9755's rule that identity is never reconstructed from a stream id still governs all of those.

⚠️ First load after this lands deletes those leftover shell transcripts. Their command and output already live in the parent conversation, and nothing can resume them.

## 2. Workflow runs got no session description

`generateSessionDescription` returned early unless `agentCategory === ToolUse`, so every workflow-script `agent()` step was labelled by nothing but its agent name — a column of identical `generic` rows, exactly the population a reader can least tell apart. There are only two categories, so the gate is deleted rather than widened. The agent's purpose is now looked up under the run's *own* category, since a tool-use lookup cannot see a workflow agent at all.

Delegated tool-use subagents were already generating one; only the workflow half was excluded.

## Verification

- `npm run typecheck` — clean
- `eslint` + `prettier --check` on every changed file — clean
- `npm run check:dead-code-ratchet` — 16 vs 16 baselined, no new findings
- `src/test-kernel/{progressView,controllers,agent/runtime,tools,transcript,session}` — 2220 tests pass
- New coverage: the sweep by stamped `{ kind: 'process' }` identity *and* by minted id, with a real session in the same load as the control; workflow-run description generation with the category-correct roster lookup

---

> [!NOTE]
> **Medium Risk**
> Startup sweeps delete persisted stream/execution data (leftover shells and orphans) on load; incorrect identification could remove real sessions, though shells are keyed to a dedicated minted prefix and active leases block deletion.
> 
> **Overview**
> Fixes two Sessions rail issues: **orphaned `bash@tool#…` streams** cluttering the list after a host exits mid-background command, and **workflow runs** that never received AI session descriptions.
> 
> **Startup sweep for leftover background shells.** `SessionStores` adds `sweepLeftoverStreams()`, which first deletes persisted background-shell streams (`sweepEphemeralStreams` via `isBackgroundShellStream` / the shared `bash@tool` prefix), then runs the existing orphan sidecar sweep. CLI, desktop, and extension `SessionState.load` call this instead of only `sweepOrphanedStreams`. Still-running shells are kept when deletion is refused (execution lease); failures are logged per stream.
> 
> **Shared `createSessionStores(session)`** centralizes goal-forget and `onCanonicalStreamDeleted` wiring used by CLI, desktop, and `SessionState`.
> 
> **Workflow session descriptions.** `generateSessionDescription` no longer skips non–tool-use categories. Agent purpose for the helper prompt comes from `resolveAgentForLaunch` (category + optional pinned `agentSource`), not `getAgent` limited to tool-use.
> 
> **Note:** First load after upgrade may delete leftover shell transcripts; command output already lives in the parent run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b535ff8a6192b1f1a7dce0482c1dd416dc12290d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/runtime/transcriptSession.ts
+++ b/packages/cli/src/runtime/transcriptSession.ts
@@ -61,6 +61,11 @@ async function initializePersistentSession(
       releaseStreamResources(stream, result.session);
     },
   });
+  // Leftover background shells go first: they leave the transcript index the
+  // orphan sweep then reads as its live set.
+  await stores.sweepEphemeralStreams(
+    new Set(result.session.transcripts.keys()),
+  );
   await stores.sweepOrphanedStreams(new Set(result.session.transcripts.keys()));
   return result;
 }
--- a/packages/cli/src/runtime/transcriptSession.ts
+++ b/packages/cli/src/runtime/transcriptSession.ts
@@ -1,9 +1,9 @@
-import { SessionStores } from '@agent/storage';
 import {
   initializeDefaultSession,
   tryDefaultSession,
   type SessionHandle,
 } from '@agent/runtime/SessionHandle';
+import { createSessionStores } from '@controllers/session/sessionStores';
 import { texraResponseTextProcessing } from '@latex/texraResponseTextProcessing';
 import { releaseStreamResources } from '@tools/approval';
 import { GoalStore } from '@tools/goal';
@@ -49,24 +49,7 @@ async function initializePersistentSession(
       responseTextProcessing: texraResponseTextProcessing,
     }),
   );
-  const stores = new SessionStores({
-    streamLogs: result.session.transcripts,
-    snapshots: result.session.snapshots,
-    goalEntries: {
-      forget: (stream) => GoalStore.forget(stream, result.session),
-      forgetMany: (streams) => GoalStore.forgetMany(streams, result.session),
-    },
-    onCanonicalStreamDeleted: (stream) => {
-      result.session.status.clearStream(stream);
-      releaseStreamResources(stream, result.session);
-    },
-  });
-  // Leftover background shells go first: they leave the transcript index the
-  // orphan sweep then reads as its live set.
-  await stores.sweepEphemeralStreams(
-    new Set(result.session.transcripts.keys()),
-  );
-  await stores.sweepOrphanedStreams(new Set(result.session.transcripts.keys()));
+  await createSessionStores(result.session).sweepLeftoverStreams();
   return result;
 }
 
--- a/packages/cli/src/runtime/transcriptSession.ts
+++ b/packages/cli/src/runtime/transcriptSession.ts
@@ -5,8 +5,6 @@ import {
 } from '@agent/runtime/SessionHandle';
 import { createSessionStores } from '@controllers/session/sessionStores';
 import { texraResponseTextProcessing } from '@latex/texraResponseTextProcessing';
-import { releaseStreamResources } from '@tools/approval';
-import { GoalStore } from '@tools/goal';
 import { ephemeralTranscriptWarning, StreamLogStore } from '@transcript';
 
 export type InteractiveTranscriptPolicy =
--- a/packages/desktop/src/main/desktopProcessStores.ts
+++ b/packages/desktop/src/main/desktopProcessStores.ts
@@ -23,6 +23,9 @@ export async function initializeDesktopProcessStores(session: SessionHandle) {
       releaseStreamResources(stream, session);
     },
   });
+  // Leftover background shells go first: they leave the transcript index the
+  // orphan sweep then reads as its live set.
+  await stores.sweepEphemeralStreams(new Set(transcripts.keys()));
   await stores.sweepOrphanedStreams(new Set(transcripts.keys()));
 
   const detachStreamRemoval = session.events.subscribe(
--- a/packages/desktop/src/main/desktopProcessStores.ts
+++ b/packages/desktop/src/main/desktopProcessStores.ts
@@ -1,32 +1,15 @@
 import { createChannelTrace } from '@agent/trace';
-import { SessionStores } from '@agent/storage';
 import type { SessionHandle } from '@agent/runtime/SessionHandle';
-import { releaseStreamResources } from '@tools/approval';
-import { GoalStore } from '@tools/goal';
+import { createSessionStores } from '@controllers/session/sessionStores';
 import { toLogData } from './desktopLogUtils.js';
 
 /**
  * Load the process-owned desktop stores and attach their lifecycle hooks.
  */
 export async function initializeDesktopProcessStores(session: SessionHandle) {
-  const { transcripts, snapshots } = session;
   const logger = createChannelTrace('DesktopProcessStores');
-  const stores = new SessionStores({
-    streamLogs: transcripts,
-    snapshots,
-    goalEntries: {
-      forget: (stream) => GoalStore.forget(stream, session),
-      forgetMany: (streams) => GoalStore.forgetMany(streams, session),
-    },
-    onCanonicalStreamDeleted: (stream) => {
-      session.status.clearStream(stream);
-      releaseStreamResources(stream, session);
-    },
-  });
-  // Leftover background shells go first: they leave the transcript index the
-  // orphan sweep then reads as its live set.
-  await stores.sweepEphemeralStreams(new Set(transcripts.keys()));
-  await stores.sweepOrphanedStreams(new Set(transcripts.keys()));
+  const stores = createSessionStores(session);
+  await stores.sweepLeftoverStreams();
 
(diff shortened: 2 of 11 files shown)


LionSR:
Addressed all four Codex findings in 7e7fd9e.

**P2 — sweep skipped session-owned backends (desktop).** Correct, and the fix is bigger than the guard: the sweep now lives in `SessionStores` beside `sweepOrphanedStreams`, and every *process owner* calls it — the extension via `SessionState.load`, the desktop via `initializeDesktopProcessStores`, the CLI via `initializePersistentSession`. That's where the orphan sweep already ran, so the two are now siblings at the same boundary instead of one being a presentation concern.

**P2 — sweep not retried after a crashed lease goes stale.** The in-memory phase guard is gone; liveness is the execution lease, which is durable and host-agnostic: a running shell holds it, so `deleteStream` answers `'active'` and keeps the stream. That removes the restart-repair interaction you described (repair marking a crashed stream `RUNNING` can no longer suppress the sweep). The residual case — a host that crashed inside the lease's staleness window — is retained with a warning until the next launch rather than silently, and the doc comment says so. I did not add a delayed retry: the next launch is the natural trigger and a second scheduler for one row is not worth its own failure mode.

**P1 — record the retirement condition.** Added beside the reader: introduced 2026-08-08, retire from 2026-11-04 (three months after #9705 shipped, per AGENTS.md's compatibility window), after which the sweep reads `RunIdentity` alone.

**P2 — resolve the description agent by pinned source.** Fixed: `getAgent(config.agentSource ? agentKey(source, agent) : agent, config.agentCategory)`, matching what launch resolves through `getAgentPath(..., agentSource)`. You're right that the category argument only orders source priority.

Also folded in: sweeping now happens *before* metadata hydration, so no stream state is built for a stream the same load then deletes.

Verification: root + desktop + CLI `tsc` clean; eslint and prettier clean on every changed file; dead-code ratchet 16 vs 16; `src/test-kernel/{progressView,controllers,agent,cli,desktop,transcript,tools}` — 6781 pass, 5 skipped. New `SessionStores` coverage pins that only the shell is deleted and that a refused deletion is warned about, not swallowed.

LionSR:
Round two, in 705a22b — the three new findings plus the duplication they exposed.

**Consolidated the wiring and the sweep order.** The three process owners were each restating two things: how a session's `SessionStores` is wired, and the order the sweeps run in. Both are single-owner now. `SessionStores.sweepLeftoverStreams()` holds the order *and* why it matters (the ephemeral sweep leaves the transcript index the orphan sweep then reads as its live set), and `createSessionStores(session)` holds the wiring. Each host is down to two lines; the CLI and desktop lost their copies of the goal/approval hook wiring entirely.

`createSessionStores` lives in `src/controllers/session/` rather than as a `SessionStores` static: `@agent/storage` sits *underneath* the tool layer those hooks come from, and reaching back up from there closed an import cycle that left the default session uninitialized at load. `ExecutionRegistry` and the TUI host-interaction suites caught it — 5 failures that pass on a clean tree and fail with the factory in `@agent/storage`, so it was mine, not flake.

**P2 — isolate failures while sweeping.** Fixed, and it was worse than "skips the rest": in the extension the sweep is awaited by `SessionState.load`, so a rejecting deletion would have failed the whole load. Each deletion is now caught, warned, and counted as retained. New test asserts the sweep resolves and still reaches the orphan half after a deletion throws.

**P2 — resolve unpinned descriptions with the launch resolver.** Switched to `getRosterAgent(category, identifier)`, which *filters* by category rather than ordering source priority, and still takes the pinned `source:name` key when one exists. I did not thread the launch resolution itself down: `getAgentPath` + `loadAgentSettingAndPrompts` is async file I/O, and `AgentLaunchContext` does not carry the resolved entry, so exposing it for a display label would mean widening the launch context for a purpose string. The residual gap — roster priority vs the launch resolver for an unpinned name that exists in more than one source — affects only which same-named agent's purpose seeds the prompt, never which agent runs.

**Medium — deep schemas import.** `agentKey` now comes from the `@shared/schemas` barrel.

Verification: root + CLI + desktop `tsc` clean; eslint and prettier clean; dead-code ratchet 16 vs 16; architecture ratchets 82 pass; **full `src/test-kernel` — 8776 pass, 5 skipped, 0 fail.**

LionSR:
Good catch — fixed in the follow-up commit. `releaseStreamResources` and `GoalStore` were only used by the option literal `createSessionStores` replaced, and with `@typescript-eslint/no-unused-vars` off repo-wide nothing failed on them. I audited every file this PR touches for the same residue; the CLI file was the only one. CLI + root typecheck, eslint, prettier clean; `src/test-kernel/cli` 2294 pass.

LionSR:
Right on both counts — `deletionQueue` is one `PQueue({ concurrency: 1 })` for the whole instance, not per-stream lanes. Corrected in 1b26f62: the loop awaits one at a time because the queue already serializes every deletion, so firing them together would only queue them. The old wording would have invited exactly the parallelism the queue re-serializes.

LionSR:
Round three, in b535ff8.

**Unpinned descriptions vs the launch resolver — fixed properly this time.** You were right to keep pushing on this line. Rather than approximate the launch rule with a third resolver, the helper now calls `resolveAgentForLaunch(config.agentCategory, config.agent, config.agentSource)` — the function `getAgentPath` is a thin wrapper over (`AgentLaunchContext.ts:161`). The described purpose therefore belongs to the entry that actually ran *by construction*, and this line no longer has a resolution rule of its own to get wrong. That closes the pinned-source case, the category case, and your visible-roster-vs-priority case in one move.

**Missing coverage on the pinned branch — added.** Fair point given the history: `configFor` now takes an `agentSource`, and a new test asserts the resolver receives `('workflow', 'correct', 'custom')`. The existing workflow test asserts the unpinned call passes `undefined` explicitly, so both branches are pinned.

**Execution ownership when the sidecar FK is missing — not taking this one, and here's why.** The behavior you describe is real but it belongs to `deleteStream`, not to the sweep: `executionIdForStream` (`SessionStores.ts:266-274`) documents the invariant deliberately — *"A stream without either has no owned execution — name resemblance is never ownership, so no suffix derivation exists."* Recovering ownership from the execution row's stamped `streamId` is the reverse index that comment refuses, and it would need a scan of every execution dir. It also isn't specific to this sweep: the rail's delete button, `autoClose`'s `removeStream`, and `clearAll` all reach the same path with the same outcome, so fixing it here would leave three other callers leaking and add a second ownership rule beside the documented one.

The right shape is an execution-side orphan sweep keyed on `ExecutionMeta.streamId`, which would serve every caller. Happy to file that as a follow-up issue if you agree — it is a separate change with its own risk profile, not a fix to fold into this PR.

Verification: `tsc` clean; eslint and prettier clean; `src/test-kernel/{agent/runtime,agent/storage,progressView/ProgressBackendCleanup}` — 611 pass. (Local `progressView` Lit suites show 8 JSDOM `beforeAll` hook timeouts under load — identical counts with this branch's changes stashed, so pre-existing on this machine, and green in CI.)

PR #9887, opened 2026-08-08
fix(progress-view): offer the inline diff only where it can render

### LINKED ISSUE #9883: Fix <texra-diff-view> rendering empty in extension webview and trace viewer
## Context

Follow-up from #9881 (fix(build): stop shipping 12MB of unreachable Monaco workers in the VSIX).

`<texra-diff-view>` is a custom element registered only by the desktop renderer (`desktop/src/renderer/main.ts:24`). `ToolEditRequestPanel` renders that tag in the shared component, so in the **extension webview and trace viewer** expanding the inline diff yields an unregistered, empty element. This has been latent since the component landed in May 2026.

## Details

Two possible approaches:
1. **Register Monaco in the extension webview** — adds megabytes to the extension bundle, and `vscode.diff` already covers this host.
2. **Hide the toggle in hosts that cannot render it** — lighter-weight; the extension already has a working diff path through VS Code's own `vscode.diff`.

A design decision is needed before implementation.

## References

- PR: #9881
- Author's note: "so it wants its own issue rather than a drive-by here"

LionSR:
Closes #9883. Follow-up to #9881, which surfaced this while tracing where Monaco was reachable from.

## The bug

`<texra-diff-view>` is registered by the desktop renderer alone (`desktop/src/renderer/main.ts:24`). `ToolEditRequestPanel` renders that tag from the shared component library, so in the **extension webview** and the **trace viewer** — neither of which ships Monaco — expanding the inline diff produced an empty, unregistered element.

The worse half was silent. `hasInlineDiff` was true whenever the payload carried `originalContent`/`proposedContent`, and that branch *wins* over the `openDiff` action:

```ts
private handleDiffAction = (): void => {
  if (this.hasInlineDiff(data)) { this.inlineDiffOpen = !this.inlineDiffOpen; return; }
  this.emitAction({ action: 'openDiff' });   // <- VS Code's own diff, unreachable
};
```

So in the extension the diff button stopped opening `vscode.diff` and started toggling an empty box — for exactly the requests that had content to show. The button label followed suit ("Open inline diff (d)").

## The fix

The extension does not need a second diff surface; `vscode.diff` is the right one there. Gate the inline branch on the element actually being registered:

```ts
function inlineDiffSupported(): boolean {
  return customElements.get('texra-diff-view') !== undefined;
}
```

Desktop keeps the inline diff. The extension and trace viewer fall through to `openDiff` — which is what the button did before payloads started carrying content.

A function, not a module-scope constant, and the comment says why: this module is pulled in through the progressView barrel, which the desktop imports one line *before* it registers the element, so evaluating once at import time reads `false` on every host.

## Tests

The existing suite covered only the desktop shape without saying so — it asserted the inline diff renders while never registering `<texra-diff-view>`, i.e. it asserted the bug (`querySelector` happily returns an unregistered tag, and Lit sets properties on it either way). It now registers a stub and reads as the desktop suite it always was.

The extension/trace-viewer behavior gets its own file, `ToolEditRequestPanelWithoutMonaco.vitest.ts`, because a custom-element registry cannot un-define a tag once the other file has defined it. It asserts the button delegates to `openDiff`, renders no `texra-diff-view`, keeps delegating on a second press rather than toggling hidden state, and is labelled "Open diff".

I removed the gate and re-ran: all three new tests fail. Restored, and they pass.

## Verified

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm run check:dead-code-ratchet` — 16 findings vs 16 baselined, no new
- Full progressView suite — 57 files, 460 tests, all pass

## Note

This is behavior-visible in the extension: the diff button now opens the VS Code diff tab in cases where it previously toggled an empty inline panel. That is the intended restoration, not a regression.

---

> [!NOTE]
> **Low Risk**
> Shared UI branching change with clear host detection and dedicated tests; extension behavior is an intentional fix restoring VS Code diff, not a silent regression.
> 
> **Overview**
> **Fixes extension/trace-viewer diff button** when edit payloads include `originalContent`/`proposedContent`. The panel used to treat that as inline-diff mode and toggle an empty `<texra-diff-view>` instead of emitting `openDiff` for VS Code’s diff tab.
> 
> `hasInlineDiff` in `ToolEditRequestPanel` now returns false unless `customElements.get('texra-diff-view')` is defined (desktop only), so hosts without Monaco keep the “Open diff” label and delegate every press to `openDiff`.
> 
> Tests are split: the existing suite registers a stub element and documents desktop behavior; **`ToolEditRequestPanelWithoutMonaco.vitest.ts`** asserts delegation, labels, and no inline toggle when the tag is never registered.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc59af727812bd2bd77dcec4a462d2ec98f5f3ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/extension/src/progressView/frontend/components/ToolEditRequestPanel.ts
+++ b/packages/extension/src/progressView/frontend/components/ToolEditRequestPanel.ts
@@ -31,6 +31,25 @@ import { BaseBypassApprovalPanel } from './BaseBypassApprovalPanel';
 // Local imports - styles
 import { toolEditRequestPanelStyles } from './ToolEditRequestPanel.styles';
 
+/**
+ * Whether this host can render the inline Monaco diff.
+ *
+ * `<texra-diff-view>` is registered by the desktop renderer alone
+ * (`desktop/src/renderer/main.ts`). The extension webview and the trace viewer
+ * deliberately ship no Monaco — the extension opens the real diff through VS
+ * Code's own `vscode.diff`, which the `openDiff` action below reaches. Without
+ * this check the diff button toggled an unregistered, empty element in those
+ * hosts *and* shadowed the working path, since the inline branch wins whenever
+ * the payload carries content.
+ *
+ * Checked per call rather than once at module scope: this module is pulled in
+ * through the progressView barrel, which the desktop imports one line *before*
+ * it registers the element.
+ */
+function inlineDiffSupported(): boolean {
+  return customElements.get('texra-diff-view') !== undefined;
+}
+
 @customElement('tool-edit-request-panel')
 export class ToolEditRequestPanel extends BaseBypassApprovalPanel<'toolEdit'> {
   static override styles = [
@@ -205,6 +224,7 @@ export class ToolEditRequestPanel extends BaseBypassApprovalPanel<'toolEdit'> {
   };
 
   private hasInlineDiff(data: ToolEditPermission): boolean {
+    if (!inlineDiffSupported()) return false;
     return data.originalContent != null || data.proposedContent != null;
   }
 }
--- a/packages/extension/src/progressView/frontend/components/ToolEditRequestPanel.ts
+++ b/packages/extension/src/progressView/frontend/components/ToolEditRequestPanel.ts
@@ -31,25 +31,6 @@ import { BaseBypassApprovalPanel } from './BaseBypassApprovalPanel';
 // Local imports - styles
 import { toolEditRequestPanelStyles } from './ToolEditRequestPanel.styles';
 
-/**
- * Whether this host can render the inline Monaco diff.
- *
- * `<texra-diff-view>` is registered by the desktop renderer alone
- * (`desktop/src/renderer/main.ts`). The extension webview and the trace viewer
- * deliberately ship no Monaco — the extension opens the real diff through VS
- * Code's own `vscode.diff`, which the `openDiff` action below reaches. Without
- * this check the diff button toggled an unregistered, empty element in those
- * hosts *and* shadowed the working path, since the inline branch wins whenever
- * the payload carries content.
- *
- * Checked per call rather than once at module scope: this module is pulled in
- * through the progressView barrel, which the desktop imports one line *before*
- * it registers the element.
- */
-function inlineDiffSupported(): boolean {
-  return customElements.get('texra-diff-view') !== undefined;
-}
-
 @customElement('tool-edit-request-panel')
 export class ToolEditRequestPanel extends BaseBypassApprovalPanel<'toolEdit'> {
   static override styles = [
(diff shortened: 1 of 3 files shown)


LionSR:
Addressed in cc59af7278.

Both reviewers flagged `inlineDiffSupported()` as a single-caller extraction, and they're right — CLAUDE.md bans those outright, and checklist §13 repeats it. The justification I gave for the helper doesn't survive scrutiny either: as `claude` pointed out, the per-call timing constraint holds either way, since `hasInlineDiff` runs at interaction time rather than module scope. The extraction bought nothing.

Folded into `hasInlineDiff`, with the docstring cut from 10 lines to the two facts a reader actually needs (who registers the element, and why it's read per call):

```ts
private hasInlineDiff(data: ToolEditPermission): boolean {
  // Only the desktop registers <texra-diff-view>; the extension and trace
  // viewer ship no Monaco and fall through to openDiff (VS Code's own diff).
  // Read per call, not at module scope: the desktop imports the progressView
  // barrel — and so this module — one line before it registers the element.
  if (customElements.get('texra-diff-view') === undefined) return false;
  return data.originalContent != null || data.proposedContent != null;
}
```

No behavior change. Re-ran after the edit: typecheck clean, lint clean, both panel suites pass (11 tests).

LionSR:
**Correcting this PR: the bug it describes cannot occur. I got it wrong.**

The repo owner pushed back that the extension's VS Code diff tab already worked, and they're right. Tracing it properly:

`prepareToolEditApprovalPrompt` (`src/tools/approval/toolEditApproval.ts:140-150`) is the only builder of the `ToolEditPermission` payload the progress view receives, and it never sets the content fields:

```ts
return {
  requestId, path, relativePath, sourceTool,
  allowBypass, streamId, addedLines, removedLines, isLatex,
};
```

Both hosts pass it through unchanged (`ProgressViewProvider.ts:174`, `desktopAgentExecution.ts:362`). `originalContent` appears in all of `src/controllers/` exactly once — on the controller's *internal* entry (`ToolEditApprovalController.ts:174`) — and never reaches the webview.

So `hasInlineDiff` is **always false, in every host**. Consequences:

1. The diff button always emitted `openDiff`, which works in both hosts by different routes — extension: `VscodeDiffViewHost` → `vscode.diff` tab; desktop: `desktopDiffHost` → `desktop:showDiff` → the Review workbench's Monaco view. Nothing was ever shadowed.
2. #9883's empty element is **unreachable** — the inline branch has never executed.
3. The gate this PR adds is therefore a runtime no-op.

The inline `<texra-diff-view>` inside `ToolEditRequestPanel` is an unfinished feature, not a broken one: optional schema fields exist, the UI was written, the producer was never wired.

I should have traced the payload producer before writing the bug description rather than reasoning from the branch logic alone. Holding this PR pending a call on the right cleanup — most likely deleting the unreachable inline path outright, which would supersede this change entirely.

PR #9892, opened 2026-08-08
refactor: delete the unreachable inline diff, move TexraDiffView to desktop

### LINKED ISSUE #9883: Fix <texra-diff-view> rendering empty in extension webview and trace viewer

LionSR:
Supersedes #9887. Closes #9883.

## The inline diff never ran

`ToolEditRequestPanel`'s inline Monaco diff was gated on `hasInlineDiff`, which needs `originalContent`/`proposedContent` on the `ToolEditPermission` payload. The only builder of that payload — `prepareToolEditApprovalPrompt` (`src/tools/approval/toolEditApproval.ts:140-150`) — never sets them:

```ts
return {
  requestId, path, relativePath, sourceTool,
  allowBypass, streamId, addedLines, removedLines, isLatex,
};
```

Both hosts pass it through untouched (`ProgressViewProvider.ts:174`, `desktopAgentExecution.ts:362`), and `originalContent` appears in all of `src/controllers/` exactly once — on the controller's internal entry, which never reaches the webview. So the branch was dead in every host from the day it landed.

**What actually works, and is untouched here:** the diff button emits `openDiff`, and each host answers with its own surface — the extension opens a VS Code diff tab (`VscodeDiffViewHost` → `vscode.diff`), the desktop posts `desktop:showDiff` to its Review workbench (`desktopDiffHost`). That's the whole story. The inline panel was a third path nobody could reach.

This corrects the premise of #9887, which I opened against a bug that cannot occur: it added a registration guard around an unreachable branch. Deleting the branch supersedes it.

## Changes

- Delete `inlineDiffOpen`, `renderInlineDiff`, `hasInlineDiff` and its label/tooltip branching. The button is now unconditionally "Open diff" and always delegates.
- Drop `originalContent`/`proposedContent` from `ToolEditPermissionSchema` — no producer, and the deleted branch was their only consumer. (The identically-named fields on the tool-side `ToolEditApprovalRequest`, which the CLI and `VscodeToolEditApprovalHost` do read, are a different type and untouched.)
- With the last shared reference gone, `TexraDiffView` is desktop-only (`renderer/main.ts`, `reviewPane.ts`), so it moves to `packages/desktop/src/renderer/`, test alongside it in `test-kernel/desktop/`.

## The one new file

`packages/desktop/src/renderer/tsconfig.json`. Per-file transpilers — esbuild under Vite and Vitest — resolve a tsconfig by walking up from each source file. The desktop package's `tsconfig.json` scopes `include` to `vite.config.ts`, and the repo-root tsconfig **excludes** `packages/desktop` outright, so a Lit `@customElement` class living there reached the runtime with its decorator syntax untranspiled and failed to parse (`SyntaxError: Invalid or unexpected token`).

The new file is a nearest-tsconfig marker that extends `tsconfig.renderer.json` — the config tsc already uses for these sources. Build scripts still pass their own configs with `-p`, so nothing else changes. This is the first decorator-based component in that package, which is why the gap never surfaced.

I tried two narrower fixes first and reverted both: explicit `experimentalDecorators` on the package tsconfig (not picked up — its `include` doesn't cover `src/`), and a global `esbuild.tsconfigRaw` in the Vitest config (no effect on the module-runner path, and a blunt global override).

## Verified

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm run check:dead-code-ratchet` — 16 vs 16 baselined
- `npx prettier --check` on every touched file — clean
- `node scripts/check-guidance-refs.mjs` — OK (the moved file isn't referenced in docs)
- progressView + desktop suites — 114 files, 1032 tests, all pass
- `pnpm --filter @texra/desktop run build:renderer` — succeeds with the component in its new home

Net −164 lines.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to removing dead UI and unused schema fields; host `openDiff` handling is unchanged. Desktop renderer tsconfig only affects transpilation of new Lit components in that folder.
> 
> **Overview**
> Removes the **inline Monaco diff** from `ToolEditRequestPanel` because approval payloads never included `originalContent`/`proposedContent`, so that UI path was unreachable. The diff control is now always **Open diff** and emits `openDiff` for each host (VS Code diff tab or desktop Review workbench).
> 
> **Schema and shared code:** `originalContent`/`proposedContent` are dropped from `ToolEditPermissionSchema`; tool-side approval types that still use those fields are unchanged.
> 
> **Desktop-only Monaco:** `TexraDiffView` moves from the progressView package to `packages/desktop/src/renderer/` with imports updated in `main.ts` and `reviewPane.ts`. A new `renderer/tsconfig.json` extends the renderer tsconfig so Lit decorators transpile under Vite/Vitest when walking up from renderer sources.
> 
> **Tests:** Inline-diff and no-Monaco panel tests are consolidated; the smoke webview fixture no longer seeds inline diff content.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 050b39a6ef42f856d7871ca082ea34d6e5497a81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
Both findings addressed in bde223dfe7 — both were real.

**Orphaned `state` import.** `@state()` was only on the deleted `inlineDiffOpen`, so `import { customElement, state } from 'lit/decorators.js'` kept a dead specifier. The reviewer's point about why it slipped through is the useful part: `@typescript-eslint/no-unused-vars` is off repo-wide and no tsconfig here sets `noUnusedLocals`, so neither `npm run lint` nor `npm run typecheck` — the two gates I cited — would ever have flagged it. Dropped.

**Stale ratchet entry.** `config/ratchets/shared-schemas-deep-import-baseline.json` still listed `TexraDiffView.ts` at its old extension path. The ratchet stayed green because it only fails on *new* entries, so a pure rename leaves behind a path that quietly no longer matches the file it describes. Repointed to `packages/desktop/src/renderer/TexraDiffView.ts`.

One note on how: I ran the documented `TEXRA_UPDATE_SHARED_SCHEMAS_BASELINE=1` regen first and reverted it — it rewrote 120 lines, sweeping unrelated drift from other branches into this PR. The one-line hand edit is the honest change. Entry order follows the file's locale-aware sort (`main.ts` before `TexraDiffView.ts`), which the ratchet's own sorted-and-distinct assertion checks.

Verified after the change: typecheck clean, prettier clean, architecture ratchets pass (12 files, 82 tests), panel suite passes.

Also worth recording from the earlier CI round: the webview smoke failure was `ToolEditPermissionSchema` being a `z.strictObject` — the removed fields made the fixture's whole `updatePermission` message fail validation, so the panel never mounted. That is the schema doing its job, and it is the reason nothing can quietly keep sending those fields.

PR #9906, opened 2026-08-09
refactor(runtime): centralize failure-aggregation rethrow in throwAggregated

LionSR:
## Summary

Several shutdown/flush paths independently hand-rolled the same tail: given a
list of collected failures, rethrow the lone error unwrapped or wrap several in
an `AggregateError`. This consolidates that idiom into two leaf helpers in
`@utils/core` and adopts them at each site. Behavior-preserving.

- `aggregateError(failures, message)` — the *reducer*: returns the value to
  throw (lone error unwrapped, else an `AggregateError`). Use as
  `throw aggregateError(...)` where a failure is guaranteed, or to build a `cause`.
- `throwAggregated(failures, message)` — the *statement*: no-op on empty, else
  throws `aggregateError(...)`. For "collected N≥0 failures, surface them" tails.

This PR is the outcome of a repo-wide tech-debt scan. The scan's headline finding
is that TeXRA is already unusually well-factored — the obvious duplication
(`classifyMediaEntry`, `mutateWithOverlay`, `applyTokenCountLimit`,
`handleStreamingFailure`, `ResidentStreamRegistry`, …) has been extracted, and
most remaining look-alike code is either irreducibly provider/field-specific or
guards subtle race/format invariants. This failure-aggregation idiom was the one
clean, safe, net-positive consolidation that survived close reading. See
**Scheduled refactoring** for the larger candidates that were deliberately *not*
taken and why.

## Issue acceptance gates

No tracking issue — this originates from a scheduled tech-debt audit. No
acceptance gates apply.

## Single source of truth

`aggregateError` / `throwAggregated` in `@utils/core` (`src/utils/core/index.ts`)
now own the "reduce a collected failure list to the correct throw" decision. This
absorbs the previously-private `toFailure` reducer in `src/transcript/runTrace.ts`
(identical logic, 4 call sites), which is deleted in this PR — so both the
reduce-shape and the throw-shape have exactly one home.

Removed: the `if (len === 1) throw failures[0]; if (len > 1) throw new
AggregateError(...)` idiom from `SessionHandle.flushPendingTraces`,
`SessionHandle.flushArtifactsOnce`, `SessionHandle.dispose`, and
`StreamSnapshotStore.flush`; the redundant `if (cleanupErrors.length > 0)` branch
plus a now-unreachable `throw error;` from the `childRunLoop` setup-rollback; and
the private `toFailure` reducer from `runTrace.ts`.

The new abstractions are leaf pure functions, not pass-through layers.
`aggregateError` owns *non-empty → unwrapped-or-aggregate*; `throwAggregated`
adds the *empty → no-op* statement wrapper.

## Net elements (R6)

- Files: 5 changed. Non-test LoC roughly flat; the win is removing ~6 copies of
  an easy-to-get-subtly-wrong idiom, not raw line reduction.
- Exported symbols (`^[+-]export` over the diff): +2 (`aggregateError`,
  `throwAggregated`); 0 removed. One private function (`toFailure`) deleted.
- Added: direct unit tests for both exports in `UtilsCore.vitest.ts`.

## Consumer counts (R8)

No emit path or existing export was deleted or re-routed. Two new exports added,
both with in-PR consumers (`check:dead-code-ratchet` passes: 16 vs 16 baselined):
`throwAggregated` — 4 (`SessionHandle` ×3, `StreamSnapshotStore`);
`aggregateError` — 5 (`runTrace` ×4, `childRunLoop`). n/a otherwise.

## Host impact

Extension: none — shared core only; no behavior change.

Desktop: none — shared core only; no behavior change.

CLI: none — shared core only; no behavior change.

## Scheduled refactoring

The audit's larger candidates were analyzed and **deliberately deferred**, not
missed. None is filed as a follow-up issue yet; recommend the two high-value ones
be done interactively with the maintainer given their risk profile:

- **Unify the two terminal-finalization state machines** (`executionRegistry`
  waiting-stop cluster vs `AgentRunLifecycle.finalizeRunTerminal`, ~80–120 LoC).
  Load-bearing shutdown path with lease-lost / generation-handoff arms; needs
  dedicated test coverage before merging — unsafe to land autonomously.
- **Templatize the per-provider `createResponseImpl` pipeline** (~120–180 LoC).
  Touches all five provider hot paths with subtle retry/compaction invariants;
  same caveat.
- **Table-drive the `StreamSnapshotStore` accumulators** — on close reading the
  store is already factored via `mutateWithOverlay`, and the residual enumeration
  guards a pre-await usage-overlay capture (#8226-class race) plus a
  `missingOutputs` reset special case. A faithful table saves little for real
  regression risk; not recommended as-is.
- **`createMediaContent` / `appendTextToLastAssistantMessage` base templates** —
  same *shape* across providers but the bodies differ in every guard, block
  constructor, and pop condition. A base template nets ~even on line count while
  reducing per-provider readability (shallow indirection). Not recommended.
- **Zod `.catch()` logging on persisted state** — the flagged sites
  (`streamSnapshot.ts`, `RunUsageAccumulator.ts`, `taskGroup.ts`) are documented,
  deliberate per-field resilience in the VS Code-free `src/shared/schemas` zone;
  adding a logger there is a layering concern and the "defect" is borderline. Left
  for a focused decision.

## Validation

- `tsc --noEmit --checkers 8` (workspace + `tsconfig.test-kernel.json`): pass.
- `vitest run` over `utils`, `transcript`, `agent/runtime`, and `agent/storage`
  suites: 765 passed (includes new `aggregateError` / `throwAggregated` tests).
- `eslint` on all changed files: pass.
- `npm run check:dead-code-ratchet`: pass (no new unused exports).

---

> [!NOTE]
> **Low Risk**
> Refactor-only consolidation of error-throwing logic with direct unit tests; no new runtime behavior beyond matching the previous aggregation rules.
> 
> **Overview**
> Adds **`aggregateError`** and **`throwAggregated`** in `@utils/core` to replace repeated “one failure → throw it; many → `AggregateError`” tails across session shutdown, snapshot flush, run trace, and child-run rollback.
> 
> **`SessionHandle`** (trace flush, artifact flush, dispose), **`StreamSnapshotStore.flush`**, and **`runTrace`** (deleting private **`toFailure`**) now call these helpers. **`childRunLoop`** setup rollback uses **`aggregateError`** so the primary error always leads the list, removing a redundant branch and unreachable **`throw error`**.
> 
> Unit tests cover both exports in **`UtilsCore.vitest.ts`**. Behavior is intended to stay the same.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 76e75902ddd273cfebe2568a9f8a7cb0334d203b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/agent/runtime/SessionHandle.ts
+++ b/src/agent/runtime/SessionHandle.ts
@@ -50,6 +50,7 @@ import {
 import type { RunTraceFlushEntry } from '@transcript/runTrace';
 import type { StreamLogStore } from '@transcript/StreamLogStore';
 import { StreamSnapshotStore } from '@transcript/StreamSnapshotStore';
+import { throwAggregated } from '@utils/core';
 import { getRunContextSession, tryUseRunContext } from './RunContext';
 import { ExecutionRegistry } from './executionRegistry';
 import { ExecutionSubscriptionBinder } from './ExecutionSubscriptionBinder';
@@ -673,13 +674,7 @@ export class SessionHandle {
         failures.push(error);
       }
     }
-    if (failures.length === 1) throw failures[0];
-    if (failures.length > 1) {
-      throw new AggregateError(
-        failures,
-        'Multiple session trace writers failed to flush',
-      );
-    }
+    throwAggregated(failures, 'Multiple session trace writers failed to flush');
   }
 
   /** Register a session-owned durable writer such as a snapshot store. */
@@ -749,13 +744,10 @@ export class SessionHandle {
     const failures = results.flatMap((result) =>
       result.status === 'rejected' ? [result.reason] : [],
     );
-    if (failures.length === 1) throw failures[0];
-    if (failures.length > 1) {
-      throw new AggregateError(
-        failures,
-        'Multiple session artifact writers failed to flush',
-      );
-    }
+    throwAggregated(
+      failures,
+      'Multiple session artifact writers failed to flush',
+    );
   }
 
   /** Listeners for terminal run results in this session (the host channel). */
@@ -873,10 +865,7 @@ export class SessionHandle {
     } finally {
       liveSessions.delete(this);
     }
-    if (failures.length === 1) throw failures[0];
-    if (failures.length > 1) {
-      throw new AggregateError(failures, 'Session teardown failed');
-    }
+    throwAggregated(failures, 'Session teardown failed');
   }
 
   /** Dispose the runtime owners and drop result listeners. */
--- a/src/agent/runtime/childRunLoop.ts
+++ b/src/agent/runtime/childRunLoop.ts
@@ -56,7 +56,7 @@ import type {
   ChildStream,
   ChildStreamOutcome,
 } from '@tools/delegation/childStream';
-import { formatDuration } from '@utils/core';
+import { formatDuration, throwAggregated } from '@utils/core';
 import { toErrorMessage } from '@utils/errors/errorMessage';
 
 /** Minimal token usage shape consumed by the loop's turn summary. */
@@ -709,12 +709,12 @@ export function startChildRunLoop<TTurn>(
     cleanup(() => stopWatchingLease?.());
     cleanup(releaseChildActivation);
     cleanup(releaseSessionOwnershipOnce);
-    if (cleanupErrors.length > 0) {
-      throw new AggregateError(
-        [error, ...cleanupErrors],
-        `Child run ${executionId} setup failed and rollback was incomplete`,
-      );
-    }
+    // Non-empty by construction (the primary error is always first): one
+    // failure rethrows `error` unwrapped, several wrap into an AggregateError.
+    throwAggregated(
+      [error, ...cleanupErrors],
+      `Child run ${executionId} setup failed and rollback was incomplete`,
+    );
     throw error;
   }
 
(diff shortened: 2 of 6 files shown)


LionSR:
Fixed in `68935e8`, thanks all.

**Unreachable `throw error;`** (Cursor Bugbot, texra-ai Finding 1, claude[bot]) — confirmed. Note a bare deletion wouldn't typecheck: `throwAggregated` returns `void`, so TS would treat the `catch` as able to fall through. Resolved by adding a value-returning sibling `aggregateError(failures, message)` and rethrowing via `throw aggregateError([error, ...cleanupErrors], …)` — a single throw, no dead code, and TS sees the block terminates.

**texra-ai Finding 2** (`toFailure` "does not exist") — that was a false positive: it existed at `src/transcript/runTrace.ts:44` as a private reducer identical to this idiom. This commit folds it into the new shared `aggregateError` (4 call sites in `runTrace.ts`), so the reduce-shape now has one home too. The PR description has been updated to reference `aggregateError`.

Also added direct unit tests for both `aggregateError` and `throwAggregated` in `UtilsCore.vitest.ts` (per claude[bot]'s coverage note). Full suite green: workspace + test-kernel typecheck, 765 tests, eslint, and the dead-code ratchet.

---

Issue #9945, opened 2026-08-10
[cli] own stream-log eviction at the lifecycle boundary; drop the process-global task-group projection cache


LionSR:
Part of the 2026-08-10 long-lived-session memory investigation. Two related **net-negative-element** simplifications in the CLI transcript sync. Both are self-contained, CLI-only, no resume/recovery or data-loss surface.

## A. Own stream-log eviction at the lifecycle boundary

Today `requestEviction`'s only production caller is `subscribeStreamLog.ts:996`, buried inside the projection closure. Release is a **side-effect of rendering**: a terminal background stream only gets released if it happens to be synced. The decision (`releaseAfterSync`, `subscribeStreamLog.ts:828,946-949`) reads three facts — not-the-active-stream, status set, `!isActivePhase` — all of which the status subscriber already has in its own context.

Move the release to the status subscriber (`subscribeStreamStatus.ts`), which fires on the terminal transition. `StreamLogStore.requestEviction` (`src/transcript/StreamLogStore.ts:473`) already queues `pendingRelease` when writers/flushing block, so a direct lifecycle call is safe.

- Deletes `releaseAfterSync` init + decide + act (`subscribeStreamLog.ts:828,946-949,996`) from the renderer.
- Fixes the coupling where `syncStreamLog`'s early returns implicitly govern whether release fires.
- Terminal streams always release, not just when a sync happens to run.

## B. Drop the process-global `TASK_GROUP_PROJECTIONS` cache

`TASK_GROUP_PROJECTIONS` (`subscribeStreamLog.ts:682`) is a process-global incremental memo for task groups — but the whole-log walk is **already paid for every tick** (`log.getRange(0)` at `:810`). Replace `projectTaskGroupsIncrementally` with the already-exported from-scratch `projectTaskGroupsFromStreamLog(allEntries)` (`src/shared/streams/taskGroupProjection.ts:152`).

- Removes a process-global `Map` + its reset-hook wiring + the `{working,index,applied,snapshot}` per-stream state.
- Removes a (minor) retention leak: `applied` holds strong `StreamLogEntry` refs that survive until `/clear`/removal.
- Cost: one cheap re-walk per tick + change the `slice.taskGroups === taskGroups` ref check (`subscribeStreamLog.ts:981`) to a content compare or always-assign.

Evidence and the consumer map in `docs/dev/audits/2026-08-10-long-lived-session-memory-investigation.md` (Finding 5 + Finding 6 verification).

[...]

PR #9967, opened 2026-08-11
Store-emitted transcript delta feed + O(delta) CLI projection fold

### LINKED ISSUE #9946: [cli] incremental transcript projection to drop the O(total) per-tick rescan
Part of the 2026-08-10 long-lived-session memory investigation.

`syncStreamLog` (`packages/cli/src/chat/tui/state/subscribeStreamLog.ts:800-996`) rebuilds the entire transcript projection every 16 ms tick from `log.getRange(0)` — filter → render → sort → finalize-settled-prefix → re-filter — even when a single entry changed. Per-tick cost is **O(total transcript size)**, not O(delta). The `entriesUnchanged` deep-equal (`:967-983`) only suppresses the signal write; the O(total) render/filter/sort work already ran. This is allocation pressure proportional to resident history, the strongest temporary-allocation candidate in the investigation.

## Proposal: cursor-based incremental projection

Keep one `lastProjectedSeqNo` per active stream; each tick process only entries appended/changed since.

- Collapses O(total)/tick → O(delta)/tick.
- Deletes the O(total) `entriesUnchanged` deep-equal pass (`:967-983`) — itself a full rescan.
- Deletes the unconditional `compactWorkflowDashboardIds` Set build (`:954-959`), computed for every stream every tick but used only for inactive workflow streams.
- Folds the ~7 throwaway intermediates (`existing`, `logCandidates`, `candidates`, `ordered`, …) into one incremental pass.

**Risk: medium.** The workflow-operational branch is entangled in the pipeline (full vs compact projection, synthetic-merge ordering, settled-prefix finalization). Needs care to preserve ordering semantics.

Evidence in `docs/dev/audits/2026-08-10-long-lived-session-memory-investigation.md` (Finding 2).

LionSR:
Fixes #9946 (rescoped per the issue comments and the PRD in `docs/prds/2026-08-11-transcript-memory-architecture.md`, top-5 item 1). This PR lands the store-emitted delta feed and the O(delta) CLI fold. Per the sequencing decision, the extension `WebviewBridge` port (which deletes its private cursor and the destructive ack protocol) is the follow-up PR; the ack machinery is untouched here.

## Store side: one authoritative change channel

- `StreamLog` now tracks, per notification, the entries appended and dirtied since the previous emission. `StreamLogStore.notify` drains that exactly once into an immutable `StreamLogDelta { emissionSeq, appended, dirtied, reset }` and multicasts it to every `onChange` listener. Dirtied entries are carried by value: `StreamLog` replaces the entry object on every mutation, so the payload can never be rewritten after emission (covered by a test).
- `emissionSeq` is monotonic per log instance; `StreamLog.instanceId` distinguishes evicted-and-rehydrated instances. The disk-merge path in `ensureLoaded` (the one place seqNos get renumbered under a live consumer) emits `reset: true`.
- `StreamLogDeltaBuffer` is the consumer-side coalescer: contiguous bursts merge (a dirtied value supersedes its buffered appended value), and a gap, a reset, or a stale sequence marks the buffer as requiring resync.
- No consumer acks anything and nothing is destroyed. The existing `dirtyUpdates`/`dirtyTextDeltas` ack machinery stays as-is for the bridge until the follow-up. The `StreamLogStore` public surface is unchanged (the emission counter lives on `StreamLog`, keeping the store-surface ratchet at its baseline).

## CLI side: the projection is a fold

`syncStreamLog`'s per-tick full pipeline (`getRange(0)` rescan, per-entry render walk with a reference cache, sort, settled-prefix recompute, the `:1021` filter/slice/map/`Set` chain that was `error8`'s fatal allocation, and the O(total) deep-equal) is replaced by a fold over buffered deltas:

- Per-stream `TranscriptFoldState` lives on the stream slice, so its lifetime is exactly the stream's: it dies with removal and CLI reset, and is cleared on the existing release path. It maintains the merged final transcript order (log rows, compaction rows, synthetic rows under one ordering key with a source-rank tiebreak), the contiguous settled-prefix promotion frontier, and the latest-line cursors incrementally.
- Change detection now comes from the producer: the deep-equal is gone, and the compact workflow dashboard selection is rebuilt only when a delta actually touches a dashboard/synthetic row, not per tick per stream. Streaming text chunks on unfocused streams typically produce no signal write at all.
- Order-dependent semantics are preserved exactly: settled-prefix finalization is defined on the final sorted order, synthetic rows merge by `syntheticAfterSeq`, workflow-operational filtering is unchanged, and `releaseAfterSync` keeps working (its relocation is #9945).
- The consumer contract from the PRD: a fresh consumer, an emission gap, a reset emission, a projection-mode flip, or a mutation that bypassed the store rebuilds from `getRange(0)` through the same application code as the fold. The from-scratch build IS the production resync path.
- `projectStreamTranscript`'s finalize step now runs through the same pipeline (`forceFinal`), so promotion can no longer diverge from the fold state; the no-resident-log case keeps the previous direct `finalizeSettledPrefix` behavior.

`TASK_GROUP_PROJECTIONS` and `COMPACTION_PROJECTIONS` keep their current shape (their relocation is #9945's scope); they are now fed the delta / the log tail instead of a full `getRange(0)` array.

## Tests (the gate)

- **Fold-vs-oracle equivalence property test** (`src/test-kernel/cli/StreamLogDeltaFold.vitest.ts`): mirrored streams receive an identical seeded randomized interleaving of user/model/thinking/tool appends, in-place updates and settles, GROUP_START to GROUP_END upserts (phase, round, untyped), workflow-call transitions, malformed and valid skills snapshots, error rows, synthetic inserts, and stream-final status flips in both directions. One stream folds every emission; the other has its state invalidated before every sync so it always rebuilds from scratch. After every emission both must project identically (entries, latestLine, thinkingActive, compactingActive, activeSkills, taskGroups), across three projection modes (tool-use transcript, workflow operational feed, workflow full-log child) and the compact unfocused-workflow selection, with a counter assertion proving the fold path actually ran.
- **Gap recovery**: deltas dropped while unsubscribed produce a real gap on resubscribe; the sync must rebuild (asserted via counters) and converge on the full transcript.
- **Store emission tests** (`src/test-kernel/transcript/StreamLogDelta.vitest.ts`): appended/dirtied-by-value payloads, multi-mutation commit coalescing, no-op update silence, direct-mutation detection, and the buffer's merge/gap/stale/reset rules.

## Element accounting

Deleted: the per-tick `getRange(0)` rescan, `dedupeEntry`/`entriesEqual`/`toolUseEqual`/`inputEqual`, the `CachedLogRender`/`StreamRenderState`/`STREAM_RENDER_STATES` render cache and its `renderLogEntry` wrapper, `TranscriptCandidate`/`compareTranscriptCandidates`/`sortTranscriptCandidatesIfNeeded`, `latestLogActivityIsThinking`, the `workflowLatestLineCandidates` second pipeline with its `toReversed()` copy, the per-tick dashboard-id `Set` build, and the O(total) deep-equal block. Added: the delta type + buffer, and the fold engine (items maintenance, frontier, cursors). Net production LoC is positive in this PR (about +660 before tests): the PRD's net-negative accounting includes the WebviewBridge cursor + ack deletion (about 270 LoC) that the task explicitly deferred to the follow-up, and the ack machinery retained here is on that same follow-up's deletion list.

## Known behavior deltas (deliberate, noted for review)

- A content-equal re-render of a dirtied row produces a new row object instead of being deduped to the previous reference (one extra row-level React render on producer-reported changes; `update()` already suppresses true no-ops at the store).
- Fold state retains full rows for unfocused streams, so `finalized` stickiness survives focus flips that previously lost promotion inheritance for rows dropped from the compact selection (monotone promotion, strictly less rollback than before).
- `forceFinal` promotion now updates the latest-line cursors in the same application instead of one sync later.

## Verification

- `npm run typecheck`, `npm run lint`, `npm run check:dead-code-ratchet` all clean.
- `npx vitest run src/test-kernel/cli/ src/test-kernel/transcript/ src/test-kernel/architecture/` (174 files, 2671 tests) and `src/test-kernel/controllers/` (37 files) all pass.

https://claude.ai/code/session_01Gtc6wQMD1hedgabjQMfAun

---

> [!NOTE]
> **Medium Risk**
> Changes core transcript notification and CLI projection semantics (ordering, finalization, compact selection); regressions would show wrong or stale TUI transcripts, but the fold is guarded by property tests and explicit gap/forceFinal cases.
> 
> **Overview**
> Adds an **entry-level change feed** on `StreamLog` / `StreamLogStore`: each `onChange` now multicasts an immutable `StreamLogDelta` (`emissionSeq`, appended rows, dirtied rows by value, optional `reset` after disk merge / instance replacement). `StreamLogDeltaBuffer` coalesces bursts and flags gaps that require a full resync.
> 
> The CLI transcript projection in `subscribeStreamLog` **folds** those deltas into per-stream `TranscriptFoldState` (merged log, compaction, and synthetic order; settled-prefix promotion; latest-line cursors) instead of rescanning `getRange(0)` every tick. Rebuild from `getRange(0)` is the same code path when continuity breaks (fresh consumer, gap, reset, mode flip, direct log mutation). Compact workflow / synthetic `entries` are rebuilt only when deltas touch relevant rows, not via per-tick deep equality.
> 
> `projectStreamTranscript` turn-boundary finalize goes through `syncStreamLog` with `forceFinal`; compaction settlement still follows real lifecycle settlement, not `forceFinal` alone.
> 
> New tests: store delta emission, buffer rules, fold-vs-oracle property equivalence, gap recovery, and compaction + `forceFinal` regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f0b4d3a8e9256dbe713152ebb9db67a5fce8b76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/tui/state/cliState.ts
+++ b/packages/cli/src/chat/tui/state/cliState.ts
@@ -19,6 +19,7 @@ import {
   type Plan,
   type RoundIndexed,
   type RunIdentity,
+  type StreamLogEntry,
   type StreamPhase,
   type StreamStage,
   type StreamSubstate,
@@ -130,6 +131,70 @@ export type ConversationEntry = ConversationEntryOrigin &
       })
   );
 
+/**
+ * One transcript-projection candidate: a rendered row plus the ordering key
+ * that places it in the final merged transcript order (log rows by seqNo,
+ * compaction rows by start position, CLI-synthetic rows by their insertion
+ * anchor). `rank` preserves the relative order of equal keys across the three
+ * sources. `rendered` is replaced in place when the source row changes or the
+ * settled-prefix promotion reaches it; the item object itself is stable.
+ */
+export interface TranscriptFoldItem {
+  rendered: ConversationEntry;
+  readonly sortSeq: number;
+  readonly tieBreak: number;
+  /** Equal-key source order: 0 = compaction row, 1 = log row, 2 = synthetic. */
+  readonly rank: 0 | 1 | 2;
+  /** Compaction rows: the block `rendered` was built from; reference equality
+   *  means the row is current. */
+  block?: CompactionActivityBlock;
+}
+
+/**
+ * Mutable per-stream transcript-projection working state. Carried on the
+ * stream's slice (never rendered) so its lifetime is exactly the stream's:
+ * it dies with stream removal and CLI-state reset, and is cleared when the
+ * stream's transcript is released. `subscribeStreamLog`'s fold over
+ * store-emitted `StreamLogDelta`s maintains it in O(delta); a fresh or gapped
+ * consumer rebuilds it from `getRange(0)` through the same application path.
+ */
+export interface TranscriptFoldState {
+  /** False until a rebuild has run; cleared again when the stream's transcript is released. */
+  hydrated: boolean;
+  /** The `StreamLog` instance + emission seq `items` reflects. A mismatch
+   *  with the store's current log means fold continuity is broken: rebuild. */
+  logInstanceId: number;
+  emissionSeq: number;
+  /** Final transcript order: log, compaction, and synthetic rows merged. */
+  readonly items: TranscriptFoldItem[];
+  readonly indexById: Map<string, number>;
+  /** First index the contiguous settled-prefix promotion has not covered. */
+  finalizedFrontier: number;
+  /** Index of the last user row with text, or -1 (latest-line fallback). */
+  latestUserPos: number;
+  /** Index of the last finalized model-response row with text, or -1. */
+  latestResponsePos: number;
+  /** Projection-mode bits `items` was built under; a flip forces a rebuild. */
+  fullLogChild: boolean;
+  workflowOperationalOnly: boolean;
+  projectLifecycleToTaskGroups: boolean;
+  /** Highest-seq ACTIVE_SKILLS entry, with its parse cached by reference. */
+  activeSkillsEntry?: StreamLogEntry;
+  activeSkillsParsedFor?: StreamLogEntry;
+  activeSkills: readonly ActiveSkillSummary[];
+  /** Highest-seq live-activity entry (drives the thinking indicator). */
+  liveActivityEntry?: StreamLogEntry;
+  /** Synthetic rows reconciled into `items`, in slice order, by identity. */
+  synthetics: readonly ConversationEntry[];
+  /** Whether the last emitted `entries` was the full transcript or compact;
+   *  undefined until the first emission. */
+  lastOutputFull?: boolean;
+  /** The exact `entries` array last emitted. A slice whose entries no longer
+   *  match was patched out of band (synthetic rows), so the next application
+   *  must rebuild its output instead of reusing `slice.entries`. */
+  lastEntriesOutput?: readonly ConversationEntry[];
+}
+
 export interface SessionMeta {
   readonly agent: string;
   readonly category: AgentCategory;
@@ -224,6 +289,9 @@ export interface StreamSlice {
    *  to fall back between. */
   readonly stage?: StreamStage | undefined;
   readonly entries: readonly ConversationEntry[];
+  /** Transcript-projection working state (see {@link TranscriptFoldState}).
+   *  A mutable box owned by `subscribeStreamLog`; renderers ignore it. */
+  readonly transcriptFold?: TranscriptFoldState;
   readonly queuedFollowUpMessages: readonly string[];
   readonly activeSkills: readonly ActiveSkillSummary[];
   readonly todos: readonly TodoItem[];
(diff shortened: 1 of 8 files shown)


LionSR:
Re the review finding at `src/transcript/StreamLog.ts:190` (delta feed coexists with the ack-based dirty channel): addressed via the checklist's allowed deferral. §13/§14 requires same-PR deletion of a superseded system or a concrete removal-trigger issue; the ack/dirty protocol still has a live consumer (`WebviewBridge`), so deleting it here would break the extension's progress view.

Filed #9969 as the mandatory removal trigger: port `WebviewBridge` to the `StreamLogDelta` feed, then delete `getDirtyUpdates`/`ackDirtyUpdates`/`getDirtyTextDeltas`/`ackDirtyTextDeltas` and StreamLog's destructive drain machinery, leaving the delta feed as the single change-feed surface.

PR #9970, opened 2026-08-11
build: format on save so formatting stops being a manual step

LionSR:
## Summary

Formatting was only ever applied by hand (`npm run format`) or by the pre-commit hook, which aborts the commit and asks you to re-stage. Neither is automatic, and `.vscode/settings.json` carried **no formatting configuration at all** — so code edited in VS Code stayed unformatted until something downstream complained.

This turns on `editor.formatOnSave` and adds `esbenp.prettier-vscode` to the workspace extension recommendations. Code edited in the editor is formatted by the time it is staged, and `format:check` stops failing on it.

The formatter is bound **per language** rather than through a global `editor.defaultFormatter`: a global binding would also claim languages Prettier cannot parse, turning save into an error toast for those files. The extension resolves the workspace Prettier from `node_modules` and honours `.prettierrc` and `.prettierignore`, so the editor and `npm run format` agree on version, config, and exclusions.

### Why not auto-stage at commit time

That was the obvious answer, it was tried, and it was reverted in #9953. Having the pre-commit hook `git add` its own output defeats pre-commit's rollback: the rollback is `git checkout -- .` (working tree ← index), so once the formatted content is in the index it restores the formatter's output instead of undoing it, and a rewrite conflicting with unstaged hunks **destroys them** — they survive only in a cache file named in a log line. [Reproduction and resulting state](https://github.com/LionSR/TeXRA/pull/9953#issuecomment-5250396928).

So the hook stays abort-on-rewrite, and remains the safety net for commits the editor never touched: agent sessions, github.dev, scripted commits. Format-on-save covers the interactive path, which is where the friction actually was.

`.prettierignore`'s note for `.github/workflows/` cited the `format-pr` workflow as its reason. That workflow no longer exists — this was the last reference to it — so the note now stands on its own: the exclusion is kept because `ci.yml` relies on hand-laid comment blocks and long `${{ }}` expressions that Prettier rewraps.

## Issue acceptance gates

No issue — direct maintainer request to stop handling formatting manually.

## Single source of truth

Prettier's configuration remains the single authority. Format-on-save adds no second definition of "formatted": the VS Code extension reads the same `.prettierrc` and `.prettierignore` and resolves the same pinned Prettier from `node_modules` that `npm run format` and CI's `format:check` use. No settings duplicate a Prettier option.

No abstraction added; no rules restated. The per-language `editor.defaultFormatter` entries are ten repetitions of one binding, which is how VS Code expresses language scoping — the alternative (one global binding) is fewer lines but wrong, because it claims languages Prettier cannot parse.

## Net elements (R6)

| Metric | Value |
| --- | --- |
| Files changed | 4 (`+37 / -8`, net **+29** LoC) |
| Exported symbols | 0 added, 0 removed |
| Classes / interfaces / enums | 0 added, 0 removed |
| npm scripts | 0 added, 0 removed |
| Workflows / CI jobs | 0 changed |

No production code, no test code, and no CI configuration changed — the diff is two `.vscode/` files, `.prettierignore`, and `AGENTS.md`.

## Consumer counts (R8)

n/a — no export, emit path, or module boundary touched. The one cross-file reference removed is prose: `.prettierignore`'s mention of the `format-pr` workflow, grepped repo-wide as the last surviving reference to a workflow that no longer exists.

## Host impact

Extension: none at runtime. `.vscode/` configures the editor for people working *on* this repo, not the shipped extension; `packages/extension/package.json` contributes no `configuration` and is untouched.

Desktop: none.

CLI: none.

## Scheduled refactoring

None. Two things deliberately left alone:

- **Non-VS-Code editors** get nothing from this. If you want parity, the equivalent is each editor's own Prettier integration; there is no repo-level mechanism that covers all of them without re-introducing a commit-time rewrite.
- **CI auto-fix** (a job that formats and pushes a fixup) remains unbuilt. It was offered and declined in favour of keeping this local.

## Validation

| Check | Result |
| --- | --- |
| `pnpm run format:check` (repo-wide) | pass |
| `.vscode/settings.json` / `extensions.json` parse as JSONC | pass |
| `prettier --check` on all changed files | pass |
| `node scripts/check-guidance-refs.mjs` | pass — 51 files |
| `.github/workflows/` still excluded from Prettier | confirmed unchanged |

Both `.vscode/` files were parse-checked explicitly because a malformed one fails silently in VS Code rather than erroring anywhere CI would notice. The commit itself was made with the pre-commit hook active (`prettier --write staged files....Passed`).

Not run: `test`, `typecheck`, `lint` — the diff contains no TypeScript, so none of them has anything new to evaluate.

---
_Generated by [Claude Code](https://claude.ai/code/session_011GHsNPhEj4SGfonqcGiWG2)_

---

> [!NOTE]
> **Low Risk**
> Editor and documentation-only changes plus a targeted test polling fix; no runtime extension, CI job, or formatting-authority changes beyond aligning local saves with existing Prettier config.
> 
> **Overview**
> Enables **format-on-save in VS Code** for every language Prettier handles (TS/JS, JSON, markdown, Vue, YAML, CSS, HTML), using **per-language** `editor.defaultFormatter` bindings so saves do not invoke Prettier on unsupported file types. Workspace recommendations add **`esbenp.prettier-vscode`** (workspace Prettier from `node_modules`) and **`Vue.volar`** so `.vue` files get the `vue` language id the settings target.
> 
> **`AGENTS.md`** documents editor formatting as the primary path, keeps the pre-commit hook as an abort-on-rewrite safety net (with rationale tied to #9953), and reorders the dev workflow steps accordingly. **`.prettierignore`** updates the `.github/workflows/` exclusion comment to reflect that the old `format-pr` workflow is gone while the exclusion remains for hand-laid CI YAML.
> 
> Separately, **`waitForFile`** in `SystemUtils.vitest.ts` now waits until the pid file exists **and has non-zero size**, avoiding a race where `existsSync` passes before `echo $!` finishes writing and `parseInt('')` flakes on loaded CI runners.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35e78e11b690d1f6b422d82f0b94094cb62adf87. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -19,8 +19,9 @@ packages/extension/resources/templates/
 # Local git worktrees (contain unstaged template files prettier cannot parse)
 .worktrees/
 
-# GitHub Actions workflows — the format-pr workflow's GITHUB_TOKEN lacks
-# `workflows: write` permission, so any auto-formatting of workflow YAML
-# triggers a "refusing to update workflow" push rejection. Skip them
-# entirely from format runs.
+# GitHub Actions workflows. Originally excluded because the (since removed)
+# format-pr workflow's GITHUB_TOKEN lacked `workflows: write`, so auto-
+# formatting workflow YAML triggered a "refusing to update workflow" push
+# rejection. The exclusion is kept deliberately: `ci.yml` relies on hand-laid
+# comment blocks and long `${{ }}` expressions that Prettier rewraps.
 .github/workflows/
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -6,6 +6,11 @@
     // workspace Prettier from node_modules, so the editor and `npm run format`
     // agree on version and config.
     "esbenp.prettier-vscode",
+    // Required for the `[vue]` block in settings.json to do anything. VS Code
+    // ships no built-in `vue` language, so without this the 104 SFCs under
+    // docs/.vitepress/components/ never carry that language id and the
+    // format-on-save binding silently never matches.
+    "Vue.volar",
     "dbaeumer.vscode-eslint",
     "amodio.tsl-problem-matcher",
     "ms-vscode.extension-test-runner",
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,6 +2,10 @@
   // See http://go.microsoft.com/fwlink/?LinkId=827846
   // for the documentation about the extensions.json format
   "recommendations": [
+    // Backs the format-on-save settings in settings.json. It resolves the
+    // workspace Prettier from node_modules, so the editor and `npm run format`
+    // agree on version and config.
+    "esbenp.prettier-vscode",
     "dbaeumer.vscode-eslint",
     "amodio.tsl-problem-matcher",
     "ms-vscode.extension-test-runner",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -62,7 +62,10 @@
   },
   // The 104 VitePress doc components under docs/.vitepress/components/ are in
   // the repo-wide `format:check`, so omitting this would leave exactly the gap
-  // this change exists to close: save does nothing, CI fails.
+  // this change exists to close: save does nothing, CI fails. This block needs
+  // the `Vue.volar` recommendation in extensions.json to have any effect --
+  // VS Code ships no built-in `vue` language, so without it .vue files never
+  // carry the language id this key matches on.
   "[vue]": {
     "editor.formatOnSave": true,
     "editor.defaultFormatter": "esbenp.prettier-vscode"
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,5 +12,28 @@
   // Use the workspace TypeScript (6.0.3) so the editor matches the build; the
   // bundled VS Code TS server is older and rejects `ignoreDeprecations: "6.0"`.
   "typescript.tsdk": "node_modules/typescript/lib",
-  "texra.ui.showLoginBanner": false
+  "texra.ui.showLoginBanner": false,
+
+  // Format on save so `format:check` never fails on work that went through the
+  // editor. The pre-commit hook stays the safety net for anything that did not
+  // (agent sessions, github.dev, `git commit` from a script) -- it deliberately
+  // aborts rather than re-staging, because staging inside pre-commit's stash
+  // lifecycle can destroy unstaged hunks on a rewrite conflict.
+  //
+  // The formatter is set per language rather than globally: a global
+  // `editor.defaultFormatter` of Prettier would also claim languages Prettier
+  // cannot parse, turning save into an error toast for those files. Prettier
+  // reads `.prettierrc` and honours `.prettierignore`, so `.github/workflows/`
+  // stays excluded here exactly as it is on the command line.
+  "editor.formatOnSave": true,
+  "[typescript]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[typescriptreact]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[javascript]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[javascriptreact]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[json]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[jsonc]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[markdown]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[yaml]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[css]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[html]": { "editor.defaultFormatter": "esbenp.prettier-vscode" }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,24 +16,67 @@
 
   // Format on save so `format:check` never fails on work that went through the
   // editor. The pre-commit hook stays the safety net for anything that did not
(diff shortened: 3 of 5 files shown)


LionSR:
All three review findings were right and are fixed in a8b5697.

**Global `editor.formatOnSave` (@chatgpt-codex-connector)** — the strongest of the three, because the config contradicted its own comment. The comment said format-on-save applied to "the languages Prettier handles"; the setting was global, so it also fired for the repo's tracked `.tex`, `.sty`, `.sql`, and `.wl` files, invoking whatever formatter that contributor happened to have installed. `formatOnSave` now lives inside each language block next to `defaultFormatter`, and the comment explains why the two must stay together — a global `formatOnSave` over-fires, a global `defaultFormatter` points unparseable languages at Prettier.

**Missing `[vue]` binding (@texra-ai and @chatgpt-codex-connector)** — confirmed: 104 tracked `.vue` files, and `prettier --check "docs/.vitepress/components/*.vue"` matches them, so they are inside the repo-wide `format:check`. Saving one would not have formatted it while CI still demanded it — precisely the gap this PR exists to close. Added.

I also checked whether Vue was the only omission rather than fixing just what was reported. Against every tracked extension:

| Extension | Count | VS Code language | Covered |
| --- | --- | --- | --- |
| `.ts` / `.mts` | 2486 | `typescript` | yes |
| `.md` | 299 | `markdown` | yes |
| `.vue` | 104 | `vue` | **added here** |
| `.yaml` / `.yml` | 85 | `yaml` | yes |
| `.tsx` | 66 | `typescriptreact` | yes |
| `.mjs` / `.js` / `.cjs` | 64 | `javascript` | yes |
| `.json` | 45 | `json` / `jsonc` | yes |
| `.css` | 9 | `css` | yes |
| `.html` | 5 | `html` | yes |
| `.tex` `.sty` `.sql` `.wl` | 15 | — | not Prettier-supported; deliberately unbound |

`.mjs`/`.cjs` resolve to `javascript` and `.mts` to `typescript`, so they need no entry of their own. Vue was the only real gap.

**github.dev fallback (@chatgpt-codex-connector)** — correct, and my wording was actively unhelpful: it told contributors to run `npm run format` in an environment with no terminal and no Node. AGENTS.md now scopes github.dev to changes Prettier does not reformat and points anything larger at a Codespace or local checkout.

Verified after the fix: all 11 language blocks carry both keys, no global `formatOnSave` or `defaultFormatter` remains, repo-wide `format:check` passes, and guidance-refs passes.

---

LionSR:
@cursor is right, and this one is worth stating plainly: the `[vue]` block I added in a8b5697 was **inert**. Fixed in d33c175.

VS Code ships no built-in `vue` language contribution, and `extensions.json` recommended no Vue extension, so `.vue` files never carried the language id the block keys on. The binding silently never matched — the 104 SFCs under `docs/.vitepress/components/` were exactly as unformatted-on-save after my "fix" as before it. I asserted in [my previous comment](https://github.com/LionSR/TeXRA/pull/9970#issuecomment-5252512651) that the gap was closed; it was not.

`Vue.volar` is now in the workspace recommendations, since that is what supplies the language id. Both files note the dependency, so neither half gets deleted as unused without the other.

**What I can and cannot verify here.** No VS Code runs in this environment, so this rests on VS Code shipping no `vue` language rather than on an observed save. That is the same class of gap that produced the bug: the previous commit's verification checked that all 11 language blocks carried both keys — a check of the file's *shape*, which passes whether or not any of those ids ever binds to a real file. It could not have caught this, and I presented it as if it had.

Flagging the general case for whoever reviews the rest: every non-TypeScript language block here carries the same assumption. `typescript`, `javascript`, `json`, `jsonc`, `markdown`, `yaml`, `css`, and `html` are all built-in VS Code languages, so those bind without an extension. `vue` was the only entry in the list that did not — which is also why it was the only one that needed a recommendation.

---

LionSR:
`kernel tests (linux, shard 1/2)` failed on d33c175 (and took `validate` with it). It is a real bug in the test, not this branch's diff, and it is fixed in 35e78e1 rather than re-run.

**The failure**

```
AssertionError: The expression evaluated to a falsy value:
  strict.ok(Number.isInteger(childPid) && childPid > 0)
 ❯ startSleeper src/test-kernel/utils/system/SystemUtils.vitest.ts:119:12
```

1 failed, 5496 passed. This branch changes only `.vscode/*.json`, `.prettierignore`, and `AGENTS.md`, so it cannot have caused a `SystemUtils` assertion — but "unrelated" is a reason to diagnose, not to hit re-run.

**Root cause**

```js
const SLEEPER_SCRIPT = 'sleep 60 & echo $! > "$PID_FILE"; wait';
...
function waitForFile(filePath: string) {
  return waitFor(filePath, () => existsSync(filePath));   // existence only
}
await waitForFile(pidFile);
const childPid = Number.parseInt(readFileSync(pidFile, 'utf8'), 10);
```

The shell's `>` **creates the file before the writer's bytes land**. `existsSync` can therefore return true while the file is still empty; `readFileSync` returns `''`, `parseInt('')` is `NaN`, and the assertion fails. The window is narrow on an idle machine and widens on a runner saturating both vCPUs at `--maxWorkers=2` — so it presents as a random flake on whichever PR happens to land in it.

**Fix** — require content, not just existence:

```js
() => existsSync(filePath) && statSync(filePath).size > 0
```

`waitForFile` has exactly one caller, this pid file, so tightening the predicate affects nothing else. The same-named helper in `LeanTools.vitest.ts` is a separate file-local function and is untouched.

**On the evidence.** The suite passes 5/5 locally after the change — but it passed most of the time before, so repeated green runs are not proof for a race. The argument here is causal: the predicate now excludes the exact state that produced the `NaN`. I would rather say that than present five passes as a demonstration.

This is also a small CI-cost win beyond this PR: each hit of this flake previously cost a full ~10-minute re-run on an unrelated change.

---

PR #5687, opened 2026-06-09
refactor(storage): keep KVStore on Keyv internally

LionSR:
## Summary

Refactors `KVStore` to use Keyv v5 internally while keeping the existing StorageFS file layout and plain-JSON on-disk format. The previous exported `StorageFSKeyvAdapter` layer is gone; `KVStore` now creates the private Keyv store object itself.

## Issue acceptance gates

- Addresses the review/user request to remove the adapter middle layer and make the Keyv use more native.
- Keeps backward compatibility: each key still maps to one percent-encoded `.json` file, with no Keyv `{ value, expires }` envelope written to disk.
- Keeps the public `KVStore` API unchanged for existing execution, stream log, and stream snapshot callers.

## Single source of truth

`src/common/storage/KVStore.ts` now owns the StorageFS path mapping through one `keyToPath(dir, key)` helper and one `filenameToKey(filename)` helper. Keyv owns only key-value semantics; StorageFS layout is not split across files.

## Duplication and abstraction budget

- Removed the exported `StorageFSKeyvAdapter` module.
- Removed duplicated key-to-path derivation in `modifiedAt()`.
- Kept the Keyv store boundary private to `KVStore` because StorageFS persistence still requires a store object for Keyv.
- No new pass-through public abstraction was added.

## Host impact

Extension: behavior-neutral persistence refactor.

Desktop: behavior-neutral persistence refactor.

CLI: behavior-neutral persistence refactor.

## Scheduled refactoring

None.

## Validation

- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm vitest run --config vitest.config.mjs src/test-kernel/transcript/StreamLogStoreLoad.vitest.ts`
- `npx tsc --noEmit -p tsconfig.test-kernel.json`
- `npx eslint src/common/storage/KVStore.ts src/test-kernel/transcript/StreamLogStoreLoad.vitest.ts`
- `git diff --check`
- `npm run typecheck -- --pretty false`
- `npm run lint -- --quiet`
- `npm run compile:fast`

---

> [!NOTE]
> **Medium Risk**
> Touches shared persistence used by execution, stream logs, and snapshots; behavior is intended to be neutral but any serialization or missing-key handling drift could affect on-disk data reads.
> 
> **Overview**
[...]

PR #9971, opened 2026-08-11
refactor(storage): remove Keyv from KVStore internals

LionSR:
Undo #5687's internals: KVStore calls StorageFS directly instead of
routing through Keyv and a custom KeyvStoreAdapter. The on-disk format
(one plain-JSON file per key, percent-encoded names) is unchanged, and
the public API is identical except that the throwOnErrors option is
gone — throwing is now the only error policy.

Keyv earned nothing here: its TTL, namespacing, stats, and hooks were
all unused, serialization was already overridden to preserve the plain
JSON format, and listKeys/modifiedAt/deleteDir already bypassed it. In
its default mode it also converted I/O read failures into silent
misses — on StreamSnapshotStore's read handle, an EIO/EPERM sidecar
read was indistinguishable from an absent file, feeding the
empty-base-overwrite data-loss path the 2026-08-10 memory audit ranks
highest.

Behavior changes, all in the loud-failure direction:
- Sidecar reads that fail with an I/O error now warn and degrade to
  missing (previously silent); corrupt-JSON handling is unchanged.
- StreamSnapshotStore's dual per-record handles (kv/writeKv, differing
  only in error policy) collapse to one.
- deleteDir failures now throw instead of being swallowed.

Verified: typecheck, lint, dead-code ratchet, and the transcript +
agent/storage vitest suites (405 tests) pass.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01BxY1Fpz1vmvvpvCFjxyr7y

---

> [!NOTE]
> **Medium Risk**
> Changes file-backed KV error semantics and how stream status reaches transcripts/UI; failures are louder (throws, skipped corrupt sidecars) rather than silent misses, but both areas are core to resume and progress projection.
> 
> **Overview**
> **Storage:** Drops the `keyv` dependency and reimplements `KVStore` on `StorageFS` directly (same one-file-per-key JSON layout). Missing keys still read as `undefined`; other I/O failures throw instead of being treated as cache misses. `throwOnErrors` and `StreamSnapshotStore`'s separate strict write handle go away; child-detach now skips streams with unreadable sidecar meta instead of failing the whole sweep. `RelativeFS.writeJson` is removed in favor of explicit `write` + `JSON.stringify`.
> 
> **Status & events:** `status` is no longer an `AgentEvent` or duplicated on run traces. `StreamStatusMachine` publishes only session facts; transcript recording consumes them through **`RunTrace.handleStatus`** (bridged at launch/child-stream attach). Call sites stop passing `{ trace }` into status transitions.
> 
> **Session plumbing:** `SessionHandle.onResult` subscribes to run facts on the hub (keeping missed-terminal replay). `onFollowUpSent` observers are removed—`listenForFollowUp` and tests use the session `followUpSent` fact instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e0a4b0f66bc32d7996eb9bc122a472e1fd38c53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
bugbot run

Re-requesting review after the rebase onto main (adopting #9976's textChunks delta design), the review-fix commit, the status single-rail change, and the formatting pass.

---

Issue #9956, opened 2026-08-11
[agent-runtime] Per-record disk provenance in StreamSnapshotStore: delete canMutateSynchronously / hasAuthoritativeStreamSet


LionSR:
Blocker PR extracted from #9947; ships before everything else in the memory-architecture plan (docs/prds/2026-08-11-transcript-memory-architecture.md).

`canMutateSynchronously` (`StreamSnapshotStore.ts:430-440`) flips `seeded = true` without a disk read when `hasAuthoritativeStreamSet === true` and no seed is in flight; a subsequent mutation merges onto an empty base and persists empty+delta over the real sidecar. A global "I know the stream set" flag is being used to assert the per-record fact "this record is hydrated".

Fix: replace the seeded flag pair with a per-record `diskState` tri-state (`unknown | verified-absent | loaded`). `verified-absent` may be set only at stream minting or from an explicit stat probe; mutations on `unknown` queue behind the existing seed chain. Required test: the clobber regression (mutate a stream whose sidecar exists on disk but is absent from `transcripts.keys()`; assert merge, not overwrite).

Issue #9947, opened 2026-08-10
[agent-runtime] stop eagerly seeding all historical stream sidecars at startup (lazy hydrate + LRU)


LionSR:
Part of the 2026-08-10 long-lived-session memory investigation. **Highest-confidence retained-heap candidate.**

## Problem

At CLI startup, `SessionHandle.repairStoresAfterRestart` (`src/agent/runtime/SessionHandle.ts:408`) calls `this.snapshots.load(this.transcripts.keys())`. `StreamLogStore.keys()` returns **every** persisted stream summary — ~1,923 in the affected workspace — unfiltered. `load()` (`src/transcript/StreamSnapshotStore.ts:1572`) reads 6 sidecar files per stream (`streamSnapshotRead.ts:144`) and retains them all in `records`. A bounding filter already exists (`getUnfinishedStreamIds`, `StreamLogStore.ts:441`) but is wired only into post-load status/repair phases, never the snapshot load.

## Proposal

At startup, preload only the unfinished set + active lineage via `preload()` (which leaves `hasAuthoritativeStreamSet = false`); hydrate finished streams lazily on first access with an LRU cap. Mirror the existing safe model at `packages/cli/src/chat/tui/state/subscribeStreamArtifacts.ts:66` (`await store.preload([streamId])` then read).

## This is not a small change — safety audit

Verified 2026-08-10 (full detail in the audit doc, Finding 1 safety audit):

- **`awaitSeeded` (`StreamSnapshotStore.ts:1521`) is a status check, not self-heal** — it only awaits an in-flight `seedChain`, returns `false` otherwise without touching disk. The lazy fallback in `read`/`readOutputFiles` works only because they carry an explicit disk read afterward.
- **9 synchronous accessors assume the record is seeded** (`getRunMetadata`, `getOutputFiles`, `getMissingOutputs`, `getCompileFailures`, `getRunUsage`, `getKnownFilePaths`, `getWorkPlan`, `getParentStreamId`, `getExecutionIdMap`) and return empty/default for an unseeded stream.
[...]

Issue #9434, opened 2026-07-30
Design: one boundary-only migration system for persisted formats, with authored vs derived data tiers


LionSR:
Umbrella for the legacy-format work surfaced by the 2026-07-30 seam audit (#9422, #9424, #9430, #9431, #9432, #9433). Those issues each delete one legacy arm. This issue is about the thing that made them accumulate: there is no single way to migrate a persisted format, so every author invents one, and the inventions live forever because nobody can tell which are still load-bearing.

## The problem: five idioms for one job

A grep for legacy handling hits 128 non-test files under `src/`. The distinct mechanisms in use today:

1. **Zod union with a legacy arm that transforms** into the canonical shape. This is the documented house style (CLAUDE.md, "normalize legacy formats once at the entry point"). Ten files declare a `Legacy*Schema`.
2. **Per-field `.catch(default)` on the persisted schema**, e.g. `src/shared/schemas/streamSnapshot.ts:83-86` and `src/transcript/StreamLogStore.ts:55-60`. Deliberate there, and documented, but it is the idiom our own rules warn about on persisted data: a corrupt value silently becomes a default, and if that value later feeds a whole-file write it becomes permanent loss.
3. **A `schemaVersion` literal**, itself spelled two ways: `.literal(V).catch(V)` (streamSnapshot) versus `.literal(V).prefault(V)` (`src/tools/inquiry/externalInquiryStorage.ts:171-173`), with newer-version gating done imperatively upstream of the parse in one case and not the other.
4. **Imperative one-shot migrations** that are not one-shot: `src/platform/defaults/legacyDataMigration.ts`, and `executionListing`'s `migrateIfNeeded` which re-attempts on every first listing per process (#9424).
5. **Perpetual dual-layout read probes** with no migration at all, so the read path speaks both vocabularies forever (#9430's `taskRuns/` probing, #9422's six normalization arms).

Consequences we can already point at: nobody knows when an arm is safe to delete, so sunsets need archaeology (every one of the six issues above needs a separate "how old is old enough" judgement call); the legacy names leak into the `packages/agent` public surface and risk freezing there (#9431); and migration work runs on hot paths (#9424, #9430).

## Proposal A: one migration mechanism, at the boundary only

One helper owns versioned persisted reads. Sketch, not final:

```ts
// versionedPersisted.ts — the only place a legacy format is understood
definePersistedFormat({
  name: 'resultMeta',
[...]

PR #9986, opened 2026-08-11
Loud unhydrated snapshot access + summary metadata mirror (#9947, part 1 of the remainder)

LionSR:
Part of #9947 — top-5 item 2's second half in `docs/prds/2026-08-11-transcript-memory-architecture.md`, after the tri-state disk provenance blocker (#9956, shipped as #9965). Ships the loud-accessor release and the summary widening; deliberately holds the startup bounding (gate rationale below).

## What ships

**1. Loud unhydrated access (PRD: "the 9 silent-empty accessors go loud one full release before the bounding lands").** All 9 synchronous `StreamSnapshotStore` accessors (`getRunMetadata`, `getOutputFiles`, `getMissingOutputs`, `getCompileFailures`, `getRunUsage`, `getKnownFilePaths`, `getWorkPlan`, `getParentStreamId`, `getExecutionIdMap`) warn at `warn` level — deduped per (stream, accessor), reset on evict — whenever they serve a record whose per-record `diskState` is unestablished, instead of silently returning empty defaults. `getExecutionIdMap` warns per resident-but-unknown record; wholly non-resident streams are that accessor's documented contract (callers merge `readPersistedExecutionId`).

**2. Summary metadata widening (derived tier per #9434).** The persisted stream summary gains an optional `meta` block: `identity`, `executionId`, `parentStreamId`, `userFollowUpSupport`, `agentCategory`, `description`, `model`, `workingDirectory`, and `command`. Field set comes from the `buildStreamTabInfo` caller audit the PRD demands: it consumes identity, category, description, model, workingDirectory, follow-up capability, and — for process streams only — `config.instruction` as the command chip. `command` therefore carries only a process run's command line; an agent run's full instruction text never enters the always-resident summaries (that was the one unbounded field). A stale-shaped summary is discarded **loudly** and rebuilt from the stream log — never migrated, no `.catch`.

Feed: `StreamSnapshotStore` publishes a whole replacement meta object on every metadata mutation (`run.start`, `run.config`, parent edge, description) and at the end of every sidecar hydration; `SessionHandle` wires the sink to `StreamLogStore.recordSummaryMeta` (deep-equal no-op gate, so the startup hydration sweep republishing unchanged metadata costs zero writes; first startup after upgrade backfills every summary once, and legacy rows backfill lazily). Metadata that projects before `ensureStream` waits in a pending antechamber and lands at registration — it never mints a phantom tab.

**3. The sidecar-reading sidebar path dies in this PR (no dual system).** `SessionState` now overlays the summary mirror at read time in `getStreamMetadata`. Deleted outright: `buildSnapshotMetadataPatch`, `applySnapshotMetadata`, `refreshStreamMetadataFromSnapshot` (and both host call sites), `SessionState.load`'s all-streams metadata loop, and the desktop presentation's refresh loop + snapshot read in its child-activity pass (now reads the summary-backed category). The sidebar/all-streams metadata paths touch no sidecar accessor anymore.

**4. Preload-then-read conversions (bounding enablers, safe today).** Extension `setActiveStream` awaits `snapshots.preload([streamId])` before content sync (the CLI focus path's model); extension `resumeFromResumeData` preloads on a metadata miss, matching the desktop resume path exactly. Together with item 3, three of the issue's four known-breaking call sites are converted; the fourth (`SessionState.load` loop) is deleted.

## Bounding gate decision: NOT shipped, and why

The task gate: land the bounding only if **every** production reader of the 9 accessors is verified to preload, use summaries, or tolerate lazy hydration. The audit could not establish that for:

- `SessionHandle.repairWaitingIfResumable` (`:321`) — reads `executionId` synchronously on follow-up submission. A parked WAITING stream is **not** in `getUnfinishedStreamIds()`, so under bounding a follow-up sent after restart but before repair marks the stream WAITING routes to `no_session`. Needs a persisted-executionId fallback first.
- Extension `ProgressViewMessageHandler` wiring (`:190/:378-390/:559-564/:629-634/:755`) and `ProgressViewCommandHandlers` (`:485`) — controllers act on webview-supplied stream ids; not verified active-stream-only in every message arm.
- Desktop equivalents in `desktopAgentExecution.ts` (`:476-479/:553-556/:756/:774-776/:908-920/:1199-1200`).
- CLI `sessionSignalsAdapter.onStreamMetadataChanged` (`:112`) — synchronous full-config read; safe for live streams, unverified for cold fan-outs.
- `ProgressBackend.hasDeletableStreamData` (`:240`) — under bounding, a non-resident stream with sidecar-only data reads `config: undefined` and the deletion sweep would skip orphan sidecars.

That is exactly the situation the PRD's sequencing exists for: this PR makes every one of those readers *loud* when it would have been served empty data, so one release of warnings identifies any reader the audit missed before the bound can silently break it. Remaining #9947 work after this release: convert/verify the readers above, then bound `repairStoresAfterRestart` to `getUnfinishedStreamIds()` + active lineage (with `evictAll()`-then-`preload` on the storage-root-change paths) and add the finished-stream lazy-hydration startup test.

## Element accounting

- Deleted: 3 SessionState methods (`buildSnapshotMetadataPatch`, `applySnapshotMetadata`, `refreshStreamMetadataFromSnapshot`), 2 host refresh loops, the `SessionState.load` all-streams loop, and the desktop second-pass snapshot read.
- Added: 1 schema (`StreamSummaryMetaSchema` + `meta` field), 3 `StreamLogStore` members (`recordSummaryMeta`, `getSummaryMeta`, pending antechamber + adopt), 3 `StreamSnapshotStore` members (`attachSummaryMetaSink`, `publishSummaryMeta`, `warnIfUnseeded` + dedupe set), 1 SessionState overlay (`applySummaryMetadata`), sink wiring in `SessionHandle`, 2 preload conversions.
- Net: +497/−82 LoC. Net-positive by design — this is the PRD's observability release plus the summary mirror; the deletions it unlocks (the ~15k-read startup sweep) land with the bounding.

## Verification

- `npm run typecheck` — clean (all 6 projects).
- `npx vitest run src/test-kernel/transcript src/test-kernel/progressView` — 726/726 pass, including new tests: unseeded-access warning fires once per accessor+stream and serves the loud default; summary meta round-trips through the persisted cache; a stale-shaped summary warns and rebuilds from the log (full log read, no migration); pre-registration metadata never mints a phantom tab and lands on `ensureStream`; the sink publishes on run facts and republishes on hydration.
- `npx vitest run src/test-kernel/controllers src/test-kernel/agent src/test-kernel/cli src/test-kernel/desktop` — pass.
- `npm run check:dead-code-ratchet` — OK (16 vs 16 baselined).
- `npm run lint` on touched files — clean.

https://claude.ai/code/session_01Gtc6wQMD1hedgabjQMfAun

---

> [!NOTE]
> **Medium Risk**
> Touches session metadata authority, active-stream hydration, resume, and deletion UI refresh paths across extension, desktop, and shared controllers; bounded by tests but behavior changes are broad.
> 
> **Overview**
> **Stream display metadata** now lives in the always-resident **stream log summary** (`meta` on persisted summaries), fed by `StreamSnapshotStore` through `SessionHandle`’s `summaryMetaSink` → `StreamLogStore.recordSummaryMeta`. Sidebar/all-streams paths read via **`getStreamMetadata` overlaying the summary at read time**, replacing snapshot refresh loops, `SessionState.load` per-stream seeding, and explicit `refreshStreamMetadataFromSnapshot` plumbing.
> 
> **Hydration and lifecycle:** activating a stream **preloads transcript and sidecar in parallel**; partial failures surface after sync. Extension resume **preloads on metadata miss** (desktop parity). Stream deletion flows **await** `rebuildRenderedStreams` (now `Promise`-based). **`StreamLogStore.openReadOnly`** is removed in favor of **`openReadOnlyForStream`**.
> 
> **Observability:** nine synchronous `StreamSnapshotStore` accessors **warn once per (stream, accessor)** when disk provenance is unestablished, instead of silently returning empty defaults—prep for a later startup bounding change.
> 
> **Persistence:** stale-shaped summary caches are **discarded with a warning** and rebuilt from the log; metadata can sit in a **pending antechamber** until `ensureStream` without minting phantom tabs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b4bb553f9f89ccf424c6b0f3c23e983fc8e50e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LionSR:
CI fixes pushed (af6b78172f), taking the surface-neutral route on the store ratchet rather than a budget bump:

- `attachSummaryMetaSink` is gone — the sink rides the existing `attachSessionEvents(events, { summaryMetaSink })`. StreamSnapshotStore is back at 28 units. The SessionHandle read-only guard around it was dead code (the constructor rejects read-only transcript stores), so the wiring is now one unconditional call — which also fixes the RunExecution test double.
- `StreamLogStore.openReadOnly` is deleted: zero production callers (test-only + generated dist types), and `openReadOnlyForStream` covers the semantic; its four test usages converted directly. With the two new summary members that lands StreamLogStore exactly on 24. The baseline name list is refreshed to the real surface (it still listed the long-deleted `clearDirtyUpdates`), so the phantom slack is gone too.
- `ConfigForm.vitest.ts` prettier fix inherited from origin/main (identical one-line deletion to the sibling branch; merges cleanly either way).

Verified: full typecheck clean; `src/test-kernel/architecture` + `RunExecution.vitest.ts` 118/118; transcript + progressView + SessionHandle suites 740/740; dead-code ratchet 16/16.

PR #10005, opened 2026-08-12
refactor(logger): add createLog(channel) factory, migrate 58 files off CHANNEL threading

LionSR:
## Summary

A tech-debt audit flagged the single highest-frequency ad-hoc pattern in the tree: ~99 modules each declared `const CHANNEL = '...'` and threaded it as the first argument through every `logger.warn/error/info/debug(CHANNEL, …)` call — ~144 single-line plus hundreds of multi-line call sites. The channel string was copy-pasted per file (and at least one module historically carried the *wrong* channel, invisible at the call site).

This PR adds a channel-bound logger factory and migrates the clean subset:

- **`createLog(channel)`** in `@logger/logUtils` returns a `Log` with the four level methods bound to one channel. It emits through the **same `writeLine` sink** as the free `debug/info/warn/error` functions, so the output line, secret redaction, and debug-data handling are byte-for-byte unchanged — this is a behavior-preserving transform.
- Call sites become `const log = createLog('X'); log.warn(msg)` — the channel is named once instead of on every call.

## Issue acceptance gates

n/a — no tracked issue; this is a proactive tech-debt refactor from a codebase audit.

## Single source of truth

`createLog` (in `@logger/logUtils`) is the one place that binds a channel to the level writers. A module now names its channel exactly once at the `createLog('X')` call instead of repeating the constant on each log invocation.

Removes the `const CHANNEL` declaration + repeated `CHANNEL,` first-argument from 57 modules. The new abstraction (`createLog`) owns one invariant: *a channel is named once and bound to the four level writers*; it adds no pass-through layer — it's a thin binding over the existing `writeLine` sink, parallel to the existing `createChannelWriter`.

Scope was deliberately bounded to files where `CHANNEL` was used **only** in `logger.LEVEL(CHANNEL, …)` calls (verified by codemod post-check: no dangling `logger.`/`CHANNEL` tokens remain). Files that pass the channel or the logger namespace as a *value* — the LaTeX cluster's shared `LATEX_COMMANDS_CHANNEL`, `SupabaseAuthProvider`'s `log: logger` — are left untouched.

## Net elements (R6)

- **Files:** 59 changed (57 migrated modules + `logUtils.ts` + `LogUtils.vitest.ts`).
- **Exported symbols:** +2 (`export function createLog`, `export interface Log`); 0 removed. Both have same-PR consumers (57 modules + tests), satisfying the exports-are-contracts ratchet.
- **Classes/interfaces/enums:** +1 interface (`Log`).
- **Net LoC:** **−96** (391 insertions, 487 deletions) — and that is *after* adding the ~35-line factory and ~24-line test, so call-site reduction is ~150+ lines.

## Consumer counts (R8)

No emit path or export was deleted or re-routed. `createLog` is a new export with 57 production consumers in this PR (+ 2 unit tests). The free `debug/info/warn/error` functions are unchanged and retain their remaining callers (78 modules still use the namespace import, the intended follow-up surface).

## Host impact

Extension: Migrated modules under `packages/extension/src/frontend/**`, `common/**`, `commands/**`. Behavior identical.

Desktop: No files under `packages/desktop` changed directly; shared-core modules it consumes (`src/**`) are behavior-identical.

CLI: `packages/cli/src/chat/tui/panes/transcriptViewport.ts` migrated. Behavior identical.

## Scheduled refactoring

Follow-up (not in this PR): 78 modules still use `import * as logger` — mostly files that pass the channel as a value (helper-threaded channels, `channel:` object fields). These need per-file judgment rather than a mechanical codemod. The audit also surfaced larger, higher-effort targets (parallel server-chain logic in the OpenAI/Google model handlers; the `StreamSnapshotStore` round-accumulator descriptor table; CLI resume/retry duplication) tracked separately.

## Validation

- `npm run typecheck` — green across all 6 workspaces (workspace, test-kernel, agent, cli, trace-viewer, desktop).
- `eslint` on all changed files — clean.
- `npm run check:dead-code-ratchet` — OK, no new unused files/exports.
- Logger unit tests — 15 passing, including 2 new cases asserting `createLog` binds the channel and forwards debug data identically to the free `debug` fn.
- Full `vitest run` suite was executing locally at push time (slow, output-buffered); CI is the authoritative full-suite gate.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01GB4BmgUZVtvUM5TokGpvZd)_

---

> [!NOTE]
> **Low Risk**
> Behavior-preserving logging refactor: same sink, redaction, and output format. Risk is mostly mechanical migration breadth and test-mock updates.
> 
> **Overview**
> Adds **`createLog(channel)`** in `@logger/logUtils`, returning a channel-bound `log.debug/info/warn/error` API so modules name the channel once instead of threading `CHANNEL` through every call.
> 
> Migrates ~57 production modules across agent, extension, CLI, tools, transcript, and utils from `logger.warn(CHANNEL, …)` to `const log = createLog('…'); log.warn(…)`. Output still goes through the existing free-function writers (via a per-call namespace lookup so spies keep working), so formatting and redaction stay the same.
> 
> Also updates logger unit tests and a few vitest mocks to cover/export `createLog`. Files that pass the logger or channel as a value are intentionally left for a follow-up.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19993814e324820a9cef101afc3f498721490512. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/chat/tui/panes/transcriptViewport.ts
+++ b/packages/cli/src/chat/tui/panes/transcriptViewport.ts
@@ -1,6 +1,6 @@
 // Pure viewport math for bounded pending transcript panes.
 
-import * as logger from '@logger/logUtils';
+import { createLog } from '@logger/logUtils';
 import type { ExecutionLabels } from '@shared/tools/executionsDisplay';
 import { toErrorMessage } from '@utils/errors/errorMessage';
 import {
@@ -11,7 +11,7 @@ import { isRenderableTranscriptEntry } from './transcriptEntries';
 import type { ConversationEntry } from '../state/cliState';
 
 const FAILED_ENTRY_ESTIMATE_ROWS = 1;
-const CHANNEL = 'transcriptViewport';
+const log = createLog('transcriptViewport');
 /** Entry ids already reported, so a persistently-throwing entry is logged once
  *  instead of once per stream-sync tick (the estimate path runs per frame). */
 const brokenEntryIdsReported = new Set<string>();
@@ -29,8 +29,7 @@ function estimateEntryRows(
   } catch (error) {
     if (!brokenEntryIdsReported.has(entry.id)) {
       brokenEntryIdsReported.add(entry.id);
-      logger.warn(
-        CHANNEL,
+      log.warn(
         `Failed to estimate rows for ${entry.role} entry ${entry.id}; assuming ${FAILED_ENTRY_ESTIMATE_ROWS}: ${toErrorMessage(error)}`,
       );
     }
--- a/packages/extension/src/commands/agent/executeCommand.ts
+++ b/packages/extension/src/commands/agent/executeCommand.ts
@@ -7,10 +7,10 @@ import { AgentConfigSchema } from '@agent/core/definition/AgentConfig';
 import { ModelHandlerCompatibilityKeySchema } from '@agent/runtime/modelHandlerCompatibilityKey';
 import { runAgent } from '@agent/runtime/runAgent';
 import { openFinalOutputIfAvailable } from '@frontend/agents/finalOutputOpener';
-import * as logger from '@logger/logUtils';
+import { createLog } from '@logger/logUtils';
 import type { ExecutionId } from '@shared/schemas';
 
-const CHANNEL = 'ExecuteCommand';
+const log = createLog('ExecuteCommand');
 
 interface WrappedExecuteInput {
   config: unknown;
@@ -62,7 +62,7 @@ export async function runExecuteCommand(input: unknown): Promise<void> {
   } catch (error) {
     if (error instanceof ZodError) {
       const message = `Invalid agent configuration. ${z.prettifyError(error)}`;
-      logger.warn(CHANNEL, message, { data: error });
+      log.warn(message, { data: error });
       void vscode.window.showErrorMessage(message);
       return;
     }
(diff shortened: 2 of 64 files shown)


LionSR:
Thanks — the spy-observability finding was correct and is now fixed in `fc69d12`.

**Root cause (as several of you identified):** `createLog` called the module-private `writeLine` directly, so `vi.spyOn(logger, 'warn')` — which patches the *exported* binding — never intercepted `log.warn(...)`. This broke 19 spy suites and crashed 5 suites that fully mock `@logger/logUtils` without a `createLog` export.

**Fix — reviewer-endorsed option (b):** `createLog` now delegates to the exported `debug/info/warn/error` **through the module's own namespace** (`loggerSelf`), so the exports remain the single observable seam. One subtlety worth calling out: the namespace lookup is done **per-call**, not captured at bind time — a module-level `const log = createLog(...)` is bound at import, *before* a test installs its spy, so an eagerly-captured reference would still miss it. `options` is forwarded only when present so the spied argument list matches a direct `warn(channel, message)` call. Both paths still terminate in the same `writeLine` emission, so production output is unchanged.

**Result:**
- All 19 spy suites pass **unchanged** — no assertion rewrites needed (the `setOutputChannelFactory` sink-capture migration is therefore not required).
- The 5 full-mock suites get a `createLog` stub; none assert on it.
- `Log` is now file-local (thanks @cursor — no external importer).

Locally verified: the 19 spy + 12 mock suites, full workspace + agent typecheck, and lint are green; the full Vitest suite is re-running here and in CI.

---

PR #10007, opened 2026-08-12
refactor: tighten desktop IPC, ExecutionsTool schema, and OpenRouter message casts

LionSR:
## Summary

Follow-up from a scheduled data-structure-design audit of the codebase. Fixes the highest-confidence findings — mechanical, behavior-preserving tightenings, each verified with a full workspace typecheck and targeted tests:

1. **`packages/desktop/src/main/desktopFileSelection.ts` (High)** — routed inbound messages through the shared `MainViewInboundMessageSchema` discriminated union at the entry point, matching every sibling desktop IPC adapter (`desktopShellIpc.ts`, `desktopWorkspaceIpc.ts`, `desktopSettingsIpc.ts`, ...). Previously this file alone read the loose `DesktopCommandMessage` envelope with hand-rolled `typeof x === 'string'` guards, so the desktop and extension hosts could silently diverge on what's valid for the same wire command.

2. **`src/tools/ExecutionsTool.ts` (Medium)** — converted the flat `z.strictObject` input schema (5 independently-optional fields whose validity actually depends on `action`) into a `z.discriminatedUnion('action', [...])`, matching the pattern every sibling multi-command tool already uses (`PlanTool`, `MemoryTool`, `ExternalInquiryTool`). Each branch now only declares the fields that action actually uses. A `z.preprocess` step preserves the pre-existing "omitted `action` defaults to `view`" call shape. Previously-silent invalid combinations are now rejected instead of quietly ignored (e.g. `action=kill` with no execution id, or `wait`/`kill`/`subscribe`/`unsubscribe` against a sub-resource path like `/executions/{id}/output`).

3. **`src/agent/modelHandlers/openrouter/modelHandlerOpenRouterNative.ts` (Medium)** — removed all 12 `as ChatMessages` / `as ChatContentItems` / `as ChatAssistantMessage` / `as ChatContentText` casts. Each one turned out to be unnecessary once traced against the OpenRouter SDK's actual (already-discriminated) types — `ChatChoice.message` is already typed `ChatAssistantMessage`, and `ChatMessages`/`ChatContentItems` are `role`/`type`-discriminated unions that TypeScript narrows correctly through plain `Array.isArray`/equality checks. No behavior change; the casts were pure noise hiding correct narrowing that already worked.

## Issue acceptance gates

No linked issue — this is a direct implementation of the audit's High/Medium findings (desktop IPC bypass, ExecutionsTool grab-bag schema, OpenRouter cast noise) reported earlier in the session that requested this refactor.

## Single source of truth

- Desktop file-selection dispatch now defers to `MainViewInboundMessageSchema` (`src/shared/schemas/mainView/inbound.ts`) as the sole validator for main-view inbound commands, same as every other desktop IPC adapter.
- `ExecutionsTool`'s action-to-valid-fields mapping is now owned by the schema itself (`ExecutionsToolActionSchema`), not re-derived by hand in `execute()`.

Removed duplication: `desktopFileSelection.ts` no longer re-implements ad hoc field guards that `MainViewInboundMessageSchema` already encodes. No new cross-file abstractions were added — the new `ViewActionSchema`/`WaitActionSchema`/`KillActionSchema`/`SubscribeActionSchema`/`UnsubscribeActionSchema` are module-private schema literals local to `ExecutionsTool.ts`, following the same in-file pattern as `PlanTool.ts`'s discriminated union.

## Net elements (R6)

3 files changed, +174/-79 (diffstat), net +95 lines. No new exported symbols (`git diff | grep '^[+-]export'` is empty). No new classes/interfaces. `ExecutionsTool.ts` gains 6 module-private `const` schema declarations replacing 1 flat one — the growth is the discriminated-union branches carrying their own descriptions (previously interpolated into one shared description string), not new abstraction layers.

## Consumer counts (R8)

n/a — no emit path or export was deleted or re-routed.

## Host impact

Extension: none (dispatch path untouched; already used the shared schema correctly).

Desktop: `desktopFileSelection.ts`'s `handleMessage` now validates via `MainViewInboundMessageSchema.safeParse` before dispatch, same as `desktopShellIpc.ts`. Behavior-preserving for all four handled commands (`SELECT_MULTIPLE_FILES`, `REQUEST_BASE_FILE`, `REFRESH_ALL_FILES`, `REQUEST_EDITED_FILE`).

CLI: none.

## Scheduled refactoring

None. The audit also flagged a larger, riskier finding (the `UserVars = Record<string, unknown>` template-variable grab-bag in `src/agent/utils/userVars.ts`, ~50+ keys across 5 functions) that was intentionally left out of this PR — it needs a broader restructuring across `PromptBuilder.ts` and template-substitution call sites that deserves its own focused pass rather than being bundled here.

## Validation

- `npm run typecheck` (full workspace: root, test-kernel, agent package build, cli, trace-viewer, desktop) — clean.
- `npx vitest run` targeted suites: `ExecutionsTool*.vitest.ts` (4 files), `ToolInputPreview.vitest.ts`, `ElectronFileSelection.vitest.ts`, `ElectronMainViewIpc.vitest.ts`, and all OpenRouter/model-handler suites (`ModelHandlerOpenRouterNative.vitest.ts` + 8 related files) — 253 tests, all passing.
- `npx eslint` and `npx prettier --check` on all three changed files — clean.
- `npm run check:dead-code-ratchet` — no new unused exports/files.
- Full `npm test` was attempted but exceeds this session's time budget (853 test files); not run to completion. The full workspace typecheck plus the targeted suites above cover the changed code paths directly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DfodXxBPqVo8wEhr8zB3vB)_

---

> [!NOTE]
> **Medium Risk**
> Touches desktop IPC validation and the ExecutionsTool input schema (agent-facing tool contract). Mostly behavior-preserving, but schema tightening can newly reject previously-ignored invalid inputs.
> 
> **Overview**
> Aligns three audit findings with existing codebase patterns — behavior-preserving type and schema tightenings.
> 
> **Desktop file selection** now validates inbound IPC through `MainViewInboundMessageSchema` at the entry point (matching `desktopShellIpc` and siblings), replacing ad-hoc `typeof` field guards so desktop and extension hosts share one wire contract.
> 
> **`ExecutionsTool`** converts its flat grab-bag input schema into a `z.discriminatedUnion` on `action`, so each branch only accepts fields that action uses. Invalid combos (e.g. `kill` without an id, or non-`view` actions on sub-resource paths) are now rejected instead of silently ignored. A preprocess step preserves omitted/`null` `action` defaulting to `view`.
> 
> **OpenRouter native handler** drops unnecessary `as ChatMessages` / `as ChatAssistantMessage` casts; SDK types already narrow correctly without them.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0b311aefc3dfee2c485554cfa5b89e31399979a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/desktop/src/main/desktopFileSelection.ts
+++ b/packages/desktop/src/main/desktopFileSelection.ts
@@ -13,12 +13,25 @@ import { listWorkspaceFiles } from '@common/files/workspaceFileListing';
 import { platform } from '@platform/platform';
 import { relativeToRoot } from '@platform/defaults/nodeWorkspace';
 import { MAIN_VIEW_COMMANDS } from '@shared/ipc';
+import {
+  MainViewInboundMessageSchema,
+  type MainViewInboundMessage,
+} from '@shared/schemas';
 import { normalizeFilePath } from '@utils/core';
 
 import type {
   DesktopCommandMessage,
   DesktopMessageHandler,
 } from './desktopIpcTypes.js';
+
+type MessageFor<C extends MainViewInboundMessage['command']> = Extract<
+  MainViewInboundMessage,
+  { command: C }
+>;
+
+type SelectMultipleFilesMessage = MessageFor<
+  typeof MAIN_VIEW_COMMANDS.SELECT_MULTIPLE_FILES
+>;
 
 interface DesktopFileSelectionDialogOptions {
   title: string;
@@ -168,22 +181,19 @@ export function createDesktopFileSelection(
   }
 
   function isDesktopMultiFileType(
-    value: unknown,
+    value: string,
   ): value is DesktopMultiFileType {
-    return (
-      typeof value === 'string' &&
-      Object.hasOwn(MULTI_SET_COMMAND_BY_FILE_TYPE, value)
-    );
+    return Object.hasOwn(MULTI_SET_COMMAND_BY_FILE_TYPE, value);
   }
 
-  async function selectMultipleFiles(message: DesktopCommandMessage) {
+  async function selectMultipleFiles(message: SelectMultipleFilesMessage) {
     if (!options.showOpenFileDialog) return;
     if (!isDesktopMultiFileType(message.fileType)) {
       // The shared webview posts this for 'output' too, which the desktop has
       // no picker for. Say so rather than dropping it: handleMessage already
       // reported the message as handled.
       createChannelTrace('DesktopFileSelection').warn(
-        `Unsupported multiple file selection: ${String(message.fileType)}`,
+        `Unsupported multiple file selection: ${message.fileType}`,
       );
       return;
     }
@@ -192,8 +202,7 @@ export function createDesktopFileSelection(
 
     const fileType = message.fileType;
     const listConfig = getFileListConfig(fileType, loadFileListSettings());
-    const currentFile =
-      typeof message.currentFile === 'string' ? message.currentFile : undefined;
+    const currentFile = message.currentFile;
     const defaultPath =
       currentFile != null
         ? resolveWorkspaceFile(workspacePath, currentFile)
@@ -223,7 +232,7 @@ export function createDesktopFileSelection(
     });
   }
 
-  function handleMessage(message: DesktopCommandMessage): boolean {
+  function dispatch(message: MainViewInboundMessage): boolean {
     switch (message.command) {
       case MAIN_VIEW_COMMANDS.SELECT_MULTIPLE_FILES:
         runAsync(selectMultipleFiles(message));
@@ -235,15 +244,20 @@ export function createDesktopFileSelection(
         runAsync(refreshDiskBackedDropdowns());
         return true;
       case MAIN_VIEW_COMMANDS.REQUEST_EDITED_FILE:
-        runAsync(
-          updateEditedFiles(
-            typeof message.baseFile === 'string' ? message.baseFile : undefined,
-          ),
-        );
+        runAsync(updateEditedFiles(message.baseFile));
         return true;
       default:
         return false;
     }
+  }
+
+  // Single discriminated-union parse at the entry point, matching every
+  // other desktop IPC adapter (see desktopShellIpc.ts) — the switch above
+  // then operates on the narrowed variant with no per-case guessing at field
+  // types or presence.
+  function handleMessage(message: DesktopCommandMessage): boolean {
+    const parsed = MainViewInboundMessageSchema.safeParse(message);
+    return parsed.success ? dispatch(parsed.data) : false;
   }
 
   return { handleMessage };
--- a/src/agent/modelHandlers/openrouter/modelHandlerOpenRouterNative.ts
+++ b/src/agent/modelHandlers/openrouter/modelHandlerOpenRouterNative.ts
@@ -73,10 +73,8 @@ import type {
   ChatStreamChunk,
   ChatUsage,
   ChatMessages,
-  ChatAssistantMessage,
   ChatToolCall,
   ChatContentItems,
-  ChatContentText,
   ChatRequest,
 } from '@openrouter/sdk/models';
 
@@ -462,7 +460,7 @@ export class ModelHandlerOpenRouterNative extends ModelHandler<
   }
 
   createAssistantMessage(text: string): ChatMessages {
-    return { role: 'assistant', content: text } as ChatMessages;
+    return { role: 'assistant', content: text };
   }
 
   extractAssistantText(message: ChatMessages): string | undefined {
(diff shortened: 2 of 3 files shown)


LionSR:
Thanks for the reviews. Pushed 6e44fb9 addressing the two real findings:

**Exhaustiveness switch (claude[bot] + texra-ai[bot], `ExecutionsTool.ts:653`)** — fixed. Added `default: return assertNever(input, 'Unrecognized executions action');`. Note: a bare `input.action satisfies never` doesn't actually compile here — I verified it separately — because Zod's inferred discriminated-union output type makes property access on the narrowed-to-`never` `input` a type error (`Property 'action' does not exist on type 'never'`), unlike a hand-written TS union. Asserting on `input` itself (not `input.action`) avoids that, and matches the repo's existing `assertNever(value, message)` helper already used by `TextEditorTool.ts` and others.

**`action` required-in-schema regression (codex P2, `ExecutionsTool.ts:584`)** — fixed, and this one was real: I'd checked `toJSONSchema`'s `required` output earlier but with the wrong `io` mode. Re-checked against the actual `TOOL_JSON_SCHEMA_OPTIONS` (`io: 'input'`) and confirmed the original flat schema's `.prefault('view')` produced `required: ["path"]`, while my discriminated-union version produced `required: ["path", "action"]` on every branch — a real violation of AGENTS.md's "design for the model's first call" rule. Fixed by making the `view` branch's `action` field `.optional().default('view')` instead of a bare literal (verified `z.discriminatedUnion` still matches an omitted `action` to this branch and populates the default in the parsed output), which let me drop the top-level `z.preprocess` entirely — also resolves claude[bot]'s minor note about `.preprocess` not covering an explicit `action: undefined`, since there's no more preprocess step to have that gap.

**Flattened-schema field permissiveness (codex P2, `ExecutionsTool.ts:539`)** — not changing. Checked `toolConversion.ts:471-504`: OpenAI tool calls in this repo run with `strict: false` specifically because several existing tools (including the already-shipped `PlanTool`/`MemoryTool`, which use this exact discriminated-union-with-per-branch-only-fields pattern) have optional/branch-specific fields incompatible with strict mode's "all properties required, null-fill the rest" behavior. So a provider isn't forced to null-fill cross-branch fields like `timeout` on a `kill` call — the concern would need a fix to the shared `flattenTopLevelUnion`/strict-mode architecture, and would apply equally to `PlanTool`/`MemoryTool` today, not something scoped to this PR's `ExecutionsTool` change.

Re-ran full workspace typecheck, the `ExecutionsTool*`/`ToolInputPreview`/`AgentPromptContracts`/`ToolConversion` suites (827 tests), lint, and prettier after the fix — all clean.

---

LionSR:
Good catch — that was real, and I should have caught it myself given AGENTS.md's explicit `.nullish()`-not-`.optional()` rule. Pushed 6afb89e: `action` on the `view` branch is now `z.literal('view').nullish().transform((v) => v ?? 'view')`. Verified directly:

- `{path}` (omitted), `{path, action: null}`, and `{path, action: 'view'}` all parse to `action: 'view'`.
- The view branch's JSON-schema `required` is still just `["path"]` (checked against the real `io: 'input'` `TOOL_JSON_SCHEMA_OPTIONS`).

Re-ran full workspace typecheck + the same 827-test suite from the last round, lint, and prettier — all clean.

---

LionSR:
Confirmed and fixed in b0b311a — this was a real P1, independently caught by both Codex and Cursor Bugbot, and worse than the bug my previous fix was solving. `.nullish()` on the `view` branch's `action` field made Zod emit `anyOf: [{const: 'view'}, {type: 'null'}]` for that property; `flattenTopLevelUnion`'s `schemaLiteralValue` only recognizes a bare `const`/single-value `enum` as a discriminator literal, so it fell back to that raw `anyOf` shape for the merged property — silently dropping `wait`/`kill`/`subscribe`/`unsubscribe` from the OpenAI/Anthropic/Google-facing schema.

Fix: `action: z.literal('view').optional().default('view')` (confirmed via `toJSONSchema` this emits a clean `const: 'view'`, not `anyOf`) plus a narrow top-level `z.preprocess` that only strips an explicit `action: null` down to "key absent" before parsing — the branch's own default then takes over exactly as it does for a truly omitted key.

Verified end-to-end against the actual `ExecutionsTool` instance's `convertToolSchema(tool.definition)` output (not just an isolated repro): `action` now shows `enum: ["view","wait","kill","subscribe","unsubscribe"]` with `required: ["path"]`. Also re-verified omitted/`null`/`'view'`/`'wait'`/`'kill'` all still parse correctly, full workspace typecheck, the same 827-test suite, lint, and prettier.

---

PR #10036, opened 2026-08-12
chore(deps): update compatible dependencies

LionSR:
## What changed

- update llm-zoo from 1.25.0 to 1.27.0 across the workspace and relay
- update Claude Agent SDK to 0.3.228 and other dependencies that remain within their current major versions, including OpenRouter SDK 1.2.29 and pdf-lib 2.8.2
- update the relay Hono dependency from 4.12.33 to 4.13.1 and regenerate its Deno lockfile
- retarget the unchanged response.metadata compatibility patch from OpenAI 7.3.0 to 7.4.0
- remove the redundant Claude SDK stream cast and simplify background-task card ownership
- sweep retired models from persisted enabled lists on every startup, including non-preferred models, without restoring preferred defaults that users deliberately disabled
- preserve the established preferred-model version hash so ordinary catalogue growth and non-preferred retirements do not reset user choices

Major-version upgrades remain separate because they require independent compatibility review. The deprecated identifiers-arxiv package is also unchanged because replacing it is a migration rather than a version update.

## Why

The model catalogue and Claude integration should track the current compatible vendor releases. Keeping the workspace and relay on the same llm-zoo version prevents model availability from differing between local and hosted execution.

The refreshed catalogue retires several non-preferred entries. Retired entries are now removed independently of the preferred-model version, while the default-addition pass remains restricted to actual preferred membership or lifecycle changes.

## Validation

- corepack pnpm install --frozen-lockfile
- npm run check:extension-package-invariants
- npm run lint
- npm run typecheck
- npm run compile:fast
- deno check supabase/functions/relay/index.ts
- 70 focused model-catalogue and registry tests on llm-zoo 1.27.0
- focused dependency, relay, and CLI tests
- full npm test: 9,941 passed, 5 skipped on the preceding compatible-release head
- exact-head GitHub Actions CI and automated review

---

> [!NOTE]
> **Medium Risk**
> Dependency and llm-zoo catalogue updates affect model availability and provider behavior; the new every-startup retired-model sweep can silently remove enabled models from shared `~/.texra` state.
> 
> **Overview**
[...]

PR #10058, opened 2026-08-12
fix: align usage pricing model catalog

### LINKED ISSUE #10060: Bump the supabase log-usage llm-zoo pin to match the workspace and relay
## Context

Follow-up from #10036 (chore(deps): update compatible dependencies).

The workspace and the relay function moved to `llm-zoo@1.27.0`, but `supabase/functions/log-usage/deno.json` still pins `npm:llm-zoo@1.25.0` (its `deno.lock` was not regenerated either). `log-usage/equivalentCost.ts:21` derives subscription-usage list prices straight from `MODEL_CONFIGS`, so any list-price change across the two skipped minor versions leaves hosted notional-cost accounting (`subscription_usage_logs.cost`) on a stale catalogue while the client and relay use the new one — the same local/hosted catalogue divergence the PR's own "Why" section says the bump is meant to prevent. Flagged on multiple review passes (claude[bot], texra-ai[bot]) and still open at merge.

## Details

- Bump `llm-zoo` in `supabase/functions/log-usage/deno.json` to `npm:llm-zoo@1.27.0` and regenerate its `deno.lock`.
- Update the `llm-zoo@1.25.0` price comment in `equivalentCost_test.ts` if it references the version.
- If the pin is intentional, document why instead (PR body or an adjacent comment).

## References

- PR: #10036
- Review comment: https://github.com/LionSR/TeXRA/pull/10036#discussion_r3769673918

LionSR:
Closes #10060.

## Summary

- update the `log-usage` Edge Function to the same `llm-zoo` 1.27.0 catalogue used by the workspace and relay
- regenerate its Deno lockfile and keep the price-regression documentation synchronized
- extend the existing catalogue-parity invariant and release checklist to cover both server-side consumers

## Why

The compatible dependency refresh in #10036 updated the client and relay catalogue, but `log-usage` remained on 1.25.0. Because that function derives subscription usage costs from `MODEL_CONFIGS`, the stale pin could make hosted cost accounting disagree with the current application catalogue. This follow-up incorporates the final review finding that arrived as #10036 merged and prevents the same drift from recurring.

## Validation

- `deno test --minimum-dependency-age=0 --lock=deno.lock equivalentCost_test.ts` (5 passed)
- `pnpm vitest run --config vitest.config.mjs src/test-kernel/supabase/RelaySharedConfigParity.vitest.ts` (6 passed)
- exact-head GitHub Actions CI and automated review

### DIFF
--- a/.claude/skills/releasing/SKILL.md
+++ b/.claude/skills/releasing/SKILL.md
@@ -45,10 +45,10 @@ Releases. No local `vsce` / `ovsx` / `npm publish` invocation and no OTP.
    `true`.
 
 5. **`llm-zoo` pin.** If the release changes `llm-zoo`, also update the exact
-   pin in `supabase/functions/relay/deno.json` and refresh
-   `supabase/functions/relay/deno.lock`. This is called out in the
-   `version-bump.yml` PR body but is easy to miss, since it isn't part of the
-   automated bump.
+   pins in `supabase/functions/relay/deno.json` and
+   `supabase/functions/log-usage/deno.json`, then refresh both adjacent
+   `deno.lock` files. This is called out in the `version-bump.yml` PR body but
+   is easy to miss, since it isn't part of the automated bump.
 
 ## Gotchas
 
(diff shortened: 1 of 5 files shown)


LionSR:
needs to deploy again too

PR #10077, opened 2026-08-12
fix(claude): fail fork missing session id

### LINKED ISSUE #10037: Pin fork behavior for success results missing session_id
## Context

Follow-up from #10034 (fix: retire unreachable Claude retry state).

Removing the unreachable fork-retry branch left `forkSession` passed only when `isFirstTurn && params.forkSession` (`src/tools/claudeAgent.ts:448-454`). `TurnResult.sessionId` is still modeled as optional (`string | undefined`) with a defensive `'session_id' in raw` read, but no test pins what happens when a success result omits `session_id`.

## Details

- The pinned SDK type (`SDKResultSuccess.session_id: string`, required) excludes a missing `session_id`, so today the path relies on the type contract alone. If a legacy or malformed SDK message ever reported `subtype: 'success'` without `session_id`, the second turn would call `resume: <source-id>` without `forkSession`, mutating the source conversation instead of re-forking.
- Add a test pinning the behavior for a success result missing `session_id`, and optionally an explicit guard or failure in that path so it does not silently rely on the SDK type contract.

## References

- PR: #10034
- Review comment (texra-ai): https://github.com/LionSR/TeXRA/pull/10034#pullrequestreview-4918950009

LionSR:
## Summary

A successful Claude fork must acquire a distinct session identity before any later turn may resume. This change treats a fork success without `session_id` as an explicit failed turn, preventing follow-ups from resuming and mutating the source conversation.

The missing-ID diagnosis is applied only when the turn would otherwise be successful. If the provider later reports a real error, that error remains authoritative.

Closes #10037.

## Issue acceptance gates

- A successful fork without `session_id` fails loudly.
- The failed turn satisfies the child-loop `isTurnError` boundary, so the loop stops before another provider turn can resume the source session.
- A normal fork with a distinct ID still resumes the forked session on later turns and never repeats `forkSession`.
- A provider error is not overwritten by the missing-ID guard.

## Single source of truth

`runStreamedTurn` owns interpretation of the provider stream and returns one `TurnResult`. The shared child-run loop owns whether a failed turn terminates the loop. The Claude session registry is updated only by the existing successful-turn path, after a valid session identity exists.

No abstraction or compatibility path is added. The change adds one post-stream invariant to the existing turn interpreter and focused regression coverage.

## Net elements (R6)

- 2 files changed.
- No exports, classes, interfaces, enums, schemas, or public methods added or removed.
- One boolean turn invariant added; tests expanded for missing identity and diagnostic precedence.

## Consumer counts (R8)

`runStreamedTurn` has two production consumers in `src/tools/claudeAgent.ts`: first-turn launch and subsequent-turn execution through the same strategy. Both receive the same `TurnResult`; the existing `isTurnError: turn => turn.isError` boundary has one strategy registration and is unchanged.

## Host impact

Extension, desktop, and CLI share the same Claude-agent tool and child-run loop. All now reject an unidentified fork before any host can submit a follow-up against the source session.

## Scheduled refactoring

None. A fork that echoes the source ID rather than omitting it is not generalized here without evidence that the provider can produce that response.

## Validation

- `corepack pnpm vitest run src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts src/test-kernel/agent/followUp/ClaudeAgentProgressEvents.vitest.ts` — 16 passed.
- `corepack pnpm exec eslint src/tools/claudeAgent.ts src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts` — clean.
- `corepack pnpm exec tsc --noEmit --checkers 8 -p tsconfig.test-kernel.json` — clean.
- `corepack pnpm exec prettier --check ...` and `git diff --check` — clean.

### DIFF
--- a/src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts
+++ b/src/test-kernel/tools/ClaudeAgentResumeFallback.vitest.ts
@@ -555,17 +555,56 @@ describe('claude_agent tool launch and resume fallback', () => {
       session_id: 'source-session',
       fork_session: true,
     });
-    const turn = await captured.strategy?.launch?.(
+    const turn = (await captured.strategy?.launch?.(
       { notify: () => {}, recordCost: () => {} },
       new AbortController(),
-    );
+    )) as
+      | { isError: boolean; sessionId?: string; errorMessage?: string }
+      | undefined;
 
-    expect(turn).toMatchObject({
-      isError: true,
-      sessionId: undefined,
-      errorMessage: expect.stringContaining('missing session_id'),
-    });
+    expect(turn?.isError).toBe(true);
+    expect(turn?.sessionId).toBeUndefined();
+    expect(turn?.errorMessage).toContain('missing session_id');
+    expect(captured.strategy?.isTurnError?.(turn)).toBe(true);
     expect(mocks.query).toHaveBeenCalledOnce();
+  });
+
+  it('preserves a provider error after an earlier fork success result', async () => {
+    mocks.query.mockReturnValue(
+      (async function* () {
+        yield {
+          type: 'result',
+          subtype: 'success',
+          result: 'Intermediate success without a session id.',
+          modelUsage: {},
+          total_cost_usd: 0,
+        };
+        yield {
+          type: 'result',
+          subtype: 'error_during_execution',
+          errors: ['provider failed after success'],
+          modelUsage: {},
+          total_cost_usd: 0,
+        };
+      })(),
+    );
+    const captured = captureStrategy();
+
+    await new ClaudeAgentTool().call({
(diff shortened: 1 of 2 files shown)


LionSR:
Superseded by merged PR #10082 (`a86b38ee36`). Current `main` enforces the stronger fork-identity invariant: it rejects both a missing `session_id` and an ID equal to the source session, preserves any provider diagnostic by appending the distinct-session failure, retains usage/cost, and pins the `isTurnError` boundary plus the no-resume/no-repeat-fork consequence in `ClaudeAgentResumeFallback.vitest.ts`. I verified these acceptance gates directly against current `main`; combining this branch would duplicate and weaken the landed implementation.

PR #10139, opened 2026-08-13
fix(agent): stop exposing tool-use turns as rounds

### LINKED ISSUE #10138: CLI: tool-use status bar jumps across hidden model rounds
## Problem

The CLI status bar presents the outer tool-use cycle as round progress even though that cycle is one user/session turn containing an arbitrary number of model invocations. The badge therefore stays at the starting value while tools run, then jumps by the hidden invocation count on the next user message.

This makes a resumed or continued chat appear to have skipped user-visible rounds. The underlying session and transcript are intact; the defect is the progress classification.

## Reproduction

Environment:

- macOS, `TERM=xterm-256color`
- TeXRA built from `c70fd52515792d0c87469fa3c4fc8a3dd2f3fcff`
- terminal size: 60 columns × 20 rows
- included API mode, approval policy `ask`

Steps:

1. Run `texra orchestrate` in a real PTY.
2. Submit one research request that causes several delegation/tool calls.
3. Observe the root status bar throughout the first user turn.
4. After completion, resume the session (or submit another user message).
5. Submit a short follow-up.

Observed in a three-agent quantum-channel calculation:

- the first user turn remained at `r1` while the orchestrator delegated two research agents and one reviewer;
- the first post-resume follow-up displayed `r11`;
- the follow-up itself used no tools and completed normally;
- the mathematical answer and transcript rehydration were correct.

## Cause

`ToolUseCycleNode` documents itself as the outer session step, and `ToolUseRoundFlow` documents each model invocation plus tool dispatch as the actual round. Nevertheless, `ToolUseCycleNode` opens its whole outer span with `kind: 'round'` and the starting value of `runState.totalRounds`. The inner flow increments `totalRounds` after each model invocation without opening a new typed round stage.

Thus the sole progress projection says that one long-lived stage is round 1, while the persisted counter advances to 10; the next outer stage begins at round 11.

## Expected behavior

A structural tool-use session turn should not be exposed as workflow round progress. Genuine fixed-round/reflection workflows should retain their `rN/M` progress.

## Impact

The badge is especially confusing after resume, where a healthy conversation appears to have skipped ten turns. It also gives no live information during the model/tool cycles it purports to count.

## Acceptance criteria

- The outer tool-use cycle is not emitted as a typed round-progress stage.
- Structural grouping and canonical completed/failed/cancelled outcomes remain recorded.
- A continued or resumed tool-use chat does not display a round jump caused solely by internal model/tool invocations.
- Genuine workflow/reflection round progress and its planned total remain unchanged.
- A boundary test fixes the stage classification and outcome contract.

LionSR:
## Summary

- classify the outer tool-use cycle as the existing structural `session` stage instead of a progress `round`;
- preserve stage grouping and canonical completed, failed, and cancelled outcomes;
- keep genuine reflection/workflow `rN/M` progress unchanged;
- add a boundary regression test and a user-visible changelog entry.

Closes #10138.

## Root cause and ownership

`ToolUseRoundFlow` defines a round as one model invocation plus its tool dispatch. `ToolUseCycleNode`, by contrast, is the outer session turn and may execute many such rounds.

The outer node nevertheless opened one long-lived `kind: 'round'` stage using the initial `runState.totalRounds` value. The inner flow then incremented that counter after every model invocation without replacing the displayed stage. A tool-heavy first turn therefore remained at `r1`, while the next or resumed turn began at (for example) `r11`.

The fix uses the stage vocabulary's existing distinction: `session` stages structure a trace but are not projected as round progress. No new counter, display exception, or parallel representation is introduced.

## User impact

Continued and resumed tool-use chats no longer show round jumps caused solely by hidden model/tool invocations. Workflow agents with genuine fixed rounds continue to show their planned progress.

## PTY evidence

Reproduced on macOS with `TERM=xterm-256color`, a 60×20 PTY, included API mode, and approval policy `ask`:

- a three-agent quantum-channel workflow remained at `r1` through ten internal model/tool cycles;
- the first post-resume follow-up displayed `r11`;
- transcript rehydration and the mathematical result were otherwise correct.

After this change, the same persisted session was resumed and given another follow-up. The running bar showed elapsed time, access, quota, and context occupancy without a false round badge; the response completed normally.

## Validation

- `npm test` — 10,081 passed, 5 skipped
- `npm run typecheck`
- `npm run lint`
- `npm run compile:fast`
- `npm run texra-local:build`
- targeted stage/status suites — 149 passed
- real 60×20 PTY resume/follow-up verification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@ All notable changes to this project will be documented in this file.
 - **Read-only completion checks stay focused** — end-of-session review no
   longer asks to inspect project plans, Git history, or pull requests when the
   current task and its delegated results already contain the evidence needed.
+- **Tool-use chats no longer show misleading round jumps** — internal model
+  and tool calls no longer appear as skipped user-visible rounds after a chat
+  continues or resumes.
 - **Concurrent app starts refresh bundled agents safely** — simultaneous TeXRA
   processes sharing one data directory no longer fail while updating built-in
   agent definitions.
--- a/src/agent/implementations/flows/tooluse/nodes/ToolUseCycleNode.ts
+++ b/src/agent/implementations/flows/tooluse/nodes/ToolUseCycleNode.ts
@@ -128,14 +128,14 @@ export class ToolUseCycleNode<C> extends Node<
       kind: 'session',
     });
 
-    let roundOutcome: RunOutcome = RUN_OUTCOME.FAILED;
+    let sessionOutcome: RunOutcome = RUN_OUTCOME.FAILED;
     try {
       await sessionStage.within(() => flow.run(roundShared));
 
       const lastError = roundShared.shouldStop
         ? roundShared.lastError
         : undefined;
-      roundOutcome = deriveRunOutcome({
+      sessionOutcome = deriveRunOutcome({
         failed: lastError !== undefined,
         cancelled: roundShared.shouldStop && !roundShared.endTurn,
       });
@@ -147,7 +147,7 @@ export class ToolUseCycleNode<C> extends Node<
           response: roundShared.latestAssistantText,
         };
       }
-      if (roundOutcome === RUN_OUTCOME.CANCELLED) {
+      if (sessionOutcome === RUN_OUTCOME.CANCELLED) {
         return {
           outcome: 'cancelled',
           response: roundShared.latestAssistantText,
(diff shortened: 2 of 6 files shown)


LionSR:
Addressed the actionable terminology feedback in `c15455e197`: the outer-stage status is now `sessionOutcome`, and the stream schema, usage monitor, and transcript-recorder comments consistently distinguish reflection rounds from tool-use session turns. I kept the existing boundary test rather than adding a second mocked turn: it already fails on the old classification by requiring a session group and rejecting every persisted `kind: 'round'` entry, while the PR records the real continued/resumed PTY verification. Revalidated the two affected suites (53 tests), workspace and test-kernel type checks, Prettier, diff check, and the pre-push changed-file lint after rebasing onto current `main`.

LionSR:
@codex review

@texra-ai re-review

PR #10190, opened 2026-08-14
refactor(replacement): use shared createPatterns helper in maxRules

LionSR:
## Summary

`src/replacement/helpers.ts` already defines a `createPatterns(items, mapper)` factory documented as "the core abstraction for all pattern generators," but it was not exported, so other modules in the replacement package open-coded the identical `Object.fromEntries(items.flatMap/map(...))` idiom for their bespoke pattern groups.

This PR exports `createPatterns` and routes all 16 open-coded sites in the package through it:
- **`maxRules.ts`** — 9 sites (math operators, text commands, symbol operators, fraction differentials, tilde Greek, tilde-boldsymbol Greek, hat Greek, hat-boldsymbol Greek, function names).
- **`rules.ts`** — 6 sites (letter-notation fixes, environment-end fixes, legacy-font normalization, mathcal-font normalization, and two XML tag-fix groups).
- **`rulesRegex.ts`** — 1 site (spaced text-command normalization).

The transformation is behavior-preserving by construction: `createPatterns(items, m)` *is* `Object.fromEntries(items.flatMap(m))`, and each former `.map(x => [k, v])` becomes a mapper returning `[[k, v]]` (flattened identically). Key insertion order and duplicate-key last-wins semantics are unchanged because both paths feed the same entry sequence to `Object.fromEntries`.

This is the cleanly-safe, self-contained item from a broader tech-debt audit of the replacement/model-handler/host/tool layers. The other audited areas turned out to be already factored to a shared kernel with genuinely-divergent remainders, or to be larger behavior-sensitive refactors better suited to their own reviewed PRs — so they are intentionally out of scope here. (The extension from `maxRules.ts` alone to the rest of the package follows a reviewer suggestion on this PR.)

## Issue acceptance gates

No issue — proactive tech-debt cleanup. No acceptance gates.

## Single source of truth

`createPatterns` in `src/replacement/helpers.ts` is now the single owner of the "build a pattern dictionary from items via a mapper" construction across the whole package — the in-module generators, `maxRules.ts`, `rules.ts`, and `rulesRegex.ts`. After this PR, no `Object.fromEntries` open-coding of pattern maps remains in `src/replacement/` outside the helper's own definition.

Removed all 16 open-coded `Object.fromEntries(items.flatMap/map(...))` pattern-map sites in the package, replacing them with calls to the pre-existing `createPatterns` factory. No new abstraction introduced — the factory already existed with multiple in-module callers; this only widens its visibility to same-package consumers. No pass-through layer added. A singular-pair helper variant was deliberately not added (it would be an unjustified new abstraction); single-pair sites use the `[[k, v]]` mapper shape inherent to `PatternMapper`.

## Net elements (R6)

- Files: 4 changed (`helpers.ts`, `maxRules.ts`, `rules.ts`, `rulesRegex.ts`), 0 added, 0 deleted.
- Exported symbols: +1 (`createPatterns`), −0. (`PatternMapper` deliberately kept unexported — a type export with no consumer would trip the dead-code ratchet.)
- Classes/interfaces/enums: no change.
- Net LoC: **−14** (88 insertions, 102 deletions).

## Consumer counts (R8)

No emit path or export was deleted or re-routed. The one new export, `createPatterns`, gains consumers in the same PR: `maxRules.ts` (9 call sites), `rules.ts` (6), `rulesRegex.ts` (1) — plus its existing in-module callers in `helpers.ts`. This satisfies the dead-code-ratchet's "new export needs a consumer" rule.

## Host impact

Extension: none (shared `src/` core only; no host wiring touched).

Desktop: none.

CLI: none.

## Scheduled refactoring

None. The broader audit's remaining items (KaTeX-macro drift guard, model-handler `createResponseImpl` template method, per-stream KV-store base class) are noted for potential follow-up PRs but are not tracked by an issue here.

## Validation

- `npm run typecheck` — full monorepo type check, green (exit 0); root `tsconfig.json` project re-checked green after the `rules.ts`/`rulesRegex.ts` additions.
- `npx vitest run src/test-kernel/replacement/` — 65/65 pass.
- Confirmed generated `max_style` pattern maps unchanged (1232 `max_style` + 13 `max_style_regex`).
- `eslint` + `prettier --check` on all changed files — clean.
- Grep confirms no `Object.fromEntries` pattern-map open-coding remains in `src/replacement/` outside `helpers.ts`, and no `vscode` import exists anywhere in the package (VS Code-free zone preserved).

### DIFF
--- a/src/replacement/helpers.ts
+++ b/src/replacement/helpers.ts
@@ -15,7 +15,10 @@ type PatternMapper<T> = (item: T) => [string, string][];
  * Creates a pattern dictionary from items using a mapper function.
  * This is the core abstraction for all pattern generators.
  */
-function createPatterns<T>(items: T[], mapper: PatternMapper<T>): PatternDict {
+export function createPatterns<T>(
+  items: T[],
+  mapper: PatternMapper<T>,
+): PatternDict {
   return Object.fromEntries(items.flatMap(mapper));
 }
 
--- a/src/replacement/maxRules.ts
+++ b/src/replacement/maxRules.ts
@@ -6,6 +6,7 @@ import type {
 import { GREEK_LETTERS } from './constants';
 import { NonRegexReplacementCategory, RegexReplacementCategory } from './types';
 import {
+  createPatterns,
   generateDecoratedMathShortcuts,
   generateDecoratorShortcuts,
   generateNestedDecoratorShortcuts,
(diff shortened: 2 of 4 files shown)


LionSR:
The `kernel tests (linux, shard 1/2)` red (and the `validate` gate that aggregates it) was a flaky failure **unrelated to this PR**, not a regression:

- The only failing test was `src/test-kernel/utils/system/SystemUtils.vitest.ts > executeCommand > unblocks timed-out array-form await when a descendant holds stdio` — a subprocess-spawn timing test that asserts `Number.isInteger(childPid) && childPid > 0` after reading a spawned sleeper's PID file. The PID file simply wasn't ready in time on the runner (a process race), 1 failed / 5651 passed.
- This diff touches only `src/replacement/` (LaTeX replacement-pattern generation) — nothing in `src/utils/system/` or any process-spawning path.
- Ran that test file locally 3×: 23/23 pass each time.

I've re-run the failed jobs; the shard should go green on the retry. The earlier `validate` red on the superseded commit `fa69658` was a separate artifact — GitHub auto-cancelled that commit's in-flight `test` job when the follow-up commit was pushed.

---

PR #10204, opened 2026-08-14
Fold agent SDK package behind @agent/runtime barrel (10 → 7 specifiers)

LionSR:
## Summary

Applies the same barrel fold-in pattern from #10011 (which sealed the three hosts) to the SDK package itself, reducing its pinned runtime specifiers from 10 to 7. This decouples the package from the runtime's internal file layout while maintaining the hard constraint that public types must come from narrow modules to avoid provider-type leaks into the published surface.

Also adds a re-verification audit documenting the Agent-SDK readiness state, reconciling prior passes, and establishing the permanent bound on further specifier reduction.

## Issue acceptance gates

Addresses the standing Agent-SDK readiness audit (scheduled 2026-08-14):
- ✓ Reconciles the `-08-12` baseline discrepancy (resolved: two merged commits, not a history gap)
- ✓ Lands the SDK package fold-in (10 → 7 specifiers)
- ✓ Documents the hard constraint on barrel-carried public types (declaration emit leak check)
- ✓ Confirms two previously-open items are already done (phantom contract arms, embedder smoke test)
- ✓ Corrects the `-08-12` "higher-level entry" proposal as conflicting with the north-star plan

## Single source of truth

The curated `@agent/runtime` barrel (`src/agent/runtime/index.ts`) now owns the runtime's public surface for the SDK package. Public types that appear in the package's declarations (`AgentFlowResult`) remain sourced from their own narrow modules to prevent declaration-emit graph expansion into the published `.d.ts` surface.

No new abstraction added. This is a barrel fold (same pattern as #10011), not a wrapper layer — no pass-through functions, no facade, no new exports. The deep specifiers it replaces are deleted in the same change.

Notably, this does **not** implement the "higher-level public-typed entry" floated by the `-08-12` pass: that is precisely the `runSession()` facade the north-star lists under "do not relitigate" (§5). §6 of the audit records that correction.

## Net elements (R6)

**Files changed:** 4
- `packages/agent/src/index.ts` — imports routed through the barrel + the hard constraint documented at the import site
- `src/test-kernel/agent/AgentPackage.vitest.ts` — mocks the one barrel door instead of each module path
- `config/ratchets/host-agent-import-baseline.json` — baseline shrunk from 10 to 7 specifiers
- `docs/proposals/2026-08-14-agent-sdk-readiness-reverify.md` — new audit document

**Exported symbols:** 0 net change (re-routed, not added/deleted)

**Specifiers reduced:** 10 → 7
- `@agent/runtime/runAgent` → `@agent/runtime`
- `@agent/runtime/SessionHandle` → `@agent/runtime`
- `@agent/runtime/ExecutionHandle` → `@agent/runtime`
- `@agent/runtime/HostInteractions` → `@agent/runtime`
- `@agent/index/agentRegistry` → `@agent/index`
- `@agent/runtime/AgentFlowResult` — kept deliberately (public type, cannot route through the barrel)

## Consumer counts (R8)

The fold is internal to `packages/agent/src/index.ts` and its test mock. The SDK package was the sole consumer of the removed deep specifiers.

## Host impact

No host is touched by this diff. Current checked-in baselines (post-#10011/#10118, as recorded in §2 of the audit):

| Package | Specifiers |
| --- | --- |
| Extension | 17 (unchanged) |
| CLI | 18 (unchanged) |
| Desktop | 13 (unchanged) |
| Agent package | 10 → 7 (this PR) |

## Scheduled refactoring

None. The audit concludes the remaining 7 specifiers are near the realistic floor without redesigning the barrel's type graph — bounded by the hard constraint in §4.

## Validation

- ✓ `npm run typecheck` — clean
- ✓ Agent package build validates 720 declarations / 69 external packages (this is the gate that caught the provider-type leak)
- ✓ `npx eslint` and `npx prettier --check` on all changed files — clean
- ✓ `npm run check:dead-code-ratchet` — no new unused exports
- ✓ 3,107 tests pass across `src/test-kernel/architecture/` and `src/test-kernel/agent/`, including the R-b deep-import ratchet and the `AgentPackage` consumer-contract suite

https://claude.ai/code/session_019q8XgkGNRPdK9BVqBgUmJg

### DIFF
--- a/config/ratchets/host-agent-import-baseline.json
+++ b/config/ratchets/host-agent-import-baseline.json
@@ -59,12 +59,9 @@
       "@agent/core/definition/AgentConfig",
       "@agent/core/definition/AgentDataclass",
       "@agent/core/tools/ToolTypes",
-      "@agent/index/agentRegistry",
+      "@agent/index",
+      "@agent/runtime",
       "@agent/runtime/AgentFlowResult",
-      "@agent/runtime/ExecutionHandle",
-      "@agent/runtime/HostInteractions",
-      "@agent/runtime/runAgent",
-      "@agent/runtime/SessionHandle",
       "@agent/trace"
     ]
   }
--- a/docs/proposals/2026-08-14-agent-sdk-readiness-reverify.md
+++ b/docs/proposals/2026-08-14-agent-sdk-readiness-reverify.md
@@ -75,8 +75,9 @@ itself at 10 specifiers. This pass applies the same pattern to that package:
 | `@agent/runtime/AgentFlowResult`  | ✓      | **kept deliberately** — see §4 |
 
 Net: **10 → 7**, baseline shrunk in `config/ratchets/host-agent-import-baseline.json`.
-The package no longer pins five internal runtime module paths; runtime files can
-move or split without breaking it.
+The package stops pinning five internal module paths — four under `runtime/` plus
+`index/agentRegistry` — so those files can move or split without breaking it.
+`@agent/runtime/AgentFlowResult` stays pinned by design (§4).
 
 This is a **barrel fold, not a new layer** — no wrapper function, no facade, no
 new abstraction. The touched files are `packages/agent/src/index.ts`,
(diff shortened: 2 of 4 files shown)


LionSR:
All four review findings were correct and are now fixed in `130ede4f` (doc) plus a PR-body edit. Each cited line was re-verified against the tree before changing anything.

**1 & 2 — §7 evidence was wrong** (raised independently by both reviewers). `runAgent.ts:78-85` is the function's JSDoc block, not the implementation; and "`executeAgent` has ≥3 production entries" was carried forward from the `-08-04` note without rechecking — there are exactly two direct callers. Corrected to:

> `runAgent` validates-then-runs: it assigns the `executionId` (`runAgent.ts:103`), registers the execution (`:121`), and calls the engine (`:157`), which forwards `openWorkflowOutput` so the host can surface a workflow result (`executeAgent.ts:474`). The `executeAgent` module's two engine entries have ≥4 production consumers: `executeAgent` itself from `runAgent.ts:157` and `nativeSubagentStrategy.ts:254`, and its sibling `resumeToolUseFromResumeData` (`executeAgent.ts:522`) from `resumeQueuedToolUse.ts:134`, `nativeSubagentStrategy.ts:306`, and the CLI resume path (`chatSessionController.ts:573`) — the resume entry CLAUDE.md deliberately distinguishes from `runAgent`.

The layering conclusion is unchanged; only the evidence was wrong. Worth fixing precisely because these notes chain forward — the `-08-12 → -08-14` correction in §6 exists for the same reason.

**3 — stale host numbers in the PR body.** Confirmed: the body carried the pre-#10011 figures (34/25/31), contradicting §2 of the doc in the same PR. Replaced with the current baselines (extension 17, CLI 18, desktop 13). The truncated Validation section is also restored in full.

**4 — §3 wording.** Both overstatements tightened: the five unpinned paths are four under `runtime/` plus `index/agentRegistry`, and the "runtime files can move freely" claim now carries the `AgentFlowResult` qualifier from §4 inline rather than only later.

On the third reviewer's non-blocking observation that the suite now mocks both barrels at the door: agreed, and left as-is deliberately. It matches the #10011 host-test pattern, and the real barrel graph is covered at build time by `validate-artifacts.mjs` walking the emitted declarations — which is what caught the provider leak in the first place.

---

PR #10216, opened 2026-08-14
test(cli): pin cancelled non-empty output command path

LionSR:
## Summary

Adds a command-level `mockWorkflowExecution` regression test for a cancelled workflow run with non-empty provisional outputs and a requested `--output` destination. The test pins the full path that #10182's fix moved to `resolveWorkflowOutput`: the destination is never created, the emitted result carries no `copiedOutput`/`copiedOutputs`, the text result reports the run-storage directory, and the recovery hint is still printed.

## Issue acceptance gates

- #10186: one command-level test covers the cancelled run with non-empty outputs plus a requested destination, asserting all four acceptance points (no destination write, no copy metadata, run-directory text result, recovery hint).

## Single source of truth

n/a (test-only)

n/a

## Net elements (R6)

n/a (test-only)

## Consumer counts (R8)

n/a

## Host impact

Extension: none

Desktop: none

CLI: test-only

## Scheduled refactoring

None

## Validation

- `npx vitest run src/test-kernel/cli/WorkflowRunCommand.vitest.ts` passed (30 tests)
- Full CI runs on this PR

### DIFF
--- a/src/test-kernel/cli/WorkflowRunCommand.vitest.ts
+++ b/src/test-kernel/cli/WorkflowRunCommand.vitest.ts
@@ -669,6 +669,61 @@ describe('CLI workflow run command', () => {
     });
   });
 
+  it('keeps cancelled non-empty outputs in run storage and still prints the run directory and recovery hint', async () => {
+    await withTempRoot(async (root) => {
+      const generated = await writeGeneratedOutput(root);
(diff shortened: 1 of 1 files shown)


LionSR:
Superseded by #10215, which covers both --output and --output-dir for the same cancelled non-empty-output command path.

PR #10347, opened 2026-08-15
Move chatgpt-subscription fallback provider policy into ProgressApiKeyRetryController

LionSR:
## Summary

An abstraction-layer audit found that `packages/extension/src/progressView/frontend/eventHandlers.ts` (a UI/render-layer webview handler) was computing credential-routing *policy* inline: mapping `exhaustionReason === 'chatgpt-subscription'` to the `'openai'` fallback provider before sending the `USE_OWN_API_KEY` message. That decision belongs to `ProgressApiKeyRetryController` (`src/controllers/progressView/`), which already owns the rest of this retry/credential policy (`shouldDisableIncludedModelAccess`, `isSubscriptionExhausted`). Because `packages/desktop/src/renderer/main.ts` imports this exact webview file, both the extension and desktop hosts shared the drift risk: if a new exhaustion reason were added, only the controller would get updated and the webview's copy would silently go stale.

This PR moves the mapping into the controller and simplifies the webview to a pure pass-through of the raw provider hint, matching the pattern it already uses for `viaRelay`.

## Issue acceptance gates

No tracked issue — found and fixed during a scheduled abstraction-layer audit. Gate: the `chatgpt-subscription` → `openai` fallback-provider decision has exactly one owner, and both retry entry points (generic webview retry and the Copilot-fallback path) resolve it consistently.

## Single source of truth

`ProgressApiKeyRetryController.resolveProvider` (new private method) is now the sole place that maps `exhaustionReason: 'chatgpt-subscription'` to the `'openai'` fallback provider. It is used inside `ensureOwnApiKey`, the single choke point both `useOwnApiKey` (generic webview retry) and `runCopilotFallbackWithRouting`'s caller (`ProgressViewMessageHandler.startCopilotFallbackRun`) route through before checking/prompting for a key.

Removed the duplicate copy of the provider-fallback mapping from the webview render layer; no new abstraction layer was added — `resolveProvider` is a private helper on the existing policy-owning class, not a new module.

## Net elements (R6)

Diffstat: 3 files changed, 43 insertions(+), 11 deletions(-) (includes a new regression test). Net new exported symbols: 0 (`resolveProvider` is `private`). No new files, classes, or interfaces.

## Consumer counts (R8)

n/a — no emit path or export was deleted or re-routed; this only relocates where an existing value is computed.

## Host impact

Extension: `progressView/frontend/eventHandlers.ts` no longer computes the fallback provider; behavior unchanged (controller resolves it instead).

Desktop: shares the same webview file (`packages/desktop/src/renderer/main.ts` imports it verbatim) and the same controller, so it gets the same fix with no separate change needed.

CLI: unaffected — the CLI's parallel retry flow already resolves this in `packages/cli/src/runtime/approval/approvalPrompts.ts`, not in a render component.

## Scheduled refactoring

None.

## Validation

- `npx vitest run src/test-kernel/controllers/ProgressApiKeyRetryController.vitest.ts src/test-kernel/progressView/PermissionActionEventHandlers.vitest.ts` — 41 passed, including a new test asserting the controller resolves `openai` from `exhaustionReason` alone when `provider` is omitted.
- `npm run typecheck` — full workspace typecheck (workspace, test-kernel, agent, cli, trace-viewer, desktop) passes clean.
- `npx eslint` on the three changed files — no errors.
- `npx prettier --check` on the three changed files — formatted correctly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KA9tVd2ALake7Fmyw1AhRa)_

### DIFF
--- a/packages/extension/src/progressView/frontend/eventHandlers.ts
+++ b/packages/extension/src/progressView/frontend/eventHandlers.ts
@@ -306,12 +306,7 @@ export function handlePermissionAction(
           requestId: data.requestId,
           model: data.model,
           exhaustionReason,
-          // Subscription quota exhaustion always means the OpenAI key is the
-          // fallback credential, regardless of how the error tagged provider.
-          provider:
-            exhaustionReason === 'chatgpt-subscription'
-              ? 'openai'
-              : data.errorDetails?.provider,
+          provider: data.errorDetails?.provider,
           viaRelay: data.errorDetails?.isRelayError === true || undefined,
         });
         break;
--- a/src/controllers/progressView/ProgressApiKeyRetryController.ts
+++ b/src/controllers/progressView/ProgressApiKeyRetryController.ts
@@ -123,9 +123,8 @@ export class ProgressApiKeyRetryController {
   async ensureOwnApiKey(
     request: Omit<ProgressApiKeyRetryRequest, 'stream' | 'requestId'>,
   ): Promise<boolean> {
-    const providersToCheck = request.provider
-      ? [request.provider]
-      : this.deps.providers;
+    const provider = this.resolveProvider(request);
+    const providersToCheck = provider ? [provider] : this.deps.providers;
     const requireChange = request.exhaustionReason === 'upstream-credit';
 
     // The gate depends on which credential failed:
@@ -137,7 +136,7 @@ export class ProgressApiKeyRetryController {
     // credential, which would allow retrying without a usable direct key.
     if (requireChange) {
       const before = await this.readKeys(providersToCheck);
-      await this.deps.promptForApiKey(request.provider);
+      await this.deps.promptForApiKey(provider);
       return this.hasChangedUsableKey(providersToCheck, before);
     }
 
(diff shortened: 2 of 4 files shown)


LionSR:
CI red on `static checks (linux)`, but it's failing on `main` too — `prettier --check` flags formatting in `docs/proposals/2026-08-14-delegation-flow-substrate-consolidation.md`, a pre-existing file this PR doesn't touch (added in #10301, already present on `main` before this branch). None of the three files changed here fail `prettier --check` locally. Will re-run once the base branch recovers.

---

LionSR:
Fixed in a512ce4, per the review threads (both `texra-ai` and `claude[bot]` independently raised the same gap):

- **Non-discriminating controller test**: `ProgressApiKeyRetryController.vitest.ts`'s new test used `keys: { openai: 'stored-openai' }`, so the old multi-provider fallback (`request.provider ? [request.provider] : this.deps.providers`) also found the OpenAI key and produced an identical result — the test passed with or without `resolveProvider`. Switched the fixture to hold only a non-OpenAI (`anthropic`) key; verified locally by running the test against the pre-refactor controller and confirming it now fails there (`proceeded: true` instead of the expected idle result).
- **Missing webview-layer coverage**: added a case to `PermissionActionEventHandlers.vitest.ts` asserting `eventHandlers.ts` now passes a tagged `anthropic` provider straight through un-overridden for a `chatgpt-subscription` exhaustion — locking in that the mapping moved, not just that it still works somewhere.
- **Comment density**: condensed the 4-line `resolveProvider` docstring to one line per the checklist.

---

PR #10348, opened 2026-08-15
fix(latex): consolidate duplicate \criticize brace-matching, fix multi-level nesting

LionSR:
## Summary

This came out of a scheduled structural-debt audit looking for confusing or overlapped module responsibilities. The most significant finding: `\criticize{message}{severity}{confidence}` annotations (inserted by critique-style agents) were parsed by **two independent, drifted implementations**:

- `src/replacement/advanced.ts` (`stripCriticizeAnnotations`) used a fixed-depth regex (`(?:[^{}]|\{[^{}]*\})*`) to strip these annotations from output.
- `src/latex/criticismParser.ts` (`parseCriticismAnnotations`) had its own hand-written brace-balanced parser for the *same* macro, used to render editor decorations — its own docstring explained it existed "in addition to the regex in `replacement/advanced.ts` because that one only handles a single level of brace nesting."

That drift was a real correctness bug, not just duplicate maintenance: a `\criticize` message containing a **doubly-nested** macro (e.g. `\criticize{see \sqrt{\frac{a}{b}}}{2}{3}`) doesn't match the regex's single-level nesting at all, so `stripCriticizeAnnotations` silently left the raw, unstripped `\criticize{...}` annotation in compiled/exported LaTeX output — while the editor-decoration parser handled the same input correctly.

This PR extracts the brace-balanced scanning both files actually needed into one shared utility and has both consumers use it, fixing the bug and removing the duplicate implementation.

## Issue acceptance gates

No linked issue — this is a scheduled autonomous audit PR. Self-imposed gates:
- [x] Identify the most significant overlapping-responsibility issue in the codebase (documented above; four other candidates were surveyed and are lower severity/riskier to auto-fix — see note below).
- [x] Produce a behavior-preserving consolidation, plus a regression test proving the previously-broken case is fixed.
- [x] No new architecture-edge violations, no browser-safety regressions, no dead-code ratchet regressions.

Other candidates surveyed but not acted on here (left for a human/follow-up call, since they're either bigger redesigns or touch security-sensitive code an autonomous run shouldn't change unreviewed): duplicate provider-message-block switch statements in `src/agent/storage/conversationFormat.ts` vs `src/agent/export/normalizeConversation.ts` (already cross-documented as a known coupling); `replacement/engine.ts` invoked from four uncoordinated call sites with no single policy owner; and — most notably — the per-host auth "in-flight sign-in" callback guard being reimplemented three times (extension/desktop/CLI) with drifted security properties (desktop has a nonce anti-hijack check the extension lacks). That last one is a real finding worth a maintainer's attention but isn't something this run should silently patch.

## Single source of truth

`findBraceBalancedMacroCalls` (`src/utils/text/braceBalancedMacro.ts`) is now the one place that knows how to scan a brace-balanced, fixed-arity LaTeX macro call. `criticismParser.ts` owns severity/confidence validation and line/column positioning on top of it; `advanced.ts`'s `stripCriticizeAnnotations` owns unconditional removal (including malformed annotations) on top of it.

Removed: the fixed-depth `CRITICIZE_ARG`/`CRITICIZE_INLINE_RE`/`CRITICIZE_WHOLE_LINE_RE` regexes in `advanced.ts`, and the private `readBraceGroup`/`skipWhitespace` pair duplicated in `criticismParser.ts`. Added: one shared scanner with two real call sites (satisfies the repo's "factories need multiple callers" bar — this isn't a single-caller extraction). No new abstraction layers beyond that.

## Net elements (R6)

Source files only (excludes tests/docs): `+149/-95` lines, net `+54`. New exports: `findBraceBalancedMacroCalls`, `BraceMacroCall` (both consumed immediately by the two existing call sites). No new classes; `criticismParser.ts`'s private `readBraceGroup`/`skipWhitespace` functions were deleted, not added.

## Consumer counts (R8)

`stripCriticizeAnnotations` and `parseCriticismAnnotations` keep their existing public signatures and call sites unchanged (verified via grep — 5 files reference either symbol, all within `src/` plus one extension frontend file; none needed edits). n/a for emit-path/export re-routing.

## Host impact

Extension: `packages/extension/src/frontend/latex/inlineCriticism.ts` calls `parseCriticismAnnotations` — signature and behavior unchanged, now also correctly handles doubly-nested annotation messages.

Desktop: none.

CLI: none (both touched modules are host-agnostic `src/` code with no CLI-specific caller).

## Scheduled refactoring

None. The other audit findings noted above (provider-message-block duplication, `replacement/engine.ts` multi-entry-point policy, and especially the per-host auth callback-guard drift) are worth separate, human-reviewed follow-ups but aren't filed as issues here.

## Validation

- `npm run typecheck` (full workspace: core, test-kernel, agent, cli, trace-viewer, desktop) — clean.
- `npm run lint` — clean.
- `npx vitest run` on `src/test-kernel/latex`, `src/test-kernel/replacement`, and `src/test-kernel/tools/AcceptRunFilesTool.vitest.ts` — 176/176 passing, including two new regression tests for the doubly-nested-brace case (one per consumer).
- `node scripts/check-browser-safe-utils.mjs` — updated the module-count docs in `CLAUDE.md`/`AGENTS.md` to match the new file; guard passes (62 total, 7 browser-reachable, unchanged reachable set — the new file isn't browser-reachable).
- `npm run check:dead-code-ratchet` — clean, no new unused exports.
- Verified `src/replacement/` has no existing dependency on `src/latex/` (and vice versa is already allowed) via `config/ratchets/architecture-edges-baseline.json`; the new shared code was placed in `src/utils/`, which both sides already depend on, specifically to avoid introducing a new `replacement → latex` edge that would have created a layering cycle.

---
_Generated by [Claude Code](https://claude.ai/code/session_018LLfZnKH4ctfEkJSrujutG)_


LionSR:
CI red on `static checks (linux)` (`prettier --check .`), but it's failing on the base branch too: the flagged file, `docs/proposals/2026-08-14-delegation-flow-substrate-consolidation.md`, isn't touched by this PR's diff at all and fails `prettier --check` identically on unmodified `main` (introduced by #10301). Not fixing it here since it's unrelated to this PR's change — will re-run when the base branch recovers.

---

LionSR:
Thanks for the review. Addressed both points:

- Added `src/test-kernel/utils/text/BraceBalancedMacro.vitest.ts` (fd57bba) with dedicated coverage for `findBraceBalancedMacroCalls`: nesting, whitespace tolerance, `argCount` mismatches, the partial-macro-name guard, and resume-after-short-match — independent of the two consumers' tests.
- Confirmed the whitespace-tolerance observation is correct: the old `stripCriticizeAnnotations` regex required `\criticize{` with no space before the first `{`, and no space between `}{`, so e.g. `\criticize {a}{b}{c}` previously passed through unstripped. The shared scanner (matching `criticismParser`'s pre-existing whitespace-tolerant behavior) now strips it too. This is a widening, not a regression — no previously-passing case changed behavior (verified by the full existing `it.each` suite still passing unmodified) — but the PR description's "unchanged for every previously-passing case" phrasing undersold that. Correcting it here for the record.

---

LionSR:
Addressed the remaining feedback from both review passes in 0ed0dab:

- **Efficiency** (`src/replacement/advanced.ts:396`): replaced the per-call backward `lastIndexOf` scan with a forward-only cursor tracked across the loop, since `calls` is already in ascending `start` order. Bounds the whole-line check to O(N) total instead of O(N×M) on a pathological single-line document with many annotations.
- **Test gaps**: added stripper regression cases for whitespace-separated arguments (`\criticize {x} {1} {2}`) and an escaped brace inside the message (`\criticize{a\}b}{2}{3}`) — both were already correctly handled by the new shared scanner (matching `parseCriticismAnnotations`' pre-existing behavior) but untested until now.

One accuracy note on my own earlier framing, since `texra-ai[bot]` correctly called it out: the new `CriticismParser.vitest.ts` doubly-nested-brace case isn't a true regression test — `parseCriticismAnnotations` already handled arbitrary nesting before this PR via its own `readBraceGroup`. Only `stripCriticizeAnnotations` (the regex-based one) had the actual bug. That test exists to confirm the shared-scanner refactor didn't change `criticismParser`'s behavior, not to pin a fix. The `ReplacementRules.vitest.ts` doubly-nested case is the real regression test.

All 81 tests in the three touched suites pass; typecheck, lint, and format are clean.

---

PR #10665, opened 2026-08-15
test(agent): cover no-ALS tool dispatch

### LINKED ISSUE #10652: Add a no-ALS test exercising a full tool-dispatch path
## Context

Follow-up from #10647 (refactor(agent): inject tool-policy fields into cycle flows).

The new no-ALS tests (`ResponseCycleTools.vitest.ts`, `ToolUseWaitNode.vitest.ts`, `RunScopedToolOverlay.vitest.ts`) drive the cycle flows without an AsyncLocalStorage frame, but each stops at a model error before any tool actually executes. They pin that tool-list/setup decisions no longer read the ambient `RunContext`, but nothing exercises a full tool-dispatch path without a frame — the property an SDK embedder driving `runToolUseFlow` directly relies on.

## Details

Add a test that runs `runToolUseFlow` without an ALS frame through an actual tool execution (dispatch + completion), asserting the path never falls back to `currentSession()`/`useRunContext()`. This may surface remaining ambient reads in the `src/tools/` layer (still load-bearing at `proposalFlow.ts:190`, `subagentExecution.ts:145`, `agentCliShared.ts:298`); where full no-ALS dispatch is not yet reachable, document which layer still needs the frame.

## References

- PR: #10647
- Review: https://github.com/LionSR/TeXRA/pull/10647#pullrequestreview-4944727782 (texra-ai[bot] residual-risk note, merged without reply)

LionSR:
## Summary

Adds the missing end-to-end no-ALS regression for `runToolUseFlow`: a real `TodoWriteTool` call now dispatches, updates the live work-plan state, returns its tool result to the model, and reaches normal flow completion without a `RunContext` frame or process-default session.

The test also pins the two injected authorities at this boundary:

- the model-facing tool list follows the supplied `ToolPolicy` by filtering an approval-gated `bash` tool;
- the attached flow context exposes the supplied `runScope.session` as its owner.

No production change was needed; the existing tools-layer contract supports this no-ALS dispatch path.

Closes #10652

## Regression coverage

- Asserts there is no default session and no ambient `RunContext` before or after the flow.
- Observes `tryUseRunContext() === undefined` synchronously from `TodoWriteTool.call()` via its todo progress callback, covering the actual dispatch window.
- Asserts the real tool completes with `status: executed`, its todo update reaches progress state, the model receives the tool result, and the flow completes on the second model turn.
- Negative control: removing `approvalPromptsUnavailable` makes the test fail because `bash` reappears in the model-facing list.

## Net elements (R6)

- Files: 1 modified, 0 added, 0 deleted.
- Exported symbols: 0 added, 0 removed.
- Net LoC: +172 / -1 = +171 (test-only).

## Consumer counts (R8)

No exports or emit paths were added, removed, or rerouted; consumer-count analysis is not applicable.

## Host impact

- Extension: none (test-only).
- Desktop: none (test-only).
- CLI: none (test-only).

## Validation

- Focused/surrounding Vitest: 4 files, 39 tests passed (run-scoped overlay, dispatch parallelism, wait node, response-cycle tools).
- Final focused Vitest after rebase: 1 file, 3 tests passed.
- Full typecheck matrix: workspace, test-kernel, agent, CLI, trace-viewer, desktop passed.
- Architecture suite: 17 files, 102 tests passed (serial worker; parallel run hit the existing 10s scan timeout in `subsystemEdgeRatchet`, while the test passed standalone and in the serial full suite).
- ESLint and Prettier for the changed file passed; `git diff --check` passed.
- `check:dead-code-ratchet`, `check:guidance-refs`, and `check:browser-safe-utils` passed.

### DIFF
--- a/src/test-kernel/agent/RunScopedToolOverlay.vitest.ts
+++ b/src/test-kernel/agent/RunScopedToolOverlay.vitest.ts
@@ -9,8 +9,15 @@ import {
 import { AgentConfigSchema } from '@agent/core/definition/AgentConfig';
 import { MapToolRegistry, type ITool } from '@agent/core/tools/ToolTypes';
 import { runToolUseFlow } from '@agent/implementations/flows/tooluse/runToolUseFlow';
-import { createRunContext, withRunContext } from '@agent/runtime/RunContext';
+import {
+  createRunContext,
+  tryUseRunContext,
+  useRunContext,
+  withRunContext,
+} from '@agent/runtime/RunContext';
 import { createRunScope } from '@agent/runtime/RunScope';
+import { tryDefaultSession } from '@agent/runtime/SessionHandle';
+import type { SdkToolCall } from '@agent/types/ModelHandlerContracts';
 import {
   AgentCategory,
   type ExecutionId,
@@ -18,7 +25,9 @@ import {
 } from '@shared/schemas';
 import { setupPlatform } from '@test/support/setupPlatform';
 import { createTestSession } from '@test/support/sessionTestUtils';
+import { TodoWriteTool } from '@tools/todo/TodoTool';
 import { testModelCell } from './modelCellTestUtils';
+import { roundModelHandler } from './toolUseRoundTestUtils';
 
 const CONFIG = AgentConfigSchema.parse({
   agent: 'chat',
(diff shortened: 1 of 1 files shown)


LionSR:
Closing per maintainer direction: we are overtesting, and these test-only additions do not need to land.

PR #10732, opened 2026-08-16
docs: over-defensive patterns top 10 — let it break

LionSR:
Eighth doc of the consolidation program (`docs/proposals/2026-08-16-overdefensive-top10.md`), executing the maintainer's 2026-08-16 ruling: *"a lot of the complexity comes from being over-defensive on so many fronts — some things we can just let them break."*

Method: 7-agent workflow — four quantified censuses (parse/validation, control-flow, state/copy, error-handling), a ranker, an adversarial verifier (5 CONFIRMED / 5 ADJUSTED / 0 refuted — corrections folded in), and synthesis.

**Scoreboard: ~230 defensive sites censused → ~180 removable across 10 families (est. net −700..−800 LoC of pure defense), gated by ~25 LoC of root-level catches so every removal lands loud.**

Highlights:
- **§0 — the let-it-crash layer map**, verified per layer with file:line catchers: run-path, tool-call, CLI, and extension-command throws are already loud today; desktop-main IPC/background (`reportAsyncError = console.error`), extension view-dispatch (log-only), and the renderer are the three silent rows — gates **G1–G3** (~25 LoC) land first, making REMOVE verdicts executable rather than reckless.
- **#1**: `z.unknown()` on `StreamLogEntry.data` forces 27 downstream `safeParse`+skip branches (~300-400 LoC of "legacy or corrupt?" reasoning) — DEFINE-AWAY via a discriminated union validated once at the two real boundaries (verifier corrected the discriminant to `messageType`).
- **#2**: the `try*` platform accessors — 42 post-init call sites where user settings/goal writes **silently vanish** (§15 M2 at family scale) — REMOVE, sequenced per leg behind the gates.
- **#5**: Zod `.catch()` on our own durable data — the policy violation the schema guide already names.
- **§2 — the boundary charter**: where defensiveness *remains correct* (true boundaries + §15 exceptions + the fenced irreducibility registers), so sweeps don't overshoot into the ~350 legitimate boundary sites.
- **§4 — runners-up recorded as explicit KEEP** so later waves don't re-audit them.

https://claude.ai/code/session_01VtxTTx1y9xMJsDNozz4CR1

### DIFF
--- a/docs/proposals/2026-08-16-overdefensive-top10.md
+++ b/docs/proposals/2026-08-16-overdefensive-top10.md
@@ -1,0 +1,433 @@
+# Over-defensive patterns: the top 10 — let it break (2026-08-16)
+
+> **Status:** Adjudicated audit + removal plan, pinned to origin/main
+> `25fe7d354b`. Executes the maintainer ruling: _defensive code that
+> tolerates states a single owner already excludes is not safety — it is
+> silent-failure manufacturing; the smart move is to let it break and let
+> the layer's catcher make it loud._ Method: four census workflows
+> (parse/validation, control-flow, state/copy, error-handling) over all
+> prod code, merged into a top-10 ranking, then every entry re-verified
+> adversarially at exact file:line. Verifier verdicts: 5 CONFIRMED,
+> 5 ADJUSTED, 0 refuted outright — all corrections are folded in below,
+> and two inflated sub-claims (#5's flagship catch story, 2 of #10's four
+> FS sites) are rewritten to what the code actually does. R6 net-element
+> accounting applies to every PR; R8 consumer-greps are named per entry.
+
+Scoreboard: **~230 defensive sites censused → ~180 removable across 10
+families (est. net −700..−800 LoC of pure defense + one whole
+silent-failure class), gated by ~25 LoC of root-level catches (§3), with
+a boundary charter (§2) so sweeps don't overshoot into the ~350 sites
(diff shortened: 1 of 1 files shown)


LionSR:
Review round addressed in the latest commit — every finding accepted, including five P1 demotions that materially improve the plan: the resume double-read is now DEFINE-AWAY-by-acquire-then-read (or KEEP until that reorder — the retrieval/lease window is real); work-plan strict-parse retains per-field recovery (whole-object-failure→empty would let the next write erase valid siblings); `orchestrate.ts:132` and `diffComputation.ts:48` demoted to KEEP-with-warn (product remediation flow / display enrichment of completed output); `modelAccessRoute.ts:134` reclassified as a user-input boundary; `getServerSideKeyService()`'s `tryGlobalState` recorded as deliberate platformless-SDK support; the UsageMonitor fix's honest outcome corrected to louder-accounting-not-run-fatal (the outer `recordUsage` catch also swallows); and **G1 rewritten so it can never suppress Node's default-throw crash** — surface-then-rethrow is now the stated contract. Also fixed: the dispatcher citation (`src/shared/utils/dispatcher.ts` + the log-only sink split to `BaseViewMessageHandler.ts:146-148`, at all three uses), headline reconciled to the per-entry sum (−660..−710), engine defaults to one retry, prefault-not-partial, Homebrew-not-npm, and both tally labels. Merging per maintainer instruction.

PR #10734, opened 2026-08-16
Fix abstraction-layer violations from architecture audit

LionSR:
## Summary

A scheduled architecture audit swept the codebase for abstraction-layer violations (business logic in the wrong layer, infra concerns leaking into domain logic, VS Code coupling). Most areas came back clean; this PR fixes the one real violation found plus three smaller hygiene items surfaced in the same pass.

## Issue acceptance gates

No tracking issue; this PR is the direct output of the audit. Gates addressed:
- Provider SDK usage types no longer live inside `src/agent/core/`, which `core/README.md` documents as a pure domain model.
- The two independent "no vscode import" guards (eslint + vitest ratchet) cover the same zone list again.
- The one undocumented `platform()` bypass among `fs` usages now carries the same kind of rationale comment its siblings already have.
- The duplicated `isDevAssertionMode()` helper has one definition instead of two.

## Single source of truth

- `src/agent/types/ProviderUsage.ts` is now the one place `ProviderUsage`/`NativeUsagePayload`/`ExtendedCompletionUsage` are defined (moved from `src/agent/core/usage/ResponseUsage.ts`), living beside `ModelHandlerContracts.ts` — the file that already documents itself as holding the provider-SDK-shaped data contracts between `ModelHandler`/`IModelHandler` and each provider's SDK.
- `src/shared/utils/devAssertionMode.ts` is now the one definition of `isDevAssertionMode()`, imported by both `SessionEventHub.ts` and `shared/utils/dispatcher.ts` instead of each carrying an identical copy.

- Removed: two copies of `isDevAssertionMode()` → one shared export. No new abstraction layer — it's a plain function move to the zone (`src/shared/utils/`) both existing call sites could already import from.
- Removed: raw provider-SDK usage types (5 SDKs' wire shapes) from the domain kernel's type surface. `core/` only ever consumes the already-normalized `NormalizedUsage`; the raw union now lives where the other provider-shaped contract types (tool-call SDK types in `ModelHandlerContracts.ts`) already live, rather than inventing a new module.
- No pass-through wrappers added.

## Net elements (R6)

- Files: +1 new (`src/shared/utils/devAssertionMode.ts`), 1 renamed (`core/usage/ResponseUsage.ts` → `types/ProviderUsage.ts`, content unchanged aside from an added doc comment), net LoC +33/-19 (mostly the two deleted `isDevAssertionMode` bodies vs. one new shared one, plus rationale comments).
- Exported symbols: +1 (`isDevAssertionMode`, previously module-private in two places, now one shared export). No new classes/interfaces/enums.

## Consumer counts (R8)

Not applicable — no emit path or existing export was deleted or re-routed; `ProviderUsage`/`NativeUsagePayload`/`ExtendedCompletionUsage` kept their exact names and shapes, only their import path changed (5 call sites updated: `ModelHandlerContracts.ts`, `openAIUsage.ts`, `modelHandlerOpenAI.ts`, `ResponseCycleFlow.ts`, `CommonCycleTypes.ts`).

## Host impact

Extension: none — internal core/shared refactor only, no host-facing behavior change.

Desktop: none, same reason.

CLI: none, same reason.

## Scheduled refactoring

None — this closes out the audit's findings; no further follow-up planned.

## Validation

- `npm run typecheck` — full workspace (core, test-kernel, agent SDK build, cli, trace-viewer, desktop) passes.
- `npx eslint` on all changed files — clean (two import-order errors caught and fixed by hand after a full-repo `--fix` OOM'd).
- `npx vitest run src/test-kernel/agent` — 203 files, 3114 tests passed (1 pre-existing skip).
- `npx vitest run src/test-kernel/architecture` — 17 files, 102 tests passed, including the updated `dependencyDirection.vitest.ts`.
- `npm run check:dead-code-ratchet` — no new unused files/exports/types (10 baselined, 10 current).
- `node scripts/check-browser-safe-utils.mjs` — unaffected (62 total, 7 reachable, unchanged).

---
_Generated by [Claude Code](https://claude.ai/code/session_01TpXkwSUtKD74rDRiq6UAZj)_


LionSR:
Addressed the review feedback in ec6f93b:

- **High (confirmed):** the "`core/` never sees an unnormalized usage shape" claim was wrong — `core/flows/CommonCycleTypes.ts` still holds `ExtractedModelResponse.usage: ProviderUsage` and normalizes it in-place by calling `ModelHandler.normalizeUsage`. Reworded both `ProviderUsage.ts`'s doc comment and `core/README.md` to say what's actually true: the raw shape crosses into `core/flows/` at the extraction boundary and is normalized there; only `core/usage/` (the value-object layer, i.e. `RunUsageAccumulator`) is raw-shape-free. Fully eliminating `ProviderUsage` from `core/flows/` too (pushing extraction+normalization entirely inside each `ModelHandler`) would be a materially larger change touching every provider handler — out of scope for this PR, and not what the original audit finding was about (that was specifically the type *definition* living under `core/usage/`).
- **Medium:** dropped the explicit `import process from 'node:process'` in `devAssertionMode.ts` — `dispatcher.ts` (one of its two consumers) sits in the webview frontends' Vite bundle closure, so an explicit Node-builtin import there was a real (if currently tree-shaken-away) risk. Now reads the ambient `process` global, matching `dispatcher.ts`'s pre-refactor code exactly. Verified with `VITE_WEBVIEW=settingsView vite build` — no `node:process` resolution issue.
- **Low:** reworded the `claudeAgentConfig.ts` comment — the real reason to bypass `platform()` is that `FileSystemProvider` has no access/existence-check primitive (only `stat`, which needs the same try/catch), not that the path is workspace-scoped (`platform().fs` isn't workspace-scoped, and it's `node:fs` underneath everywhere anyway).
- Also reattached the orphaned "Dev/test-only outbound-message assertions" doc comment in `dispatcher.ts` to `assertOutboundMessage`, which it had floated away from after `isDevAssertionMode` moved out.

Full workspace typecheck, targeted lint, and the affected test suites (SessionEventHub, dispatcher, architecture ratchet) all pass on the new commit.


---

Issue #10773, opened 2026-08-16
[transcript] Replace whole-file stream log rewrites with append-only JSONL; defer SQLite to a derived summary index


LionSR:
## Problem

`StreamLogStore`'s persistence primitive is "re-serialize the entire stream and atomically rewrite it" — `writeStream` calls `kv().write(streamId, logInstance.toPersistedEntries())` on every 300ms save window (`src/transcript/StreamLogStore.ts`). The semantics layered on top (writer authority, derived-tier discard+rebuild, delta/emission resync, loud degradation) are solid, but a large share of the store's machinery exists specifically to compensate for this primitive:

- **Write amplification** — a long transcript pays O(n) serialization+IO per window, so keeping streams dirty is expensive → residency/eviction (`requestEviction`, leases, `ResidentStreamRegistry`) → merge-on-hydrate ("disk history under live appends") → seqNo renumbering → the `reset: true` emission protocol so fold consumers rebuild.
- **Clobber risk** — because every save rewrites history, a save after a failed load could destroy the authoritative disk copy → `loadFailed` quarantine, append rejection, tombstones, write generations, the iterative `flush()` retry loop.
- **Crash window** — a crash loses up to one full save window of appends, not just a torn tail.

Two smaller weaknesses in the same area:

- **Summary staleness by mtime** — `readSummary` compares `summaryMtime < logMtime` to invalidate the cache; clock skew or a copy/restore defeats it.
- **Scan-all listing** — `readPersistentSummaries` lists and reads every summary file on open/reload. Fine at dozens of streams, degrades at thousands.

## Evidence from comparable systems

Surveyed openai/codex and sst/opencode (2026-08 snapshots):

- **Codex** persists raw session history as append-only JSONL (`codex-rs/rollout/src/recorder.rs` opens with `OpenOptions::append(true)`, writes one line per item, flushes pending lines). An append is O(1 line); a crash loses at most a torn last line; append-only cannot clobber history, so the residency/quarantine/merge machinery largely has no equivalent there. SQLite appears only as a **derived, rebuildable metadata mirror** (`codex-rs/state`: "extracts rollout metadata from JSONL rollouts and mirrors it into a local SQLite database"), with local storage still working when SQLite is unavailable.
[...]

PR #10778, opened 2026-08-16
refactor(lease): replace heartbeat liveness with proven instance presence

LionSR:
## Problem

Live runs were killed by their own gating: a run "suddenly disconnects", then every follow-up fails with `No active session for follow-up on stream <id>`. Root cause (confirmed in code and on an affected machine): execution-lease liveness was **inferred from wall-clock heartbeat age** (`Date.now() - heartbeatAt > 120s ⇒ owner dead`). System sleep freezes heartbeats while wall-clock advances, so on every wake all live owners looked dead — exactly when `RestartRepairRetryScheduler` timers queued during sleep fired and reclaimed their executions (mark FAILED + delete flow record + delete lease). A second path (`handleHeartbeatFailure`) self-declared loss on a single transient heartbeat error after any >2 min sleep. Two TUIs sharing a workspace made a reaper timer pending across every sleep.

Observed in the wild as:

```
! Error executing agent leanOrchestrator: Execution add97114dcbe is no longer owned by this TeXRA process.
! Execution add97114dcbe failed and its final artifacts could not be persisted
```

## Design: liveness is proven, never inferred

PRD: `docs/prds/2026-08-16-prd-execution-liveness-presence.md`

- Each process binds **one presence socket** (Unix domain socket / Windows named pipe) under `<globalStorage>/instances/`; every lease record it writes points at that instance (`{instanceId, socketPath, pid, hostname}`), written once, never renewed.
- **Death is a kernel fact**: `ECONNREFUSED`/`ENOENT` on the recorded socket, or a different TeXRA instance answering on a reused path. **Alive** = connected with matching banner. Everything else is **unprovable → treated as alive, loud** — we never reclaim on ambiguity.
- **Reclamation is event-driven**: restart repair skips a live foreign owner and holds a watch connection; the kernel closes it when the owner dies, which re-triggers repair. No clock anywhere.
- The `ownerToken` **write fence is unchanged** (that layer was correct).
- Consequence: a live owner can never be displaced automatically; `ExecutionLeaseActiveError` now carries the owner identity. A hung owner is reclaimed the moment its process exits.

## Retired completely (no compat arms)

Heartbeat timer/renewals, `EXECUTION_LEASE_STALE_MS`, `handleHeartbeatFailure` self-kill, `renewOwnedExecutionLease` (→ `validateOwnedExecutionLease`, a pure fencing check), `RestartRepairRetryScheduler` + `nextLeaseCheckAt`, wall-clock freshness, and the v1 record schema. Leftover v1 lease files are **tombstones**: any locked code path deletes one on contact and proceeds as if it were absent (loud warn), so upgrades self-heal — no manual cleanup. The unlocked classifier reports them as orphaned and leaves the delete to the next locked path. **Restart all TeXRA hosts after upgrading** (a still-running pre-upgrade host's live lease is a tombstone to the new code and will be reclaimed).

## Verified

- `npm run typecheck` clean (incl. `@texra-ai/agent` build), `npm run lint` clean, dead-code ratchet OK.
- Lease/repair/lifecycle/delegation suites rewritten against the new oracle with real sockets (alive / refused / banner-mismatch / owner-killed-mid-watch); all pass.
- Full vitest suite: green except unrelated files that also fail under machine contention (worker-start timeouts); those 18 files pass in a clean rerun.
- The lease-release leak observed on the affected machine (115 still-owned records spanning 3 days) is orthogonal and tracked separately; under this design it is inert load rather than a heartbeat stampede.

https://claude.ai/code/session_01RhUVea8MCbiKyDe9esF4kK

### DIFF
--- a/docs/prds/2026-08-16-prd-execution-liveness-presence.md
+++ b/docs/prds/2026-08-16-prd-execution-liveness-presence.md
@@ -87,19 +87,27 @@ What the old system conflated, and what happens to each job:
 - `probeInstance(owner)` → `'alive' | 'dead' | 'unprovable'`: connect with a
   short timeout, read the banner, verify the instance id, destroy.
   - banner matches → `alive`
-  - `ECONNREFUSED` / `ENOENT`, or a _valid_ banner naming a different instance
-    (path reused by a newer TeXRA process) → `dead`
+  - `ECONNREFUSED`, or a _valid_ banner naming a different instance (path
+    reused by a newer TeXRA process) → `dead`
+  - `ENOENT` (socket file unlinked — graceful exit, or a tmp cleaner racing a
+    live owner) → `dead` only together with a pid proof: `kill(pid, 0)`
+    throwing `ESRCH` is a kernel fact; success proves nothing (pid may be
+    recycled) → `unprovable`
   - timeout, other errors, non-TeXRA banner, or `owner.hostname` differing
-    from `os.hostname()` (shared home directory across machines) →
-    `unprovable` → treated as alive, logged at `warn`
+    (case-insensitively) from `os.hostname()` (shared home directory across
+    machines) → `unprovable` → treated as alive, logged at `warn`. The
+    hostname guard applies on every platform and equally to exit watches.
 - `watchInstanceExit(owner, listener)` → shared per-socket-path held
   connection; the kernel closes it when the owner process dies, firing the
   listeners. This replaces polling entirely: death is pushed, not sampled.
+  Inconclusive outcomes reconnect for a fresh verdict; a repeatedly
+  unprovable owner's watch gives up loudly, leaving a **bounded liveness
+  gap** (no automatic retrigger until the next natural repair pass) — a
+  deliberate fail-safe, never a reclaim hazard.
 - If the socket path would exceed the platform `sun_path` limit (~104 bytes on
-  macOS), fall back to a directory under `os.tmpdir()` with a `warn` log. In
-  that fallback `ENOENT` remains a death verdict; the residual risk (tmp
-  cleanup unlinking a live socket file) is bounded by the retained write
-  fence, and the fallback should be rare enough to be an incident, not a mode.
+  macOS), fall back to a directory under `os.tmpdir()` with a `warn` log; the
+  pid-proof rule above makes a cleaner-unlinked live socket `unprovable`
+  rather than dead.
 
 ### 3.2 Lease record v2
 
@@ -112,7 +120,7 @@ What the old system conflated, and what happens to each job:
   "owner": {
     "instanceId": "…",
     "socketPath": "…",
-    "pid": 12345, // diagnostics only — never a liveness input
+    "pid": 12345, // death-proof-only tiebreak (ESRCH) for unlinked sockets — never an aliveness input
     "hostname": "…", // cross-host guard → unprovable, never a death proof
   },
 }
--- a/docs/prds/2026-08-16-prd-execution-liveness-presence.md
+++ b/docs/prds/2026-08-16-prd-execution-liveness-presence.md
@@ -1,0 +1,214 @@
+# PRD: Proven execution liveness (presence sockets), full retirement of heartbeat leases
+
+**Date:** 2026-08-16
+**Status:** Approved for implementation
+**Owner:** Ray
+**Replaces:** the wall-clock heartbeat/staleness layer of `src/agent/storage/executionLease.ts`
+
+## 1. Problem
+
+Live runs are being killed by their own gating system. Observed repeatedly as:
+a run "suddenly disconnects", then every follow-up fails with
+`No active session for follow-up on stream <id>. Status: failed`.
+
+Root cause, confirmed in code and on an affected machine:
+
+- Execution ownership is a lease file renewed by a 15 s heartbeat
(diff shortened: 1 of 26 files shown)


LionSR:
Review findings addressed in 83e68d216c (and 52dc2e0c74 for the tombstone change):

**Fixed:**
- **Cross-host guard on the watch path + Windows** (claude finding at `instancePresence.ts:248`/`:198`; texra-ai #1): `watchInstanceExit` now applies the same hostname guard as `probeInstance` (never-fire disposer + warn), and the guard is platform-independent with case-insensitive comparison. The busy repair loop (watch says dead → repair says unprovable → re-watch) is gone.
- **Owner-exit repair fan-out** (texra-ai #2): `restartRepairWatches` is keyed by `owner.socketPath`, so one owner exit enqueues exactly one repair pass regardless of how many executions it held.
- **ENOENT as death proof / write-fence bound overstated** (texra-ai #3): `ENOENT` now proves death only together with `kill(pid, 0)` ⇒ `ESRCH` (a kernel fact; success stays unprovable since pids recycle). This is uniform — no tmpdir path heuristic — and also removes the permanent-wedge where a dead owner's unlinked socket could never be reclaimed. The overstated write-fence comment is gone. The local-ownership short-circuit in `runWithInactiveExecutionLease` now reports this process's own presence identity instead of the record's copy, so a tampered record naming a dead instance can't seed a self-firing watch loop.
- **FakePlatform tmpdir litter** (claude finding): the fake global-storage dir is now lazy and worker-shared — only suites that actually touch presence create one directory per worker.
- **Stale heartbeat-era fixtures/mocks** (texra-ai #4): `heartbeatAt` removed from `DeleteExecutionResult`/inspection mocks in the five untouched suites; the `EXECUTION_LEASE_STALE_MS` mock keys and their now-false module-load comments are deleted.
- **Coverage overclaim** (texra-ai #4): added real tests for banner-mismatch-on-reused-path (→ orphaned) and cross-host owner (→ unprovable, never reclaimed).
- **Residual liveness gap documented** (texra-ai closing note): the PRD now states explicitly that a repeatedly-unprovable owner's watch gives up loudly and the execution has no automatic retrigger until the next natural repair pass — a bounded fail-safe, never a reclaim hazard.
- **Platform-decoupling exception** (claude question at `:87`): documented in the module header as a deliberate exception — presence is made of kernel facts (real sockets, 0o700 dir, sync exit unlink) that a Platform port would only wrap for a single caller.

**Declined:**
- **Per-pass probe memoization** (claude finding at `restartRepair.ts:216`): a probe of a live local owner resolves in ~1ms; the 1.5s cost exists only on the unprovable timeout path, which is a rare corner rather than the steady state, and threading a per-pass cache through the lease API adds a parameter surface for a latency win that startup profiles haven't asked for. The related amplification (N watches per owner) is fixed by the socketPath keying above. Happy to revisit if a profile shows repair-pass latency dominated by probes.

LionSR:
Audit riders from the 2026-08-17 verified storage audit (full context: audit comment on #10773; freeze list: #10809). Two findings attach to this PR's presence model:

1. **The waiting-termination recovery arms collapse with THIS PR, not with the JSONL swap.** Adversarial verification established that `persistTerminalExecution` never throws (it marks the lease undurable internally), the transcript-group-close failure is absorbed inline at executionRegistry.ts:853-864, and the lease-release failure is warn-swallowed at :904-912 — so the outer three-arm catch (:798-843) plus the `settleTerminal` `untrackMode:'ifCurrent'` variant (~110 LOC) fire only on lease-infrastructure failures (`ExecutionLeaseLostError` et al.). Once presence-proven liveness replaces heartbeat staleness, that failure class shrinks and the apparatus can collapse — schedule it as this PR's follow-up, and treat the machinery as frozen (bugfix-only) meanwhile.
2. **Refuted — do not delete with the heartbeat retirement:** the `releaseRootExecutionLease` injected exit boundary (executionRegistry.ts:110-118,148-150,804-812,904-912) is NOT heartbeat apparatus. It drains two durable writer families (trace flushers AND snapshot artifact flushers) before ownership ends, and a successor's resume admission reads the authoritative transcript under the lease lock (`hasAuthoritativeStream`) — 'drain durable writers before ending ownership' is substrate- and liveness-model-independent single-writer exit discipline. This PR may reshape the lease internals it calls into, not the boundary itself.

LionSR:
Audit riders acknowledged:

1. **Recovery-arm collapse scheduled as this PR's follow-up.** Agreed with the analysis: with presence-proven liveness, `ExecutionLeaseLostError` can no longer be manufactured by clock skew — it requires actual displacement (foreign takeover of a provably-dead owner's record) or file-level tampering — so the three-arm catch + `settleTerminal` `untrackMode:'ifCurrent'` apparatus guards a failure class this PR shrinks structurally. Treating it as frozen (bugfix-only) until the follow-up lands; not filing the tracking issue from this session — will file on maintainer go-ahead.
2. **Exit-boundary discipline respected.** This PR reshapes only the lease internals the boundary calls into: `SessionHandle.releaseExecutionLease` keeps its exact choreography (validate → drain trace + snapshot flushers → validate → post-drain publish → complete), with the two validations changed from heartbeat write-renewals to pure fencing reads (`validateOwnedExecutionLease`). The injected `releaseRootExecutionLease` boundary in executionRegistry is untouched.

PR #10795, opened 2026-08-16
fix: settle transcript groups on completion

### LINKED ISSUE #10774: [transcript] Enforce "settled ⇒ immutable" in StreamLog; settle group rows at stage.end
## Problem

`transcriptFold.ts` already treats settlement as an immutability guarantee — `transcriptFold.ts:402` states it as fact ("An entry is 'settled' once its content can no longer change") and the settled-prefix promotion builds on it (`finalized: entry.settlementSeqNo !== undefined`, lines 271–389). But `StreamLog.updateWithSettlement` enforces nothing: any patch applies to any entry regardless of `settlementSeqNo`. A consumer is betting on an invariant the producer doesn't guarantee — the wrong-but-quiet contract gap our silent-degradation rule exists to close.

The de facto discipline in `TexraTranscriptRecorder` (the sole `TranscriptWriter` holder) is already settle-on-terminal everywhere except one path:

- streams: `append` → `appendText` → `settle` on `stream.end` ✔
- tools: `append` → `update` while in-progress → `settle` on terminal; boundary sweep settles orphans as FAILED ✔
- workflow calls: `settle` on terminal, typed-bridge cleanup owns the rest ✔
- model responses: `settle(correlatorId)` on `response.finalized` ✔
- **groups: `stage.end` terminalizes via `writer.update()` (`TexraTranscriptRecorder.ts:347`) — GROUP_END rows never get a `settlementSeqNo` on the live path.** Only the crash-recovery sweep (`endRunningGroupsForStreams`) settles groups.

The recorder's own boundary-sweep comment already names settlement as "one durable settlement coordinate" for CLI promotion; group rows just missed the memo.

## Proposed change

1. **`stage.end` settles instead of updates.** `settlementSeqNo` is schema-optional, so persisted format and trace-viewer replay tolerate the new field. Care point: `StreamLog.update` replaces `data` wholesale and the `stageMetadata` carry-over at `TexraTranscriptRecorder.ts:191-196` exists to preserve stage presentation on that rewrite — keep that behavior identical under `settle`.
2. **Audit late re-emissions.** A `workflow.call` event arriving after its terminal settle would today mutate the settled row (`:452-455` chooses `update` for non-terminal); a duplicate terminal `tool.end` re-settles. Under enforcement both become loud no-ops — verify no live path depends on post-settle mutation.
3. **Guard in `StreamLog.updateWithSettlement`: warn-and-drop, NOT throw.** A throw inside the recorder's subscriber hits `recordFailure` (`:184-190`), which disables the entire recorder — a single late duplicate event would silently kill transcription for the rest of the run. The guard logs at `warn` and returns `undefined` (the existing no-op contract).

~3 sites plus one regression test at the `StreamLog` boundary.

## Why it pays

The mutable set becomes the unsettled tail, which is small and bounded:

- Fold consumers may drop fold state for the settled prefix — the prerequisite for scrollback-commit rendering in the CLI (heap scales with the active tail, not history; the structural fix behind the long-session OOMs tracked in the 2026-08-11 transcript-memory PRD).
- #10773 (append-only JSONL) gets its clean encoding: settled entries are final lines, patches only ever target the tail, close-time compaction yields a canonical file.
- Delta consumers (`StreamLogDeltaBuffer` `dirtied` handling, `logSlice`) get a bounded invalidation window instead of "any of 50k entries may change."

Sequencing: independent of the Wave A–C consolidation campaign (#10669, #10672); shares only the `StreamLogDelta` seam, which this does not change. Should land before the wire contract hardens so settlement semantics are inside it from the start.

LionSR:
Closes #10774.

Makes settled transcript entries immutable: late `update` or duplicate `settle` calls warn and leave the row unchanged. Group starts remain mutable while running and `stage.end` settles the same row, establishing one durable terminal coordinate.

Also makes late `response.finalized` reconciliation a loud, contained no-op once the stream entry has already settled.

Validation:
- `corepack pnpm run typecheck:workspace`
- `corepack pnpm run typecheck:test-kernel`
- focused ESLint
- `corepack pnpm vitest run src/test-kernel/transcript/StreamLog.vitest.ts src/test-kernel/transcript/TexraTranscriptRecorder.vitest.ts --config vitest.config.mjs`

### DIFF
--- a/src/test-kernel/transcript/StreamLog.vitest.ts
+++ b/src/test-kernel/transcript/StreamLog.vitest.ts
@@ -96,6 +96,26 @@ describe('StreamLog', () => {
     expect(delta.textChunks).toEqual([]);
   });
 
+  it('warns and drops late updates to settled entries', () => {
+    const log = new StreamLog();
+    const settled = log.appendSettled({
+      id: 'complete',
+      type: STREAM_LOG_ENTRY_TYPES.LOG,
+      level: LOG_LEVELS.INFO,
+      timestamp: 1,
+      text: 'done',
+      messageType: MESSAGE_TYPES.DEFAULT,
+    });
+    log.drainEmission();
+
+    expect(log.update('complete', { text: 'late change' })).toBeUndefined();
+    expect(log.settle('complete', { text: 'late change' })).toBeUndefined();
+    expect(log.getRange(0).find((entry) => entry.id === 'complete')).toEqual(
+      settled,
+    );
+    expect(log.drainEmission().dirtied).toEqual([]);
+  });
+
   it('assigns one durable settlement order when rows become printable', () => {
     const log = new StreamLog();
     const header = log.appendSettled({
@@ -138,7 +158,7 @@ describe('StreamLog', () => {
 
     expect(header.settlementSeqNo).toBe(1);
     expect(completed?.settlementSeqNo).toBe(2);
-    expect(revised?.settlementSeqNo).toBe(2);
+    expect(revised).toBeUndefined();
     expect(later.settlementSeqNo).toBe(3);
   });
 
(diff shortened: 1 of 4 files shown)


LionSR:
Closing this implementation rather than adding settlement exceptions. The exact-head CI has six deterministic kernel failures: finalized responses remain raw, persisted settled/running groups cannot recover, suspended-group teardown regresses, and CLI transcript ordering changes. The current review requests identify the same root issue. This conflicts with the agreed removal of divergent StreamLog/History logic; #10774 remains the place for a later coherent Stream-tab transcript design.

PR #10808, opened 2026-08-17
feat(transcript): journal stream log mutations

### LINKED ISSUE #10773: [transcript] Replace whole-file stream log rewrites with append-only JSONL; defer SQLite to a derived summary index

LionSR:
Closes #10773

Replaces whole-stream rewrite persistence for new transcript mutations with an append-only JSONL journal, while retaining existing JSON arrays as immutable legacy bases at the read boundary. Journal records preserve append, update/settle, and streaming-text semantics with UUID-backed idempotent replay.

The public transcript writer, lease, and emission interfaces remain unchanged. Deletion, stream discovery, authoritative existence, and summary freshness account for both legacy and journal files.

Validated with:
- `corepack pnpm exec tsc --noEmit -p tsconfig.json`
- `corepack pnpm exec tsc --noEmit -p tsconfig.test-kernel.json`
- `corepack pnpm exec eslint src/common/storage/KVStore.ts src/transcript/StreamLogStore.ts src/test-kernel/transcript/StreamLogStoreLoad.vitest.ts`
- `corepack pnpm vitest run src/test-kernel/transcript/StreamLogStoreLoad.vitest.ts` (70 tests)

### DIFF
--- a/src/common/storage/KVStore.ts
+++ b/src/common/storage/KVStore.ts
@@ -14,12 +14,12 @@ import { isFile } from '@utils/files/fsEntryType';
 import { hasExtension } from '@utils/core/pathCore';
 import { StorageFS } from '@utils/files/storageFS';
 
-function keyToPath(dir: string, key: string): string {
-  return path.join(dir, `${encodeURIComponent(key)}.json`);
+function keyToPath(dir: string, key: string, extension = '.json'): string {
+  return path.join(dir, `${encodeURIComponent(key)}${extension}`);
 }
 
-function filenameToKey(filename: string): string {
-  return decodeURIComponent(filename.replace(/\.json$/, ''));
+function filenameToKey(filename: string, extension = '.json'): string {
+  return decodeURIComponent(filename.slice(0, -extension.length));
 }
 
 async function withMissingFallback<T>(
@@ -61,6 +61,24 @@ export class KVStore {
     return JSON.parse(raw) as T;
   }
 
+  /** Read a non-JSON storage value without imposing a serialization format. */
+  async readText(key: string, extension: string): Promise<string | undefined> {
+    return withMissingFallback(
+      () => StorageFS.read(keyToPath(this.dir, key, extension)),
+      undefined,
+    );
+  }
+
+  /** Append text to one storage value; callers own record framing. */
+  async appendText(
+    key: string,
+    extension: string,
+    content: string,
+  ): Promise<void> {
+    await StorageFS.ensureDir(this.dir);
+    await StorageFS.appendFile(keyToPath(this.dir, key, extension), content);
+  }
+
   async write<T = unknown>(key: string, value: T): Promise<void> {
     // Ensured on every write, not latched once: `dir` is storage-root
     // relative, so a cached store outlives the root it first saw (workspace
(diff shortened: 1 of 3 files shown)


LionSR:
Closing this draft: the JSON-plus-JSONL overlay introduced a second persistence path. We are simplifying the History deletion surface instead of extending that dual-system design.

PR #10813, opened 2026-08-17
fix(runtime): log normalized usage provider

LionSR:
## Summary

- restore the per-round normalized usage provider as the backend billing provider
- keep model configuration limited to model metadata, so OpenRouter-routed rounds are logged as OpenRouter

## Root cause

#10790 derived billing provider from the model configuration. That configuration identifies the underlying model family and can be `others`, while normalized round usage records the canonical provider that served the request.

## Validation

- `pnpm vitest run src/test-kernel/agent/runtime/UsageMonitorLastTotals.vitest.ts`
- `pnpm run typecheck:test-kernel`
- `pnpm run typecheck:workspace`
- `pnpm eslint src/agent/runtime/UsageMonitor.ts src/test-kernel/agent/runtime/UsageMonitorLastTotals.vitest.ts`

### DIFF
--- a/src/agent/runtime/UsageMonitor.ts
+++ b/src/agent/runtime/UsageMonitor.ts
@@ -9,7 +9,7 @@ import type {
   StreamTabId,
   UsageRoute,
 } from '@shared/schemas';
-import { AgentCategory, UsageProviderSchema } from '@shared/schemas';
+import { AgentCategory } from '@shared/schemas';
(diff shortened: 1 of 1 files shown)


LionSR:
Superseded by #10796: the normalized-provider billing fix is now folded into the larger model-boundary refactor and validated with its affected flow fixtures.

PR #10832, opened 2026-08-17
refactor: retire terminal output preview cap

### LINKED ISSUE #10831: Retire redundant per-tool output truncation now that the recorder bounds entries
## Context

Follow-up from #10798 (feat: spill oversized transcript output).

#10775 (resolved by #10798) explicitly deferred this: "No per-tool truncation logic is added or changed; existing per-tool caps become redundant and can be retired separately." The recorder is now the single choke point (2,000 lines / 50 KB per entry), so display-oriented per-tool truncation is duplicated policy.

## Details

Audit the per-tool caps named in #10775 and remove the ones the recorder bound supersedes — display-only truncation such as memory-preview self-capping. For each remaining cap (e.g. the grep `maxBuffer` ceiling, which guards subprocess capture memory before the recorder ever sees the result), either retire it if the recorder bound suffices, or document why it survives as process-level protection. End state: no per-tool display truncation remains; the recorder bound is the single truncation policy.

## References

- PR: #10798
- Resolved issue: #10775
- Umbrella: #10829

LionSR:
Closes #10831.

## What changed
- Removed the 4,000-character display-only truncation from `send_to_terminal`.
- The complete captured terminal result now reaches the transcript recorder, which is the shared 2,000-line / 50 KB bound and spill-artifact boundary.

## Retained limits
- `grep` retains its `maxBuffer` ceiling because it constrains subprocess capture before any tool result exists.
- Memory previews remain bounded because they are settings-file previews, not transcript tool results.

## Validation
- `corepack pnpm vitest run src/test-kernel/tools/setup/SendToTerminalTool.vitest.ts`
- `corepack pnpm run typecheck:workspace`
- `corepack pnpm run typecheck:test-kernel`
- `corepack pnpm run lint`
- `git diff --check`

### DIFF
--- a/src/test-kernel/tools/setup/SendToTerminalTool.vitest.ts
+++ b/src/test-kernel/tools/setup/SendToTerminalTool.vitest.ts
@@ -150,9 +150,9 @@ describe('SendToTerminalTool', () => {
     assert.equal(runs.length, 0);
   });
 
-  it('truncates from the head, not the tail — the success/error line is at the end', async () => {
-    // Output longer than the tool's preview cap. We sentinel the start
-    // and end so we can see which side survived truncation.
+  it('returns complete terminal output for recorder-owned bounding', async () => {
+    // The transcript recorder owns output bounds and spill artifacts. Keep both
+    // sentinels so the tool does not discard either half before that boundary.
     const head = 'BEGIN_MARKER\n' + 'x'.repeat(8_000);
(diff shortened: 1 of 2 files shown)


LionSR:
Superseded by the larger batched PR #10833. Its substantive commit (087b51a950) is an ancestor of #10833 head 0798242372; the remaining head change here is only a merge from main.

Issue #10842, opened 2026-08-17
[transcript] Retire the residency tier after the JSONL cutover


LionSR:
Gate 1 of the #10773 family-completion criteria (audit trail: the 2026-08-17 ledger comment on #10773; freeze inventory: #10809). Blocked on the JSONL cutover PR landing.

Delete from `StreamLogStore` once appends are eager per-stream journal records:
- eviction/rehydration apparatus: `requestEviction`/`tryRelease`, `hydrateStream`, presentation leases, `releaseRequests`, `loadFailed` quarantine (the verified store-core verdicts: appends no longer need whole-log residency; a failed read no longer blocks saves because saves cannot clobber)
- `writeGeneration` + `writeTombstones` residue that the journal design still carries (verified: per-stream append queues + the admission guard own those races)
- the `pendingSummaryMeta` antechamber — fix by making the run-fact projection path register the stream first (define-out-of-existence), then delete the waiting room
- reload/root-switch machinery shrinks to drain-appenders + rescan headers (keep the `stateRevision` check and `pendingReload` dedup)
- consumer guards that exist only for residency: released-stream writer/append throws re-derived to the smaller post-#10774 requirement (unsettled-tail residency only)

Estimated −500..−700 prod LOC in `StreamLogStore.ts` plus test retirement (suites pinned to eviction/rehydrate behavior). Acceptance: `StreamLogStore` net LOC below its pre-campaign main baseline (1,841), and no `ensureLoaded`-before-writer choreography remains at consumer call sites.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

LionSR:
Retiring this follow-up: its prerequisite JSONL cutover has been closed as divergent from the single Stream-tab authority direction. We will not preserve or extend a JSONL/residency migration chain.

PR #10845, opened 2026-08-17
feat: cut transcripts over to append-only JSONL with single-coordinate settlement

### LINKED ISSUE #10774: [transcript] Enforce "settled ⇒ immutable" in StreamLog; settle group rows at stage.end

LionSR:
## Summary

Re-lands the JSONL cutover (closed #10819) onto current main, redesigned per the directive that closed the previous stack ("redesigned from the single Stream-tab authority"):

- one append-only `.jsonl` journal per stream is the canonical transcript format: a `seed` record from one-shot legacy conversion, then per-mutation records folded on read; torn tails repaired before retry; malformed lines isolated loudly
- the store-wide 300 ms rewrite throttle and global write queue are replaced by immediate, independently serialized per-stream appends
- **`presentationSeqNo` is deleted — settlement is the single persisted ordering coordinate.** A mutable group heading's slot is reserved in the settlement counter when it opens (in-memory) and consumed by its terminal settlement, so cold replay keeps open-time order with no parallel field; the recorder reserves at `GROUP_START` append so headings order before their children
- settled ⇒ immutable enforcement, group settle at `stage.end` with the `GROUP_END` metadata fold, and the `logHead` wire-field removal carry over from the closed #10817, minus its parallel ordering state — closes #10774
- `delete()` reconciled with #10805's supersession semantics on main: a superseded durable delete re-seeds the canonical journal from the surviving log (`writeSeedJournal`) so a re-claimed workflow identity keeps its transcript across restarts, and re-persists its summary once the tombstone lifts

Refs #10773 (stays open — the family umbrella; summary retirement and the #10842/#10843/#10844 retirement gates follow). The temporary pre-#10774 group-start reader is dated 2026-08-17, retire after 2026-11-17.

## Crash-loss semantics change (ratified here, per the review note on #10819)

Streamed chunks are memory-only until settlement. This fixes redaction permanence — a secret split across streaming chunks can never persist in the raw, user-shareable file — and changes crash loss from "≤300 ms of streamed tail" to "the in-flight streamed entry." Settlement persists the materialized, redacted whole entry.

## Recovery-ordering note

A crash before a group settles drops its in-memory slot reservation; the recovery sweep settles such orphans with fresh slots in sweep order. Crash-only, and strictly better than the previous stack's behavior on every non-crash path while carrying one less persisted field.

## Validation

- `npm run typecheck` (workspace, clean)
- `npm test`: 875 files passed, 1 skipped; 10,724 tests passed, 5 skipped
- focused: transcript suites (279), `RunTraceDispose`
- changed-file ESLint clean; `git diff --check` clean
- 17 files, +1,678/−380 vs main (the substrate lands whole; net-negative arrives with #10842/#10843/#10844 per the #10773 completion gates)

### DIFF
--- a/packages/cli/src/chat/tui/panes/transcriptEntries.ts
+++ b/packages/cli/src/chat/tui/panes/transcriptEntries.ts
@@ -113,18 +113,26 @@ function isStaticTranscriptEntryAt(
   );
 }
 
-/** `(settlement order, synthetic-after-source tiebreak)` sort key for one entry. */
+/**
+ * `(printable order, synthetic-after-source tiebreak)` for one entry.
+ * Settlement order is the single printable coordinate: a group heading's
+ * slot is reserved when it opens and consumed by its terminal settlement,
+ * so headings sort at their open position without a second field.
+ */
 function transcriptOrderKey(
   entry: ConversationEntry,
   index: number,
-): readonly [seq: number, synthetic: number] {
+): readonly [seq: number, sourcePosition: number] {
   const seq =
     entry.settlementSeqNo ??
     entry.syntheticAfterSettlementSeqNo ??
     entry.sourceSeqNo ??
     index + 1;
-  const synthetic = entry.syntheticAfterSettlementSeqNo !== undefined ? 1 : 0;
-  return [seq, synthetic];
+  const sourcePosition =
+    entry.syntheticAfterSeq !== undefined
+      ? entry.syntheticAfterSeq + 0.5
+      : (entry.sourceSeqNo ?? index + 1);
+  return [seq, sourcePosition];
 }
 
 function compareTranscriptOrderKeys(
@@ -137,15 +145,10 @@ function compareTranscriptOrderKeys(
 /**
  * Printable rows in their append-only scrollback order.
  *
- * Source-backed rows use the durable order in which they became immutable.
- * Synthetic rows retain the settlement cursor captured when the CLI appended
- * them, with their original array position as the final stable tie-breaker.
- * Consumers that place rows relative to `<Static>` output must use this same
- * order rather than the stream's mutable storage order.
- *
- * Runs on every stream-sync tick, and entries arrive in settlement order on
- * all but the rare reorder — so it skips the O(n log n) sort whenever the
- * filtered slice is already ordered, at the cost of one O(n) pass over it.
+ * Source-backed rows use the durable order in which their rendered view became
+ * immutable. Synthetic rows retain the settlement cursor captured when the
+ * CLI appended them. Consumers that place rows relative to `<Static>` output
+ * must use this same order rather than mutable storage order.
  */
 export function orderedStaticTranscriptEntries(
   entries: readonly ConversationEntry[],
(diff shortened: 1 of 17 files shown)


LionSR:
Closing as divergent from the settled single Stream-tab authority direction. This reintroduces the retired JSONL cutover and settlement/compatibility stack, including one-shot legacy conversion and a temporary reader, instead of simplifying the current transcript surface. We are not maintaining a second persistence migration chain; any future transcript work must begin from a newly specified single-authority design after the duplicate-history proposal family is retired.

PR #10875, opened 2026-08-18
refactor: sweep local simplifications across agent, shared, and host code

LionSR:
Behavior-preserving cleanup collected from a local simplification sweep, rebased onto current `main`. Net **-306 lines across 182 files**.

## Recurring shapes

- Hoist repeated `createLog(channel)` / `getExecutionStore(id)` calls into a single const instead of re-deriving them per call site.
- Collapse flag-then-act control flow (set `closeFailedGroup`, act at the bottom) into direct early-return branches.
- Replace hand-rolled shapes with schema-derived types (`z.infer`, `.pick()`) and drop interfaces that only restated an existing base.
- Fold single-use locals and redundant object copies.

## Dropped during the rebase

Several hunks in the raw local sweep were not behavior-preserving and were left out rather than shipped:

- `SessionHandle` — `hasPendingWorkspaceStorageChange?.() === false` is not `!…`; the optional call returns `undefined` when the host does not implement it, and the falsy form skipped the reload entirely.
- `restartRepair` — a stream with no in-memory phase closes its group as FAILED but must *not* terminalize the execution; the rewrite conflated that with the `retryFailedStreams` path.
- `AgentLaunchResources` — `attach()` must run after ownership is recorded so a throwing attach still leaves the raw trace disposable.
- `streamApprovalQueue` — looping `Object.values(bypassByKind)` changed the observable emission order of bypass updates.
- `streamData` — the collapsed `rawKind` ternary trips the repo's `no-nested-ternary` rule.

Two files touched by the local sweep do not exist on `main` (`ChatExportController.ts`, `packages/cli/.../workflowPhase.ts`) and were excluded.

## Verified

- `npm run typecheck` — pass
- `npm run lint` — pass
- `npm run check:dead-code-ratchet` — 10 findings, same as baseline
- `npm test` — 876 files / 10718 tests pass


LionSR:
Thanks — all three findings confirmed against the base tree and fixed in bbafbb8cdb.

1. **Google output-token accounting** — you're right that `Math.max` is not the old ternary. `derivedOutput` subtracts input from `total_tokens`, which can carry cached input, so preferring the larger value double-counts billed usage. Restored `directOutput > 0 ? directOutput : derivedOutput` with a comment saying why `max` is wrong here.
2. **Suspended-run teardown** — restored the async IIFE rather than relaxing the comment; the documented synchronous-transition contract is the load-bearing part.
3. **Boolean normalization** — restored `=== true`, with a note that `get<boolean>` casts rather than coerces.

Also taking the residual-risk note: the equivalences that hold only because `shouldDelete?: () => boolean`, `figures: string[]`, and `StateStore.get<T>: T` are exact are now the kind of thing that would need revisiting if those types widen.

LionSR:
Second round confirmed and fixed in 8ff14ae1a0 — all three reproduce against the base tree.

1. **`classifyProviderMessageBlockType`** — correct, and the sharpest of the three. Replacing the `switch` with a bare index into a frozen object literal made inherited `Object.prototype` keys reachable, so a block typed `'toString'` or `'constructor'` returned a *function* into consumer switches that `assertNever` on anything unrecognized. Guarded with `Object.hasOwn`.
2. **`ServerSideKeyService`** — restored `?? true`. Same non-coercing-`get` hazard as `detachSubagentsOnStop` last round; a persisted `null` flipped included access off.
3. **`quotaFallbackRoutes.runtimeFor`** — restored the explicit three methods, so the spread no longer overwrites the route descriptor with the larger `CodingPlanSubscription` one.

Worth naming the pattern: across both rounds every finding is the same species — a collapse of `?? `, `Math.max`, `=== true`, or an exhaustive `switch` into a shorter form that is equivalent only for well-formed input. Six now, four of them on persisted or accounting data. Thanks for catching them.

Residual-risk notes acknowledged and not actioned: the `SupabaseSession` `options.log` spread, `terminalPersistence` empty-string `agentName`, and `streamStatusDisplay` `null` handling all sit on edge inputs no current caller produces.

LionSR:
Confirmed and fixed in 8eaa2bd9ac. The hoist is a real aliasing hazard, not just a convention point: `runtimeSkills.skills` is handed to the `skills.snapshot` logger consumer, so a shared array would carry across calls. Restored the per-call object with a comment saying why it stays per-call.

That's seven findings across three rounds, and all seven are one species — a collapse of `??`, `Math.max`, `=== true`, an exhaustive `switch`, or a per-call literal into a shorter form that holds only for well-formed input or a single caller. Four touched persisted or accounting data. Typecheck and 10721 tests were green through every one of them, so this review caught what the local gate structurally cannot.

Taking the visual-confirmation note as well: `settingsView/frontend/styles.ts` (~400 lines) and the new progress-view export toolbar are structural-only checks on my side and warrant a look in the running app before merge.

PR #10927, opened 2026-08-19
refactor(executions): share one run-file walk, gate resumable listings on lease liveness

LionSR:
## What

Two independent implementations of the same depth-2 walk over an execution's run storage directory — `ExecutionsTool.walkRunDirectory` / `listRunDirectoryFiles` and the CLI history's `listGeneratedFiles` — collapse into one home, `src/tools/executions/runGeneratedFiles.ts`. Separately, resumability **as displayed to a person** now consults execution-lease liveness, so a run that is executing right now stops being advertised as resumable.

## R6 — why here, not `@agent/storage`

The shared walk lives beside `executionKvFiles.ts` in `src/tools/executions/` because the predicate it is built around, `isKVFile`, imports `@agent/workflowScript` and `@tools/delegation`. Hoisting the walk into `@agent/storage` would drag the predicate with it and invert the layering (`src/tools` consumes `agent/core`, not the reverse) — and no lint rule or ratchet would have caught it, since `@tools/*` is not a fenced edge. The CLI already imported `@tools/executions/executionKvFiles`, so the host edge is unchanged in kind.

## R8 — the four divergences

The CLI copy was the stricter and richer side in every case, so it wins each one.

| Divergence | Tool copy (before) | CLI copy (before) | Shared module |
| --- | --- | --- | --- |
| Output order | unsorted (readDir order) | sorted by path | **sorted by path** |
| `stat` failure | bare `catch {}` — every failure silently skipped | skip ENOENT/ENOTDIR, rethrow the rest | **skip ENOENT/ENOTDIR, rethrow the rest** |
| `readDir` failure | warn-and-continue, empty listing | rethrow non-ENOENT | **warn-and-continue**, now stated as deliberate |
| KV-named **directory** | filtered by basename *after* the walk | skipped *before* recursion | **skipped before recursion** |

On `readDir` the tool's policy wins rather than the CLI's, and the module now says why in a comment: an unreadable *sub*directory must not blank out the rest of a run's listing, which is a different failure mode from a single entry vanishing between `readDir` and `stat`. Matches the `outputDiscovery` / `externalInquiryStorage` precedent (#10630).

## The silent `catch {}` is gone

```ts
    } catch {
      // Skip entries we can't stat
    }
```

That block turned an EACCES or an I/O fault into a file that simply is not there, with nothing logged — the canonical silent-degradation defect. It is deleted. The shared module skips only the two benign races (`ENOENT`, `ENOTDIR` — the entry disappeared, or its parent stopped being a directory, between `readDir` and `stat`) and propagates everything else.

## User-facing fix: files visible to the agent, invisible to the human

The tool filtered KV entries by basename **after** the walk. So a directory whose name matched the KV predicate — e.g. a generated `meta.json/` directory — was itself dropped from the listing, but the walk had already descended into it and its children were kept. The CLI skipped the directory before recursing and never saw them.

Net effect: a generated file nested under a KV-named directory appeared in the agent's `/executions/{id}/files` listing but never in `texra history`. The agent could cite a path the user could not find. Both surfaces now skip before recursion and agree.

## Lease liveness in resumability display

`inspectExecutionLease` had four production callers and not one of them was a list or status surface. Meanwhile `deriveResumability` reads durable state only — and durable state cannot tell "interrupted with a checkpoint" from "running right now": the flow record is written at flow start and removed only in `runToolUseFlow`'s finally block, while `meta.outcome` is stamped at finalization. A live run therefore has a flow record and no outcome, which is exactly the resumable shape. Every listing that rendered the durable decision advertised running work as resumable.

`deriveOfferableResumability` sits at the same durable-state boundary, wrapping the existing derivation: it asks liveness **once**, through `inspectExecutionLease`, with no second probe path. An `owned` or `foreign` lease yields the new `active-lease` cause; an unclassifiable lease fails closed and loudly (`unreadable-lease` + a `warn`) — without proof that nobody owns the run, offering it invites a double resume.

Display surfaces switched: `detectWaitingStreams` (extension/desktop WAITING marking) and `readCliResumeDataForListing` (both `texra history` entry points).

**Admission semantics are unchanged, deliberately.** `texra resume` already refuses an owned or foreign lease before loading resume state, and the in-process resume paths (`retrieveSessionResumeData`, restart repair, shutdown checkpoint detection, `SessionHandle`) run while the current host legitimately owns the lease — switching those to the offerable derivation would make a host unable to resume its own execution. They keep `deriveResumability`.

Two `ExecutionsTool` call sites also keep the durable derivation on purpose: at lines ~970 and ~1246 the decision is an *existence* probe ("does this execution exist at all"), not an offer, and a live run must not 404 there.

## Also

`mergeHistoryFiles` was the only thing left in `packages/cli/src/runtime/history/generatedFiles.ts` once the walk moved out — a single-caller module — so it is inlined into `history.ts` and the file is deleted. `SizedEntry` stays in `ExecutionsTool`: `/workspace-files` still renders through `formatSizedEntryLines`.

## Tests

No new test files. `src/test-kernel/cli/GeneratedHistoryFiles.vitest.ts` is retargeted and renamed to `src/test-kernel/tools/RunGeneratedFiles.vitest.ts` — it is the only place the ENOENT / ENOTDIR / EACCES stat policy is pinned, and that policy is now the shared one. One fixture is added to it (`meta.json/buried.tex`) so the existing exact-listing assertion also pins the KV-named-directory skip; verified to fail (both `meta.json` and `meta.json/buried.tex` appear) when the skip is removed. That is the one regression assertion this PR's bug fix earns under AGENTS.md testing discipline.

### Review points not taken

- **Unit-testing `deriveOfferableResumability`'s four branches.** Declined. It is a three-line fail-closed gate over an existing derivation plus `inspectExecutionLease`, whose own classification is already pinned in `ExecutionLease.vitest.ts`. Pinning the four branches means mocking the lease module at a seam that churns, for assertions that restate the `if`. Testing budget goes to the file-walk regression above.
- **Double durable read per resumable history row.** Real and accepted, not fixed here. `readCliResumeDataForListing` now derives resumability, then `retrieveSessionResumeData` derives it again internally. Threading the already-read `flowRecord` through would mean widening the resume-retrieval signature to carry pre-fetched state — a dual-ownership shape, for a KV read that listings already perform per row. Left for a follow-up that changes the retrieval boundary deliberately.
- **`active-lease` wording.** Taken: `inspectExecutionLease` returns `foreign` for any owner it cannot prove dead, so the cause comment no longer claims the run is demonstrably running — it now says the lease is held by a live *or* unprovable owner, both failing closed for the same reason.

## Verification

- `npm run typecheck` — clean
- `FORCE_COLOR=0 npx vitest run src/test-kernel/cli src/test-kernel/tools src/test-kernel/agent` — 6609 passed (one load-induced timeout in `ResumeCommand.vitest.ts`, green in isolation)
- `src/test-kernel/architecture` (subsystem-edge, host-agent deep-import, and mock ratchets) — pass
- `npm run format:check`, `eslint` on the changed files, `npm run check:dead-code-ratchet` — clean


LionSR:
Review pass addressed in `138d1a2a68`:

- **KV-named-directory regression (texra-ai)** — taken. Added a `meta.json/buried.tex` fixture to `RunGeneratedFiles.vitest.ts`; the existing exact-listing assertion now pins the skip. Verified it fails (both `meta.json` and `meta.json/buried.tex` appear) with the `isKVFile` skip removed. Note the example in the original description was wrong: `todos` is not a reserved KV key — the reserved set is `meta`, `config`, `report`, `workspace-files`, `result-meta`, plus `child-*`, `flow_*`, and the workflow/subagent checkpoint keys. Commit message and body corrected to `meta.json`.
- **`active-lease` overclaims liveness (texra-ai)** — taken. `inspectExecutionLease` returns `foreign` for any owner it cannot prove dead, so the cause doc now covers both a demonstrably live host and an unprovable owner, and says why both fail closed.
- **Unit tests for `deriveOfferableResumability` (claude)** — declined, per AGENTS.md testing discipline. It is a fail-closed gate over an existing derivation plus `inspectExecutionLease`, whose classification is already pinned in `ExecutionLease.vitest.ts`; pinning the four branches means mocking a churning module seam for assertions that restate the `if`. The one regression assertion this PR earns went to the file-walk defect.
- **Double durable read per resumable history row (claude)** — real, accepted, not fixed. Threading the already-read `flowRecord` into `retrieveSessionResumeData` would widen that signature to carry pre-fetched state — a dual-ownership shape — for a KV read listings already do per row. Left for a follow-up that moves the retrieval boundary deliberately.

Also stale-reference cleanup: `executionKvFiles.ts`'s header pointed at the now-deleted `packages/cli/src/runtime/history/generatedFiles.ts`.

PR #10948, opened 2026-08-19
feat(transcript): one shared row model so the CLI renders what the webview does

LionSR:
## Summary

The CLI TUI and the progress-view webview each built their own transcript rows
from the same stream log, and disagreed about what a row even contains. The
maintainer ruling for this track is that the terminal should render the same
state the extension and desktop have — one data source, per-host rendering,
with terminal economy expressed as row **density** rather than as dropped data.

`src/shared/transcript/` is now the single owner of that model:

- `transcriptRow.ts` — the typed row union (`assistant`, `log`, `user`,
  `error`, `tool`, `fileList`, `phase`, `workflowTask`, `compactionActivity`,
  `thinking`, `scratchpad`, `webSearch`, `webFetch`, `missingOutputs`,
  `latexdiff`, `statistics`, `contextManagement`, `progressStatus`,
  `contextState`).
- `projectTranscriptRow.ts` — the one projection from a `StreamLogEntry` to a
  row.
- `toolRowModel.ts` — the one fold from a normalized tool-use payload to
  header label, header preview, typed `ToolSection`s, output, error and the
  elision measurements a host needs.
- `transcriptText.ts` — the untruncated-text + measurement contract. Nothing
  in the model truncates; each host spends its own budget.

Both hosts were converted to paint from it. The Lit and Ink layers are the
only per-host code left.

## Issue acceptance gates

Track work, no issue number. The gates carried into this PR were the
maintainer ruling above plus the five open questions the two adopters flagged;
each is resolved below under **Adopter decisions**.

## Single source of truth

`src/shared/transcript/` owns the transcript row model. Specifically:

| Fact | Owner now |
| --- | --- |
| Which stream-log entries become rows, and of what kind | `projectTranscriptRow` |
| The error detail field set and its display order | `ERROR_DETAIL_FIELDS` in `projectTranscriptRow.ts`, typed `satisfies readonly (keyof ErrorLogData)[]` |
| Tool header label, preview, sections, output suppression | `toolRowModel` |
| Untruncated text + line/char metrics + head/tail elision | `transcriptText` |
| Whether a row is settled / promotes only on a typed terminal state | `isSettledRow`, `promotesOnlyOnTypedTerminalState` |

The barrel `src/shared/transcript/index.ts` is the module's public surface and
exports only what a host actually names: union members and internal shapes stay
unexported because a painter reaches them by narrowing `TranscriptRow` or
`ToolSection`.

Removed, not added:

- Two independent transcript projections collapsed into one. The webview
  parsed delegation/workflow-script/MCP inputs into typed sections while the
  CLI fell through to `JSON.stringify(input)`; the CLI showed tool output only
  for `bash` and MCP while the webview showed it for everything else. Both
  halves are now one ruleset.
- `toLogMessage` + `LogMessageDataSchema` + `LogMessageOf` **deleted** — the
  webview's parallel message shape was a second projection of the same entries.
- The CLI's `TRANSCRIPT_MESSAGE_TYPES` / `CHILD_STREAM_LOG_MESSAGE_TYPES`
  allowlists **deleted** — membership is the projector's, not the painter's.
- The CLI's hard-coded 240-char error truncation **deleted** — the row carries
  the full typed payload and the painter elides.
- `codexToolTemplates.ts` (281 lines of host-only Lit) **deleted** — see
  decision (a).
- `parseUtils.ts` / `stringifyWithLanguage` **deleted** — superseded by
  `stringifyPayload` in the shared model.
- `loadedImageDisplay.ts` **deleted** — folded into the `fileList` row.

The one new abstraction is the row model itself, and the invariant it owns is
exactly the host boundary: **the model carries data and measurements; a host
carries markup and width.** No section, glyph, or truncation decision crosses
that line in either direction.

## Net elements (R6)

- **Files:** 71 changed, +8 added, −3 deleted (`codexToolTemplates.ts`,
  `parseUtils.ts`, `loadedImageDisplay.ts`).
- **LoC:** +4,736 / −3,634 overall (net +1,102); production-only
  (excluding `src/test-kernel/**` and `packages/cli/scripts/**`):
  +3,938 / −3,063, net +875. Where it lands:

  | Area | +/− | net |
  | --- | --- | --- |
  | `src/shared/transcript/` (new module) | +2,149 / −0 | **+2,149** |
  | `packages/extension/src/progressView/frontend/` | +917 / −2,132 | **−1,215** |
  | `packages/cli/src/` | +781 / −798 | **−17** |
  | `src/shared/` (excl. transcript) | +59 / −121 | **−62** |
  | `src/test-kernel/` | +750 / −530 | +220 |

  The two hosts shed 1,232 lines between them; the new module costs 2,149. The
  excess is the typed vocabulary — 19 row kinds and 10 section kinds that
  previously existed only as untyped shapes reconstructed at each paint site.
- **Exported symbols:** 68 added / 29 removed over the diff. The added count is
  dominated by the row-union interfaces that now exist once instead of being
  re-derived per host; the shared barrel itself was shrunk to 6 value exports
  and 28 type exports after `check:dead-code-ratchet` showed the first cut
  exported more than either host consumed.
- **Interfaces/classes/enums:** +47 / −6. 40 of the 47 are `TranscriptRow`
  members and `ToolSection` members — the vocabulary, not new machinery.
- **Net direction:** this PR adds LoC. It is a consolidation of two divergent
  projections into one typed model plus a genuine feature increase on the CLI
  side (see Host impact), not a pure dedupe, so a net-negative count was never
  the target. Duplication removed is stated above.

## Consumer counts (R8)

Grepped over `packages/` + `src/` on `origin/main` vs this branch:

| Deleted / re-routed export | Files referencing on main | Now |
| --- | --- | --- |
| `toLogMessage` | 1 | 0 |
| `LogMessageDataSchema` | 4 | 0 |
| `LogMessageOf` | 10 | 0 |
| `TRANSCRIPT_MESSAGE_TYPES` | 1 | 0 |
| `CHILD_STREAM_LOG_MESSAGE_TYPES` | 1 | 0 |
| `codexToolRenderers` | 2 | 0 |
| `stringifyWithLanguage` | 5 | 1 (a historical comment in `NormalizeToolUse.vitest.ts`) |
| `loadedImageDisplayLines` | 2 | 0 |

No event-bus emit path was added, removed, or re-routed. `AgentEvent` /
`SessionFact` channels are untouched; this PR reads the stream log, it does not
emit.

## Adopter decisions

**(a) Codex tool vocabulary — added to the shared model (the preferred
option).** `toolRowModel` had no builder for `codex` / `codex_patch` /
`codex_thread` / `codex_todo` / `codex_turn`, so they fell to a generic YAML
section and the CLI would have lost the webview's patch file lists, todo
checklists and turn badges. Rather than keep the host override, the vocabulary
moved into the model: five section builders, plus two additions —
`ToolSectionFile.note` (the per-file change kind) and a new
`ToolChecklistSection` carrying `{ text, done }` items as data so each host
draws its own check mark. `codexToolTemplates.ts` is deleted; the CLI now
renders codex prompts, sandbox/follow-up badges, patch file lists with change
kinds, todo checklists with progress, and turn state/duration for the first
time.

**(b) Round headings in plain chat — fixed, and the rule is now stated once.**
The reported symptom was real, but "set the flag unconditionally" would have
been wrong: it broke `keeps lifecycle headings for full-log SDK children`. The
pre-track behaviour was actually a *composite* of the flag and the deleted
`TRANSCRIPT_MESSAGE_TYPES` allowlist, which is why deleting the allowlist
exposed it. The one rule, now written in one place:

```
projectLifecycleToTaskGroups = workflowStream || !fullLogChild
```

- plain chat → task-group surface → **no round headings** (regression fixed)
- workflow stream / workflow-script child → task-group surface (unchanged)
- full-log child that is not a workflow run (detached process, external-CLI
  session) → transcript rows, because it has no task-group renderer and its
  verbatim log is the point of opening it

The webview passes `projectLifecycleToTaskGroups: true` unconditionally, which
is what it already did implicitly — it returns before row projection for every
non-`LOG` entry. Phase headers are unaffected on both hosts.

**(c) Inline thinking in the CLI — kept, and it is already density-compact.**
It follows from the one-allowlist ruling and matches the webview. Body lines go
through the shared 6-head/3-tail elision budget with a
`… +N lines (Ctrl-T to view full output)` marker, so a long thinking block costs
at most 10 rows at any width while the untruncated text stays on the row and in
the Ctrl-T reader. That is terminal economy as density, not as dropped data.

**(d) `delegate_multi_agents` output — real bug, fixed.**
`buildWorkflowScriptSections` builds sections from the call's *input* (agent,
script, args, files) and carries no output, yet `dispatchSections` marked the
tool `carriesOutput: true`, suppressing the output block as
`rendered-by-sections`. The label was simply false, and it hid the script's
actual result. The special case is removed; `carriesOutput` is now only MCP
(which really does push a `Result:` section) and edit diffs. The webview gains
the workflow script's result; the CLI keeps the behaviour it had before this
track.

**(e) Carried-but-unpainted fields — both resolved, neither left dangling.**

- `ErrorRow.quotaHint` (and `ErrorRow.relayLabeled`) are **deleted**. Their
  copy source `INCLUDED_ACCESS`, the `isRelayError` field, and the CLI's
  `appendCliApiSwitchHint` were all removed by the relay-removal work that
  landed on `main` during this track, so the hint had no text to show and no
  consumer left. `isRelayError` is also out of `ERROR_DETAIL_FIELDS`.
- `WebFetchRow.content` is now **painted by the webview** as a `Content:`
  section (scroll-boxed past 20 lines), matching the CLI, which already
  elides it. Neither host showed the fetched text before; the model made the
  omission visible.

## Host impact

**Extension:** the progress view paints `TranscriptRow`s instead of
`LogMessage`s. It gains the wider error detail set (`operation`, `model`,
`missingApiKey`, `contextWindow`, `exhaustionReason`, `requestId`,
`rawMessage`, `rawErrorBody`, typed against `ErrorLogData` so a schema rename
is a compile error), the shared bash-header rule, the workflow-script output
block (d), the web-fetch `Content:` section (e), and codex cards rendered from
shared sections. It loses `codexToolTemplates.ts`, `parseUtils.ts`, the
relay error label and the `banner-details--relay-error` class.

**Desktop:** no renderer feature change, but see *Desktop renderer typecheck*
below — this PR closes an accidental module-graph leak into the renderer's
browser TypeScript project and fixes the 19 latent Node-global errors that leak
had been masking.

**CLI:** the terminal now renders, for the first time:

- full typed error details instead of a 240-char truncation of the summary
- attachment partial-load rows (`Files (2/3 loaded, 1 not found)`)
- user-feedback text on a rejected or redirected tool call
- delegation and workflow-script sections (agent, script, args, file groups)
- MCP tool output (structured and raw)
- seven message types it previously dropped entirely: `WEB_SEARCH`,
  `WEB_FETCH`, `STATISTICS`, `LATEXDIFF`, `MISSING_OUTPUTS`,
  `CONTEXT_MANAGEMENT`, `PROGRESS_STATUS`
- thinking and scratchpad content inline (previously they only fed the
  liveness indicator)
- codex prompt/mode/patch/thread/todo/turn cards

Security: `render/transcriptRowLines.ts` passes every body line through
`redactSecrets` after `safeTerminalText`. The newly shown `rawErrorBody` /
`rawMessage` fields would otherwise leak credential shapes that the old
truncation path hid only incidentally.

## Desktop renderer typecheck

Worth recording, because the symptom points the wrong way. Before this PR the
desktop **renderer** TS project (`tsconfig.renderer.json`, `types:
["vite/client"]`, no Node types by design) transitively pulled in
`@tools/WriteTool` through the progress-view formatters, and from there the
whole agent runtime, `src/tools/**`, `src/latex/**`, `src/auth/**` and finally
`@anthropic-ai/sdk` → `@types/node`. Its program held ~3,657 files. That
accidental leak silently supplied `process` and `NodeJS` to browser-context
code that had no business using them.

This PR removes that import edge — the row model carries a typed `ToolSection`
instead of reaching into a tool's input type — so the renderer program drops to
~1,423 files and `@types/node` is correctly absent. The 19 errors that appeared
were therefore **pre-existing latent violations, unmasked, not introduced**.
Fixed at the source rather than by adding Node types to the renderer:

- `src/shared/commands/accelerators.ts` now declares
  `export type DesktopPlatform = 'darwin' | 'win32' | 'linux'` and no longer
  reads `process.platform`; `NodeJS.Platform` is gone from
  `shortcutPreferences.ts`, `desktopCommandSurface.ts`, `rendererPlatform.ts`,
  `desktopCommandPalette.ts`, `desktopShortcutRegistry.ts`.
- `desktopMenuTemplate.ts` (main process, Node-typed) is the single boundary
  where Electron's wider `process.platform` is narrowed into that union.
- `src/shared/utils/devAssertionMode.ts` reads the ambient `process` off
  `globalThis`, keeping its documented "no `node:process` import" property
  while compiling in a project with no Node types and returning `false` in a
  browser instead of throwing.

## Scheduled refactoring

None.

## Validation

- `npm run typecheck` — **0 errors** across all six projects (was 19 in the
  desktop renderer before the leak diagnosis above).
- `FORCE_COLOR=0 npx vitest run` (full suite) — **854 files, 10,361 passed,
  5 skipped, 0 failed.**
- `packages/cli$ node scripts/validate-tui.mjs` — **22 of 164 scenarios fail,
  and the failing set is byte-identical to a baseline captured on `origin/main`
  in a separate clean worktree.** No new failure, no fixed failure, and no
  scenario expectation needed updating: the scenarios assert on content that
  survived the row-model conversion. Both sets are exactly:
  `backslash-goal-help`, `child-event-order-completion-remove`,
  `compact-model-form`, `config-category-back-responsive`,
  `escape-returns-focused-child-to-parent-list`,
  `escape-root-restores-prompt-history`, `escape-stops-focused-main-only`,
  `focused-stopped-subagent`, `focused-stopped-subagent-submit`,
  `focused-stopped-subagent-tab-focus`, `model-form`, `model-form-selectable`,
  `model-switch-compatible-only`, `no-color-model-form`,
  `orchestrate-no-runnable-models`, `process-child-composer-hidden`,
  `retry-approval`, `retry-approval-switch-api`,
  `slash-palette-ctrl-u-clears-raw-control`, `subagent-focused-submit`,
  `subagents-narrow-navigation`, `workflow-running-composer-hidden`.
- `npx eslint <68 changed files>` — clean.
- `npm run check:dead-code-ratchet` — **8 current vs 8 baselined, OK.** The
  first cut reported 47 findings; every one was fixed by shrinking the shared
  module's exported surface (un-exporting union members and internal shapes,
  deleting `isEmptyTranscriptText`). The knip baseline was **not** widened.
- `node scripts/check-browser-safe-utils.mjs` — OK, 62 total / 7 reachable.
  The shared transcript module stays browser-safe.

Test policy: obsolete blocks were deleted, not repaired. The one added fixture
module (`src/test-kernel/support/transcriptRowFixtures.ts`) replaces hand-built
entries in five existing suites and adds no cases. `TranscriptRowProjection.
vitest.ts` is the new module's own boundary suite. One case was added to the
existing `ToolRenderers.vitest.ts` for the codex section vocabulary (a), and
three existing expectations were updated to the decided behaviour: the round
heading under (b), the codex sections under (a), and the workflow-script output
block under (d).


LionSR:
further try to consolidate to eliminate the projector

PR #10974, opened 2026-08-19
Relay removal/attic and recovery doc

LionSR:
## Host impact

Extension:

Desktop:

CLI:


LionSR:
Closing rather than reviving this 190+ commit stale mixed-scope branch. It has five conflicts and two unresolved correctness risks: Grok can pass setup preflight without a selectable launch model, and deleting the legacy relay-limit schema arm can reject persisted canonical records. Any still-desired pieces should be proposed independently against current main.

PR #10991, opened 2026-08-19
fix(platform): degrade a broken project config on every host; one composition root for all three

LionSR:
## Summary

**Lead fix (C4, a live user-facing defect).** All three hosts open the same
project config store, but only the two GUI hosts guarded it. The VS Code
extension (`frontend/vscode/texraConfig.ts`) and the Electron main process
(`main/platform/index.ts`) each carried their own copy of the same walk-up —
same `constants.W_OK | X_OK` loop, same `isFileNotFoundError`, same
`parent === dir` termination, same `hasProjectConfig || canCreateOrWrite(path)`
rule, same internal-store fallback, near-identical warning text; only the log
sink differed. The CLI had **neither** copy: `initPlatform.ts` opened
`workspaceTexraConfigPath(cwd)` unguarded.

`JsonStore.open()` only reads, so a *missing* file was already an empty store.
But a **malformed or unreadable** `.texra/config.json` threw straight out of CLI
initialization — so every `texra` command hard-failed on the one host that has no
settings UI to fix the file with. An **unwritable** project directory failed later,
on the first command that persisted configuration. The GUI hosts degraded both
cases to the internal workspace store; the CLI degraded neither.

`openTexraConfigStores(storage, workspaceRoot, warn)` in
`src/platform/defaults/nodeStores.ts` now owns the rule once, with three call
sites. The degradation is loud, never silent: the host-supplied `warn` sink
carries the cause (`log.warn` in the extension, `console.warn('[desktop] …')` in
Electron, `[warn] [cli.config]` on stderr in the CLI). Folding the desktop copy
in also retired its hard-coded `'config.json'` literal in favour of
`TEXRA_CONFIG_FILE_NAME`.

Then, on the same plane:

- **C15** — desktop re-derived `join(storage.getStoragePath(), 'state.json')`
  that `createCliStateStores` already owned, and in production both hosts write
  the *same* `~/.texra/workspace-storage/<id>/state.json`. Now
  `openNodeWorkspaceStateStore(storage)`, two callers. The *global*-state path
  divergence is untouched — that is §3 V1, a ruling with a data migration.
- **C5** — `MemoryConfigProvider` restated `JsonConfigProvider`'s entire
  layered-resolution rule (workspace shadows global → `getCoreSettingDefault` →
  caller default) byte for byte; only the store accessor differed.
  `JsonConfigProvider` now reads through a three-method `ConfigStore` surface
  that both a `JsonStore` and a `Map` satisfy, and `MemoryConfigProvider` *is*
  that provider over in-memory stores.
- **C7** — the extension re-implemented two shared bootstrap helpers by
  admission. `copyDefaultAgents` duplicated `bootstrapNodeAgentDirectories`
  minus the re-entry guard, and the runtime-skills block carried a comment
  saying it was inlined so the extension bundle would not also pull the Lean
  direct-adapter import that `nodeHost.ts` carried. That blocker was a
  module-boundary artifact: `initNodeAgentRuntime` — the only Lean importer —
  moved to `nodeAgentRuntime.ts`, `nodeHost.ts` became Lean-free, and both
  inline copies died.
- **C8** — three near-identical `Platform` literals became one
  `createNodePlatform` call each.

**C6 (secrets) is refuted, not deferred** — see "Duplication and abstraction
budget".

## Issue acceptance gates

Rows C4/V10, C5, C7, C8, C15 of
`docs/proposals/2026-08-15-cross-host-consolidation.md` §2c. Re-verified at HEAD;
two §2c citations were stale and are corrected here:

- §2c cites `nodeHost.ts` at a path that has moved — the module is
  `src/platform/defaults/nodeHost.ts`, not `src/hosts/`.
- C5 lists `watch` among the duplicated members. Neither provider has a `watch`
  method; the real overlap is `get` / `inspect` / `isExplicitlySet`.

C6 is deliberately not landed (below). C14 is out of scope for this PR.

## Single source of truth

| Fact | Owner now |
| --- | --- |
| Which store backs the workspace config target, and when the project file is unusable | `openTexraWorkspaceConfigStore` (`src/platform/defaults/nodeStores.ts`) |
| Where global config lives | `openTexraConfigStores`, same module |
| Where workspace state lives | `openNodeWorkspaceStateStore`, same module |
| Layered config resolution (workspace → global → schema default → caller default) | `JsonConfigProvider` (`MemoryConfigProvider` is now an instance of it, not a peer) |
| The `Platform` skeleton a Node-family host assembles | `createNodePlatform` (`src/platform/defaults/nodeHost.ts`) |
| Bundled-agent reconciliation, incl. its failure policy | `bootstrapPlatformAgentDirectories` / `bootstrapNodeAgentDirectories` |
| Runtime skill-source precedence | `initializeNodeRuntimeSkills` |
| Direct Lean language-service registration | `initNodeAgentRuntime` (`src/platform/defaults/nodeAgentRuntime.ts`) |

Removed: two hand-rolled `canCreateOrWrite` walk-ups → one; two project-config
selection blocks → one (and the CLI, which had zero, now shares it); two
workspace-state path derivations → one; one drifted `'config.json'` literal;
`MemoryConfigProvider`'s three duplicated resolution methods; `copyDefaultAgents`;
the extension's inlined `setRuntimeSkillSources(defaultSkillSources(…))`; two
hand-assembled `Platform` literals.

New names, each meeting the §5b bar (a name is legitimate only when the same
change deletes ≥2 hand-rolled equivalents):

- `openTexraConfigStores` / `openTexraWorkspaceConfigStore` — deletes the
  extension's and desktop's copies of the selection walk-up.
- `openNodeWorkspaceStateStore` — deletes the CLI's and desktop's copies of the
  path derivation.
- `ConfigStore` — the seam that lets one provider replace two.
- `nodeAgentRuntime.ts` is a *file split*, not a new layer: the function moved
  verbatim so `nodeHost.ts` could stop dragging the Lean adapter into every
  importer. That deleted two inline copies in `extension.ts`.

`createNodePlatform`'s config input widened to *stores-or-provider*. Widening the
declared type of `configStores` would not have been enough: the factory
unconditionally constructed a `JsonConfigProvider`, while the SDK deliberately
supplies a `MemoryConfigProvider`, so the SDK's process-local semantics only
survive if the factory can take an already-constructed provider. `packages/agent`'s
exported signature is unchanged — `nodePlatform(options): Platform` — and its
build validates the identical 716 declarations.

**Rider (review-caught) honoured.** `copyDefaultAgents` *caught* reconciliation
errors so VS Code activation survived an unreadable agent directory, while the
shared `bootstrapPlatformAgentDirectories` *propagated*. The shared helper gained
catch-and-report (`log.error` with the cause) **before** the extension copy was
deleted, so a broken agent dir still cannot fail activation — and desktop and CLI
startup gain the same survival, which they did not have.

**C6 (secrets) refuted, not landed.** The review had already refuted its headline
bug (Electron's "missing PQueue" is not a race: `JsonStore.set()` synchronously
enqueues flushes into a module-wide per-path `PQueue` and the flush takes a
cross-process file lock; the CLI needs its outer queue only because it re-opens a
fresh store per mutation). What remains does not clear the net-≤0 bar this
§5b-FLAGGED extraction was conditioned on. Re-measured at HEAD: the genuinely
shared code between `CliSecrets` and `ElectronSecrets` is four trivial lines
(`get` = env ?? stored, `delete`, `listStoredKeys`, `getEnv`), while `getStored`
and `set` diverge completely (Electron carries a keychain-disabled shim, a
decrypt-failure latch, a `safeStorage` mode machine, and a warn-once dialog). A
`JsonStoreSecrets` base would have to abstract the store accessor as well, since
the CLI re-opens per operation and Electron holds one. `withEnvOverride(store)`
would wrap 1–3 lines per site across three real sites (the fourth,
`packages/agent`'s `environmentSecrets`, has no store at all) and would hide the
env-priority rule behind an indirection. Both come out net-positive. Not landed.

Platform ports were not touched — 14 ports / ~49 methods remain a never-collapse
fence row.

## Net elements (R6)

- **Files:** 17 changed, +2 added (`src/platform/defaults/nodeStores.ts`,
  `src/platform/defaults/nodeAgentRuntime.ts`), 0 deleted.
- **Exported symbols** (`^[+-]export` over the diff): +7 / −3. Two of the seven
  are relocations, not additions (`initNodeAgentRuntime` moved file;
  `MemoryConfigProvider` re-declared as a subclass). Net *new* exports: **+5**
  (`ConfigStore`, `openTexraConfigStores`, `openTexraWorkspaceConfigStore`,
  `openNodeWorkspaceStateStore`, and the moved `initNodeAgentRuntime`);
  net *deleted*: **−1** (`copyDefaultAgents`).
- **Classes/interfaces:** +1 interface (`ConfigStore`), +1 private class
  (`MemoryConfigStore`), −0 classes; `MemoryConfigProvider` went from a full
  `ConfigProvider` implementation (57 L) to a 3-line subclass.
- **Net LoC:** +416 / −380 = **+36** overall; **+5 in production code**, +31 in
  tests. The production figure is flat because ~110 of the added lines are the
  doc comments on the two new modules that replace the comments deleted from the
  four call sites; the win here is element count (nine duplicated
  implementations removed, one missing degradation path added), not line count.
- **Ratchet:** `config/ratchets/host-agent-import-baseline.json` **shrinks by
  one** — deleting the extension's `copyDefaultAgents` removed its last
  `@agent/index/AgentDirectorySync` import, and `hostAgentDeepImportRatchet`'s
  stale-headroom check required the baseline entry to go.

## Consumer counts (R8)

| Symbol | Consumers before → after |
| --- | --- |
| `copyDefaultAgents` | 1 (`extension.ts`) → 0, deleted; re-routed to `bootstrapNodeAgentDirectories` |
| `initNodeAgentRuntime` | 3 (`cli/runtime/initPlatform.ts`, `desktop/main/platform/index.ts`, `packages/agent/src/index.ts`) → same 3, new module path |
| `createNodePlatform` | 2 (CLI, desktop) → 4 (+ extension, + `packages/agent/src/node.ts`) |
| `initializeNodeRuntimeSkills` | 2 (CLI, desktop) → 3 (+ extension) |
| `bootstrapNodeAgentDirectories` | 2 (CLI, desktop) → 3 (+ extension) |
| `MemoryConfigProvider` | 1 (`packages/agent/src/node.ts`) → 1, unchanged construction |
| `openTexraConfigStores` | new → 3 (extension, desktop, CLI) |
| `openTexraWorkspaceConfigStore` | new → 2 (`openTexraConfigStores`, the extension's workspace transition) |
| `openNodeWorkspaceStateStore` | new → 2 (desktop, `createCliStateStores`) |

No emit path or event key was added, removed, or re-routed.

## Host impact

**Extension:** no user-visible change to config selection (it already had the
guard). Its `Platform` literal, its inlined skills wiring, and `copyDefaultAgents`
are gone in favour of the shared helpers; the bundle still does not pull the Lean
direct adapter, which is what the file split existed to preserve.
`bootstrapNodeAgentDirectories` supplies the re-entry guard the extension copy
lacked, and the shared reconciliation now catches failures the way the deleted
copy did, so activation survives a broken agent directory exactly as before.

**Desktop:** same config selection, now from the shared helper; the hard-coded
`'config.json'` for the global store becomes `TEXRA_CONFIG_FILE_NAME` (same
filename, no migration). Bundled-agent reconciliation failures now warn instead
of rejecting startup. Global-state path untouched.

**CLI:** the fix. A malformed or unreadable `.texra/config.json` no longer aborts
`texra`; it warns on stderr with the cause and degrades to the internal workspace
store, as the GUI hosts always did. An unwritable project directory likewise
degrades instead of failing the first command that persists a setting. No change
when the project config is present and well-formed.

## Scheduled refactoring

C6 is refuted above and should be struck from §2c rather than re-queued. C14
(`registerCoreShutdownHandlers`) remains open on this plane and is not addressed
here.

## Validation

- `FORCE_COLOR=0 npx vitest run src/test-kernel/architecture src/test-kernel/cli
  src/test-kernel/settings src/test-kernel/platform src/test-kernel/desktop
  src/test-kernel/frontend src/test-kernel/skills` — 279 files, 3469 passed,
  1 skipped, 0 failed.
- `NODE_OPTIONS=--max-old-space-size=8192 npx eslint <changed files>` — clean.
- `npm run check:tsconfig-paths` — in sync.
- `packages/agent` `npm run build` — 716 declarations and 69 external packages
  validated (identical to `origin/main`), confirming the frozen SDK surface did
  not move.
- One new test, the regression pin for the C4 defect
  (`WorkspaceStorage.vitest.ts`: a malformed project config degrades to the
  internal store and warns with the cause). No other tests added; the obsolete
  `jsonStore` / `jsonConfigProvider` mocks in `CliInitPlatform.vitest.ts` were
  deleted rather than adapted.
- Pre-existing, not caused here: `DesktopAgentExecution.vitest.ts` has 3 failures
  when run as a single file in isolation. Reproduced identically on a clean
  `origin/main` worktree; passes in the full-suite run above.

### DIFF
--- a/config/ratchets/host-agent-import-baseline.json
+++ b/config/ratchets/host-agent-import-baseline.json
@@ -32,7 +32,6 @@
       "@agent/followUp",
       "@agent/implementations/agentCreator/agentCreatorFlow",
       "@agent/index",
-      "@agent/index/AgentDirectorySync",
       "@agent/index/platformAgentDirectories",
       "@agent/review/reviewDiff",
       "@agent/review/reviewIssues",
--- a/packages/agent/src/index.ts
+++ b/packages/agent/src/index.ts
@@ -28,7 +28,7 @@ import type { AgentFlowResult } from '@agent/runtime/AgentFlowResult';
 import { AgentConfigSchema } from '@agent/core/definition/AgentConfig';
 import { createLog } from '@logger/logUtils';
 import { initPlatform, tryPlatform, type Platform } from '@platform/platform';
-import { initNodeAgentRuntime } from '@platform/defaults/nodeHost';
+import { initNodeAgentRuntime } from '@platform/defaults/nodeAgentRuntime';
 import type { ProgressPermissionKind as PendingInteractionKind } from '@shared/schemas';
 import { AgentCategory } from '@shared/schemas';
 import {
--- a/packages/agent/src/node.ts
+++ b/packages/agent/src/node.ts
@@ -5,18 +5,14 @@ import type { Platform } from '@platform/platform';
 import type { PlatformSecrets } from '@platform/secrets';
 
 // Local imports - platform defaults
-import { NO_TOOL_AVAILABILITY_HOST } from '@platform/interfaces';
-import { UNAVAILABLE_LANGUAGE_MODEL_PORT } from '@platform/languageModel';
 import { createLifecycleHost } from '@platform/defaults/lifecycleHost';
 import { MemoryConfigProvider } from '@platform/defaults/memoryConfigProvider';
 import { MemoryStateStore } from '@platform/defaults/memoryState';
-import { nodeFileLocks } from '@platform/defaults/fileLocks';
-import { nodeFilesystem } from '@platform/defaults/nodeFilesystem';
+import { createNodePlatform } from '@platform/defaults/nodeHost';
 import {
   createNodeStorageProvider,
   DEFAULT_NODE_STORAGE_ROOT,
 } from '@platform/defaults/nodeStorage';
-import { createNodeWorkspace } from '@platform/defaults/nodeWorkspace';
 
 /** Filesystem locations used by the default Node platform. */
 export interface NodePlatformOptions {
@@ -46,25 +42,18 @@ const environmentSecrets: PlatformSecrets = {
  */
 export function nodePlatform(options: NodePlatformOptions): Platform {
   const workspaceDir = options.workspaceDir ?? process.cwd();
-  const storageDir = options.storageDir ?? DEFAULT_NODE_STORAGE_ROOT;
-  const globalState = new MemoryStateStore();
-  const workspaceState = new MemoryStateStore();
-  const storage = createNodeStorageProvider({
-    storageRoot: storageDir,
-    workspacePath: workspaceDir,
-  });
-  const lifecycle = createLifecycleHost();
-
-  return {
+  return createNodePlatform({
+    // Process-local configuration: an embedder's settings must not be read
+    // from, or written to, the user's `.texra/config.json`.
     config: new MemoryConfigProvider(),
-    globalState,
-    workspaceState,
-    fs: nodeFilesystem,
-    workspace: createNodeWorkspace(() => workspaceDir),
-    storage,
-    fileLocks: nodeFileLocks,
+    globalState: new MemoryStateStore(),
+    workspaceState: new MemoryStateStore(),
+    storage: createNodeStorageProvider({
+      storageRoot: options.storageDir ?? DEFAULT_NODE_STORAGE_ROOT,
+      workspacePath: workspaceDir,
+    }),
     secrets: environmentSecrets,
-    lifecycle,
+    lifecycle: createLifecycleHost(),
     agentResume: {
       tryResumeStream: async () => false,
     },
@@ -73,7 +62,6 @@ export function nodePlatform(options: NodePlatformOptions): Platform {
       builtIn: async () => '',
       builtInToolUse: async () => '',
     },
-    languageModel: UNAVAILABLE_LANGUAGE_MODEL_PORT,
-    toolAvailability: NO_TOOL_AVAILABILITY_HOST,
-  };
+    getWorkspacePath: () => workspaceDir,
+  });
 }
--- a/packages/cli/src/runtime/cliStateStores.ts
+++ b/packages/cli/src/runtime/cliStateStores.ts
@@ -2,6 +2,7 @@ import * as path from 'node:path';
 
 import type { StateStore, StorageProvider } from '@platform/interfaces';
 import { JsonStore } from '@platform/defaults/jsonStore';
+import { openNodeWorkspaceStateStore } from '@platform/defaults/nodeStores';
 import { createNodeStorageProvider } from '@platform/defaults/nodeStorage';
 
 export interface CliStateStores {
@@ -36,7 +37,7 @@ export async function createCliStateStores(
   });
   const [globalStore, workspaceStore] = await Promise.all([
     openCliGlobalStateStore(storage),
-    JsonStore.open(path.join(storage.getStoragePath(), 'state.json')),
+    openNodeWorkspaceStateStore(storage),
   ]);
 
   return {
--- a/packages/cli/src/runtime/initPlatform.ts
+++ b/packages/cli/src/runtime/initPlatform.ts
@@ -1,6 +1,3 @@
-// Node imports
-import * as nodePath from 'node:path';
-
 // Local imports
 import {
   initializeBundledPrompts,
@@ -16,7 +13,6 @@ import { refreshModelListStateIfNeeded } from '@model/modelListRefresh';
 import { SHUTDOWN_PHASE } from '@platform/interfaces';
 import { initPlatform, platform, tryPlatform } from '@platform/platform';
 import type { LifecycleHost } from '@platform/interfaces';
-import { JsonStore } from '@platform/defaults/jsonStore';
 import { createLifecycleHost } from '@platform/defaults/lifecycleHost';
 import {
   bootstrapNodeAgentDirectories,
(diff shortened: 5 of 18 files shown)


LionSR:
**CI root cause (first push, `kernel tests` shard 2/2) — one cause, not ten.**

All ten failures came from a single file, `src/test-kernel/agent/AgentPackage.vitest.ts`, and none of them was a real behavior change.

That file stubs `initNodeAgentRuntime` with `vi.mock('@platform/defaults/nodeHost', …)`. C7's file split moved that function to `@platform/defaults/nodeAgentRuntime` (so `nodeHost.ts` stops dragging the Lean direct adapter into every importer — which is the whole point of the split), so the stub stopped intercepting it and the **real** `registerAgentFeatures()` ran once per test case. The second case onwards threw `Duplicate conditional tool injection: memory`, which cascaded through every remaining case in the file — including `treats bare and prefixed configuration keys as equivalent`, which lives in the same file and shares its `beforeEach`, not in a config suite.

Fix: re-point the mock's module specifier. No assertion was edited, no production code changed, and the config-key normalization path is untouched — `canonicalConfigKey` is called at exactly the same four sites it was before, now on `JsonConfigProvider` only instead of on two copies.

Ruling out the two hypotheses raised:
- **C5 did not change the resolution ladder.** `MemoryConfigProvider` now *is* `JsonConfigProvider` over in-memory stores, so workspace → global → `getCoreSettingDefault` → caller default is the one implementation both used to spell out identically.
- **C4's degradation does not affect `initPlatform` fixtures.** It changes only which `JsonStore` backs the workspace target when the project file is unusable; the platform is fully built either way, and it is reached only from the three composition roots, none of which this suite exercises.

Verified locally with the full suite (`FORCE_COLOR=0 npx vitest run`): **852 files passed, 1 skipped; 10359 tests passed, 5 skipped, 0 failed.** Rebased onto current `main` (which also needed a restored `toErrorMessage` import in the desktop composition root, where an incoming model-list `catch` met my import cleanup). `gh pr checks` is now fully green.

PR #10994, opened 2026-08-19
refactor(session): give the status machine the run-start stamp and the roster its own elapsed

LionSR:
## Summary

Two facts about a run's timing were computed by whichever host happened to need
them, so the hosts could not agree and one of them could not show the fact at
all. This PR gives each one a single owner in the substrate and deletes the host
copies.

**`runStartedAt`** was CLI-only. `streamSliceWithStatus` (`cliState.ts`) read a
clock on every status transition — `Date.now()`, or the stream log's last
timestamp when the applier's renderer callback supplied one — and folded the
result into `StreamSlice`. The webview had no run-start fact at all, and the
CLI's answer depended on when its own process observed the transition. It now
lives on `StreamPhaseState`, the session status machine's own record: stamped
when a stream enters an active phase, held across substate changes and
active→active transitions, cleared the moment the phase is no longer active. It
rides the canonical `status` fact and `BackendOwnedFieldsSchema`, so the
progress view gains a live elapsed segment in the stream header and the CLI
mirrors the value instead of deriving one.

**The child roster's `elapsed`** was the dual derivation audit C22 names.
`ExecutionRegistry` stamped a `formatDuration` string onto every
`ActiveChildInfo`; the webview rendered it verbatim while the CLI recomputed
from `startedAt` with a different formatter and a live tick. The roster now
carries timestamps only. `childElapsedMs` is the one derivation; format and
tick rate stay host modality. A retained row measures to its `finishedAt`,
which is exact — the stamped string froze at whatever the last active emission
read, which is why a finished row's figure was always slightly short.

### Refutations (verified at HEAD, not shipped)

- **D18, the CLI's dual exit teardown**
  (`docs/proposals/2026-08-16-define-out-of-existence.md` §1e) is **already
  landed** and needs no PR. #10800 replaced `exitNow` + the `exiting` re-entry
  guard with exactly the adjudicated shape: one cause-parameterized
  `beginTeardown(cause: ExitCause)` memoized through `teardownPromise ??=`
  (`sessionExitController.ts:312-385`). `exitNow` has no remaining occurrence in
  `packages/cli`. The load-bearing rider still holds — the signal arm does
  `ink.unmount()` + `cleanupTerminalModes()` + the resume hint synchronously
  before its first `await` (`:319-330`).
- **`plannedRounds` is not promotable to `SessionStreamConfigDetails`** without
  minting a second owner. The substrate already publishes planned rounds as
  `StreamStage.total`: `runReflectionFlow.ts:249` emits
  `total: computeRoundStageTotal(shared.totalRounds, roundIndex)` on
  `stage.start`, and `stage` is already on `BackendOwnedFieldsSchema`, so GUI
  hosts receive it. `runProgressRenderer`'s `config.plannedRounds` is a
  registry lookup (`getAgent(agent, Workflow)?.rounds`) filling the window
  before round 1 opens. Promoting it beside `StreamStage.total` is the same
  species as the `compactingActive` row the plan withdrew.
- **Run input files stop on the wire-vocabulary rule.** `SessionStreamConfigDetails`
  is written only by `SessionState.applySummaryMetadata` from
  `StreamSummaryMeta`, whose contract is "bounded scalars only" — an input-file
  list is neither, and it is persisted in every stream's always-resident
  `summary.json`. And landing it there closes nothing for the GUI on its own:
  `config` never reaches the webview, which sees the flattened
  `StreamTabInfoSchema` (`model`, `modelLabel`, `command`) — so the promotion
  additionally needs a new tab-info wire field. Per the stop rule, reported
  rather than invented.

## Issue acceptance gates

`docs/proposals/2026-08-15-single-substrate-hosts-as-renderers.md` §4, the
`runStartedAt` row: promoted to the substrate, host copy deleted, webview gains
the live elapsed source it did not have, and the `elapsed` dual derivation
(audit C22) collapses to one owner. The two rows the plan already withdrew
(`compactingActive`) or parked (`contextState`, `thinkingActive`) are untouched.

## Single source of truth

- **Run start**: `StreamPhaseState.runStartedAt`, owned by
  `StreamStatusMachine` (`src/agent/runtime/StreamStatusService.ts`) — the only
  writer of the phase it derives from. Published on `StatusEvent.runStartedAt`
  and `BackendOwnedFieldsSchema.runStartedAt`; every host reads, none stamps.
- **Child elapsed**: `childElapsedMs` (`src/shared/streams/childElapsed.ts`)
  over the roster row's own `startedAt`/`finishedAt`. No formatted duration on
  the roster wire.

Removed: a per-host clock read for run start (CLI) plus the absence of one
(webview); a formatted-duration wire field with two divergent readers; and the
CLI's second write of every status transition — the applier's
`onStreamStatusChanged` callback re-applied a transition the fact path had
already written one call earlier in the same handler, and existed only to
supply the timestamp the machine now stamps.

Added: one 27-line pure module, `childElapsedMs`, with three consumers in this
PR. No pass-through layer, no new port, no new event channel; `runStartedAt`
extends the existing `status` fact rather than opening a rail.

## Net elements (R6)

- **Files**: 20 changed, 1 added (`src/shared/streams/childElapsed.ts`), 0
  removed.
- **Exported symbols** (`^[+-]export` over the diff): +1 (`childElapsedMs`),
  −0.
- **Classes / interfaces / enums**: 0 added, 0 removed. `StreamEntry`'s
  `reserved` arm gains a `runStartedAt` field and the `RESERVED_STATE` module
  constant is deleted (its value is now per-entry).
- **Wire/state fields**: +3 the same fact (`StreamPhaseState.runStartedAt`,
  `StatusEvent.runStartedAt`, `BackendOwnedFieldsSchema.runStartedAt`), −1
  (`ActiveChildInfoSchema.elapsed`). One `nowMs` parameter deleted from
  `setStreamStatusInCliState`, five parameters deleted from the CLI's
  `onStreamStatusChanged`.
- **Net LoC**: +197 / −102 across the diff; production-only, excluding the new
  module and its doc comment, is roughly net-flat — the additions are the one
  stamp site plus the webview render that closes the gap.

## Consumer counts (R8)

- **`ActiveChildInfo.elapsed` (deleted)** — grepped before deletion: 1 producer
  (`executionRegistry.getActiveChildren`, `executionRegistry.ts:692`), 2
  production readers (`BackgroundTasksPanel.ts:478`,
  `postCompactionContext.ts:53`). CLI readers: **0** — `childElapsed`'s
  `child.elapsed` fallbacks were unreachable, because its single production
  caller (`SubagentList.tsx:146`) passes `{ status, startedAt: slice.runStartedAt }`
  and never an `elapsed`. Both readers now derive through `childElapsedMs`.
  `ExecutionStatusInfo.elapsed` (`ExecutionHandle.ts:27`) is a different
  surface — tool-result status text with one producer and no second
  derivation — and is untouched.
- **`setStreamStatusInCliState` (one call site deleted)** — production callers
  before: 2, both in `sessionSignalsAdapter.ts` (the raw `status` fact
  subscription at `:254`, then the applier's renderer callback at `:119`, which
  fired for the same fact one call later). After: 1. Test callers: unchanged;
  `runStartedAt` is optional, so a fixture that omits it renders no elapsed
  segment exactly as a fixture that omitted `nowMs` did.
- **`SessionRendererPort.onStreamStatusChanged`** — 3 implementers
  (`LitSessionRenderer`, `TuiSessionRenderer`, the CLI headless port). The port
  signature is unchanged: the Lit renderer still consumes `logHead` and
  `lastTimestamp` for `UPDATE_STREAM_STATUS`. Only the CLI implementation's
  body shrinks to its roster invalidation.
- **`childElapsedMs` (new export)** — 3 consumers in this PR
  (`postCompactionContext.ts`, `BackgroundTasksPanel.ts`, `childControls.ts`).

## Host impact

Extension: gains a live elapsed segment in the stream header, driven by
`runStartedAt` through the existing `<tool-timer>`; a background-task row's
duration is now derived from its own timestamps, and a retained row reads
`finishedAt − startedAt` instead of the last figure the backend happened to
stamp while it was still active.

Desktop: same progress-view frontend, same gains.

CLI: no visible change. It stops reading a clock for `runStartedAt` and stops
re-applying every status transition a second time; the value it renders is now
the backend's transition stamp rather than the stream log's last entry
timestamp.

## Scheduled refactoring

None. The two §4 rows this PR could not take (`plannedRounds`, run input files)
are refuted above rather than deferred.

## Validation

- `FORCE_COLOR=0 npx vitest run src/test-kernel/cli src/test-kernel/controllers
  src/test-kernel/architecture src/test-kernel/progressView
  src/test-kernel/shared` — the only failures are pre-existing on `origin/main`
  (`ResumeCommand`, `RunChatSignalOwnership` ×2, `TranscriptSessionPolicy`),
  confirmed by running the same files from a detached `origin/main` in this
  worktree; `AppEscapeRouting` / `RunExecution` failures appear only under full
  parallel load and pass in isolation on both revisions.
- `npm run format:check` — clean. `npx eslint` over the changed files — clean.
- `npm run check:dead-code-ratchet` — 8 findings vs 8 baselined, no new.
- `packages/cli` `node scripts/validate-tui.mjs` — baseline captured from a
  detached `origin/main` in this worktree (22/164 failing) and compared against
  the branch. Six scenarios differed; five re-ran green in isolation (the suite
  is load-flaky — two runs of identical branch code reported 23 and 28
  failures). The sixth, `subagents-with-todos-narrow-status`, was a real,
  deterministic delta and is fixed: the harness stamped the root run twice with
  different backdates, and the old CLI derivation happened to keep the first
  (`slice.runStartedAt ?? nowMs`) while the promoted value is taken verbatim.
  The harness now mirrors the status machine's own rule — stamp once, hold it
  across later active phases — so the frame is byte-identical again.

Obsolete test blocks were deleted rather than repaired: `childElapsed`'s
stamped-string assertion is gone with the branch it covered, and the
`elapsed:` roster fixtures across five suites and the TUI harness were dropped
with the field.

### DIFF
--- a/packages/cli/scripts/tui-harness.tsx
+++ b/packages/cli/scripts/tui-harness.tsx
@@ -1227,7 +1227,7 @@ if (HARNESS_INITIAL_STREAM_STATUS) {
     streamId: STREAM_ID,
     status: HARNESS_INITIAL_STREAM_STATUS,
     // Backdated so the status bar shows a plausible elapsed time.
-    ...(HARNESS_RUN_ACTIVE ? { nowMs: Date.now() - 42_000 } : {}),
+    ...(HARNESS_RUN_ACTIVE ? { runStartedAt: Date.now() - 42_000 } : {}),
   });
 }
 
@@ -1436,7 +1436,6 @@ function seedRunningWorkflow(): void {
       identity: { kind: 'agent', agent: 'correct' },
       agentName: 'correct',
       childStreamId: firstAgentStreamId,
-      elapsed: '18s',
       workflowPhase: 'Proofread',
     },
     {
@@ -1444,7 +1443,6 @@ function seedRunningWorkflow(): void {
       identity: { kind: 'agent', agent: 'correct' },
       agentName: 'correct',
       childStreamId: secondAgentStreamId,
-      elapsed: '17s',
       workflowPhase: 'Proofread',
     },
   ] as const satisfies readonly ActiveChildInfo[];
@@ -1496,8 +1494,8 @@ function seedRunningProcessChild(): void {
 // in `attachSessionSignalsAdapter` wiring must be able to fail these
 // scenarios, matching the "PTY ordering tests" section of
 // docs/proposals/2026-07-10-cli-child-stream-state-consolidation.md.
-// No `startedAt` is set on the roster row, so `childElapsed` returns the
-// static `elapsed` string instead of a live-ticking duration
+// No `startedAt` is set on the roster row, so `childElapsed` reports no
+// duration at all rather than a live-ticking one
 // (packages/cli/src/chat/tui/state/childControls.ts); `setActiveStream`
 // always sets `suppressViewSwitch: true` so the harness's own active/focused
 // stream (and its live wall-clock `runStartedAt` elapsed display) stays on
@@ -1525,7 +1523,6 @@ function childEventOrderRosterRow(): ActiveChildInfo {
     identity: { kind: 'agent', agent: CHILD_EVENT_ORDER_AGENT_NAME },
     agentName: CHILD_EVENT_ORDER_AGENT_NAME,
     childStreamId: CHILD_EVENT_ORDER_STREAM_ID,
-    elapsed: '1m 4s',
   };
 }
 
@@ -1805,7 +1802,7 @@ if (SHOW_CHILDREN) {
       agentName: 'leanSolver',
       childStreamId: 'harness-child-lean-stream',
       status: STREAM_PHASE.WAITING,
-      elapsed: '2m 3s',
+      startedAt: startedAt - 123_000,
     },
     {
       executionId: 'harness-child-review',
@@ -1821,7 +1818,6 @@ if (SHOW_CHILDREN) {
           ...child,
           status: STREAM_PHASE.FAILED,
           startedAt: undefined,
-          elapsed: null,
         }
       : child,
   );
@@ -1839,7 +1835,10 @@ if (SHOW_CHILDREN) {
   setStreamStatusInCliState({
     streamId: STREAM_ID,
     status: STREAM_PHASE.RUNNING,
-    nowMs: startedAt,
+    // The status machine stamps a run window once and holds it across every
+    // later active phase, so a scenario that already seeded an initial RUNNING
+    // keeps that backdated start rather than restarting the clock here.
+    runStartedAt: streams.get().get(STREAM_ID)?.runStartedAt ?? startedAt,
   });
   for (const child of childStreams) {
     const streamId = child.childStreamId;
@@ -1862,7 +1861,7 @@ if (SHOW_CHILDREN) {
     setStreamStatusInCliState({
       streamId: streamId,
       status: child.status,
-      nowMs:
+      runStartedAt:
         child.status === STREAM_PHASE.RUNNING ? child.startedAt : undefined,
     });
   }
@@ -1886,7 +1885,7 @@ if (SHOW_CHILDREN) {
     setStreamStatusInCliState({
       streamId: 'harness-nested-local-checker-stream',
       status: STREAM_PHASE.RUNNING,
-      nowMs: nestedStartedAt,
+      runStartedAt: nestedStartedAt,
     });
   }
 }
@@ -2098,7 +2097,7 @@ function markHarnessInterrupted(): void {
   setStreamStatusInCliState({
     streamId: STREAM_ID,
     status: STREAM_PHASE.CANCELLED,
-    nowMs: undefined,
+    runStartedAt: undefined,
   });
   for (const streamId of childStreamIds) {
     defaultSession().status.transition(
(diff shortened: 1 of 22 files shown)


LionSR:
Validator delta closed: a full `validate-tui` run on this branch and one from a detached `origin/main` in the same worktree now report the **same 22/164 failing scenarios, with an empty set difference in both directions**.

Also fixed after the first CI run: `StreamStatusService.vitest.ts` pins the published `status` fact and the reservation state by exact shape (including an explicit sorted key list), so the promoted `runStartedAt` had to be added to the seven RUNNING-phase expectations and to that key list. Those assertions are the reason the field's presence is now pinned rather than incidental.

LionSR:
Second CI fix, and it is worth recording because it is a genuine consequence of the promotion rather than a fixture typo.

`AppEscapeRouting.vitest.ts`'s `setRunning` helper seeds a RUNNING status straight into the CLI slice, bypassing the status machine. Under the old derivation that seed *implied* a run start (`nowMs` defaulted to `Date.now()` and `isActivePhase` folded it in), so the status bar always carried a live elapsed segment and repainted at 1 Hz — which is what the narrow-width layout assertions were actually measuring. With the stamp promoted, a seed that does not state `runStartedAt` gets none, the segment disappears, and the 1 Hz repaint stops. The helper now states the run window explicitly, mirroring what the machine does on a real transition.

Full local gate after the fix: `src/test-kernel/{cli,controllers,architecture,progressView}` — 257 files / 3551 tests, all passing.

PR #10996, opened 2026-08-19
fix(lifecycle): settle owned execution leases when a host exits

LionSR:
## Summary

Only the CLI settled executions when a host exited. Desktop and the extension registered `flushArtifacts()` plus disposal and nothing else, so quitting mid-run left the execution with no terminal outcome and its lease record on disk; the CLI TUI settled only an `isResumableIdle()` session.

**The defect's framing needed correcting at HEAD before fixing it.** The proposals describe the harm as "quitting mid-run blocks `texra resume` for up to the 120 s stale horizon". That symptom is gone: #10778 replaced heartbeat liveness with kernel-proven presence, `executionLease.ts` reads no clock at all (the `{version:1}` tombstone is the only heartbeat residue), and `acquireExecutionLease` reclaims a dead owner's record on contact — a quit host's socket is unlinked in its `process.once('exit')` (ENOENT + `kill(pid,0)`→ESRCH → `dead`) and a crashed one leaves the file (ECONNREFUSED → `dead`). `deriveOfferableResumability` likewise sees `orphaned` immediately, so listings offer the run right away.

What survives is the **durable record**, and it is worse than "merely unsettled":

1. `meta.outcome` is never written, so the execution reads as neither running nor finished in every cross-host surface (`ExecutionsTool`, `texra ls`, GUI history) until its *own* host relaunches — repair candidates come from each session's own transcript/snapshot stores, so no other host ever visits it.
2. When that host does relaunch, an execution with no resumable flow record goes down `repairRestartedStream`'s non-waiting branch: `FAILED` + `finalizeExecution({flowRecord:'delete'})`. A deliberate quit is recorded as a failure on the GUI hosts where the CLI records `CANCELLED`.
3. The lease file stays on disk, deleted only by a later acquisition of that same execution — for a run never resumed, never. This is the ongoing source of #10778's lease-release leak (the ~115-record backfill is explicitly *not* attempted here).

`settleLiveSessionExecutions` (`src/agent/runtime/SessionHandle.ts`) is registered as each host's **first ON-phase** shutdown handler — after every BEFORE handler, which is where run drivers settle what they own, and before any synchronous disposal. For each execution a live session still owns it reuses `SessionHandle.releaseExecutionLease` — the session's existing exit choreography — to persist `CANCELLED` with the flow record preserved and release the lease. It is bounded by that phase's abort signal (#10893's abort-then-advance deadline; no second deadline mechanism is introduced) and, once aborted, names in the log every execution it could not reach rather than proceeding silently.

## Issue acceptance gates

`docs/proposals/2026-08-15-lifecycle-ownership.md` §3 fix 2 / §6 step 4 (the doc's own "widest-impact fix"), and `docs/proposals/2026-08-15-cross-host-consolidation.md` §3 V5. Both status blocks are updated in this PR, V5 with the corrected harm statement above.

Gates satisfied:

- Awaited, session-owned drain rather than sync `dispose()` — required because `completeOwnedExecutionLease` is async under cross-process file locking. ✔
- All four hosts covered by one obligation; the CLI-UI copy deleted as sole implementation. ✔
- Bounded, with a loud report on timeout. ✔
- `executionLease.ts` untouched (§7.1 KEEP fence) — the drain composes existing verbs (`captureOwnedExecutionLeaseIfPresent`, `releaseExecutionLease`, `finalizeExecution`). ✔

Not attempted (stated per the brief): the ~115-record lease backfill and the non-destructive-repair item from #10778. This change stops new leaked records accruing; it does not clean up existing ones.

## Single source of truth

`SessionHandle.releaseExecutionLease` remains the one exit choreography for an execution — the drain calls it, it does not restate it. `settleLiveSessionExecutions` is the single owner of the *quit-time* obligation, sitting beside `killAllSessionBackgroundProcesses` / `forEachLiveSession` in the same file, i.e. the existing home for process-shutdown sweeps over live sessions.

Removed: `packages/cli/src/chat/tui/sessionExitController.ts`'s `persistAndReleaseResumableIdleLease` — the idle-gated `completeOwnedExecutionLease` copy — and the `resumableIdle` argument it forced through `persistBeforePlatformShutdown` and the signal branch of `beginTeardown`. Its test file `src/test-kernel/cli/SessionExitLease.vitest.ts` retires whole (−175 test LoC): its entire subject was the CLI-only mechanism.

New name: `settleLiveSessionExecutions`. It is not on the §5b ledger because the docs framed this as a *registration*, not a concept: the invariant it owns is "no execution is left owned when its host exits", it introduces no object or coordinator, and the verb it performs (`releaseExecutionLease`) already exists. Its paired deletion is the CLI-UI copy; declared honestly, that is one hand-rolled equivalent, not two.

Deliberately **not** promoted: the CLI headless settlement in `runExecution.ts`. Read critically, it is not the model — it must release the lease *before* it inspects it (`inspectExecutionLease` expecting `missing`/`orphaned`) in order to advertise recovery, so its settlement has to stay in BEFORE. With that ordering it settles first and the ON-phase drain finds no owned lease left, which is why the two coexist without a second write. Its CLI-facing `finalizeCliExecution` also reports through `session.interactions.emit('requestShowError')`, which core must not do.

## Net elements (R6)

- Files: 10 changed; 1 deleted (`src/test-kernel/cli/SessionExitLease.vitest.ts`), 0 added.
- Exported symbols (`^[+-]export` over the diff): **+1 / −0** → `settleLiveSessionExecutions`. Re-export lines changed inside existing blocks: +1 in `src/agent/runtime/index.ts` (the barrel the three hosts already import from, so no new `@agent/*` deep-import specifier and no `host-agent-import-baseline` widening), −1 in `src/agent/storage/index.ts` (`completeOwnedExecutionLease`, whose only consumer was the deleted CLI copy — the dead-code ratchet caught it).
- Net LoC: production **+96** (SessionHandle +86, three host registrations +8 each, barrels +1/−1, sessionExitController +10/−24); tests **−175**. Total **−57**.

A leak fix that closes a four-host gap is a net-add by construction; the offsetting deletions are the CLI copy and its suite.

## Consumer counts (R8)

No emit path or subscribe surface is added, moved or deleted.

- `completeOwnedExecutionLease` removed from the `@agent/storage` barrel: grepped consumers of the barrel specifier = **1** (`packages/cli/src/chat/tui/sessionExitController.ts:15`), deleted in this PR. The symbol itself keeps its 2 production consumers via the direct module path (`SessionHandle.ts` and, transitively, its own `completeOwnedExecutionLease` call inside `releaseExecutionLease`) plus its own suite; `src/agent/storage/executionLease.ts` is unchanged.
- `settleLiveSessionExecutions`: 3 consumers in the same PR (`packages/extension/src/extension.ts`, `packages/desktop/src/main/index.ts`, `packages/cli/src/runtime/initPlatform.ts`).
- Looks-orphaned-but-isn't guard: `getActiveIds()` on `ExecutionRegistry` gains a consumer rather than losing one. A parked WAITING execution stays tracked — `untrack` is reached only from `AgentRunLifecycle`'s terminal-result path and `runAgent`'s finally, neither of which runs for a flow suspended at its WAIT node — so the drain still covers the case the deleted CLI copy handled.

## Host impact

Extension: gains quit-time settlement it never had. `lifecycle.onShutdown(ON, …)` added before `clearStoreCache()`. Closing a window mid-run now records CANCELLED and releases the lease instead of leaving the next activation's `repairRestartedStreams` to write FAILED.

Desktop: same, added before the `agentResumeHandler`/`processResources` disposal. The `ElectronCompositionRoot` ordering assertion still holds (BEFORE handlers → flushArtifacts → an ON registration → `processResources.dispose()`).

CLI: headless unchanged in behaviour — its own handler still settles in BEFORE and the drain no-ops. The TUI loses its `isResumableIdle()` gate: a parked WAITING flow now also gets a durable `CANCELLED` with its flow record preserved (still `resumable: true` via `INTERRUPTED_WITH_FLOW`), and the flush-before-release ordering the deleted test pinned is preserved — `ctx.flushArtifacts()` runs before `runCliPlatformShutdownSequence`, which then drains.

## Scheduled refactoring

None from this PR. Pre-existing and untouched: #10778's lease-release backfill (~115 records) and its non-destructive-repair item.

## Validation

- `FORCE_COLOR=0 npx vitest run src/test-kernel/{agent/runtime,agent/storage,cli,desktop,platform,architecture}` — 276 files, 3886 passed, 1 skipped.
- `NODE_OPTIONS=--max-old-space-size=8192 npx eslint` over the six changed source files — clean.
- `npm run check:dead-code-ratchet` — OK (8 findings vs 8 baselined, after dropping the now-dead barrel re-export it flagged).


LionSR:
Both findings addressed / adjudicated.

**🔴 The clobber race — real, fixed (7242f9cd81).** The interleaving holds: `releaseExecutionLease` only reaches `completeOwnedExecutionLease` (which sets `releasing`) *after* `afterArtifactsDrained`, so a driver that already wrote its terminal outcome still passes `captureOwnedExecutionLeaseIfPresent`, and `enqueueMetaUpdate` serializes without comparing. A drain write landing second downgraded a COMPLETED run to CANCELLED.

Fixed at the durable boundary rather than at the caller: `finalizeExecution` gains `keepExistingOutcome`, evaluated **inside** the same locked read-modify-write cycle (`executionLifecycle.ts`), so "already settled" cannot go stale between the read and the write — a caller-side read-then-skip would only have narrowed the window. The drain sets it; every other caller is unchanged (`restartRepair`'s FAILED write over a resumable CANCELLED, and `clearTerminalExecutionState`, both still need to overwrite, which is why this is opt-in rather than a blanket refusal).

**Declining the suggested `isInFlightPhase` gate.** It would skip exactly the case this PR exists to close. `docs/proposals/2026-08-15-lifecycle-ownership.md` §3 fix 2 names the CLI's `isResumableIdle()` gate as *part of the defect* ("a four-host gap of varying width … CLI TUI settles only when `isResumableIdle()`"), and the headline harm is quitting the desktop app or VS Code **mid-run** — an in-flight stream by definition. Gating on phase would reduce the fix to "generalize the CLI's idle-only settlement", leaving a GUI quit mid-run to be recorded FAILED with its flow record deleted by the next launch's `repairRestartedStreams`.

For an exiting process CANCELLED *is* the truth for an in-flight run; the only thing that was wrong was overwriting a driver that had genuinely finished, which is what `keepExistingOutcome` fixes. Residual, stated plainly: if a driver survives the drain, its own `releaseExecutionLease` now finds the lease gone and reports a lease-lost artifact failure. That is a logged error during a shutdown that is milliseconds from `app.exit()` / `process.exit()`, not a durability loss.

**Test — added one (`ExecutionMetaWrites.vitest.ts`), not the suggested drain test.** A bug fix gets at most one regression test at its durable boundary (AGENTS.md testing bar), and the no-clobber property is the thing that can silently regress; the ON-phase registration ordering it depends on is already pinned by `ElectronCompositionRoot.vitest.ts`. A drain-level test would have to stand up a live session, a real lease and a presence socket to assert what the meta-write test asserts directly.

Re-ran: `typecheck`, `format:check`, eslint on the changed files, `check:dead-code-ratchet` (8 vs 8), and `vitest run src/test-kernel/agent/{storage,runtime}` (50 files, 647 passed).

LionSR:
Both findings from the `texra-ai` round were reviewed against `6110568c72`; here is where they stand.

**Finding 1 (Medium) — already fixed in `7242f9cd81`, pushed before this review posted.** Same race, same interleaving, independently found by the `claude` reviewer. `finalizeExecution` gained `keepExistingOutcome`, and the drain sets it. The suggested "read `readMetaStrict()` and skip" was deliberately *not* taken: a caller-side read-then-write is outside `metaWriteLocks`, so it narrows the window without closing it. The check runs inside `enqueueMetaUpdate`'s own locked read-modify-write cycle instead, which makes "already settled" un-staleable by construction. Opt-in rather than blanket, because `restartRepair`'s FAILED-over-resumable-CANCELLED write and `clearTerminalExecutionState` both still need to overwrite. Pinned by a case in `ExecutionMetaWrites.vitest.ts`.

The alternative you offered — "abort/await SDK runs before the ON phase" — is the larger change and is not this PR's: making the GUI hosts kill their runs at quit rewrites the §4 stop/quit policy table (`docs/proposals/2026-08-15-lifecycle-ownership.md`), where GUI quit is deliberately abandon-to-repair, and interrupting a parked WAITING flow clears the very flow record this drain preserves.

**Finding 2 (Low, docs) — accepted and fixed in `f2a9b9b08b`.** You are right that the premise was false for two of three hosts, and it was load-bearing for the ordering argument. The docstring now says what is actually true: the first-ON-phase position buys ordering against the drivers a BEFORE handler *does* reach (the CLI headless path, which kills and awaits) and against disposal; on desktop and the extension `registerAgentShutdownHandlers` reaches only background OS processes and the agent-CLI registries, so an in-process run is still being driven when the drain runs — which is why the drain exists at all, and why its terminal write yields to a driver that got there first.

Residual risk on the post-deadline laggard: agreed, and unchanged — that is #10893's abort-then-advance contract, and the drain adds a per-execution log line naming what it could not reach.

PR #10997, opened 2026-08-19
feat(desktop): wire the AppSignals the desktop was silently ignoring

LionSR:
## Summary

`docs/proposals/2026-08-15-cross-host-consolidation.md` §3 **V7** found that
`appSignals` has **zero references** in `packages/desktop` and `packages/cli`,
even though core emits on it from `src/tools/**` in every host's process. The
harm is not only the missed refreshes — it is that a host which silently never
reacts is indistinguishable from a host that deliberately doesn't.

This PR closes that by giving **every** declared signal exactly one recorded
answer at its declaration in `src/eventBus/AppSignals.ts`: which hosts consume
it, and why the others do not. Where the honest answer was "the desktop has a
surface and it is going stale", the subscription is wired; where it was "there
is no equivalent surface", it is fenced in the doc comment rather than wired as
a no-op. No new surface was built to justify a wiring.

### Per-signal decision table

| Signal | Decision | Stale surface fixed (WIRE) / rationale (FENCE) |
|---|---|---|
| `toolAvailabilityChanged` | **WIRE** desktop | Settings → **Tools dashboard**. Repainted only on an explicit Re-check click. It also never re-probed after a GitHub token write, so the `github_subscription` tool group kept rendering as unavailable (`desktopSettingsIpc.ts` set/removeGitHubToken vs `extension.ts:567` `secrets.onDidChange`). |
| `githubSubscriptionsChanged` | **WIRE** desktop | Settings → **Git tab subscription list**. Refresh-on-request only (`postGitHubSubscriptions` was reachable from startup, an explicit `getPRSubscriptions`, and a local unsubscribe), so a run that bound or released a PR/repo/issue subscription left the open panel wrong. |
| `githubTokenInvalid` | **WIRE** desktop | **No surface at all.** A token rejected by GitHub made `PollingSourceBase` fail in silence outside VS Code; the extension's `extension.ts:588` was the sole subscriber in the repo. Now an error dialog plus a Git-tab token-status re-post. |
| `workspaceFilesWritten` | **WIRE** desktop | **Files tree** (`renderer/editorPane.ts`). It caches its directory listing and there is no filesystem watcher, so files landed by `accept_run_files` stayed invisible until the user hit Refresh. `desktopProgressFileActions.ts` also explicitly dropped its `emitWritten` port with a comment saying nothing listened — that is no longer true, so it emits. |
| `languageModelsChanged` | **FENCE** | Extension-only *by construction*: the sole emitter is VS Code's `lm.onDidChangeModels` / `onDidChangeAccess`. Desktop and CLI have no editor-provided model catalogue that can change. |
| `approvalPolicyChanged` | **FENCE** | Extension-only. It exists because VS Code's status-bar tooltip repaints outside the settings webview's round-trip. The desktop's write path already re-posts the approval snapshot inline (`applyStateSettingUpdate` → `postStateSettingSnapshot`) and its only policy display is that snapshot-fed control; the CLI holds the policy in TUI state that `/approval` updates directly. A subscription on either would be a duplicate repaint. |

**CLI: fenced on all six**, and the fence text says why per signal. It has no
tools dashboard, no persistent subscription panel, no file tree, no
out-of-band notice surface, and its approval display is already live. Nothing
was wired there.

## Issue acceptance gates

Addresses audit finding **V7** in
`docs/proposals/2026-08-15-cross-host-consolidation.md` §3, which asks to
"wire desktop (and where sensible CLI TUI) subscriptions; where a host
deliberately won't react, fence it at the signal declaration." Both halves are
satisfied: 4 wirings, 2 fences, and 6/6 signals annotated with per-host
consumption.

Two claims in the audit are **refuted at HEAD** (see Validation).

## Single source of truth

- **`src/eventBus/AppSignals.ts` is now the single place that records per-host
  consumption for every signal.** Previously that fact existed only implicitly,
  as the presence or absence of a subscriber somewhere in a host tree.
- **`toolAvailabilityChanged` becomes the single repaint owner for the desktop
  Tools dashboard.** `DefaultDesktopToolingSettingsController.refreshToolDashboard`
  used to pair `refreshAvailability()` with its own `postToolDashboardData()`;
  that pass-through pair is deleted and the two call sites go straight to
  `dashboard.refreshAvailability()`. Every re-probe now repaints through one
  path regardless of which input changed, instead of only the paths that
  remembered to re-post.

No new abstraction, no new bus, no new subscribe surface — every subscriber
uses the existing `appSignals.on(...)`, and the two new emits
(`desktopProgressFileActions`) are on `workspaceFilesWritten`, which is squarely
inside `AppSignals`' documented scope (workspace-file writes) and therefore
inside the `appSignals.emit` exemption in CLAUDE.md. Nothing run-scoped or
session-scoped was added; no signal was added.

One wire message is added — `desktop:workspace:filesChanged`,
`DesktopWorkspaceFilesChangedMessageSchema` — because the desktop main process
cannot reach the renderer's tree cache any other way. It is keyless on purpose:
the tree re-lists from the main process anyway, and a path list would only tempt
the renderer into a partial update it cannot keep consistent with the
directories it has not loaded. It follows the existing
`createMessageRoutes` / `messageRoute(schema, handler)` pattern exactly.

Duplication removed: the `refreshAvailability` + `postToolDashboardData` pair
described above.

## Net elements (R6)

Measured against the merge base `3742d53`, not `origin/main` (which advanced
during this work):

- **Files**: 14 changed, 0 added, 0 deleted. `255 insertions(+), 23 deletions(-)`.
- **Exported symbols** (`^[+-]export` over the diff): **+1 / -0** →
  `DesktopWorkspaceFilesChangedMessageSchema`, consumed in the same PR by
  `renderer/messageRoutes.ts`.
- **Classes / interfaces / enums**: **+0 / -0**.
- **Functions / methods**: **+3 / -1**. Deleted: `refreshToolDashboard`
  (pass-through pair, both callers inlined). Added: three `dispose()` methods,
  one per owner of a window-scoped subscription — required by the review round,
  see Validation.
- **Net LoC**: **+232**, of which **96 lines are comment** (the per-signal fence
  annotations that are the deliverable, plus the window-lifetime rationale) and
  **136 are code**: 4 subscriptions, 3 disposers + their 3 registrations, 2
  re-probe calls, 1 emit, 1 schema, 1 route, 1 handler, 2 test pins, and 3 test
  teardown hooks.

**This is a deliberate net-add**, flagged as such in §7.4 revival row 6
("wiring is net-add, uncosted"). It is justified per surface, not per line:
each of the four wirings names a surface a user can watch go stale today, and
the two fences avoid the no-op wirings that would have made the number bigger
for nothing.

## Consumer counts (R8)

No export was deleted. One emit path was **re-routed** and one **restored**:

| Key | Subscribers before | Subscribers after |
|---|---|---|
| `toolAvailabilityChanged` | 1 (`SettingsViewMessageHandler.ts:217`) | 2 (+ `desktopToolingSettingsController.ts`) |
| `githubSubscriptionsChanged` | 1 (`SettingsViewMessageHandler.ts:210`) | 2 (+ `desktopSettingsIpc.ts`) |
| `githubTokenInvalid` | 1 (`extension.ts:588`) | 2 (+ `desktopSettingsIpc.ts`) |
| `workspaceFilesWritten` | 1 (`fileDecorations.ts:75`) | 2 (+ `desktopWorkspaceIpc.ts`) |
| `languageModelsChanged` | 1 (`SettingsViewMessageHandler.ts:224`) | 1 (unchanged, fenced) |
| `approvalPolicyChanged` | 1 (`extension.ts:718`) | 1 (unchanged, fenced) |

Re-routed emit: the desktop Tools dashboard repaint. It had exactly **1**
producer-side caller pair (`refreshToolDashboard`, called from
`recheckToolStatus` and `postStartupData`); both now call
`dashboard.refreshAvailability()` and receive the repaint through the signal.
Grepped: no other caller of `refreshToolDashboard` existed.

Restored emit: `emitWritten` on `acceptEditedFileReplace`. Ports grepped —
`compareCommands.ts:62` (extension, emits), `desktopProgressFileActions.ts`
(desktop, was `() => undefined`, now emits), and 2 test stubs. The CLI has no
caller of this path.

## Host impact

**Extension:** No behavior change. One dead `import { appSignals }` removed from
`packages/extension/src/webview/MainViewProvider.ts` — the symbol had zero uses
in that file.

**Desktop:** Four surfaces stop going stale — the Tools dashboard (now also
re-probed on a GitHub token write, which it never was), the Git tab
subscription list, the previously nonexistent invalid-token notice, and the
Files tree after a run writes into the workspace. The tree reload uses the
pane's existing `refresh()`, which re-lists from the root and drops expansion
state — the same reset the Files rail already performs each time it is opened,
so this stays consistent with the pane's behavior everywhere else rather than
introducing a second, subtler kind of refresh.

**CLI:** No code change. All six signals fenced with a stated reason. Its
`refreshModelListStateIfNeeded` call at `runtime/initPlatform.ts:290` is intact
(see below).

## Scheduled refactoring

Two follow-ups, both out of this PR's scope and neither introduced by it:

1. **Desktop open-buffer staleness.** `editorPane.open()` re-reads a file from
   disk only when the path is opened again, so a Monaco tab left open on a file
   that `accept_run_files` overwrote keeps showing the old contents. This PR
   refreshes the *tree*, not open buffers.
2. **CLI `/config` approval-policy desync** (unrelated to AppSignals, found
   while fencing `approvalPolicyChanged`): `CliConfigForm` writes
   `texra.approvalPolicy` through `writeSetting` directly rather than
   `applyStateSettingUpdate`, so it bypasses the `onApprovalPolicyChanged` port
   and leaves both the live `SessionHandle` policy and the status-bar segment on
   the old value. `/approval` is the only CLI path that updates them. Most of
   the display chain is in files owned by another track right now.

## Validation

- `FORCE_COLOR=0 npx vitest run src/test-kernel/desktop/ src/test-kernel/architecture/ src/test-kernel/tools/ src/test-kernel/latex/ src/test-kernel/frontend/` — **1687 passed**, including `src/test-kernel/architecture` per the kernel-invariant requirement.
- `NODE_OPTIONS=--max-old-space-size=8192 npx eslint <changed>` — clean.

**Tests: 2 added, both pins on genuinely-fixed stale surfaces**, per the
zero-default budget.
- `DesktopWorkspaceIpc.vitest.ts` — a write inside the workspace posts
  `FILES_CHANGED`, a write outside does not.
- `DesktopToolingSettingsController.vitest.ts` — *adapted, not added*: its
  `refreshAvailability` fakes now emit `toolAvailabilityChanged`, matching what
  the real `refreshToolAvailability` always does. Without that the fakes were
  pinning the pass-through pair this PR deletes.

### Refutations

Three claims checked and found **not to hold at HEAD** — recorded so the audit
can be corrected rather than re-actioned:

1. **"`refreshModelListStateIfNeeded` may now have no production caller on any
   host."** False. It has **two**: `packages/extension/src/extension.ts:392`
   and `packages/cli/src/runtime/initPlatform.ts:290`. It is not dead and was
   not deleted. The real gap is narrower and already tracked elsewhere (V2 /
   §7.4 row 5): **desktop** never calls it at startup. That is a bootstrap
   change in `packages/desktop/src/main/index.ts`, not an AppSignals change,
   so it is deliberately left to that track rather than folded in here.
2. **"Of 13 declared signals…"** The declaration list is down to **6**.
   `includedModelAccessChanged` and the six subscription signals V7 counted are
   already gone, as is `extensionDeactivating` (#10924). The
   tool-availability / token / file-write half of the finding was still fully
   open, which is what this PR closes.
3. **Pre-existing flake, not caused by this PR:**
   `src/test-kernel/desktop/DesktopAgentExecution.vitest.ts` ("presents a merge
   failure…") fails intermittently on clean `origin/main` too — reproduced
   there with 2 failures in one run, 3 in another, and 0 in the full-directory
   run on this branch. Not touched here.

### Review round (resolved in d84dca7)

Two findings, both real, both fixed — recorded here because the first
invalidates a claim the original commit made.

1. **The four new subscriptions were not process-lifetime.** `createWindow` is
   re-entered by `reopenMainWindow` from `app.on('activate')`, and
   `window-all-closed` deliberately does not quit on darwin — so every macOS
   dock reactivation stacked another listener holding the previous window's
   closure, one of which would have raised a dialog against a destroyed
   `BrowserWindow`. Each owner now exposes `dispose()` returning its
   unsubscriber, registered in the existing `windowResources` `DisposableStore`.
   `createDesktopWorkspaceIpc` keeps `dispose` separate from
   `disposeRendererResources`, which also runs on renderer reload where the
   subscription must survive. The pre-existing `subscribeGoalStateChanges`
   disposer in `desktopSettingsIpc` was being discarded the same way and is
   folded into the same path — so this fixes a leak that predates the PR.

   Consequently `setMaxListeners(0)` on the AppSignals emitter is **reverted**.
   It was masking this leak, not a test-only artifact. With proper disposal the
   warning does not reappear: the 1687-test run above is clean.

2. **The `postGitHubTokenStatus()` re-post after `githubTokenInvalid` was a
   no-op.** `resolveGitHubTokenSource` returns `'secret' | 'env' | 'none'` from
   token *presence*, and rejection leaves the stored secret in place, so the
   Git tab would repaint the same "token set" badge. Removed; the dialog is the
   whole fix, and the comment now says what marking a rejected token would
   actually require (a new status on the `UPDATE_GITHUB_TOKEN_STATUS` contract).

### DIFF
--- a/packages/desktop/src/main/desktopProgressFileActions.ts
+++ b/packages/desktop/src/main/desktopProgressFileActions.ts
@@ -3,6 +3,7 @@ import path from 'node:path';
 
 import { createLatexExecutionDiscovery } from '@agent/storage';
 import type { ValidatedExecutionRequest } from '@agent/core/state/executionRequests';
+import { appSignals } from '@eventBus/AppSignals';
 import { acceptEditedFileReplace } from '@latex/acceptedFileTarget';
 import { openFirstLabelMatch } from '@latex/labelSearch';
 import { LaTeXdiffService } from '@latex/latexdiff';
@@ -107,10 +108,10 @@ export class DesktopProgressFileActions {
         writeFile: (location, content) =>
           writeFile(location.absolutePath, content, 'utf8'),
         confirm: (message) => this.ui.confirmAcceptFile(message),
-        // The desktop main process has no `workspaceFilesWritten` subscriber
-        // (file decorations are a VS Code surface), so there is nothing to
-        // notify.
-        emitWritten: () => undefined,
+        emitWritten: (absolutePath) =>
+          appSignals.emit('workspaceFilesWritten', {
+            absolutePaths: [absolutePath],
+          }),
         showInfo: (message) => this.ui.showInfoMessage(message),
         deleteFile: async (location) => {
           try {
--- a/packages/desktop/src/main/desktopSettingsIpc.ts
+++ b/packages/desktop/src/main/desktopSettingsIpc.ts
@@ -5,6 +5,7 @@ import {
   listGitHubSubscriptionEntries,
   unsubscribeGitHubKey,
 } from '@controllers/settingsView/githubSubscriptions';
+import { appSignals } from '@eventBus/AppSignals';
 import { invalidateModelOptionsCache } from '@model/computeModelOptions';
 import type { ConfigProvider } from '@platform/interfaces';
 import { platform } from '@platform/platform';
@@ -26,6 +27,7 @@ import { buildSettingsSnapshotMessage } from '@shared/settingsView/handlers/sett
 import type { SettingsStores } from '@shared/config/settingsAccess';
 import type { SettingsStatePorts } from '@shared/settingsView/types';
 import { GoalStore, subscribeGoalStateChanges } from '@tools/goal';
+import { refreshToolAvailability } from '@tools/toolAvailability';
 import {
   GITHUB_TOKEN_CREATE_URL,
   GITHUB_TOKEN_STORAGE_KEY,
@@ -296,6 +298,14 @@ export function createDesktopSettingsIpc(
     });
   }
 
+  // Both writers below re-probe external tools: the GitHub token gates the
+  // `github_subscription` tool group, so without it the Tools tab keeps
+  // showing the group as unavailable until the user clicks Re-check. The
+  // extension gets this from `secrets.onDidChange`; the desktop has no
+  // secret-change event, but these two functions are the only places it
+  // writes the token, so the explicit calls cover the same ground.
+  // `refreshToolAvailability` emits `toolAvailabilityChanged`, which is what
+  // repaints the dashboard.
   async function setGitHubToken(): Promise<void> {
     const token = await options.ui.promptForSecret?.({
       title: 'GitHub token',
@@ -306,12 +316,14 @@ export function createDesktopSettingsIpc(
     await platform().secrets.set(GITHUB_TOKEN_STORAGE_KEY, token.trim());
     await options.ui.showInfoMessage('GitHub token saved.');
     await postGitHubTokenStatus();
+    await refreshToolAvailability();
   }
 
   async function removeGitHubToken(): Promise<void> {
     await platform().secrets.delete(GITHUB_TOKEN_STORAGE_KEY);
     await options.ui.showInfoMessage('GitHub token removed.');
     await postGitHubTokenStatus();
+    await refreshToolAvailability();
   }
 
   async function postGitHubSubscriptions(): Promise<void> {
@@ -322,6 +334,24 @@ export function createDesktopSettingsIpc(
       ),
     });
   }
+
+  // The same stance as the goal subscription above: a run that binds or
+  // releases a PR, repo, or issue subscription changes the list the Git tab is
+  // showing, and until now the desktop only re-read it when the user asked.
+  appSignals.on('githubSubscriptionsChanged', () =>
+    runAsync(postGitHubSubscriptions()),
+  );
+  // Outside VS Code a rejected token left the pollers failing in silence. Say
+  // so, and re-read the token status so the Git tab stops presenting the
+  // stored token as usable.
+  appSignals.on('githubTokenInvalid', ({ message }) =>
+    runAsync(
+      (async () => {
+        await options.ui.showErrorMessage(`GitHub token rejected: ${message}`);
+        await postGitHubTokenStatus();
+      })(),
+    ),
+  );
 
   /**
    * Jump from a settings entry (a goal, a PR subscription) to the run that owns
--- a/packages/desktop/src/main/desktopSettingsIpc.ts
+++ b/packages/desktop/src/main/desktopSettingsIpc.ts
@@ -97,6 +97,14 @@ export interface DesktopSettingsIpc extends DesktopMessageHandler {
     deferAgentCatalogRefresh?: boolean;
   }): Promise<void>;
   signInChatGpt(): Promise<void>;
+  /**
+   * Releases the goal and app-signal subscriptions. They are scoped to the
+   * window that built this IPC, not to the process: `createWindow` runs again
+   * on macOS dock reactivation, so an undisposed listener would post to a
+   * destroyed window's renderer — and, for `githubTokenInvalid`, raise a
+   * dialog against a `BrowserWindow` that no longer exists.
+   */
+  dispose(): void;
 }
 
 export function createDesktopSettingsIpc(
@@ -285,9 +293,11 @@ export function createDesktopSettingsIpc(
 
   // Agent runs execute in this same main process and the settings panel shares
   // the app window with run progress, so a Goals tab left open during a run
-  // needs the push. Lifetime == app, matching the extension's process-global
-  // stance, so there is no dispose to hold.
-  subscribeGoalStateChanges(options.session, () => runAsync(postGoalList()));
+  // needs the push. The session outlives the window, so the subscription is
+  // window-scoped and released in `dispose` below.
+  const subscriptions = [
+    subscribeGoalStateChanges(options.session, () => runAsync(postGoalList())),
+  ];
 
   // ── GitHub token + PR/repo/issue subscriptions (Git tab) ──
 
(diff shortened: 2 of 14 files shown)


LionSR:
Both review findings confirmed and fixed in d84dca7.

**Per-window listener leak (claude[bot], 4 comments).** Verified: `createWindow` is re-entered by `reopenMainWindow` from `app.on('activate')` (index.ts:1354-1356), and `window-all-closed` does not quit on darwin (index.ts:1367-1369). My "Lifetime == app" comments were wrong — copied from the pre-existing `subscribeGoalStateChanges` comment, which turns out to be wrong for the same reason and was discarding its own disposer.

Each owner now exposes `dispose()` returning its unsubscriber, registered in the existing `windowResources` `DisposableStore`:
- `DefaultDesktopToolingSettingsController.dispose()`
- `createDesktopSettingsIpc(...).dispose()` — covers both new subscriptions **and** the pre-existing goal subscription
- `createDesktopWorkspaceIpc(...).dispose()` — kept separate from `disposeRendererResources()`, which also runs on renderer reload where the subscription must survive

**`setMaxListeners(0)` reverted.** You were right that it was masking this rather than a test-only artifact. With proper disposal the warning does not reappear: the full desktop + architecture + tools + latex + frontend run (1687 tests) is clean with no `MaxListenersExceededWarning`. The two desktop suites now dispose their fixtures in `afterEach` instead.

**Token-status re-post is a no-op (texra-ai[bot]).** Confirmed — `resolveGitHubTokenSource` returns `'secret' | 'env' | 'none'` from token *presence* only (githubAuth.ts:64-71), and `PollingSourceBase.handleFailure` leaves the stored secret in place, so the badge would repaint identically. Removed the re-post; the dialog is the whole fix, and the comment now says so plus what marking a rejected token would actually require (a new status on the `UPDATE_GITHUB_TOKEN_STATUS` contract).

Gates re-run green: typecheck, format:check, eslint, dead-code ratchet, and the suites above.

PR #11015, opened 2026-08-19
fix(cli): apply /config approval-policy writes to the live session

LionSR:
## Summary

**Symptom:** in a `texra chat` session, changing **Approval policy** from the
in-chat `/config` panel appeared to be accepted, but the running session kept
enforcing the old policy — the status bar still showed the previous mode and
Bash/tool-edit approvals still used it. Only the next process picked the new
value up.

**Cause:** `/config` wrote every catalog row with a bare `writeSetting`,
bypassing `applyStateSettingUpdate` and its `onApprovalPolicyChanged` port that
the extension and desktop settings views go through. One setting, two write
paths, only one of which notifies. The bypass was **general** — every `/config`
row skipped the shared update path — but `texra.approvalPolicy` is the only row
that carries a live side effect today, so it was the only user-visible break.

**Fix:** the CLI `/config` write/reset now go through the same
`applyStateSettingUpdate`, with the chat TUI supplying the port that `/approval`
already drives (`setApprovalPolicy` on the session + `patchSessionMeta` for the
status bar). Non-`applied` outcomes throw, so `ConfigForm` rolls its optimistic
value back and surfaces the failure instead of a write silently reading as
applied.

`texra config edit` (standalone process, no live session) supplies no port and
persists only, as before — documented on the prop.

Verified at HEAD before fixing: with the new pin against the pre-fix files, the
store assertion passed while the live-policy hook never fired (`[]` vs
`['yolo']`).

### Sibling-surface audit

Four CLI ways to set the policy; they now agree on which one notifies:

| Surface | Persists | Updates live session |
| --- | --- | --- |
| `/config` row | yes | **yes (this PR)** |
| `/approval`, `/yolo` | no (session-scoped override) | yes |
| `--approval-policy` flag | no | yes (session seed) |
| `--no-input` fallback | no | yes (session seed) |

The `/approval` split (live-only, no persistence — same shape as the flag) is
deliberate and now documented in `approvalCommand.ts` rather than changed.
Audit item **V12**, the `--no-input` → `'never'` divergence, is likewise
**documented, not changed**: `TEXRA_APPROVAL_POLICY_NO_INPUT_DEFAULT` now
carries the headless-discipline rationale (a headless run denies rather than
hangs, and `buildCliContext` deliberately drops the persisted/env candidates so
a stored `yolo` cannot auto-approve an unwatched run).

## Issue acceptance gates

No issue — bug found during a fencing pass on a neighbouring track and confirmed
at HEAD before this change.

## Single source of truth

`applyStateSettingUpdate` (`src/shared/settingsView/handlers/stateSettingWrite.ts`)
is now the single write path for catalog-backed settings across all three hosts,
and therefore the single site that fires the approval-policy change port. The
CLI's live-session hook stays one closure (`onApprovalPolicySelect` in
`runChatTui`), shared by `/approval` and `/config`.

Removed a duplicate write path rather than adding a notifier: `/config`'s
hand-rolled `writeSetting` / `resetSetting` pair (with its duplicated
git-author + model-invalidation tail) collapses into one `applyUpdate` closure
(2 call sites) that delegates to the shared function. No new module, no new
type, no pass-through layer; the one new prop is a port the host fills with an
existing closure.

## Net elements (R6)

- Files: 6 changed, 0 added, 0 deleted (`+111 / −29`).
- Exported symbols: 0 added, 0 removed (`^[+-]export` over the diff = 0).
- Classes/interfaces/enums: 0 added, 0 removed (one optional field added to the
  existing `CliConfigFormProps`).
- Net LoC: +82 total; production +74, of which 46 lines are explanatory comments
  and doc blocks (`stateSettingWrite.ts` header, the `--no-input` rationale, the
  `/approval` scope note). Net non-comment production lines: ≈ +28, for the
  result-handling switch that replaces two throw-on-reject calls.
- Test: +28 (one regression pin, extended into the existing suite).

## Consumer counts (R8)

Re-routed one write path; no emit path or export deleted.

- `writeSetting` / `resetSetting` from `@shared/config/settingsAccess`:
  production callers before = 3 (`CliConfigForm`, `stateSettingWrite`,
  `platformSettings`); after = 2 — the CLI now reaches them through
  `stateSettingWrite`. No export removed (both still have callers).
- `applyStateSettingUpdate`: production callers 2 → 3 (extension settings view,
  desktop settings IPC, CLI `/config`).
- `onApprovalPolicyChanged` port implementations: 2 → 3 (vscode, desktop, cli).
- `setApprovalPolicy` call sites: unchanged — the CLI reuses the existing
  `runChatTui` closure, so `approvalPolicyAuthorityRatchet`'s
  `SEED_CALL_ALLOWLIST` needs no widening.

## Host impact

Extension: none (doc comments only).

Desktop: none (doc comments only).

CLI: `/config` approval-policy changes now take effect in the running session
(policy + status bar) as well as being persisted, and a `/config` write that the
shared path refuses now surfaces as an error with the optimistic value rolled
back instead of appearing to succeed. `texra config edit` unchanged.

## Scheduled refactoring

None.

## Validation

- `npm run typecheck` — clean (includes `packages/cli` `tsconfig.scripts.json`).
- `FORCE_COLOR=0 npx vitest run src/test-kernel/cli src/test-kernel/settings src/test-kernel/shared/stateSettings.vitest.ts src/test-kernel/architecture` — 179 files, 2688 passed, 1 skipped.
- Reproduction check: the new pin fails against the pre-fix files
  (`applied` = `[]`) and passes after.
- `NODE_OPTIONS=--max-old-space-size=8192 npx eslint <6 changed files>` — clean.
- `npm run check:dead-code-ratchet` — 8 current vs 8 baselined, OK.
- `npm run check:package-contributes`, `check:browser-safe-utils`,
  `check:guidance-refs` — all OK.

### DIFF
--- a/packages/cli/src/chat/tui/commands/handlers/approvalCommand.ts
+++ b/packages/cli/src/chat/tui/commands/handlers/approvalCommand.ts
@@ -10,6 +10,12 @@ import { type SlashCommandContext } from './slashContext';
 const APPROVAL_USAGE = 'Usage: /approval [ask | never | yolo]';
 export const YOLO_USAGE = 'Usage: /yolo [ask | never | yolo]';
 
+/**
+ * `/approval` is a session-scoped override, like `--approval-policy`: it moves
+ * the live policy for this session only and never writes `.texra/config.json`.
+ * The persisted default is `/config`'s row, which applies its new value to this
+ * same session through the shared write path's approval-policy port.
+ */
 export function applyCliApprovalPolicySelection(
   input: string,
   context: SlashCommandContext,
--- a/packages/cli/src/chat/tui/commands/registerBuiltins.tsx
+++ b/packages/cli/src/chat/tui/commands/registerBuiltins.tsx
@@ -703,6 +703,9 @@ export function registerBuiltinSlashCommands(options?: {
         <CliConfigForm
           stores={stores}
           availableRows={props.availableRows}
+          // Same hook `/approval` drives, so the approval-policy row updates the
+          // live session and the status bar from whichever surface set it.
+          onApprovalPolicyChanged={options?.onApprovalPolicySelect}
           openExternalForm={(formName) => openCliSlashCommandForm(formName, '')}
           onClose={() => props.onDone(undefined)}
           onError={async (error) => {
--- a/packages/cli/src/chat/tui/forms/CliConfigForm.tsx
+++ b/packages/cli/src/chat/tui/forms/CliConfigForm.tsx
@@ -13,14 +13,15 @@ import {
   saveProviderApiKey,
 } from '@cli/runtime/providerApiKey';
 import type { ApiKeyStatus, ApiProvider } from '@model/apiProviders';
-import { CLI_STATE_SETTINGS } from '@shared/schemas';
+import type { TexraApprovalPolicy } from '@shared/approvalPolicy';
+import { CLI_STATE_SETTINGS, type SurfacedSettingEntry } from '@shared/schemas';
 import {
   readSetting,
-  resetSetting,
-  writeSetting,
   type SettingsStores,
 } from '@shared/config/settingsAccess';
+import { applyStateSettingUpdate } from '@shared/settingsView/handlers/stateSettingWrite';
 import { GlobalStateKey } from '@shared/state/stateKeys';
+import { toErrorMessage } from '@utils/errors/errorMessage';
 
 import { refreshSubscriptionPreferenceViews } from '../state/subscriptionPreference';
 import { AgentRosterForm } from './AgentRosterForm';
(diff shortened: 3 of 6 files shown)


LionSR:
Re: the transcript-line suggestion — intended, and now documented at the wiring site (44696b9b4). Reusing `onApprovalPolicySelect` whole means `/config` emits the same `Approval mode: …` line `/approval` does; that line is the user-visible proof the change reached the running session rather than only `.texra/config.json`, which is exactly the failure this PR fixes. Splitting the port to suppress it would re-create two policy-change paths that differ.

LionSR:
Note: this merged at 0773eed88f, before the comment-only follow-up landed. That one-line documentation of the transcript-confirmation decision is now in #11026.

PR #11079, opened 2026-08-19
refactor(progress-view): absorb ProgressViewHost into the command handlers

LionSR:
## Summary

`src/controllers/progressView/ProgressViewHost.ts` was a container class: three
readonly fields, zero methods. Both hosts took one apart immediately —
`packages/desktop/src/main/desktopAgentExecution.ts:172-174` declared
`private agentProposalController!: ProgressViewHost['agentProposalController']`
and copied the fields out at `:405-406`; the extension reached through
`this.progressHost.<field>` at six sites.

Its only real content was a re-forward table: twelve one-line arms mapping
`ProgressViewCommandActions.file.*` onto `ProgressWorkflowFileActionsController`
methods of *identical signature*, plus `approval.handleAgentProposalAction` onto
`ProgressAgentProposalController.handleAction`. Giving
`createProgressViewCommandHandlers` the two controllers directly **deletes** the
table rather than relocating it — which is the point: forwarding the same
argument list to the same method name is what "pass-through" means.

This is the shape `docs/proposals/2026-08-03-ssot-consolidation-plan.md:349`
(C7, "Inbound forward chains") already prescribed for this file, including its
explicit carve-out: **`handleInteractionEvent` is not folded** (it holds a real
guard) and is untouched here.

## Issue acceptance gates

n/a — verified subsystem-elimination sweep, no issue. Implements the C7 row of
the SSOT consolidation plan.

## Single source of truth

`createProgressViewCommandHandlers` is now the only place a progress-view inbound
command is bound to its implementation. Previously the binding was split: the
factory named the commands, `ProgressViewHost` decided which object each one
actually reached.

`resumeStream` / `runNewStream` and the run-dependency types move into
`ProgressViewCommandHandlers.ts`, next to `resolveNativeAgentRun` — the gate they
both call, whose only other caller (`RESTORE_STATE`) was already in that file.
`resolveNativeAgentRun` therefore stops being exported.

Removed a container with no behavior plus its 12-arm forwarding table. Nothing
new was introduced: each host gains two `new Controller(...)` calls in place of
one `new ProgressViewHost(...)`, and the deps objects they pass are unchanged
character for character (only their indentation shifted).

## Net elements (R6)

Whitespace-insensitive diffstat (`git diff -w --numstat`), because both host
factories were split into three methods and re-indented, which inflates a plain
diffstat symmetrically:

| file | + | − |
|---|---|---|
| `src/controllers/progressView/ProgressViewHost.ts` | 0 | **173** |
| `src/controllers/progressView/ProgressViewCommandHandlers.ts` | 112 | 41 |
| `packages/extension/.../ProgressViewMessageHandler.ts` | 41 | 31 |
| `packages/desktop/.../desktopAgentExecution.ts` | 36 | 25 |
| `src/test-kernel/.../ProgressViewCommandHandlers.vitest.ts` | 24 | 29 |
| `src/test-kernel/.../ProgressViewRunCommands.vitest.ts` (renamed) | 21 | 21 |
| **total** | **234** | **320** |

| | Δ |
|---|---|
| Production files | **−1** (`ProgressViewHost.ts`) |
| Classes | **−1** (`ProgressViewHost`) |
| Exports | **−10**: `ProgressViewHost`, `ProgressViewHostOptions`, `resolveNativeAgentRun` un-exported, and 7 interfaces that were exported only so `ProgressViewHost` could compose them (`ProgressViewLifecycleCommandActions`, `ProgressViewFileCommandActions`, `ProgressViewFollowUpCommandActions`, `ProgressViewBypassCommandOptions`, `ProgressViewApprovalCommandActions`, `ProgressViewExternalInquiryCommandActions`, `ProgressViewRunDependencies`) |
| Interfaces deleted outright | **−3** (`ProgressViewHostCommandOptions`, `ProgressViewRunCommandActions`, `ProgressViewHostOptions`) |
| Type aliases | **−2** (`ProgressViewFileHostActions`, `ProgressViewApprovalHostActions`) |
| Interface members | **−8** (`ProgressViewFileCommandActions` 9 → 2; `handleAgentProposalAction` off `ProgressViewApprovalCommandActions`) |
| Forwarding arms | **−12** |
| Test files | **0** (one renamed) |
| Tests | **−1** (a case duplicated by the retargeted run-command suite); **0 added** |
| **Net LoC** | **−86** (prod **−81**, test **−5**) |

The verifier of this candidate estimated ≈ −40 prod. The extra came from
un-exporting the seven interfaces, which the dead-code ratchet demanded once
their only external consumer was gone.

## Consumer counts (R8)

```
git grep -n "ProgressViewHost" -- src packages
    # before: the module, 2 host imports + 11 use sites, 1 vitest
    # after:  0 hits (the one remaining match is ProgressHostInteractions, a different symbol)
git grep -n "resolveNativeAgentRun"          # after: 2, both inside ProgressViewCommandHandlers.ts
git grep -n "progressHost"                   # after: 0
git grep -n "handleAgentProposalAction"      # after: 0
grep -rn "ProgressViewHost" config/ratchets/*.json    # 0 hits
```

`createProgressViewCommandHandlers` had exactly one production caller
(`ProgressViewHost`), so its signature was free to change; it now has two (one
per host). `architecture-edges-baseline.json:32` already carries
`controllers → platform (value)`, and `ProgressViewCommandHandlers.ts:32` already
imported `RunMetadata` from `@transcript/StreamSnapshotStore`, so moving
`resumeStream` introduces no new architecture edge.

## Host impact

Extension: `ProgressViewMessageHandler` holds `workflowFileActionsController`,
`agentProposalController`, and `commandHandlers` as its own fields instead of
reaching through `progressHost`. Same commands, same wiring.

Desktop: `DesktopProgressBridge` already held two of those three as fields and
copied them out of the host; it now builds them directly. Same commands.

CLI: none.

## Sequencing note for reviewers

`docs/proposals/2026-08-03-run-classification-consolidation.md:239` documents a
live defect **inside** `resumeStream` (it branches on `config.agentCategory`
rather than `identity.kind === 'agent'`), and its Step 2 (`:715`) schedules a
rewrite of exactly those lines. This PR **moves** that function without changing
its logic — the defect is carried across verbatim, deliberately, so Step 2 edits
it at its new home. I checked the open PR list: no run-classification Step 2 work
is in flight, and no other open PR touches `desktopAgentExecution.ts`.

## Scheduled refactoring

None new. This closes the C7 row of
`docs/proposals/2026-08-03-ssot-consolidation-plan.md`.

## Validation

- `npm run typecheck` — clean (all six projects).
- `npm run lint` — clean.
- `npm run check:dead-code-ratchet` — 8 current vs 8 baselined, OK (it caught the
  seven newly-unexported interfaces first; they are now module-private, and the
  baseline is unchanged).
- `npx vitest run src/test-kernel/controllers src/test-kernel/progressView src/test-kernel/desktop`
  — 153 files, 1487 tests, all pass.
- No webview render change: this PR touches only inbound command wiring; the
  handler for every command does the same thing it did before.


LionSR:
Addressed the review suggestion: `StreamHeader.ts`'s `NATIVE_AGENT_ONLY_BUTTONS` comment now points at `resolveNativeAgentRun` in `ProgressViewCommandHandlers` (315a3c9). Also renamed the two hosts' private `createProgressViewCommandHandlers` methods to `createSharedCommandHandlers` (87577ea) — the private method and the imported function shared a name, which reads like recursion even though a class member is not in the method body's lexical scope.

PR #11159, opened 2026-08-19
fix(cli): fold the executions subagent label at projection time

LionSR:
## Summary

The webview folds the subagent execution-id→label map into a tool row's `model.headerPreview` at projection time (`logSlice.ts` passes `executionLabels` to `projectTranscriptRow`). The CLI instead re-derived the preview **at paint time**, threading an `executionLabels` map through `App` → `ConversationPane` → `ConversationRegion` → `StaticConversationTranscript` / `TranscriptReader` → `TranscriptEntry` → `ToolUseRow` → `toolRenderers`, plus the viewport estimators and a `WeakMap` line cache keyed on label-map identity.

This PR gives the CLI the same architecture as the webview: the fold samples the label map once (`subagentExecutionLabelsNow()`), and every painter reads `model.headerPreview`.

- `packages/cli/src/chat/tui/state/childExecutions.ts` — adds `subagentExecutionLabelsNow()`, a synchronous snapshot of the same rosters the existing `subagentExecutionLabels` signal exposes (the signal stays — `TuiStateAndFocus.vitest.ts` consumes it).
- `packages/cli/src/chat/tui/state/transcriptFold.ts` — passes `executionLabels: subagentExecutionLabelsNow()` into the projection.
- `packages/cli/src/chat/tui/panes/toolRenderers.tsx` — header preview is `toolRow.model.headerPreview`; the paint-time re-derivation (and its `executionLabels` option) is gone.
- Deletes the now-dead `executionLabels` plumbing from `transcriptEntryLayout`, `transcriptViewport`, `transcriptLines` (the labels-token `WeakMap` cache included — the key is just `String(cols)` now), `StaticConversationTranscript` (including `executionLabelsEqual`), `ConversationPane`, `ConversationRegion`, `TranscriptReader`, `TranscriptEntry`, `ToolUseRow`, `App`.
- `src/shared/transcript/toolRowModel.ts` — `toolHeaderPreview` loses its last outside caller (the CLI paint path) and drops back to a module-private helper; its stale "exported because labels exist only at paint time" docstring is corrected.
- Each host keeps its own label source (per the audit: do NOT unify the maps) — this only moves *when* the CLI samples it.

## Behavior changes (deliberate)

- **Pruned-roster label loss fixed**: previously a scrollback row printed before a label landed re-derived from a pruned roster at repaint and painted the raw execution id. The row now keeps the label the fold had.
- **Bounded raw-id window while a wait is in flight**: a label that lands mid-`wait` appears when the row re-projects on result settle (same re-projection trigger the webview relies on). No new subscription surface is added to close it further.

## Verified

- `npm run typecheck` — clean (all workspaces).
- `npm run lint` — clean.
- `npm run check:dead-code-ratchet` — clean (the `toolHeaderPreview` export removal is what keeps it at baseline).
- Full CLI kernel suite: 144 files, 2534 passed, 1 skipped.
- `src/test-kernel/progressView` + `src/test-kernel/shared`: 96 files, 901 passed (webview projection path untouched but exercised).
- Consumer greps: `executionLabels` in `packages/cli/src` now appears only in `state/childExecutions.ts` (definition) and `state/transcriptFold.ts` (projection call); the webview's `logSlice.ts` supply sites are unchanged.
- Test changes: 4 `ToolRenderers` executions tests rewritten to pass labels through the projection-time ctx (fixture param, same shape the fold passes); 2 `ConversationTranscript` tests deleted (they pinned the removed paint-time seam — a seam test, not a behavior test); `StaticBandResize` label test rewritten to build the labeled row through the fixture and assert on the ANSI-stripped frame (the label spans an SGR boundary in raw bytes).

## Eyeball checklist (visible change)

- [ ] Run a CLI session that dispatches subagents; during an in-flight `executions` wait the header may briefly show the raw id form, and on settle it shows `executions (wait: labelA, labelB)` — matching the progress view.
- [ ] Scroll a transcript with completed subagent dispatches: headers show labels, never raw execution ids, including after a terminal resize (static-band repaint).
- [ ] `executions (view: label/path...)` rows keep the resource path after the label.

## Net LoC

+168 / −451 across 19 files; production-only (packages/ + src/shared/): roughly +105 / −280.

Part of the CLI state/render host-parity track (C1). Sibling PRs: #11102 (merged), #11105, #11123.


LionSR:
Closing under the current explicit no-test-modification constraint. This refactor rewrites four ToolRenderers tests, rewrites StaticBandResize, changes shared test fixtures, and deletes two ConversationTranscript tests. Those changes are structural to moving label resolution from paint time to projection time; retaining the old test-facing seam would preserve much of the duplicate plumbing this PR is intended to remove. Revisit only when the affected tests may be updated.

PR #11177, opened 2026-08-19
refactor(agent): collapse modelHandlerCompatibilityInference to its guard

LionSR:
## Summary

`modelHandlerCompatibilityInference.ts` carries a wrapper around nothing.
`inferAndLogPersistedModelHandlerCompatibilityKey` calls
`inferPersistedModelHandlerCompatibilityKey`, then logs *if* the result is
truthy. The base function has exactly two outcomes today: it throws for
Google, or it returns `undefined`. So the `if` and its `logger.info` are dead
over the entire input domain, and the wrapper is a pure alias with an extra
`logger` parameter.

This PR collapses the module to the guard it actually is:

- Delete `inferAndLogPersistedModelHandlerCompatibilityKey` and its
  statically unreachable `logger.info` block. Its single production caller
  (`SessionResumeRetrieval.ts:199`) now calls the base function directly —
  same value, no logger argument.
- Delete the now-unused `AgentTrace` type import.
- Inline the one-caller `asRecord` helper as
  `if (!isObject(shared)) return undefined;`, deleting its 8-line docstring.
  The docstring justified the wrapper by "every shape-sniffing call site
  below" — there is exactly one, since `063cc83ad1` removed
  `currentModelFromRawSharedState`. `isObject` already narrows `unknown` to
  `Record<string, unknown>`, so no cast remains to hide.
- Delete the tautological `does not log when inference is inconclusive` test
  and its spy fixture; the two surviving cases (Google throw, explicit
  persisted key honored) keep their value.
- Fix the review checklist (see "Docs ruling overturned" below).

Nothing about the surviving behavior changes: the Google throw, the persisted
compatibility-key parse, and the `undefined` return are byte-identical.

## Irreducibility

Commit `e4f35c87c3` (#10872) deleted the COPILOT arm — the only arm that ever
returned a key:

```
-  if (modelConfig?.provider === ModelProvider.COPILOT) {
-    return 'ModelHandlerOpenRouterNative';
-  }
```

Since that landed, every path through
`inferPersistedModelHandlerCompatibilityKey` either throws (Google) or returns
`undefined`. There is no second behavior to document as deserved — **there is
one behavior and a wrapper around nothing.** Two independent verification
lanes confirmed this by *execution* over the full input domain (149 registered
model ids plus synthetics): 0 returned keys, 0 logger calls, 0
wrapper-vs-base divergence.

Removing the wrapper does not remove a warning or a throw: the deleted
`logger.info` is `info`-level *success* logging on a branch that cannot be
taken, and the Google `throw` is preserved verbatim in the surviving function.

## Docs ruling overturned (mandatory paragraph)

`docs/dev/audits/2026-07-21-agent-sdk-readiness-checkpoint.md:156,342` says
**"Do not re-flag"** `inferAndLogPersistedModelHandlerCompatibilityKey`. That
ruling is falsified on its own premises, and I verified both claims at
`941e36404b` before writing this:

1. **The caller count in the ruling is wrong.** It keeps the symbol because it
   "has **4** production callers across `executeAgent` /
   `SessionResumeRetrieval` / `runToolUseFlow` / `runReflectionFlow` + a
   dedicated test". A repo-wide grep at `941e36404b` finds **one** production
   caller — `src/agent/runtime/SessionResumeRetrieval.ts:199` — plus the test.
   `executeAgent`, `runToolUseFlow`, and `runReflectionFlow` contain zero
   references to the symbol.
2. **The ruling predates the change that emptied the function.** It is dated
   2026-07-21; `e4f35c87c3` (#10872) landed 2026-08-18 and removed the last
   live arm. The ruling could not have accounted for it.

The ruling was an anti-incomplete-grep guard. Its premise (a 4-caller helper
with a live logging branch) no longer describes the code, so the guard does
not apply.

Also fixed in this PR:
`.claude/skills/code-review/references/review-checklist.md:142` is **active
review guidance** that cites this module as the precedent for "message-shape
sniffing at 6 call sites". That parenthetical has been false since
`063cc83ad1` — the module reads model identity only and parses no messages
(its own surviving comment says so). The stale precedent is removed from the
bullet; the bullet's two accurate precedents (`executionStreamResolver.ts`,
the dual `resolveWorkspaceSourceDir`) remain.

## Single source of truth

`inferPersistedModelHandlerCompatibilityKey` (`modelHandlerCompatibilityInference.ts`)
is now the sole owner of "what compatibility key does a keyless persisted run
get?", with `inferPersistedFlowModelHandlerCompatibilityKey` as the one
flow-record entry point above it. The aliasing wrapper that made it look like
two owners is gone.

Two indirection layers removed, none added: a single-caller logging wrapper
and a single-caller `asRecord` type-narrowing helper. No new function, type,
or file.

## Net elements (R6)

| | |
|---|---|
| Files added | 0 |
| Files deleted | 0 |
| Exported symbols | −1 (`inferAndLogPersistedModelHandlerCompatibilityKey`) |
| Non-exported functions | −1 (`asRecord`) |
| Classes / interfaces / enums | 0 |
| Type imports | −1 (`AgentTrace`) |
| Tests | −1 case, −1 fixture (`vi.fn` spy + `beforeEach`) |
| Net LoC | **−56** (7 insertions, 63 deletions across 4 files) |

Element delta is negative in every row.

## Consumer counts (R8)

No emit path touched. Deleted export
`inferAndLogPersistedModelHandlerCompatibilityKey`:

- Production consumers, grepped across `src/`, `packages/extension`,
  `packages/desktop`, `packages/cli`, `packages/agent`, `packages/trace-viewer`:
  **1** (`src/agent/runtime/SessionResumeRetrieval.ts:199`) — migrated in this
  PR to `inferPersistedModelHandlerCompatibilityKey`.
- Test consumers: **1**
  (`src/test-kernel/agent/modelHandlers/ModelHandlerCompatibilityInference.vitest.ts`)
  — the case was tautological (`toBeUndefined()` + `not.toHaveBeenCalled()` on
  a function that can only do that) and is deleted.
- Not present in `config/ratchets/knip-baseline.json`; the dead-code ratchet
  is green (8 current vs 8 baselined).

Retained exports `inferPersistedModelHandlerCompatibilityKey` (2 callers, both
in `SessionResumeRetrieval.ts`) and `inferPersistedFlowModelHandlerCompatibilityKey`
(1 caller, `AgentLaunchContext.ts:237`) are unchanged.

## Host impact

Extension: none — behavior-identical, VS Code-free zone untouched.

Desktop: none.

CLI: none.

## Scheduled refactoring

One follow-up, **deliberately not fixed here** because it is a behavior
question rather than a deletion: the Google keyless-resume guard has two
different outcomes depending on entry point.

- `src/agent/runtime/SessionResumeRetrieval.ts:257` lets the throw propagate,
  failing the resume.
- `src/agent/runtime/AgentLaunchContext.ts:228-247` catches it, warns, and
  resumes on the default model-handler route — its own comment acknowledges
  this "can resume with the wrong provider message format".

Whether a keyless Google resume should fail or degrade-with-warning needs an
owner decision; deciding it inside a deletion PR would smuggle a behavior
change past review. Out of scope here.

Also explicitly out of scope: renaming `assertKeylessResumeSupported`-style
symbols. That would be churn.

## Validation

Run from an isolated worktree at `origin/main` = `941e36404b`
(`pnpm-lock.yaml` install artifact reverted before commit; `git status`
confirms it is not in the diff):

- `FORCE_COLOR=0 npx tsc --noEmit --checkers 8` — clean
- `FORCE_COLOR=0 npx tsc --noEmit --checkers 8 -p tsconfig.test-kernel.json` — clean
- `FORCE_COLOR=0 npx vitest run src/test-kernel/agent/modelHandlers/ModelHandlerCompatibilityInference.vitest.ts src/test-kernel/agent/runtime/` — 41 files, 526 tests passed
- `FORCE_COLOR=0 npx eslint <changed files>` — clean
- `npm run check:dead-code-ratchet` — OK (8 current vs 8 baselined)

No new tests added: this is a behavior-preserving deletion, so per the testing
discipline the existing suite passing is the evidence.


LionSR:
Closing under the active no-test-modification constraint. This refactor deletes and rewrites an existing test contract; preserving that test would require retaining the production helper the PR removes. The remaining simplification is not worth a compatibility wrapper.

PR #11179, opened 2026-08-19
refactor(settings): retire the hand-rolled agent-CLI enum readers

LionSR:
## Summary

Six agent-CLI getters — `getClaudeAgentModel`, `getClaudeAgentPermissionMode`,
`getClaudeAgentEffort`, `getCodexReasoningEffort`, `getCodexApprovalPolicy`,
`getCodexSandboxMode` — each read their workspace-state key through
`createEnumStateGetter` plus a `parseEnumSetting`-built `parse*` function. That
is a hand-rolled read path that restates key → slot → default knowledge the
setting catalog already owns. All six keys are catalogued, so
`readPlatformSetting(key)` resolves the same slot and the same `.prefault()`
default. This rewrites the six getters as one-line `readPlatformSetting` calls
and deletes the superseded path.

**This is the retirement of a superseded read path, not a dedup of two
divergent systems.** The corner inputs were executed against both readers and
there is **no value divergence**: `'ask'` on the Codex approval key,
`'claude-opus-3'` on the Claude model key, JSON `null`, and worktree-namespaced
state all produce identical values through either reader. Both paths read the
same `platform().workspaceState`, and all six catalog rows declare
`sameSlot('workspaceState')` with the same `*_DEFAULT` constant that the old
`fallback` argument passed.

### Irreducibility

The catalog rows' own `honoredBy` fields **name these two files as the
readers** — `src/shared/schemas/stateSettings.ts:695-696` defines
`CODEX_CONFIG_READER = 'src/tools/codexConfig.ts'` and
`CLAUDE_AGENT_CONFIG_READER = 'src/tools/claudeAgentConfig.ts'`, and every one
of the six rows passes them to `everyHost(...)`. And
`src/utils/config/providerConfig.ts:16-17` states the house rule verbatim:

> A key that genuinely has no catalog entry should be catalogued, not given a
> fourth read path.

So the surviving path is superseded by a replacement that is documented as its
replacement. Both citations were re-derived at HEAD before writing this.

### `normalizePersisted: parseClaudeAgentModel` goes with it

`stateSettings.ts:1070` fed `parseClaudeAgentModel` into the Claude-model row as
`normalizePersisted`. Unlike the approval-policy row's normalizer at
`stateSettings.ts:665` (which really does `trim().toLowerCase()` for
hand-edited config files), `parseClaudeAgentModel` performs **no migration
mapping at all** — it is a bare enum-membership filter. Its only effect on the
row was to convert an invalid persisted value into the default *before*
`readSetting` could warn about it, i.e. to suppress the diagnostic. The
`normalizePersisted` field itself stays; the approval-policy row still uses it.

### Louder, not quieter (user-visible note)

Marginally: a stale persisted value now emits one `log.warn` naming the key
(`readSetting`'s `Ignoring invalid persisted value for setting "<key>"`) before
falling back to **the same** default the old parser returned silently. That is
louder, not quieter — no `throw` or `warn` was removed anywhere in this diff.

### MANDATORY scope guard — this does NOT fix P1.6

`docs/proposals/2026-07-12-fallback-audit.md:1395` and its §P1.6 prescribe
**fail closed** (or block launch with a precise diagnostic) for
`CODEX_APPROVAL_POLICY`, `CODEX_SANDBOX_MODE`, and
`CLAUDE_AGENT_PERMISSION_MODE`. This PR does **not** do that: those three still
resolve an invalid persisted value to an ordinary permissive default. **P1.6's
fail-closed disposition for those three is untouched and still owed**, tracked
under #10674. Please do not read this PR as "P1.6 fixed".

### SDK alignment guards preserved

`getCodexApprovalPolicy` and `getCodexSandboxMode` keep their `ApprovalMode` /
`SandboxMode` return annotations while the *call* is typed with the schema type
(`CodexApprovalPolicy` / `CodexSandboxMode`) — so the assignability check that
catches a schema value the Codex SDK union does not accept still fires at
compile time. (Typing the call itself as the SDK type would have made the cast
unchecked and silently dropped that guard.) All six getters stay arrow
functions, unhoisted, so `platform()` resolves per call.

## Issue acceptance gates

No issue; this lands one verified item from an adversarial code-sweep ledger.
The item was pre-refuted once as "dedup of divergent systems" and re-verified
as GO-NARROWED after the corner inputs were executed and found identical.

## Single source of truth

The state-setting catalog (`src/shared/schemas/stateSettings.ts`) now owns
key → slot → default for all six agent-CLI enum settings, reached through the
one canonical runtime reader `readPlatformSetting`
(`src/utils/config/platformSettings.ts`). There is no second place that names a
default for these keys.

Removed: `createEnumStateGetter` (a one-caller-shape state-getter factory),
`parseEnumSetting` (a hand-rolled enum-membership validator that duplicates the
Zod enum already on each row), and six `parse*` functions built from it. **No
abstraction added** — the replacement is an existing repo-wide helper with 14
prior call sites.

## Net elements (R6)

- Files: 6 changed, **1 deleted** (`src/tools/support/enumConfig.ts`), 0 added.
- Diffstat: **24 insertions, 133 deletions** (net **−109 LoC**; production
  −88, test −21).
- Exported symbols: `^-export` = 12, `^+export` = 5 → **net −7 exports**
  (`createEnumStateGetter`, `parseCodexSandboxMode`,
  `parseCodexReasoningEffort`, `parseCodexApprovalPolicy`,
  `parseClaudeAgentModel`, `parseClaudeAgentPermissionMode`,
  `parseClaudeAgentEffort`). The 5 `+export` lines are the same three Claude
  getters and two Codex getters, re-spelled as arrow functions.
- Classes/interfaces/enums: 0 added, 0 deleted.

## Consumer counts (R8)

Grepped across `src/**` and `packages/**` (`.ts`/`.tsx`) before deleting:

| Deleted export | Consumers found | Disposition |
| --- | --- | --- |
| `createEnumStateGetter` | 2 (`codexConfig.ts`, `claudeAgentConfig.ts`) | both re-routed to `readPlatformSetting` |
| `parseCodexSandboxMode` | 1 (`codexConfig.ts`) | re-routed |
| `parseCodexReasoningEffort` | 1 (`codexConfig.ts`) | re-routed |
| `parseCodexApprovalPolicy` | 2 (`codexConfig.ts`, `SharedSchemas.vitest.ts`) | re-routed; test deleted as tautological |
| `parseClaudeAgentModel` | 3 (`claudeAgentConfig.ts`, `stateSettings.ts` row, `SharedSchemas.vitest.ts`) | re-routed; row normalizer deleted; test deleted |
| `parseClaudeAgentPermissionMode` | 1 (`claudeAgentConfig.ts`) | re-routed |
| `parseClaudeAgentEffort` | 1 (`claudeAgentConfig.ts`) | re-routed |
| `parseEnumSetting` (module-local) | 6, all in-file | deleted with them |

The six *getters* are unchanged as exports and keep all their consumers:
`src/tools/codex.ts:462,464,493` and `src/tools/claudeAgent.ts:526-528` (plus
the two resume-fallback suites, which stub the getters and are unaffected).

## Host impact

Extension: none — same values read from the same workspace-state slot; one
extra `log.warn` when a persisted value is already invalid.

Desktop: same as extension.

CLI: same as extension. All six rows carry `surfaces.cliConfig: true` and
`honoredBy.cli` pointing at the same two files; `/config` writes are unchanged.

## Scheduled refactoring

P1.6 fail-closed handling for `CODEX_APPROVAL_POLICY`, `CODEX_SANDBOX_MODE`,
and `CLAUDE_AGENT_PERMISSION_MODE` remains owed — tracked under #10674.

## Validation

All run from an isolated worktree at `origin/main`
(941e36404b), with `FORCE_COLOR=0`:

- `npx tsc --noEmit --checkers 8` — clean
- `npx tsc --noEmit --checkers 8 -p tsconfig.test-kernel.json` — clean
- `npx tsc --noEmit --checkers 8 -p packages/{cli,desktop,trace-viewer,extension}/tsconfig.json` — clean
- `npx vitest run src/test-kernel/{schemas,shared/stateSettings.vitest.ts,utils/config,tools,cli/ConfigForm.vitest.ts}` — 108 files / 996 tests passed
- `npx vitest run src/test-kernel/architecture` — 17 files / 104 tests passed
  (one transient failure on the first invocation of `subsystemEdgeRatchet`
  baseline read; green on 4 subsequent runs, including 3 isolated runs of that
  file alone)
- `node scripts/check-dead-code-ratchet.mjs` — OK, 8 current vs 8 baselined, no
  baseline regeneration needed
- `node scripts/check-browser-safe-utils.mjs` — OK (64 total, 6 reachable)
- `node scripts/check-guidance-refs.mjs` — OK
- `npx eslint <changed files>` — clean

`pnpm-lock.yaml` was reverted after install and is not part of this commit.

**No tests added.** Two suites in `SharedSchemas.vitest.ts` were deleted because
they exercised only the deleted parsers (`parseCodexApprovalPolicy('ask')` and
`parseClaudeAgentModel('claude-opus-3')`); those exact inputs are the corner
cases verified above to produce identical values through the surviving reader.


LionSR:
Closing under the active no-test-modification constraint. This refactor deletes existing parser-contract tests and the production helpers those tests import. Restoring the tests would require retaining compatibility wrappers, which would preserve the duplicate readers the PR aims to remove.

PR #11181, opened 2026-08-19
refactor(scripts): cut format-staged to the one Prettier config shape this repo has

LionSR:
`scripts/format-staged.mjs` carried three Prettier config-shape families — a `package.json` `"prettier"` key with pointer chasing, a `package.yaml` key, and executable JS configs — plus a ~200-line hand-rolled reimplementation of Node's CommonJS `LOAD_AS_FILE` / `LOAD_AS_DIRECTORY` search, needed only to verify a JS config's relative dependencies without executing it.

**None of that machinery has ever run in this repository.**

## Measured before deleting

| Check | Result |
|---|---|
| `prettier.resolveConfigFile` over every tracked file | **3256 / 3256** → the single root `.prettierrc` |
| Top-level `prettier` key in any workspace `package.json` | none |
| `git log --all --diff-filter=A` for `*prettier.config.*`, `.prettierrc.{js,cjs,mjs,json,yaml,yml}`, `**/package.yaml` | **empty** — no such config on any branch, ever |

The provenance explains it. Issues **#10430 → #10602** are a four-day cascade in which each issue hardened one hypothetical config shape against another hypothetical, and none fixed a contributor-reported defect.

## What survives

Standalone data configs (`.prettierrc`, `.json`, `.yaml`, `.yml`) are handled; anything else **skips loudly**, so a second config shape announces itself instead of being silently mishandled. The header comment records the measurement and points at where to add a shape back.

## Two behaviors deliberately preserved

Both were caught by adversarial verification of an earlier version of this proposal, and both are load-bearing:

1. **A clean config still goes to Prettier's own loader.** A config equal to its index blob returns `{config: null}` regardless of shape (#10504). Without that arm, a clean `package.json` or JS config would regress from *working* to *skipping*. The three #10504 regression tests that pin this are kept deliberately.

2. **`verifyConfigDeps` keeps its three separate parse branches and every throw.** Merging them looks tempting and is wrong: `parseYaml` does not reject JS source — it returns a string for `import opts from "./opts.mjs";` — so a merged branch would turn a diverged `.mjs` config's loud skip into a **silent format using unstaged rules**. That is exactly the silent-degradation defect class CLAUDE.md names. The always-true `strict` parameter is gone with all of its throws intact; throw count is unchanged for every path that still exists.

## Tests

**24 deleted, none added.** 19 covered the deleted shapes directly. The other 5 — four CommonJS-resolution cases and one autocrlf `package.yaml` case — still pass, but now exercise *Prettier's* resolver rather than this script's; autocrlf keeps seven other regression tests.

## Net

**−1,036 LoC across 2 files.** The script has zero exports and its test spawns it as a subprocess, so the blast radius is exactly these two.

## Verified

- `npx tsc --noEmit --checkers 8` — clean.
- `npx tsc --noEmit --checkers 8 -p tsconfig.test-kernel.json` — clean.
- `npx vitest run src/test-kernel/scripts/FormatStaged.vitest.ts` — **37 passed**.
- `npx prettier --check` on both changed files — clean.

One failure to name plainly: `install-local-hooks > installs a hook that auto-stages formatting without losing unstaged edits` fails in this environment. It fails **identically on unmodified origin/main** (verified by reverting the script and re-running), so it is pre-existing and unrelated.


LionSR:
Closing under the active no-test-modification constraint. This refactor deletes 698 lines from an existing test suite, including 24 contract tests for behavior the production script removes. Restoring the test file would make the refactor fail by design, so it cannot be repaired without violating the constraint.

PR #11189, opened 2026-08-19
refactor(core): sweep simplification leftovers since v0.40.2

LionSR:
## What

Fifty `our-code-simplifier` agents (Sonnet 5, high effort) swept the **618
host-agnostic files under `src/` and `packages/agent` that changed since release
`v0.40.2`** (2026-08-12) — ~149k LOC — one disjoint, directory-local slice each.
Every edit is behavior-preserving; the yield is leftovers from eight days of very
heavy refactoring churn.

**42 files, +98 / −188.**

## What it found

| Category | Count | Examples |
| --- | --- | --- |
| Dead exports / dead types | 19 | symbols with no consumer outside their declaring file lose `export`, or the declaration entirely (`BbtLibrary`, `AgentOptionsDataPayload`, …) |
| Pass-through collapse | 3 | `StreamSnapshotStore.mutate` and `writeUnbuffered` each had one caller and made no decision — inlined into it |
| Dead parameters / fields | 5 | ctor and function params that every live call site left `undefined`; Zod fields nothing reads |
| Union-arm removal | 1 | `checkToolInstalled(string \| ToolConfig)` → `checkToolInstalled(string)` |
| Repeated-literal folding | 8 | `textBlock(…)` in the Anthropic handler replaces 6 hand-written `{ type: 'text', citations: null }` literals |
| Stale comments | 3 | comments describing machinery that a prior PR already deleted |

The one non-mechanical change: `checkToolInstalled` accepted either a tool name or
an ad-hoc `ToolConfig`. Only two call sites used the config arm, both with literals
declared next to them in `externalToolDefs.ts`. Those move into `TOOL_CONFIGS` as
`texra` / `texra-local`, so every caller now goes through the one registry and the
`config === null` branch disappears.

## Honest scope note

Net −90 lines from 50 agents over 149k LOC is a small return, and that is the
finding: core `src/` has already absorbed ~200 simplification PRs plus the recent
collapse campaign. Roughly half the bins reported **zero** edits after reading their
slice in full and grepping consumers repo-wide — "already coherent" was the
dominant verdict, not a shortage of effort. The agents also recorded **306
candidates they deliberately did not take**, mostly needing a caller outside their
own slice; those are captured for a follow-up wave rather than forced in here.

A companion sweep over the host packages (`extension` / `cli` / `desktop`,
391 files) is running separately.

## Verified

- `typecheck` — all six projects (workspace, test-kernel, agent, cli, trace-viewer, desktop)
- `eslint` on every changed file, `prettier --check`
- full `vitest run` — **842 files, 10,316 tests passed**, 0 failures
- `src/test-kernel/architecture` ratchets — 104 tests passed

Three agent proposals were **rejected during integration** and reverted rather than
shipped: a deleted `AgentRosterController.allPresets()` that the CLI roster form
actually calls, a nested ternary that violates `no-nested-ternary`, and the dead
`BbtLibrary` type was finished off properly instead of left to trip the ratchet.

## Net elements (R6)

- 42 production files changed, with 98 additions and 188 deletions, for a net reduction of 90 lines.
- 29 exported declarations were removed or made file-private, and one exported declaration was added, for a net reduction of 28 exported declarations.
- No test files are changed.
- The six Anthropic `textBlock(...)` substitutions are request-equivalent rather than byte-identical: they omit the optional `citations: null` field, matching the existing helper and the other call sites in that handler.

## Consumer counts (R8)

- Every removed export was searched repo-wide. Each had zero consumers outside its declaring file.
- Every inlined helper had one caller. The consolidated tool configuration entries replace the only two production callers that used the removed ad hoc `ToolConfig` union arm.
- No event, subscription, or output path is deleted or re-routed.

### DIFF
--- a/src/agent/implementations/flows/reflection/RoundPersistedFlow.ts
+++ b/src/agent/implementations/flows/reflection/RoundPersistedFlow.ts
@@ -117,9 +117,8 @@ export class RoundPersistedFlow<
       parentStage?: StageHandle | null;
       sharedSchema?: z.ZodType<S>;
     },
-    runId?: string,
   ) {
-    super(start, kv, runId, options?.sharedSchema);
+    super(start, kv, undefined, options?.sharedSchema);
     this.callbacks = options?.callbacks ?? {};
     this.parentStage = options?.parentStage ?? null;
   }
--- a/src/agent/implementations/flows/reflection/output/LatexDiffManager.ts
+++ b/src/agent/implementations/flows/reflection/output/LatexDiffManager.ts
@@ -1,6 +1,6 @@
 import * as path from 'node:path';
 
-import type { AgentTrace, StageHandle } from '@agent/trace';
+import type { AgentTrace } from '@agent/trace';
 import { AgentWorkflowSetting } from '@agent/core/definition/AgentDataclass';
 import { LaTeXdiffResult, LaTeXdiffService } from '@latex/latexdiff';
 import { compileLatex2Pdf } from '@latex/texTools';
@@ -113,7 +113,6 @@ export class LatexDiffManager {
   async handleLatexdiffOfOutput(
     currRound: number,
     mapping: RoundFileMapping,
-    stage?: StageHandle,
   ): Promise<CompiledPdfArtifact[]> {
     const execute = async (): Promise<CompiledPdfArtifact[]> => {
       if (!(await checkToolInstalled('latexdiff'))) {
@@ -255,7 +254,7 @@ export class LatexDiffManager {
 
       return artifacts;
     };
-    return tryOperation(() => (stage ? stage.within(execute) : execute()), {
+    return tryOperation(execute, {
       logger: this.logger,
       level: 'error',
       label: 'Error during latexdiff processing',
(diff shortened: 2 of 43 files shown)


LionSR:
Confirming the one flagged item rather than changing code.

`textBlock(text)` returns `{ type: 'text', text }` — it omits `citations` rather than sending `citations: null`. So the six converted sites are not byte-identical payloads, and "behavior-preserving throughout" in the PR body is one word too strong there.

Why it stands anyway: `citations` is an optional nullable field on Anthropic's text block param, so omitted and `null` are the same request. More to the point, `textBlock` was already the convention at ~14 other sites in this same file, all of which omit `citations`. The six literals were the outliers; this makes the file consistent rather than introducing a new shape. Nothing in the repo reads `citations` off an outbound block, and no test asserts on it by exact equality (`AnthropicServerTools.vitest.ts` builds its own fixture literal, it does not compare against ours).

Accurate wording would be: request-equivalent, not payload-identical, at those six sites.

LionSR:
rebase

PR #11193, opened 2026-08-19
refactor(hosts): sweep simplification leftovers in extension/cli/desktop

LionSR:
Companion to #11189 (which covers host-agnostic `src/`). Disjoint file sets — the two can merge in either order.

## What

Fifty `our-code-simplifier` agents (Sonnet 5, high effort) swept the **391 host-package
files that changed since release `v0.40.2`** (2026-08-12) — ~99k LOC across
`packages/extension`, `packages/cli`, `packages/desktop`, `packages/trace-viewer` —
one disjoint, directory-local slice each. Every edit is behavior-preserving.

**58 production files, +342 / −663.**

The prompt was stiffened relative to #11189 after that batch came back mostly
`export`-keyword removals: whole-module deletion put in scope, wrapper *chains*
targeted rather than single wrappers, "already coherent" rejected as an answer unless
it names the candidates examined, and cross-bin candidates reported explicitly. The
yield roughly tripled (−574 vs −92 net).

## What it found

| Category | Count | Notes |
| --- | --- | --- |
| Pass-through / wrapper inline | 26 | single-caller private helpers folded into their sole call site; argument-shuffling wrappers deleted; one JSX component that only re-rendered a single child collapsed |
| Dead exports / dead types | 19 | no consumer outside the declaring file |
| Dead code | 8 | including a function a maintainer proposal already named as dead |
| Duplication | 7 | repeated patch+push blocks routed through the helper already present in the same file; identical derivations given one owner |
| Type tightening | 4 | e.g. `Awaited<ReturnType<typeof …>>` chains replaced by the real type |
| Speculative generality | 2 | |

Two changes worth a reviewer's eye:

- **`StaticConversationTranscript`** (−58) — header insertion re-ran a full relayout
  `reduce` to learn how many rows it had added. It now derives the delta directly. The
  equivalence was hand-checked against the old reduce; the exported
  `buildStaticTranscriptItems` / `advanceStaticTranscriptState` signatures and return
  shapes are unchanged.
- **`TodosPlanPanel`** (−16) — the panel opened its own `useSignal` subscriptions to
  `activeStreamId` / `streams`, which its parent `ConversationRegion` already reads in
  the same render. It now takes them as props. Its parent already re-renders on
  `streamArtifactRevision`, so repaint timing is unchanged; the test file exercises the
  pure `compactTodosPlanRows` helper and never renders the component.

## Rejected during integration

Three agent proposals were reverted rather than shipped:

- A collapse of `messageDispatcher.dispatchMessage` into `ProgressApp`. It looked like a
  pass-through but is a partial application binding the handler registry, with **four**
  consumers (`ProgressApp`, trace-viewer's `replayTrace`, two kernel tests). Removing it
  would have forced all four to re-state the registry and added a new export to do it.
- A swap of a locally-derived type for `import type { AgentOptionsDataPayload } from
  '@agent/index/agentRegistry'` in `mainViewCommands.ts` — a **new** `@agent/*` deep
  import from the extension, which the `host-agent-import-baseline` ratchet rightly
  rejected. Never widen a baseline.
- Prettier drift in two CLI files (fixed rather than shipped).

One accepted trade to note: `tui-harness.tsx` now routes three fixed-id transcript
entries (`harness-interrupted`, …) through the existing `appendHarnessTranscript`, which
generates ids from `Date.now()`. Those ids are opaque React keys, never rendered and
never asserted on, and this is a dev harness rather than production UI — but it is a
determinism reduction in the harness, so it is called out rather than buried.

## Verified

- `typecheck` — all six projects (workspace, test-kernel, agent, cli, trace-viewer, desktop)
- repo-wide `npm run lint`, `prettier --check` on every changed file
- `src/test-kernel/architecture` — 104 tests, including the host `@agent/*` deep-import ratchet
- full `vitest run` — **842 files, 10,316 tests passed**, 0 failures

The agents also recorded **371 candidates they deliberately did not take**, 33 of them
explicitly flagged as needing a caller outside their own slice. Those are captured for a
follow-up wave rather than forced in here.

## Net elements (R6)

- 58 production files changed, with 342 additions and 663 deletions, for a net reduction of 321 lines.
- 41 exported declarations were removed or made file-private. Three exported declarations were added while consolidating shared ownership, for a net reduction of 38 exported declarations.
- The sole test-file diff from the original sweep was reverted in `ddb697f97b`; this PR now changes no tests.

## Consumer counts (R8)

- Every removed export and private helper was searched repo-wide before removal. Each removed export had zero consumers outside its declaring file; each inlined private helper had one caller unless the PR routes repeated logic through an existing multi-caller helper.
- The surviving extracted helpers have multiple consumers: `carriedPromotionFrontier` has two call sites, `deferOrHandleBareEscape` has two, and `showMessageBoxOfType` has three.
- The attempted `messageDispatcher.dispatchMessage` removal was rejected because it has four consumers. The attempted new `@agent/index/agentRegistry` import was also rejected because it would widen the host-agent deep-import baseline.

### DIFF
--- a/packages/cli/scripts/tui-harness.tsx
+++ b/packages/cli/scripts/tui-harness.tsx
@@ -46,7 +46,6 @@ import {
   MESSAGE_TYPES,
   STREAM_PHASE,
   STREAM_LOG_ENTRY_TYPES,
-  USER_FOLLOW_UP_SUPPORT,
   TODO_STATUS,
   TOOL_USE_STATUS,
   type ActiveChildInfo,
@@ -606,18 +605,12 @@ function harnessModelStatus(
   }
 }
 
-function harnessModel(fixture: HarnessModelFixture): CliModelAccess {
-  return {
+function harnessOrchestrationModels(): readonly CliModelAccess[] {
+  const models = HARNESS_ORCHESTRATION_MODEL_FIXTURES.map((fixture) => ({
     model: fixture,
     available: isHarnessModelAvailable(fixture.availability),
     status: harnessModelStatus(fixture.availability),
-  };
-}
-
-function harnessOrchestrationModels(): readonly CliModelAccess[] {
-  const models = HARNESS_ORCHESTRATION_MODEL_FIXTURES.map((fixture) =>
-    harnessModel(fixture),
-  );
+  }));
   return SHOW_NO_RUNNABLE_ORCHESTRATION_MODELS
     ? models.map((model) => ({
         ...model,
@@ -936,16 +929,6 @@ function makeChildEntries(agent: string, action: string): TranscriptRow[] {
     ),
     harnessTextRow(`${agent}-assistant`, 'assistant', assistantText, 2, false),
   ];
-}
-
-function harnessMessageEntry(
-  id: string,
-  text: string,
-  kind: 'assistant' | 'error' | 'user' = 'assistant',
-  seqNo = 1,
-  settled = true,
-): TranscriptRow {
-  return harnessTextRow(id, kind, text, seqNo, settled);
 }
 
 function makeEditApprovalRequest() {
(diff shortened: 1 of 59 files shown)


LionSR:
fix CI

LionSR:
Two notes on the review, both addressed.

**The `appRoot` non-null assertion is back.** The review says removing `!` "has no runtime effect and makes the pre-existing `appRoot == null` guard meaningful" — correct about runtime, but `tsconfig.renderer.json` disagreed and failed the build: TypeScript does not carry the `== null` throw's narrowing into the function bodies that later pass `appRoot` to `render()`, so both call sites became `HTMLElement | null`. Restored in af1f6e9. The runtime guard stays; it just is not what satisfies the compiler. (The review ran against the pre-fix commit.)

**The `StaticConversationTranscript` comment is fixed rather than left cosmetic.** It claimed the fresh totals walk costs "the same one full-history walk the old inline reduce paid", which only holds on a cold start — once a previous render has emitted rows, `currentItems` is the newly appended tail. The comment now says what the walk actually covers.

Everything else in the review is a confirmation, and the `tui-harness.tsx` `Date.now()` id disclosure is unchanged from the PR body.

PR #11194, opened 2026-08-19
refactor(cli): delete relay-era model-access and auth-profile residue

LionSR:
## What

Four leftovers of `368ee4f601` (relay client plane removal), which deleted the readers these fields fed but left the fields standing. **−66 net LoC.**

## 1. `CliModelAccessStatus.texraSignedIn` and `.personalKeyProviders`

Both write-only. Their readers were the "Included access" and "own API keys" rows in `buildCliModelAccessItems`, deleted by `368ee4f601`.

**Care needed here, and it is the reason this is worth stating:** `CliAccountStatus.texraSignedIn` (`orchestration.ts:101`) is a *different interface with the same field name*, and it **is** live — read at `:216` and `:284`. A field-name grep alone would have condemned both. Only the model-access copy goes; I confirmed the two reads take `CliAccountStatus` by their parameter types.

Dropping `personalKeyProviders` also removes a `configuredApiKeyProviders(secrets)` call — a secret-store scan on every status read.

## 2. `CliAuthProfile.note`

No writer. `getCliAuthProfile` never sets it; the three sites that did (relay-token warnings at `supabaseAuth.ts:300,316,341`) went with the same commit. Four `if (profile.note)` branches were unreachable.

Not an output contract: `cliOutput.ts:47` is `z.looseObject`, and `docs/guide/texra-cli.md:152-158` documents no `note`.

## 3. `getCliSessionAccessToken`

**One occurrence repo-wide — its own declaration.** Zero consumers in production, tests, scripts, or docs. Its caller `loadIncludedUsageLine` went with the relay plane.

This is the specimen for the `packages/cli` knip blind spot recorded in #11165: a provably dead export that the dead-code ratchet does not report.

## 4. `launcherModelAccess`

A local that existed only to graft `texraSignedIn` onto `modelAccess`. With the field gone the call site passes `modelAccess` directly.

## Tests

Deleted with the behavior they pinned, per AGENTS.md "Testing discipline": the two profile-note rows, `preserves a profile note exactly once`, and `reports configured provider keys as display names`. Three fixtures re-pinned.

## Net elements (R6)

- Files: 8 changed (6 production, 2 test)
- `^[+-]export` symbols: **−1** (`getCliSessionAccessToken`)
- Interface fields: **−3** · locals: **−1** · now-unused imports: **−2**
- Net LoC: **−66** (`26 insertions(+), 92 deletions(-)` vs `origin/main`)

## Consumer counts (R8)

Grepped with `--no-ignore-vcs` (a global gitignore hides tracked files from plain `rg` here — #11165):

- `CliModelAccessStatus.texraSignedIn` — 0 readers (the 2 `status.texraSignedIn` reads are `CliAccountStatus`)
- `CliModelAccessStatus.personalKeyProviders` — 0 readers
- `CliAuthProfile.note` — 0 writers, 4 dead reader branches
- `getCliSessionAccessToken` — 1 occurrence repo-wide, its declaration
- `CliAccountStatus.texraSignedIn` — 2 readers, **retained**

## Checks

`typecheck` ✅ (exit 0) · `lint` ✅ (exit 0) · `format` ✅

`vitest run src/test-kernel/cli` ✅ — **144 files, 2537 tests, 0 failures**.

https://claude.ai/code/session_01JJmSeJm4wocNwc7AFMrJBU

### DIFF
--- a/packages/cli/src/commands/auth.ts
+++ b/packages/cli/src/commands/auth.ts
@@ -285,10 +285,7 @@ const authStatusCommand = defineCliCommand({
     emitCliResult(context, {
       json: profile,
       ndjson: { kind: 'auth-status', ...profile },
-      text: [
-        formatAuthStatusLine(profile),
-        ...(profile.note ? [profile.note] : []),
-      ].join('\n'),
+      text: [formatAuthStatusLine(profile)].join('\n'),
     });
     return CliExitCode.Success;
   },
--- a/packages/cli/src/commands/orchestrate.ts
+++ b/packages/cli/src/commands/orchestrate.ts
@@ -183,16 +183,12 @@ async function runOrchestration(context: CliContext): Promise<number> {
       grokSignedIn: modelAccess.grokSignedIn,
       grokAccountLabel: modelAccess.grokAccountLabel,
     };
-    const launcherModelAccess = {
-      ...modelAccess,
-      texraSignedIn: authProfile.authenticated,
-    };
     const items = buildCliOrchestrationItems({
       presetPlans: presetPlanSet.plans,
       history,
       toolUseAgents,
       includeMultiAgentLoginHint: !presetPlanSet.remoteCatalogRefreshAttempted,
-      modelAccess: launcherModelAccess,
+      modelAccess,
       account: accountStatus,
       presetLaunchBlockReason,
     });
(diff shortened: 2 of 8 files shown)


LionSR:
Closing because the active campaign constraint forbids adding, deleting, or modifying tests. This cleanup deletes and rewrites existing CLI test expectations, and the production deletions cannot remain compatible with those unchanged contracts. The remaining production-only residue can be reconsidered separately.

PR #11195, opened 2026-08-19
test: clear residue the campaign's deletions left in fixtures and assertions

LionSR:
## What

Seven single-line survivors of the 2026-08-17→19 campaign, each verified inert at HEAD. Small, but each one currently misleads a reader into thinking a deleted thing still matters.

| Where | What | Why inert |
|---|---|---|
| `SdkErrorUtils.vitest.ts:875` | `isRelayError: false` | not a field on any schema in `shared/schemas/errors.ts`; the window scrubbed this identifier from **four other sites in the same file** and missed this one. The only relay residue left in the suite. |
| `WorkflowScriptStreamTranscript.vitest.ts:67`, `ConversationTranscript.vitest.ts:102` | `apiMode: 'personal'` | `SessionMeta` no longer has the field. `368ee4f601` removed the identical line from `ConfigForm.vitest.ts` and missed these two. |
| `setupAssistantCommand.ts:80` | doc comment | credits the CLI with an "apiMode-aware policy" for a mode with **zero** production hits |
| `AgentHandlersDelete.vitest.ts:38`, `ExtensionAgentHandlers.vitest.ts:51` | `getUserTier` | the static went with `368ee4f601` |
| `ToolUseFormatter.vitest.ts:245` | `not.toContain('Built Mathlib')` | sits directly under an exact `toBe` on the same value |
| `Doctor.vitest.ts:227` | `not.toContain('/api personal')` | pins a subcommand the grammar dropped |
| `AgentWorkspaceWorkPlanState.vitest.ts:13` | `a retired todo/plan snapshot` row | schema became `z.looseObject`, so it now fails for the same reason as the row above it |

## Why typecheck did not catch these

Two distinct reasons, both worth knowing for the next sweep:

- **The `apiMode` fixtures** are unannotated object literals passed by variable reference. TypeScript's excess-property check applies to literals assigned directly to a typed target, not to a variable that is later passed — so a stale key survives indefinitely.
- **The `getUserTier` mocks** sit in `vi.mock` factory returns, which are not excess-property-checked at all.

The one *production* line here (`setupAssistantCommand.ts:80`) is a comment, so nothing could have caught it.

## Scope note

This is the narrow, high-confidence slice of a larger test-kernel survey. The broader findings — tests that were *rewritten* to keep passing when the behavior they pinned was deleted — are deliberately **not** in this PR. They need per-case judgment about whether the surviving assertion still earns its place, which is a different review than "this line is inert".

## Net elements (R6)

- Files: 9 changed (1 production comment, 8 test)
- `^[+-]export` symbols: 0
- Net LoC: **−7** (`3 insertions(+), 10 deletions(-)` vs `origin/main`)

No new tests, per AGENTS.md's zero-new-tests default.

## Consumer counts (R8)

Grepped with `--no-ignore-vcs` (a global gitignore hides tracked files from plain `rg` here — #11165):

- `isRelayError` — 1 occurrence repo-wide, the line removed here
- `apiMode` — 3 occurrences, all removed here; 0 production readers
- `getUserTier` — 2 occurrences, both mocks
- `OUTPUT_PREVIEW_MAX`, `truncateHeaderSummary` — 0 occurrences (already gone; their assertions are what remained)

## Checks

`typecheck` ✅ (exit 0) · `lint` ✅ (exit 0) · `format` ✅

Targeted suites ✅ — all 9 touched files plus `test-kernel/settings`: **38 files, 308 tests, 0 failures**.

https://claude.ai/code/session_01JJmSeJm4wocNwc7AFMrJBU

### DIFF
--- a/packages/extension/src/commands/setup/setupAssistantCommand.ts
+++ b/packages/extension/src/commands/setup/setupAssistantCommand.ts
@@ -77,8 +77,8 @@ async function withOpenRouterFlagOn<T>(fn: () => Promise<T>): Promise<T> {
  * Pre-flight uses the shared host-neutral predicate (adapter-level checks, so
  * blank env keys do not count as credentials and then fail later as "No model
  * is available") so this host can't drift from desktop's credential gate. The
- * CLI has its own apiMode-aware policy and doesn't call this predicate — see
- * the note on `hasUsableSetupCredential`.
+ * CLI has its own policy and doesn't call this predicate — see the note on
+ * `hasUsableSetupCredential`.
  */
 export async function hasAnyUsableSetupCredential(): Promise<boolean> {
   return hasUsableSetupCredential(platform().secrets);
--- a/src/test-kernel/agent/AgentWorkspaceWorkPlanState.vitest.ts
+++ b/src/test-kernel/agent/AgentWorkspaceWorkPlanState.vitest.ts
@@ -10,7 +10,6 @@ import {
 describe('agent workspace work-plan state', () => {
   it.each([
     ['a snapshot without workPlan', {}],
-    ['a retired todo/plan snapshot', { todos: [], plan: null }],
     ['a null workPlan', { workPlan: null }],
     [
       'invalid current workPlan entries',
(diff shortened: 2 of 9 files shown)


LionSR:
Closing because the active campaign constraint forbids adding, deleting, or modifying tests. Eight of this PR’s nine changed files are test files, so its substantive scope cannot land in this campaign.

PR #11204, opened 2026-08-20
refactor: sweep the cold surface and work the recorded candidate queue

LionSR:
Third and final round, after #11189 (core `src/`) and #11193 (host packages). **37 files, +71 / −128.**

Run in two phases, answering the two open questions left by rounds 1–2: *is there anything in the areas we didn't sweep*, and *was the declined-candidate queue worth working*.

## Phase A — the cold surface (38 agents)

Rounds 1–2 covered every production file that **changed** since `v0.40.2`. Phase A took the exact complement: the **612 files nobody had touched** during those eight days of heavy refactoring — 60k LOC, concentrated in `packages/extension` (135), `packages/cli` (98), `src/shared` (83), `src/agent` (75).

The prompt reframed the question for cold code. Not "is this tidy?" but *"does this still have a reason to exist?"* — a helper whose only caller the churn deleted, a compat reader whose writer is gone, a workaround whose cause was fixed elsewhere, a consolidation that stopped halfway.

**Result: 17 files, net −32.**

## Phase B — the recorded queue (12 agents, disjoint write scopes)

Rounds 1–2 declined 677 candidates, mostly because the fix needed a file outside the agent's own slice. I bucketed 372 of them by owning area into twelve non-overlapping write scopes so they could run concurrently without collision.

Agents were told explicitly that **a note is a lead, not evidence** — most entries are honest rejections, and the concrete ones were hours old with two PRs merged since. Re-verification was mandatory. **Twelve notes came back refuted or already fixed**, including a `preserveMarkdownPrefix` param that looked unused only because the original grep missed a multi-line call site, and a `workspaceFromSnapshot` "pass-through" that is a deliberately narrower entry point.

**Result: 20 files, net −25.**

## What it found

| Category | Count | Notes |
| --- | --- | --- |
| Dead exports / types | 18 | no consumer outside the declaring file |
| Pass-through inline | 7 | `attachRunProgressRenderer`, `setProviderHint`, two `InstructionPanel` helpers, and `packCommands`' `runRunDir` identity wrapper — whose sibling `cleanCommands` already passed the function directly, proving the wrapper was never needed |
| Dead params / options | 3 | see below |
| Speculative generality | 1 | `OAuthFormEndpoint.formHeaders`, set by neither of its two implementors |
| Stale comments | 2 | comments naming machinery a prior PR deleted |

The three dead options are the most interesting, because each had a plausible-looking reason to exist:

- `Select.onBoundaryEscape` — no caller ever supplied it, so `props.onBoundaryEscape?.(direction)` was always a no-op. Focus escape is routed through Tab at the App level instead, which the adjacent comment (also corrected) already said.
- `openFileInEditor`'s `column` — all three call sites omit it, so the cursor column was always 0.
- `DesktopCommandPaletteOptions.canOpen` — one caller, passing `() => true`. Its documented purpose (suppress the palette while a first-run walkthrough is visible) is not wired up anywhere, so the `if (canOpen?.() === false) return` guard was unreachable. **If that suppression is intended future work rather than an abandoned idea, this is the hunk to push back on.**

## Honest scope note

Net −57 from 50 agents. The notable result is that **the cold surface was cleaner than the churned surface, per file** — the opposite of the "untouched code accumulates debt" assumption. Phase A averaged one edit per 34 files; rounds 1–2 averaged one per 10. Untouched code here is untouched because it was already settled, not because it was neglected.

Taken with rounds 1–2, this closes the sweep: −92, −323, −57 across 1,621 production files. Per-file simplification of this codebase is exhausted; remaining value is in structural consolidation, not another pass.

## Post-review correction

The `texra-ai` review caught a real defect and it is reverted in 091150e: the sweep deleted `mainViewStyles`' `.view-header` rule as unreachable, on the strength of grepping `MainApp.ts`'s own template for the class name. The markup actually comes from the shared `renderViewHeader` helper (`@shared/wa/viewHeader`), which `MainApp` renders on the VS Code extension host (`!isDesktopHost`). Since `mainViewStyles` is last in that component's cascade, the rule was the only source of the header's `gap`, `padding`, and `flex-shrink` — `commonViewStyles` supplies only display/justify/align/margin. Deleting it changed the extension header's layout. I verified the reviewer's reading against the tree before reverting; the earlier "unreachable" claim in this description was wrong and is corrected above.

The review's second note is a non-defect I agree with: `taskRunStorage.ts` drops an `if (extra)` filter that was dead given `FileLocation[]` typing and the sole caller's own null-filtering. Worth remembering only if a future caller loosens that typing.

## Verified

- `typecheck` — all six projects, and **every per-package tsconfig** individually: desktop's `main`/`preload`/`renderer` and `packages/cli/tsconfig.scripts.json`. (#11193 shipped a renderer break because I had only run the top-level desktop config; that gap is closed.)
- repo-wide `npm run lint`, `prettier --check`
- full `vitest run` — **842 files, 10,305 tests passed**, 0 failures

## Net elements (R6)

- 37 production files changed, with 71 additions and 128 deletions, for a net reduction of 57 lines.
- 24 exported declarations were removed or made file-private, with no exported declarations added.
- No test files are changed.

## Consumer counts (R8)

- Every removed export was searched repo-wide and had zero consumers outside its declaring file.
- Every inlined helper had one caller, while removed callback options and fields had zero passers or readers.
- No event, subscription, or output path is deleted or re-routed.

### DIFF
--- a/packages/cli/src/chat/tui/commands/handlers/sessionCommands.ts
+++ b/packages/cli/src/chat/tui/commands/handlers/sessionCommands.ts
@@ -97,8 +97,8 @@ export async function showCliWorkPlan(
  * A tool-use run emits one ACTIVE_SKILLS snapshot at prompt-build time, so the
  * newest entry is the answer; older ones only exist across resumed lifetimes.
  * Read on demand rather than folded per delta — `/status` is a one-shot
- * command, and the projected `ConversationEntry[]` deliberately carries no row
- * for this message type (`projectTranscriptRow`), so the raw log is the source.
+ * command, and the transcript projection (`projectTranscriptRow`) deliberately
+ * carries no row for this message type, so the raw log is the source.
  */
 function activeSkillNamesFor(streamId: string | undefined): readonly string[] {
   if (streamId === undefined) return [];
--- a/packages/cli/src/runtime/cliPresentationHost.ts
+++ b/packages/cli/src/runtime/cliPresentationHost.ts
@@ -19,10 +19,7 @@ import {
   type Logger,
   type LogSink,
 } from './logSinks';
-import {
-  attachRunProgressRenderer,
-  createRunProgressRenderer,
-} from './runProgressRenderer';
+import { createRunProgressRenderer } from './runProgressRenderer';
 import { missingAgentMessage } from './agents';
 import type { CliContext } from './cliContext';
 
@@ -139,7 +136,7 @@ export function createCliRuntimeHost(context: CliContext): CliRuntimeHost {
 
   return {
     attachRunProgressRenderer: (session) =>
-      attachRunProgressRenderer(session, runProgress),
+      runProgress ? runProgress.attach(session) : () => undefined,
     prepareInteractivePrompt: () => runProgress?.preserve(),
     emitApprovalBypassState({ streamId, kind, bypassActive }) {
       if (closed || context.outputFormat !== 'ndjson') return;
--- a/packages/cli/src/runtime/runProgressRenderer.ts
+++ b/packages/cli/src/runtime/runProgressRenderer.ts
@@ -98,13 +98,6 @@ export function createRunProgressRenderer(
   });
 }
 
-export function attachRunProgressRenderer(
-  session: SessionHandle,
-  renderer: RunProgressRenderer | undefined,
-): () => void {
-  return renderer ? renderer.attach(session) : () => undefined;
-}
-
 /**
  * Everything the live line needs that no shared record carries: the run's
  * `AgentConfig` never lands on `SessionState` (the summary mirror keeps only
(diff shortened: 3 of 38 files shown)


LionSR:
Confirmed the finding and reverted it in 091150e.

I checked the reading against the tree before acting: `renderViewHeader` in `@shared/wa/viewHeader:43` emits `<div class="view-header">`, `MainApp.renderViewHeader()` returns it whenever `!this.isDesktopHost`, and `MainApp.styles` lists `mainViewStyles` last — so that rule was the only source of `gap`, `padding`, and `flex-shrink`, with `commonViewStyles` contributing only display/justify-content/align-items/margin-bottom. Deleting it would have changed the extension header's layout. The agent's justification grepped `MainApp.ts` for the literal class name and missed that the markup comes from the shared helper.

The PR description is corrected too — the table row claiming an "unreachable `.view-header` rule" is gone, and the totals are restated as 37 files / +71 −128.

On the `taskRunStorage.ts` note: agreed, non-defect. The `if (extra)` filter was dead given `linkFiles: FileLocation[]` and `collectRunSupportFiles` already filtering nulls. Flagging it as a typing-dependency rather than a bug is the right call.

PR #11213, opened 2026-08-20
fix(desktop): close refresh coalescing gap

### LINKED ISSUE #11207: Guard refresh() coalescing loop against a lost trailing pass
## Context

Follow-up from #11206 (fix(desktop): refresh clean editor buffers).

Review (texra-ai[bot], low-severity) flagged that `refresh()`'s coalescing loop has a microtask-shaped lost-update window: `refreshQueued` is read for the last time in the `while` condition, and the `.finally` that clears `refreshPromise` runs one microtask later. A `refresh()` call landing in that gap sets `refreshQueued = true` with no trailing pass left to consume it, silently dropping a refresh request.

Not reachable today — every current `refresh()` call site is a message/click/startup macrotask, so the `.finally` microtask always completes first — but the invariant is implicit and fragile. The PR author explicitly deferred the fix in review, noting the gap isn't reachable from current call sites and leaving the simpler coalescing loop unchanged.

## Details

Harden `refresh()` in `packages/desktop/src/renderer/editorPane.ts` so a `refresh()` invoked from a promise continuation (a future caller) cannot land in the microtask gap between the `while` loop's last `refreshQueued` check and the `.finally` clearing `refreshPromise`. For example, clear `refreshPromise` synchronously before the queued-check, or fold the queued-check and promise-clear into one atomic step.

## References

- PR: #11206
- Review comment: https://github.com/LionSR/TeXRA/pull/11206#issuecomment-5350201316 (texra-ai[bot], "Low" item #2)

LionSR:
## Summary

- clear the editor refresh owner inside the coalescing coroutine
- make the final queued check and owner cleanup atomic within one JavaScript job
- preserve shared-promise and failure cleanup behavior

Closes #11207.

## Validation

- Prettier
- ESLint
- `npm run typecheck:desktop`
- focused desktop control and editor-tree tests: 16 passed
- independent code review found no issues
- `git diff --check`

No tests were modified or added.

### DIFF
--- a/packages/desktop/src/renderer/editorPane.ts
+++ b/packages/desktop/src/renderer/editorPane.ts
@@ -497,13 +497,15 @@ export function createEditorPane(callbacks: EditorPaneCallbacks): EditorPane {
       return refreshPromise;
     }
     refreshPromise = (async () => {
-      do {
-        refreshQueued = false;
-        await refreshOnce();
-      } while (refreshQueued);
-    })().finally(() => {
-      refreshPromise = undefined;
-    });
+      try {
+        do {
+          refreshQueued = false;
+          await refreshOnce();
+        } while (refreshQueued);
+      } finally {
+        refreshPromise = undefined;
+      }
+    })();
     return refreshPromise;
   }
 


LionSR:
The requested regression test is deliberately not added because this issue-resolution run has a strict constraint against adding, deleting, or modifying any test. The production fix remains independently reviewed, typechecked, linted, and covered by the unchanged focused suite. The change is limited to moving cleanup into the coroutine, with no test-contract adjustment.

PR #11222, opened 2026-08-20
fix(runtime): finalize failed child launches

### LINKED ISSUE #11099: Have the launch guard own ChildStream creation (avoid stuck-RUNNING tabs)
## Context

Follow-up from #11073 (refactor(tools): run background bash through childRunLoop).

`runWithOwnedExecutionLeaseLaunchGuard` releases the lease and nothing else (`executionLease.ts:707-716`), so a `ChildStream` created inside `buildLaunch` is abandoned un-finalized if anything after its creation throws, leaving a stuck-RUNNING tab. `launchAgentCliSession:243-259` handles this case and is the more correct of the two. Reachable through `WorkflowScriptTool`'s `createRehydratedChildStream`.

## Details

Better shape than a catch: have the guard own stream creation so the creation→handoff window stops existing.

## References

- PR: #11073
- Deferred in the PR body's "Scheduled refactoring" section, item 2.

LionSR:
## Summary

- split optional child-stream creation from stream-dependent launch setup
- keep the created stream inside the owned-execution launch guard until loop handoff succeeds
- finalize failed pre-handoff streams with their flow record deleted
- aggregate setup and finalization failures without changing native streamless launches

Closes #11099.

## Validation

- Prettier
- ESLint
- `npm run typecheck:workspace`
- `npm run typecheck:cli`
- `npm run typecheck:test-kernel`
- focused unchanged child-run, delivery, Bash, and delegation tests: 65 passed
- independent lifecycle review found no correctness issues after one obsolete return property was removed
- simplification review reduced the input type without runtime changes
- `git diff --check`

No tests were modified or added.

### DIFF
--- a/src/tools/bash.ts
+++ b/src/tools/bash.ts
@@ -537,16 +537,16 @@ export class BashTool extends defineTool({
         // A background shell is an external process on no model budget, like
         // the agent-CLI children (see the child-run concurrency budget note).
         budgeted: false,
-        buildLaunch: async () => {
-          const childStream = createChildStream(executionId, parentStreamId, {
+        createChildStream: () =>
+          createChildStream(executionId, parentStreamId, {
             streamPrefix: BASH_CHILD_STREAM_PREFIX,
             run: { kind: 'process', tool: 'bash' },
             userFollowUpSupport: USER_FOLLOW_UP_SUPPORT.UNSUPPORTED,
             description: command,
             config: syntheticConfig,
-          });
+          }),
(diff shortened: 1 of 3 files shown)


LionSR:
Addressed the auto-close finding in ebcb78f763. Once launch setup returns a strategy, the guard retains its `autoCloseChildStream` policy and forwards `autoClose: true` if loop handoff then fails. Focused Bash and child-loop tests pass unchanged (52 tests).

PR #11303, opened 2026-08-23
fix(follow-up): keep the draft until the session admits it

LionSR:
## Summary

Step 1 (D5) of `docs/proposals/2026-08-23-single-owner-sessions.md`, added in this PR.

A follow-up send cleared the composer synchronously on all three hosts while the real outcome (`sent` / `queued` / `no_session`) was decided later on a fire-and-forget path with no channel back to the view. A refused send destroyed the typed text; the only notice was a toast on a different surface.

- `submitFollowUp` gains an `onAdmitted` hook that fires at admission, before any recovery resume, so the ack never waits on a model turn.
- One shared path (`submitProgressFollowUp`) owns waiting repair, admission, the composer ack, the queued-follow-ups refresh, and outcome presentation for extension and desktop. Both hosts now supply ports only; the desktop's detached copy and the extension command's duplicate toast logic are gone.
- New outbound `FOLLOW_UP_RESULT` message. The webview marks the draft in-flight on send (Send busy, re-entrancy guarded), clears text and image chips only on an accepted ack, and on refusal keeps both and refocuses.
- CLI: a refused follow-up restores the submitted line into the InputBar via a draft-restore signal and posts a sticky notice.

## Verified

- `npm run typecheck` clean (workspace, test-kernel, agent, cli, trace-viewer, desktop)
- `npm run check:dead-code-ratchet` OK
- eslint + prettier on changed files
- vitest: `ProgressViewCommandHandlers`, `ProgressViewRunCommands`, `StreamMetaState`, `ToolUseFollowUp`, `FollowUpCommandWaitingRepair` (107 passing). The waiting-repair test's "first send rejects" expectation was changed: a failed repair read is now reported to the user and the draft handed back, not thrown.

Not live-tested in a running extension/desktop/TUI.

### DIFF
--- a/config/ratchets/host-agent-import-baseline.json
+++ b/config/ratchets/host-agent-import-baseline.json
@@ -13,7 +13,6 @@
     ],
     "desktop": [
       "@agent/core/state/executionRequests",
-      "@agent/followUp",
       "@agent/index",
       "@agent/runtime",
       "@agent/storage",
@@ -22,7 +21,6 @@
     "extension": [
       "@agent/core/state/executionRequests",
       "@agent/features",
-      "@agent/followUp",
       "@agent/implementations/agentCreator/agentCreatorFlow",
       "@agent/index",
       "@agent/review",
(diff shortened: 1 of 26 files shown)


LionSR:
stack the PR

LionSR:
Implement the rest of the Proposal

PR #11334, opened 2026-08-23
refactor(runtime): one resume path, one FK index, fewer vocabularies

LionSR:
## Summary

Second deep-clean pass over the session ownership area, from a verified audit of what the seven merged campaign PRs left behind. Production net **−572 LoC** (40 files, +943/−1515 including tests).

- **One resume path.** There were eleven resume entry points (`resumeToolUseFromResumeData`, `resumeToolUseTurn`, `resumeQueuedToolUseFromResumeData`, `resolveAndResumeStream`, `resumeStreamWithRecovery`, four host adapters, `texra resume`, the native child loop) over one operation. They collapse into `resumeRun(executionId, options)` on the execution's lane. Deleted with them: `ResumeStateResolution`, `UnresolvedResumeState`, `ResumeFailureDescription`, `describeResumeStateResolution`, `describeResumeFailure`, `resolveResumeStateFromSnapshots`, `reportResumeStateResolution`, the eight-port `ResumeStreamPorts`, and the three differently worded "cannot be resumed" strings. Hosts now call one function and word its failure with the `FollowUpFailureReason` vocabulary they already use. The admission refusal for a live stream and the desktop authoritative-stream disk check are kept; the child loop keeps the unlaned turn.
- **One stream→execution index.** `meta.streamId` is awaited inside `registerExecution` before a run can emit, so the two derived mirrors (`streamData/<id>/meta.json.executionId`, the `streamLogSummaries` mirror) and the three "the mirror arm is load-bearing" fallback ladders are gone, replaced by one index built from `listExecutionStreamReferences()`. Derived tier, so discard and rebuild; no migration.
- **Fewer vocabularies.** `RESUMABILITY_CAUSE` 7→3 kinds (`classifyRunFacts` and `throwIfResumeStorageUnreadable` vanish with it, and the user-facing cause is now worded rather than an enum slug); `FollowUpSubmission` 8→5 and `QueueEntry.lifecycle` deleted (it was `owner === undefined`); `FinalizeExecutionResult.stage` and the three finalize tails collapse into one `finalizeRun`; `RUN_OUTCOME_PROJECTION` + `projectRunOutcome` + `RunOutcomeProjection` become a three-case switch; `applyExecutionOutcome` folds into its one caller; one `ChildRunOutcome` instead of two byte-identical unions; `isCurrentGeneration` and the `streamStatusDisplayKey` export dropped.
- The CLI reaches `ExecutionLeaseActiveError` through the `@agent/storage` barrel rather than a new deep import; the host mock baseline shrinks 34→32 sites; the knip baseline shrinks 429→425.

**Kept, with evidence, against the audit's own suggestion:** the two child finalization arms (their outcome precedence differs on one real cell, and collapsing either way regresses the other), `result.outcome` as a required field (it protects a producer-side FAILED downgrade that unconditional stamping would erase), `ExecutionLeaseSchema` and `hasLiveOwner` (not zero-caller), plus `runWithInactiveExecutionLease`, `admittedDeliveryIds`, the `terminalized` set, `settleLiveSessionExecutions` and `computeStartupSeedSet`.

## Verified

- `npm run typecheck` clean
- `npx vitest run` over agent, controllers, progressView, desktop, cli, tools, architecture, shared, platform, transcript: **8560 passed, 0 failed**, 5 skipped (658 files)
- eslint, prettier, `check:dead-code-ratchet` (425 vs 425), `check:browser-safe-utils` OK; all 17 architecture ratchets green with no baseline widened

Not live-tested in a running host.


LionSR:
Check the surface areas. Do more like this. Warn if the codebase has become like this

LionSR:
cli/extension/desktop should all use the same path

PR #11364, opened 2026-08-24
fix: default Google background responses off

LionSR:
## Summary

Google background responses now have their own setting and default to off. Gemini workflow generations stay on the live request unless you turn Background responses on in the Google provider settings.

OpenAI's background toggle is unchanged and no longer shares a control with Google.

## Issue acceptance gates

No tracking issue. Gate: a fresh install shows Google Background responses off, and an unset config does not send `background: true`.

## Single source of truth

`texra.model.useGoogleBackgroundResponses` in `coreSettings.ts` / `stateSettings.ts`. `ModelHandlerGoogleInteractions.useBackgroundMode` is the only runtime reader.

No new abstraction. The Google Models-tab row no longer reuses OpenAI's `useBackgroundResponses` key.

## Net elements (R6)

n/a — bug fix, not a refactor/simplify/extract PR.

## Consumer counts (R8)

n/a — no emit path or existing export was deleted. Google stopped reading `texra.model.useBackgroundResponses`; OpenAI remains the sole reader of that key.

## Host impact

Extension: Google provider card shows Background responses off by default.

Desktop: same catalog-driven toggle.

CLI: same config key, default off.

## Scheduled refactoring

None.

## Validation

- `npx vitest run --config vitest.config.mjs src/test-kernel/agent/modelHandlers/GoogleInteractionsBackground.vitest.ts src/test-kernel/controllers/SettingsProfileController.vitest.ts src/test-kernel/shared/stateSettings.vitest.ts` (67 passed, 4 live tests skipped)
- ESLint on the changed TypeScript files
- `npx tsc --noEmit -p tsconfig.json`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Shared (all surfaces)
+
+#### Bug Fixes
+
+- **Google background responses are off by default** — Gemini workflow
+  generations stay on the live request unless you turn them on in the Google
+  provider settings. OpenAI's background setting is unchanged.
+
 ## [0.40.4] - 2026-08-23
 
 ### Shared (all surfaces)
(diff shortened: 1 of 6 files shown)


LionSR:
Need to trim

LionSR:
Trimmed: restored the original gate comment, dropped the extra catalog/profile tests, shortened the changelog, and gave the Google catalog row its own title (`Google background responses`) so a failed write is not indistinguishable from OpenAI's toggle. Rebased onto current `main`.

PR #11373, opened 2026-08-24
chore(maintenance): preserve unpublished multi-agent stack

LionSR:
## Preservation status

This is a historical preservation PR. It must not be merged into `main` or another product branch in its present form.

The source checkout was an unpublished maintenance line from 13 August 2026 with no common commit ancestry with the present `main`. Comparing its tip directly with `main` would therefore produce a meaningless 2,084-file diff. This PR instead anchors the exact tree immediately before the unpublished tail and presents only the sixteen local commits that followed it: 34 files, 770 insertions, and 393 deletions.

The synthetic base has the same tree as historical commit `66796d1fd5a9ee22e62c0a811a1d2f04fb78e9b1`. That historical commit has the same stable patch ID as current-main commit `56d649559f14796a75923d9a61775b6533c18f52` from #10118. The synthetic commit exists only to avoid publishing the private email recorded on the old commit. The exact original head is also preserved at `codex/texra-maintenance-main`.

## Architectural inventory

1. **Roster and team authority.** Centralize built-in team roots, keep roster reads side-effect-free, separate default and workspace roster writes, require canonical roster resolution, and remove the unused fallback surface.
2. **Delegation and progress presentation.** Report both agent catalogs and render workflow/background tasks with the native task controls.
3. **Settings ownership.** Derive team-role availability from delegation capability, confirm the active team from the roster, and separate team-card controls.
4. **CLI presentation.** Name concurrently running subagents in progress output.
5. **Research record.** Preserve the multi-agent mathematics trial audit that explains the observed workflow behavior.

## Suggested extraction order

A future current-main adaptation should first audit the roster and team authority model, then extract the delegation catalog change, then the two progress-view changes, followed by settings behavior and CLI presentation. The audit document should be moved only with the behavior it still describes. Each group needs a fresh semantic comparison because the corresponding current-main modules have changed substantially since this stack was written.

## Representation audit

- The sixteen commit subjects have no corresponding open or closed PR title.
- Their stable patch IDs have no match in current `origin/main` or any fetched `origin` branch from the relevant period.
- The rewritten preservation head has the same Git tree as the original local head.
- The base-to-head binary diff is byte-identical to the historical sixteen-commit diff.

## Validation state

- The historical checkout was clean.
- `git diff --check` passes for the preserved range.
- No dependency installation or current test run was performed. The checkout contained only stale ignored build products from 13 August, which are not treated as validation evidence.
- This branch is not claimed to compile against current `main` and is not merge-ready.

## Scope exclusions

All earlier historical commits are excluded from this PR. They belong to the separately represented maintenance history; the chosen anchor makes only the unpublished tail reviewable.

### DIFF
--- a/docs/dev/audits/2026-08-13-multi-agent-mathematics-trials.md
+++ b/docs/dev/audits/2026-08-13-multi-agent-mathematics-trials.md
@@ -197,14 +197,16 @@ the exact remainder checks.
 
 - Team resolution, delegation, synthesis, and recovery all completed in real
   mathematical runs without weakening the mathematical acceptance criteria.
-- The compact progress status remains legible during long runs, but concurrent
-  ownership is partially hidden behind `+N`, and nested workflow structure is
-  represented by only its outer task label.
+- The compact progress status remained legible during long runs, but initially
+  hid the second concurrent owner behind `+N`. The follow-up formatter change
+  now retains two names and counts only further children. Nested workflow
+  structure is still represented by its outer task label.
 - The settings interface corrections made during the same audit are captured
   in commits `cc407df910`, `98edb7fa98`, and `b0de5a63cc`: delegation roles now
   come from backend capability data, the roster alone owns active-team state,
   and apply/delete actions use separate native controls.
 - No new recovery mechanism is justified by these trials. The intentional
   missing-agent failure was isolated and recovered by an ordinary subsequent
-  delegation. The remaining status limitation is an information-presentation
-  question, not evidence of incorrect lifecycle ownership.
+  delegation. The remaining nested-status limitation is an
+  information-presentation question, not evidence of incorrect lifecycle
+  ownership.
--- a/docs/dev/audits/2026-08-13-multi-agent-mathematics-trials.md
+++ b/docs/dev/audits/2026-08-13-multi-agent-mathematics-trials.md
@@ -1,0 +1,210 @@
+# Multi-agent mathematics trials, 2026-08-13
+
+This note records direct trials of TeXRA's CLI team runner at commit
+`b0de5a63cc`. It is intended to make the mathematical and product observations
+reproducible without access to the development conversation.
+
+## Environment and command form
+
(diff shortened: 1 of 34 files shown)


LionSR:
Closing: the still-applicable parts are extracted in #11412, and the rest is superseded on current `main`.

**Extracted into #11412**

- *Roster owns the active team.* The `@state()` in `MultiAgentTab` was optimistic, so a cancelled or blocked apply left the card badged `Active` over an unchanged roster, and a freshly opened Settings badged nothing. Now `getActiveTeamId()` → `activePresetId` on the wire → `@property`.
- *Delegation reports both catalogs.* `requireWorkflowOrToolUseAgent` rethrew the workflow-only failure.
- *Roster and team authority, the parts that survived.* `commitPresetResolution` → `commitPreset` with `commitTeamRoster` deleted, `resolveAgent` required, the legacy-selection repair serialized behind the write mutex, and the two no-op `INHERITED_AGENT_ROSTER` rewrites dropped.

**Already on `main` independently**

`fallbackTeamId` is gone from `AgentRosterControllerDeps`, and `MultiAgentTab.isOrchestratorAgent` is already capability-based rather than a name heuristic.

**Superseded**

The CLI two-name change conflicts with the description budget `formatActiveChildren` grew in bea8038680 / bf2c019411 / bfc24c6ef6: `main` now picks the RUNNING child and spends the width on its task description. The underlying observation from the audit note, that a second concurrent owner hides behind `+N`, still holds and needs a fresh design against that budget.

**Not taken**

The `div role="button"` → `<button>` conversions in `BackgroundTasksPanel` and `workflowCallFormatter`.

The audit note `2026-08-13-multi-agent-mathematics-trials.md` is a real record of four CLI team runs with execution ids and expected checks, but its Product conclusions cite commits that exist only on the preservation branches and describe the two-name formatter as shipped, so it would need editing before it could land.

The tree stays reachable at `preservation/texra-maintenance-2026-08-13` and `codex/texra-maintenance-main` regardless of this PR's state.

PR #11377, opened 2026-08-25
refactor: dedup SDK probe narrative and CLI inline-flag helper

LionSR:
## Summary

A tech-debt audit fanned five subagents across the largest production areas (`src/agent`, `src/tools`, `src/shared`+`src/controllers`, `packages/cli`, `packages/extension`, ~220k LoC). The strongest and most consistent signal was that **the codebase is already well-factored** — the meaningful shared abstractions exist and the classic debt markers are largely absent. On close inspection most of the audit's headline "duplication" findings did **not** survive scrutiny: they were either already collapsed behind an existing helper, or an intentional, heavily-documented divergence in delicate code.

This PR lands the changes that *did* survive as genuine, behavior-preserving improvements, all fully verified. It is deliberately small — see "Findings examined but not changed" below for why the larger audit items were not force-fit into this PR.

**Change 1 — `src/tools/externalToolDefs.ts`:** the Codex and Claude Code dashboard entries each spelled out the same `import SDK → classify missing-package → resolve native binary → format ready-line` sequence in their `detailCheck` bodies, kept consistent only by hand. Extracted `probeSdkBinaryStatus` so that policy lives in one place, completing the pattern `probeSdkBinaryAvailable` already established for the boolean `check`. Each entry still owns only its final "ready" line (and Claude keeps its extra auth-detection tail).

**Change 2 — `packages/cli/src/commands/_helpers/dispatch.ts`:** `knownGlobalFlagTokenCount` inlined its own copy of the "strip `=value` from a flag token" logic that `inlineFlagBase` already expressed lower in the file. Hoisted `inlineFlagBase` above its first use and call it from both sites (unconditionally — it already returns the token unchanged when there is no `=`).

## Issue acceptance gates

No linked issue — this originates from a scheduled tech-debt audit. All changes are behavior-preserving refactors with no functional acceptance gate; the gate is "checks stay green" (see Validation).

## Single source of truth

- The SDK-backed CLI **probe narrative** (missing-package classification, WSL install-hint appending, import-then-binary ordering) is now owned by `probeSdkBinaryStatus` in `externalToolDefs.ts`, alongside the existing `probeSdkBinaryAvailable` that owns the boolean form.
- The **inline flag-base split** is now owned solely by `inlineFlagBase` in `dispatch.ts`.

- Removed: the duplicated import→binary→ready sequence across the two `detailCheck` bodies; the second inline copy of the `=`-split in `knownGlobalFlagTokenCount`, plus the redundant ternary that re-tested the guard `inlineFlagBase` owns.
- New abstraction: `probeSdkBinaryStatus` — owns the shared probe policy for SDK-backed CLI integrations; not a pass-through (it encodes the missing-package/WSL/ordering decisions). Returns a discriminated `{ ok }` result so each caller keeps its own ready-line and (for Claude) auth tail. Two current callers, and the code anticipates additional providers.
- `inlineFlagBase` was moved, not newly introduced.

## Net elements (R6)

Honest accounting — this is a **single-source-of-truth consolidation, not a line reduction**:

- Files changed: 2 (`+74 / −47`, net **+27 LoC**).
- Exported symbols: **0** added/removed (`^[+-]export` over the diff is empty).
- New constructs: `probeSdkBinaryStatus` (file-local fn) + `SdkBinaryStatus` (file-local type) = **+2**; `inlineFlagBase` net 0 (relocated). Deleted constructs: 0.
- The `externalToolDefs` change is net line-positive because the parameterized config object is more verbose than the ~15 duplicated lines it replaces; the payoff is removing drift risk between the two providers' probe narratives and extending it to future providers, not LoC. The `dispatch` change is roughly line-neutral and removes two literal duplications.

## Consumer counts (R8)

n/a — no export or emit path is deleted or re-routed. Both new symbols are file-local; `probeSdkBinaryAvailable` and the `detailCheck`/`check` surfaces are unchanged. `inlineFlagBase` stays file-local within `dispatch.ts`.

## Host impact

Extension: `externalToolDefs.ts` feeds the external-tool dashboard availability display — `detailCheck` output strings are byte-for-byte unchanged, so no visible change.

Desktop: same as Extension (shares the host-agnostic `src/tools` surface).

CLI: `dispatch.ts` global-flag routing/usage behavior unchanged; covered by `CliRootArgs.vitest.ts` (all passing).

## Scheduled refactoring

The audit's larger findings were examined and deliberately **not** bundled here. Each would be its own focused, separately-verified PR if wanted:

- **Twin stateful server-response handlers** (`modelHandlerOpenAIResponse` 2,958 LoC + `modelHandlerGoogleInteractions` 2,268 LoC), kept in sync by "// Mirrors OpenAI…" comments — genuine debt, but **L / high blast radius** across two live providers (streaming + resume) that can't be verified without provider integration. Not safe to bundle.
- **`htmlMarkdownNormalize.ts` regex sieve** and the **reflection output-recovery heuristics** — correctness-sensitive; consolidation risks behavior regressions without runtime measurement of which heuristics fire.
- **`SessionFactApplier` removal settlement** and **`PRPollingSource.pollOne` seed/diff split** — examined closely and found to be *intentional, documented divergence*, not accidental duplication (fact-path vs command-path diverge on the `undefined`/`superseded` outcomes; the polling seed branch seeds terminal markers without emitting while the diff branch uses entirely different machinery). Collapsing them for ~10 LoC would risk delicate protocols against the maintainers' clear intent.
- **Extension notification ports** and **`LaTeXTab` config-row scaffold** — the meaningful shared helpers (`showInfo/showWarning/showError` fields, `renderConfigRow`, `renderSettingStatusIcon`, `renderResetButton`) already exist; the residual is trivial prelude or genuinely-divergent (`error` vs `error instanceof Error ? … : undefined`).

## Validation

- `npm run typecheck` — passes (full workspace + test-kernel + agent + cli + trace-viewer + desktop; TypeScript 6 via `corepack pnpm install`).
- `pnpm run format:check` (Prettier) — passes on both changed files.
- `npx eslint` on both changed files — clean.
- `npx vitest run src/test-kernel/tools/externalToolDefs.vitest.ts src/test-kernel/cli/CliRootArgs.vitest.ts` — passing (115, then 110 after the CLI follow-up).

### DIFF
--- a/packages/cli/src/commands/_helpers/dispatch.ts
+++ b/packages/cli/src/commands/_helpers/dispatch.ts
@@ -70,7 +70,7 @@ function knownGlobalFlagTokenCount(
   }
 
   const inline = arg.includes('=');
-  const baseFlag = inline ? inlineFlagBase(arg) : arg;
+  const baseFlag = inlineFlagBase(arg);
   if (GLOBAL_BOOL_FLAGS.has(baseFlag)) {
     return 1;
   }
--- a/packages/cli/src/commands/_helpers/dispatch.ts
+++ b/packages/cli/src/commands/_helpers/dispatch.ts
@@ -49,6 +49,12 @@ interface LeadingGlobalFlags {
   readonly stoppedOnUnknownFlag: boolean;
 }
 
+/** Strip a trailing `=value` from a flag token, leaving the bare flag. */
+function inlineFlagBase(token: string): string {
+  const equalsIndex = token.indexOf('=');
+  return equalsIndex === -1 ? token : token.slice(0, equalsIndex);
+}
+
 /**
  * How many tokens the known global flag at `index` consumes, or `undefined`
  * when the token is not a recognized global flag. Value flags consume two
@@ -64,7 +70,7 @@ function knownGlobalFlagTokenCount(
   }
 
   const inline = arg.includes('=');
-  const baseFlag = inline ? arg.slice(0, arg.indexOf('=')) : arg;
+  const baseFlag = inline ? inlineFlagBase(arg) : arg;
   if (GLOBAL_BOOL_FLAGS.has(baseFlag)) {
     return 1;
   }
@@ -495,11 +501,6 @@ async function commandFlagSpecs(
   }
 
   return specs;
-}
-
-function inlineFlagBase(token: string): string {
-  const equalsIndex = token.indexOf('=');
-  return equalsIndex === -1 ? token : token.slice(0, equalsIndex);
 }
 
 interface FlagValidation {
--- a/src/tools/externalToolDefs.ts
+++ b/src/tools/externalToolDefs.ts
@@ -220,8 +220,7 @@ async function probeSdkBinaryAvailable(
 
 /** Resolved status of an SDK-backed CLI integration for the dashboard. */
 type SdkBinaryStatus =
-  | { ok: false; message: string }
-  | { ok: true; binaryPath: string };
+  { ok: false; message: string } | { ok: true; binaryPath: string };
 
 /**
  * Human-readable probe shared by the SDK-backed CLI integrations (Codex,
@@ -252,7 +251,10 @@ async function probeSdkBinaryStatus(config: {
 
   const binaryPath = await config.findBinary();
   if (!binaryPath) {
-    return { ok: false, message: config.binaryNotFoundMessage + wslInstallHint() };
+    return {
+      ok: false,
+      message: config.binaryNotFoundMessage + wslInstallHint(),
+    };
   }
   return { ok: true, binaryPath };
 }
--- a/src/tools/externalToolDefs.ts
+++ b/src/tools/externalToolDefs.ts
@@ -218,6 +218,45 @@ async function probeSdkBinaryAvailable(
   }
 }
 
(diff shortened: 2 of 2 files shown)


LionSR:
Thanks for the thorough review — addressed both actionable findings:

- **Formatting gate** (`externalToolDefs.ts:255`): already fixed in `158fcf4` (ran `prettier --write`), which predates the review's base SHA. `format:check` now passes locally and the re-triggered `static checks` job is running against the fix. I've also added `format:check` to the PR's Validation section — that was the gap that let it slip.
- **Residual guard** (`dispatch.ts:73`): fixed in `4b5affc` — `inlineFlagBase(arg)` is now called unconditionally, dropping the ternary that re-tested the `=` guard the helper already owns. The `inline` boolean stays for the value-flag token count. CLI typecheck + `CliRootArgs` suite green.

On the `!= null` vs `!binaryPath` residual risk: deliberately leaving it. The extraction is meant to be strictly behavior-preserving (the inlined bodies both used the falsy test), so I'd rather not fold a behavior change on the empty-string path into this PR. Agreed it's a one-liner worth doing now that the policy is centralized — happy to align on `== null` in a follow-up if you'd prefer.

---

PR #11396, opened 2026-08-25
refactor(cli): move StreamArtifactProjection to its only consumer

### LINKED ISSUE #11388: StreamArtifactProjection sits in the shared session layer but has only CLI consumers
## Evidence

`src/controllers/session/StreamArtifactProjection.ts` justifies its location in the host-agnostic layer with a docstring that is false at HEAD (`:13-17`):

> The store surface a host needs to preload and read the accumulated round-artifact/usage projection. **The CLI and the shared progress-view controllers project the same fields**, so the reader shape lives here rather than on any one host.

Every consumer is the CLI. `rg` over production for `projectStreamArtifacts` / `StreamArtifactProjection` / `StreamArtifactReader`:

- `packages/cli/src/chat/tui/state/subscribeStreamArtifacts.ts:17-20,58,139,144,193,243` — the only call site of `projectStreamArtifacts` (`:144`)
- `packages/cli/src/chat/tui/commands/handlers/sessionCommands.ts:33,60`
- `packages/cli/src/chat/tui/commands/registerBuiltins.tsx:10,276`
- `packages/cli/src/chat/tui/state/sessionSignalsAdapter.ts:86,160,172` (comments) and `cliState.ts:178` (comment)

Zero extension/desktop consumers. `LitSessionRenderer` reads the snapshot store directly (`:247`, `:351`, `:359-366`) and never imports this module — so the "shared progress-view controllers" half of the claim has no referent.

## Scope

Either:

1. Move the module to the CLI beside its only consumer (`packages/cli/src/chat/tui/state/`), or
2. Keep the location and fix the docstring to say what is actually true.

(1) is preferred — it is a wrong-layer file, and `src/controllers/` is host-neutral orchestration, not a home for single-host projections.

Also in scope, same directory, same species: `StreamInfoSource` (`src/controllers/session/streamInfoUtils.ts:7`) is exported but used only at `:13` and `:18` in its own file.

## Estimated delta

≈ −10 LoC plus a 58-line file move out of the host-agnostic layer. Elements: −3 exports from `src/controllers/session/`, −1 false docstring claim.

## Risk

Low. The move rewrites `@controllers/session/...` imports in 4 CLI files plus 2 test files. Per the B3 obligation recorded in `docs/proposals/2026-08-15-single-substrate-hosts-as-renderers.md:30-32`, check `config/ratchets/architecture-edges-baseline.json` and `host-agent-import-baseline.json` for edge deltas in the same commit — a move that deletes a host's last `@agent/*` deep import must prune the baseline or CI fails on stale headroom.

Context: `docs/proposals/2026-08-25-cli-controller-seam-audit.md` §3.

LionSR:
Closes #11388.

## Summary

`src/controllers/session/StreamArtifactProjection.ts` justified its place in the host-agnostic layer with a docstring claiming:

> The CLI **and the shared progress-view controllers** project the same fields, so the reader shape lives here rather than on any one host.

That is false at HEAD. Every consumer is the CLI:

- `packages/cli/src/chat/tui/state/subscribeStreamArtifacts.ts:144` — the only call site of `projectStreamArtifacts`
- `packages/cli/src/chat/tui/commands/handlers/sessionCommands.ts:33`
- `packages/cli/src/chat/tui/commands/registerBuiltins.tsx:10`
- tests: `src/test-kernel/cli/SlashCommandDispatch.vitest.ts`, `src/test-kernel/cli/TuiStateAndFocus.vitest.ts` — both under `test-kernel/cli/`

`LitSessionRenderer` reads `StreamSnapshotStore` directly (`:247`, `:351`, `:359-366`) and never imports this module, so there is no "shared progress-view" consumer for it to be neutral between.

Moved to `packages/cli/src/chat/tui/state/streamArtifactProjection.ts`, beside `subscribeStreamArtifacts`, and the docstring rewritten to state what is actually true. `src/controllers/` is host-neutral orchestration, not a home for one host's presentation projection.

## Single source of truth

Unchanged — `projectStreamArtifacts` remains the one canonical per-stream artifact/usage projection for the TUI, and still delegates the summing rule to `sumUsageStats` (`@shared/schemas`). Only its location moves.

No new module, type, or wrapper. One file relocated, five import specifiers repointed, one false docstring corrected.

## Net elements (R6)

6 files changed, 15 insertions / 13 deletions against `origin/main` (an earlier body said 14/13; rename detection counts the docstring hunk differently, and two follow-up commits tightened it after review). Net LoC **+1** (the corrected docstring runs one line longer). Elements: **−3 exports out of `src/controllers/session/`** (`StreamArtifactReader`, `StreamArtifactProjection`, `projectStreamArtifacts`), same three added under `packages/cli`. No class/interface/enum created or deleted.

This one is honest about direction: it is a **layering fix, not a LoC win**. The gain is that the host-agnostic layer no longer carries a single-host projection behind a docstring that misdescribed it.

## Consumer counts (R8)

- `projectStreamArtifacts`: production call sites **1 → 1** (`subscribeStreamArtifacts.ts:144`), all CLI before and after.
- `StreamArtifactReader`: production consumers **3 → 3**, all under `packages/cli/src/chat/tui/`.
- `StreamArtifactProjection`: production consumers **1 → 1**.
- Extension/desktop consumers: **0 → 0**.

## Ratchets

Per the B3 obligation (`2026-08-15-single-substrate-hosts-as-renderers.md:30-32`), both edge ratchets were run rather than assumed:

- `npx vitest run src/test-kernel/architecture` — 17 files, 106 tests, all pass. No `architecture-edges-baseline.json` or `host-agent-import-baseline.json` change is required: the moved file imports `@shared/schemas` and `@transcript`, neither of which is an `@agent/*` deep import, and the CLI already reaches both.
- `npm run check:dead-code-ratchet` — 412 vs 412 baselined, unchanged.

## Host impact

CLI: three import specifiers repointed; one now relative (`./streamArtifactProjection`) since it sits in the same directory.

Extension and desktop: none — neither imported the module.

## Validation

- `npm run typecheck` (all 6 sub-projects) — pass
- `npx eslint <changed files>` — clean (import-order auto-fixed after the move)
- `npx vitest run src/test-kernel/cli src/test-kernel/architecture` — 161 files, 2656 passed, 1 skipped
- `npm run check:dead-code-ratchet` — pass, zero net baseline change

## Deferred

The issue also listed `StreamInfoSource` (`src/controllers/session/streamInfoUtils.ts:7`) as an export used only inside its own file. It is left alone here: it has three references in `src/test-kernel/progressView/streamInfoUtils.vitest.ts`, which puts it in the "a test is a consumer" species that `2026-08-19-dead-code-gate-blind-spots.md` §1 says needs per-symbol judgement rather than a blanket un-export. Unrelated to this move.

### DIFF
--- a/packages/cli/src/chat/tui/commands/handlers/sessionCommands.ts
+++ b/packages/cli/src/chat/tui/commands/handlers/sessionCommands.ts
@@ -30,7 +30,7 @@ import { hydrateStreamArtifacts } from '@cli/chat/tui/state/subscribeStreamArtif
 import { terminalCapabilities } from '@cli/chat/tui/state/terminalCapabilities';
 import { appendLocalAssistantTranscript } from '@cli/chat/tui/state/transcript';
 import { activeStreamParentOrSelfId } from '@cli/chat/tui/state/streamViews';
-import type { StreamArtifactReader } from '@controllers/session/StreamArtifactProjection';
+import type { StreamArtifactReader } from '@cli/chat/tui/state/streamArtifactProjection';
 import { activeSubscriptionUsageRoute } from '@model/codingPlanSubscriptions';
 import { formatTexraApprovalPolicy } from '@shared/approvalPolicy';
 import { MESSAGE_TYPES } from '@shared/schemas';
--- a/packages/cli/src/chat/tui/commands/registerBuiltins.tsx
+++ b/packages/cli/src/chat/tui/commands/registerBuiltins.tsx
@@ -7,7 +7,7 @@ import {
   type CliLogoutTarget,
   parseChatLoginSlashArgs,
 } from '@cli/runtime/loginOptions';
-import type { StreamArtifactReader } from '@controllers/session/StreamArtifactProjection';
+import type { StreamArtifactReader } from '@cli/chat/tui/state/streamArtifactProjection';
 import type { ApiProvider } from '@model/apiProviders';
 import type { TexraApprovalPolicy } from '@shared/approvalPolicy';
 import { type ExecutionId } from '@shared/schemas';
--- a/packages/cli/src/chat/tui/state/streamArtifactProjection.ts
+++ b/packages/cli/src/chat/tui/state/streamArtifactProjection.ts
@@ -11,10 +11,11 @@ import {
 import type { StreamSnapshotStore } from '@transcript';
 
 /**
- * The store surface a host needs to preload and read the accumulated
- * round-artifact/usage projection. The CLI and the shared progress-view
- * controllers project the same fields, so the reader shape lives here rather
- * than on any one host.
+ * The store surface needed to preload and read the accumulated
+ * round-artifact/usage projection. The TUI is the only consumer: the
+ * progress-view renderer reads `StreamSnapshotStore` directly instead
+ * (`LitSessionRenderer.getOutputFiles`/`getRunUsage`), so this projection is
+ * CLI presentation, not shared state.
  */
 export type StreamArtifactReader = Pick<
   StreamSnapshotStore,
--- a/packages/cli/src/chat/tui/state/streamArtifactProjection.ts
+++ b/packages/cli/src/chat/tui/state/streamArtifactProjection.ts
@@ -12,10 +12,11 @@ import type { StreamSnapshotStore } from '@transcript';
 
 /**
  * The store surface needed to preload and read the accumulated
- * round-artifact/usage projection. The TUI is the only consumer: the
- * progress-view renderer reads `StreamSnapshotStore` directly instead
- * (`LitSessionRenderer.getOutputFiles`/`getRunUsage`), so this projection is
- * CLI presentation, not shared state.
+ * round-artifact/usage projection. No host-neutral module imports this
+ * projection: the progress-view renderer reads the store directly
+ * (`state.snapshots.getOutputFiles`/`getRunUsage`), and the progress-view
+ * controllers reach those same accessors through their own narrower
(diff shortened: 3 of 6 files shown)


LionSR:
Closing as superseded by #11400, not as wrong.

The question that prompted this — *why does a projection layer exist at all; why can't the hosts share the data structure?* — has an answer that invalidates the fix rather than the finding.

They can't share it because **`StreamSnapshotStore` clones on read**. `getOutputFiles`, `getMissingOutputs`, and `getCompileFailures` each return `cloneRoundIndexed(...)` (a new object plus a new array per round) and `getRunUsage` returns `new Map(...)`. Ink repaints per streaming token, so reading the store directly would re-clone every round map and re-sum usage on each repaint. That is the entire reason `projectStreamArtifacts` + `artifactProjectionMemo` exist — the memo's own comment says so and cites #10731.

So `StreamArtifactProjection` is not a layer in the wrong place. It is a **memo value shape**, and the thing in the wrong place is the defensive copy upstream of it. #11400 removes that (readonly views instead of clones), which collapses the projection entirely — and also stops both GUI hosts cloning on every render pass, which this PR did nothing about.

Merging this first would move a module that #11400 deletes: a rename in history for a file that will not exist, and a second review of the same module. That is the churn-class change checklist §14 R5 rules out.

Both findings from this PR carry forward into #11400:

- the docstring's "the CLI **and the shared progress-view controllers** project the same fields" was false, and
- the module had no host-neutral consumer.

Branch `fix/seam-cleanup-4` is left in place; if #11400 stalls on the `store-public-surface-baseline.json` reconciliation, this is still a correct standalone improvement and can be reopened.

PR #11407, opened 2026-08-25
feat(cli): compact the terminal tab title

LionSR:
## What

The terminal tab title spent most of its width on the product name and two spaced em dashes:

```
TeXRA — - — TNLean
```

The `-` in the middle was not a separator — it was the spinner. `loadingFrameAt` cycles `| / - \` at 1 Hz, so a quarter of the time the running title showed a bare hyphen indistinguishable from the punctuation around it.

Now:

| state | before | after |
| --- | --- | --- |
| running | `TeXRA — - — TNLean` | `⠋ {T}·TNLean` |
| approval | `TeXRA — Approval needed — TNLean` | `⚠ {T}·TNLean` |
| idle | `TeXRA — TNLean` | `{T}·TNLean` |

The state marker leads so it survives end-truncation in a narrow tab, the brand becomes the `{T}` mark, and the separator drops to a tight middle dot. `Approval needed` was 15 characters that pushed the project name out of view exactly when you are scanning tabs to find the one asking for something; `⚠` says the same in one column.

The title keeps its animation, but through braille dots rather than the TUI's ASCII frames — in a tab title with no adjacent label, `-` and `\` read as debris. Quarter-turn steps, since the shared clock ticks at 1 Hz and a ten-frame cycle would crawl.

## Style, not a second formatter

`formatSessionTitle` is shared with the Electron window title, where the mark alone would be cryptic and there is room for words. It now takes a `SessionTitleStyle` — brand, separator, per-state markers — with `TERMINAL_TAB_TITLE` and `NATIVE_WINDOW_TITLE` as the two instances. Desktop windows read `Approval needed TeXRA · geometry`. One formatter, one shape, host-specific vocabulary.

## Verified

- `npm run typecheck` clean
- `npm run lint` clean
- `npx vitest run src/test-kernel/cli/TerminalCleanup.vitest.ts src/test-kernel/desktop/DesktopWindowTitle.vitest.ts` — 26 passed (expectations updated in place, no new test files)

### DIFF
--- a/packages/cli/src/chat/tui/terminalTitle.ts
+++ b/packages/cli/src/chat/tui/terminalTitle.ts
@@ -5,6 +5,7 @@ import { loadingFrameAt } from '@cli/tui/ui/LoadingIndicator';
 import { subscribeToSharedTick } from '@cli/tui/useLiveNowMs';
 import {
   formatSessionTitle,
+  TERMINAL_TAB_TITLE,
   type SessionTitleState,
 } from '@shared/sessionTitle';
 import { subscribeToSignalChanges } from '@shared/signals';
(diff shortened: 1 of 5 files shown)


LionSR:
Addressed the moderate finding: `TERMINAL_TAB_TITLE` and `NATIVE_WINDOW_TITLE` are now `Object.freeze`d in 3ef6c7b162, with the nested `marker` record frozen separately — a shallow freeze would have left it mutable, which is the case the AGENTS.md rule calls out.

PR #11414, opened 2026-08-25
docs: record the 2026-08-25 simplification survey (49 verified candidates)

LionSR:
## Summary

Records one dated survey doc for the 2026-08-25 simplification sweep so the
candidates, and just as importantly the refutations, live somewhere durable
instead of in a session transcript.

Thirty-four read-only survey agents ran one disjoint area each over `src/`,
`packages/*/src`, `scripts/`, `config/`, `packages/extension/resources/`,
`prompts/`, and `supabase/functions/`, capped at four candidates apiece and
required to dedupe against `label:tech-debt` (open and closed) plus `git log`
before reporting. Every surviving candidate then went to an adversarial
verifier prompted to refute it, defaulting to refuted under uncertainty, which
re-ran the greps independently across the boundaries that are easy to miss:
`packages/extension/package.json` contributions, `commands.ts` registration,
the settings Zod schemas, agent YAML, prompts, and edge functions.

91 candidates surveyed, 50 survived, 41 refuted. After merging one cross-area
duplicate, 49 remain. Aggregate if all land: -2793 LoC, -267 elements. Every
one is a bounded deletion; none needed a design-level proposal of its own.

## Why section 3 is the important half

The 41 refutations are recorded with their reasons because most were killed by
a dated ruling the claimant had missed, not by sloppy evidence. A few examples:

- The workflow-script `'starting'` status does reach disk; it is a member of
  the strictly-parsed persisted `meta.json` snapshot.
- The `@agent/workflowScript` barrel narrowing was deliberately decided the
  other way in #10826.
- The desktop Sentry path was kept on purpose in `b8df8b98fd`.
- Deriving `contributes.menus.commandPalette` from the command catalog nets
  roughly +50 LoC, which is the §13 Refactor-LOC trap rather than a win.

Without this section the same ground gets re-surveyed and the same items get
re-proposed.

## Follow-up

Section 1 carries the PR plan: seven disjoint-lane branches grouped by path
ownership, batching the small items rather than shipping them one at a time.
Those are in flight separately; this PR is docs only.


LionSR:
Both findings addressed.

**Local absolute paths (claude[bot]).** Confirmed at 23 occurrences. They came from the verifier transcripts, which ran with absolute paths; two also carried a temporary agent worktree prefix (`.claude/worktrees/wf_*/`) that resolves for nobody. All are repo-relative now, and the count is zero.

**`Record<UsageRoute, string>` (codex P1).** Correct and worth stating precisely: a total record would have to return a string for `api-key` and legacy `relay`, and it is exactly the `undefined` those routes get from the `default` arm that routes them to `usage_logs` instead of `subscription_usage_logs`.

Checked against main: this is a defect in the doc's prose, not in shipped code. #11422 kept the switch with `default: return undefined` intact (`supabase/functions/log-usage/usageValidation.ts:65-81`), matching the verifier correction that the doc already recorded below the proposal. The proposal paragraph nonetheless read as actionable on its own, so it is now marked superseded inline with the reason.

The remaining codex comments on this PR are suggestions about candidate implementations rather than the document. Those candidates have shipped in #11416 through #11422, and three of them were adopted there: the `ensureDir`/`createDir` behavior change, the mutable `UserVariableValueSchemas` export, and the `startRunWorkspacePreparation` single-caller wrapper are all fixed in #11419.

PR #11417, opened 2026-08-25
refactor(extension): fold six settings and main-view leftovers into their owning paths

LionSR:
## Summary

Six independently verified simplification candidates in the extension host, batched into one commit because each alone is too thin to file. The largest two are dual-representation folds: the `texra.memory.enabled` toggle becomes a `stateSettings` catalog row instead of a hand-built twelve-file command vertical, and the degenerate `modelOptionsByCategory` wire shape collapses to a flat list now that model options no longer vary by agent category. The rest are dead flags, a pass-through method, an unread interface member, and eleven exports that never crossed a module boundary.

Three items span `src/shared`, `src/controllers` and `packages/desktop` as well as `packages/extension`. Both briefs called that out: the memory fold needs the wire contract and both host poster maps, and the model-options collapse needs the schema, the controller relay and both desktop senders. `SettingsSnapshotPosters` is a total `Record` over `SettingsViewSnapshot`, so the new `memory` member is compile-forced into both host maps rather than silently missing from one.

## What landed

**Item 1. Memory toggle folded into the settings catalog.** One `surfacedSetting` row for `GlobalStateKey.MEMORY_ENABLED` (`z.boolean().prefault(true)`, `sameSlot('globalState')`, `honoredBy: everyHost('src/agent/features.ts')`, `surfaces: { settingsView: 'memory' }`). `MemoryTab` renders the shared `renderStateSettingToggleRow` and writes through the generic `UPDATE_STATE_SETTING` boundary. `MemoryToggle.ts` is deleted. The frozen wire literal `UPDATE_MEMORY_ENABLED` is reused as the `memory` snapshot command rather than minting a new one, matching the `multi-agent` to `UPDATE_SUPER_YOLO_ENABLED` precedent; `SET_MEMORY_ENABLED` is gone. The row declares no `cliConfig`, so the CLI reachability guardrail does not apply and the CLI `/config` roster assertion is untouched.

This fixes a live defect recorded as G6 in `docs/proposals/2026-08-04-ssot-consolidation-part-2.md`: `src/agent/features.ts` defaults the key to `true` while the frontend signal initialized to `false`, so the Memory tab first-painted "off" for users who had memory on. The signal now seeds from the row's `.prefault()`.

**Item 2. `modelOptionsByCategory` collapsed to `modelOptions`.** All five senders wrote the same array under both keys, and the one consumer that appeared to distinguish them (`getModelOptionsForSession`) had two branches returning the same array. `computeModelOptionsData` lost its `agentCategory` argument in #9951; the split has been a no-op since. `SetModelOptionsMessageSchema` now carries `optionsData: ModelOptionData[]`, `MainViewStartupOptions.modelOptions` is flat, `loadMainViewModelOptions` and `MainViewModelOptionsByCategory` are deleted, and `sessionModelOptions$` becomes `modelOptions$`. `SessionContextSchema.modelOptions` was already flat, so the store now matches the context it feeds.

Incidental repair worth flagging: `scripts/capture-walkthrough-media.mjs` already posted `{ command: 'setModelOptions', optionsData }`, which did not validate against the by-category schema on main. That script is now correct again without being edited.

**Item 3. Instruction placeholder derived instead of re-pushed.** `instructionPlaceholder$` becomes a `Signal.Computed` over `launchTarget$`, `isSelectedAgentOrchestrator$` and `sessionType$`, the same pattern `instruction$` already uses. `refreshInstructionPlaceholder` and its eight call sites go. `ONBOARDING_PLACEHOLDERS` collapses from three string arrays (12 strings) to three strings, dropping nine entries unreachable since the rotation carrier was deleted in June 2026. Deriving also left `instructionPlaceholder$` and `isSelectedAgentOrchestrator$` with in-file readers only, so both lost their `export` rather than becoming new unused exports.

**Item 4. Two settings-view leftovers.** `LaTeXTab.configLoaded` was write-only, so the whole chain behind it goes (`latexConfigValuesLoaded`, its write in `latexSlice`, and both SettingsApp lines). `LatexSettingsHandlers.sendLatexConfigValues` re-inlined the `SettingsStores` literal its only caller already owns as a private method; both call sites now use the owning class's `sendSettingsSnapshot(webview, 'latex')` and `rebroadcastSnapshot('latex')`, and the message posted is identical.

**Item 5. `ProgressToolbarButton.disabled` deleted.** One interface member plus eleven descriptor lines. The toolbar tables have exactly one consumer, `StreamHeader`, which computes every button's disabled state from `getButtonState` and never reads the field. It is a stale initial-render hint from the pre-Lit imperative-DOM toolbar.

**Item 6. Batch.** `"onColorTheme"` dropped from `activationEvents` (not a VS Code activation event, and `onStartupFinished` already covers activation unconditionally; the package-invariants snapshot is updated in the same commit). `apiKeyCommands` deleted in favour of `EXTENSION_COMMANDS.SET_API_KEY`, which is compile-checked against the command catalog by its `satisfies Record<string, CommandId>`. Eleven types that never crossed a module boundary lost their `export`.

## What was skipped

**Item 4(c), the `AIAgentsTab` `@property` initializers.** Dropped per the verifier's authoritative correction, not by failure. Removing the six initializers is `TS2564` under `"strict": true` (strictPropertyInitialization), and the escapes (`!`, or `?` plus `?? ''` at every `renderSelectRow` call site) are net-negative relocated complexity. The premise was also wrong: the six imports are references to the catalog SSOT constants that `src/tools/codexConfig.ts` and `src/tools/claudeAgentConfig.ts` also consume, not restated literals, and defaulting every `@property` is the convention across all thirteen tabs in that directory. Item 4 therefore lands as two of three sub-items by design.

## Behavior changes

Three, all deliberate and all named in the briefs.

1. The Memory tab switch no longer greys out until the backend answers. It renders the catalog default (`true`) like the other catalog-backed toggles. This is the `DEFAULT_GIT_MARK_COMMITS` precedent and it is what fixes the first-paint "off" defect. The bespoke `memoryToggleDisabled` gate was a pre-snapshot input guard, not a masking fallback.
2. The instruction placeholder now updates on `SET_AGENT_OPTIONS` and on persisted restore, where it previously went stale until an unrelated mutator fired. Item 3 is a fix, not behavior-preserving. Blast radius is one line of grey placeholder copy. Because `sessionContext$` already depended on every input of the new computed, no additional re-render is introduced.
3. Nine unreachable placeholder strings are gone. The three survivors are index 0 of each old group, the only index any code path could produce.

Everything else is behavior-preserving. The storage key is unchanged, so `src/agent/features.ts` and every persisted value keep working; the `latex` snapshot message is byte-identical; the toolbar's rendered disabled state was already computed.

## Tests

No new tests. Three suites had sections pinning retired behavior; those sections were deleted rather than rewritten, since the surviving paths are already covered.

- `SettingsViewHost.vitest.ts`: dropped the memory-enabled legs, renamed the case to what it still asserts.
- `SettingsMemoryController.vitest.ts`: dropped the `setMemoryEnabled` case and the now-unused fixture plumbing.
- `DesktopSettingsIpc.vitest.ts`: deleted "handles desktop memory toggle messages"; the generic `UPDATE_STATE_SETTING` write path already has about ten cases in that file.
- `MainViewStartupController.vitest.ts`, `ElectronMainViewIpc.vitest.ts`, `SidebarSurfaceOwnership.vitest.ts`: mechanical shape updates for the flat `modelOptions`.
- `stateSettings.vitest.ts`: `EXPECTED_DEFAULTS` gains `texra.memory.enabled: true`. The catalog-snapshot test iterates `SETTINGS_SNAPSHOT_COMMANDS`, so `memory` is now exercised end to end.

## Net elements (R6)

Measured from `git diff --stat origin/main...simplify/L2-extension` and `grep` over the full diff.

- Files changed: 54, one of them deleted (`MemoryToggle.ts`, 52 lines)
- Lines: 135 insertions, 428 deletions, net **-293 LoC**
- `^-export`: 25. `^+export`: 2. Net **-23 exported symbols**
- class / interface / enum declaration lines: 12 removed, 10 re-added without `export`. Net **-2 declarations** (`class MemoryToggle`, and the `declare global` `HTMLElementTagNameMap` augmentation that went with it)
- 11 types demoted from exported to file-local: 10 interfaces plus the `Raced<T>` type alias. These account for the 10 re-added declaration lines and 11 of the 25 removed export lines
- The 2 added exports are replacements, not additions: `memoryEnabled` (from `trackedSignal` to `settingSignal`) and `modelOptions$` (replacing `sessionModelOptions$`)

Deleted exported symbols, 13 net after excluding the 11 demotions and 2 replacements: `apiKeyCommands`, `MainViewModelOptionsByCategory`, `loadMainViewModelOptions`, `MemoryToggle`, `memoryToggleDisabled`, `latexConfigValuesLoaded`, `refreshInstructionPlaceholder`, `instructionPlaceholder$`, `sessionModelOptions$`, `getModelOptionsForSession`, `isSelectedAgentOrchestrator$`, `UpdateMemoryEnabledMessageSchema`, `SetMemoryEnabledMessageSchema`.

Non-exported members also deleted: `ProgressToolbarButton.disabled` plus its 11 descriptor lines, `SettingsMemoryController.getMemoryEnabledMessage` / `setMemoryEnabled` and its two constructor deps, `SettingsViewHost.sendMemoryEnabled` / `setMemoryEnabled`, `MemoryHandlers.sendMemoryEnabled` / `handleSetMemoryEnabled`, `LatexSettingsHandlers.sendLatexConfigValues`, `LaTeXTab.configLoaded`, `MemoryTab.toggleDisabled`, nine `ONBOARDING_PLACEHOLDERS` strings, and the `SET_MEMORY_ENABLED` wire literal.

Added: one declarative `surfacedSetting` row (15 lines), one `SettingsViewSnapshot` member, one `SETTINGS_SNAPSHOT_COMMANDS` entry, one local `snapshotMessage('memory')` arm, two one-line poster arms, one `EXTENSION_COMMANDS.SET_API_KEY` entry.

## Consumer counts (R8)

Counts are `git grep` at this branch. "After" means source files; `docs/` hits are historical PRD text, not consumers.

**Item 1, memory vertical**

| Deleted symbol or path | Consumers before | After |
| --- | --- | --- |
| `SET_MEMORY_ENABLED` / `'setMemoryEnabled'` | 7 production, 1 test | 0 source, 1 docs mention |
| `SetMemoryEnabledMessageSchema` | 2 (declaration, inbound union) | 0 |
| `UpdateMemoryEnabledMessageSchema` (old `{ enabled }` form) | 2 (declaration, outbound union) | 0; command literal retained as the `memory` snapshot arm |
| `SettingsMemoryController.getMemoryEnabledMessage` | 2 | 0 |
| `SettingsMemoryController.setMemoryEnabled` | 1 | 0 |
| `SettingsViewHost.sendMemoryEnabled` | 2 | 0 |
| `SettingsViewHost.setMemoryEnabled` | 2 | 0 |
| `MemoryHandlers.sendMemoryEnabled` | 1 | 0 |
| `MemoryHandlers.handleSetMemoryEnabled` | 1 | 0 |
| `MemoryToggle` / `<memory-toggle>` | 1 | 0 source, 10 docs |
| `memoryToggleDisabled` | 2 | 0 |

`MEMORY_VIEW_COMMANDS` is referenced only by `src/shared/ipc.ts` and `src/shared/schemas/memoryViewMessages.ts`, so the standalone memory view is long gone and the deletion is contained. The runtime reader is unchanged: `src/agent/features.ts:12` still reads `GlobalStateKey.MEMORY_ENABLED` live with a `true` fallback, matching the new row's `.prefault(true)`.

**Item 2, model options**

| Deleted symbol | Consumers before | After |
| --- | --- | --- |
| `optionsDataByCategory` / `modelOptionsByCategory` | 5 producers, 1 consumer, 5 test sites | 0 source, 4 docs |
| `loadMainViewModelOptions` | 3 callers, 1 test mock | 0 |
| `MainViewModelOptionsByCategory` | 1 | 0 |
| `sessionModelOptions$` | 2 | 0, replaced by `modelOptions$` with 4 readers |
| `getModelOptionsForSession` | 3 | 0 |
| `AgentCategory` import in `mainView/outbound.ts` | 1 | 0, import removed |

**Item 3, placeholder**

| Deleted symbol | Consumers before | After |
| --- | --- | --- |
| `refreshInstructionPlaceholder` | 8 call sites, 3 imports | 0 |
| `instructionPlaceholder$` export | 1 cross-module | 0 exported, 1 in-file reader (`sessionContext$`, rendered by `InstructionPanel`) |
| `isSelectedAgentOrchestrator$` export | 1 cross-module | 0 exported, 2 in-file readers |
| `ONBOARDING_PLACEHOLDERS` indices above 0 | 0 reachable paths (`placeholders[0]` was the sole write) | deleted |

**Item 4, settings leftovers**

| Deleted symbol | Consumers before | After |
| --- | --- | --- |
| `LaTeXTab.configLoaded` | 0 reads, 1 write (SettingsApp binding) | 0 |
| `latexConfigValuesLoaded` | 2 | 0; repo grep for `configLoaded|config-loaded|latexConfigValuesLoaded` returns 0 |
| `LatexSettingsHandlers.sendLatexConfigValues` | 2, both inside `SettingsViewMessageHandler` | 0 |

**Item 5, toolbar**

`ProgressToolbarButton` / `TOOLBAR_BUTTONS` / `NEUTRAL_TOOLBAR` have exactly one consumer, `StreamHeader.ts`, which reads `id`, `icon`, `label`, `title`, `titleActive`, `className`, `bypassKind`, `command` and `localAction`. Readers of `btn.disabled`: 0.

**Item 6, batch**

| Deleted symbol | Consumers before | After |
| --- | --- | --- |
| `apiKeyCommands` | 4 (`setupAssistantCommand.ts` x3, `ProgressViewMessageHandler.ts`) | 0, repointed to `EXTENSION_COMMANDS.SET_API_KEY` |
| `"onColorTheme"` | 0 source; 2 file references (manifest, invariants snapshot) | both updated in this commit |
| 11 file-local types | `git grep -l` returns exactly 1 path each | 0 cross-module consumers |

Command registration is untouched: `texra.setApiKey` is registered through the catalog-driven `extensionCommandHandlers.ts`, not the deleted object, and `'texra.setApiKey'` exists at `src/shared/commands/catalog.ts:178` so the `satisfies Record<string, CommandId>` constraint holds.

## Ratchets

No baseline was widened and no baseline row needed removing.

- `knip-baseline.json`: 0 rows matched any of the 25 removed export lines. The ratchet fails only on new findings; this change only removes exports, and resolved findings are reported, not enforced.
- `architecture-edges-baseline.json` covers directed edges between top-level `src/` subsystems only, so the two new host imports of `@model/computeModelOptions` are out of scope. The same edge already existed through `optionsLoader.ts`.
- `host-agent-import-baseline.json`: no new `@agent/*` deep-import specifier, type-only included.
- `store-public-surface-baseline.json` and `shared-schemas-deep-import-baseline.json`: untouched.
- `scripts/extension-package-invariants.snapshot.json` updated in the same commit, as `MANIFEST_KEYS` includes `activationEvents`.

## Verification notes for the integrator

The lane had no `node_modules`, so nothing was typechecked, linted or run. Reviewed by tracing every changed signature to every caller at this revision. Two things worth running first:

1. `vitest run src/test-kernel/webview/SidebarSurfaceOwnership.vitest.ts`. Its `@frontend/agents/optionsLoader` mock became a `@model/computeModelOptions` mock that spreads `importOriginal`, so the real module graph is now evaluated inside that suite. `src/model` has zero `vscode` imports so this should be inert, but it is the least mechanical edit in the branch.
2. `npm run typecheck`. `UpdateMemoryEnabledMessageSchema` is now a local `snapshotMessage('memory')` const declared at line 490 and consumed in the outbound union at line 657; `snapshotMessage` is a hoisted function declaration at line 220, so declaration order is fine, but this is the one spot where ordering matters.


LionSR:
Conflict resolved and the residual-risk item from the TeXRA review is now closed.

The conflict was the one predicted in the survey doc: main removed the zero-consumer `controllers.memory` injection seam (#11421) while this branch removed the bespoke memory callbacks the catalog row replaces. The resolution keeps both deletions:

```ts
this.memoryController = new SettingsMemoryController({
  prompt: options.memoryPrompt,
});
```

On the "nothing was typechecked or run" risk: the branch has now been validated with a real toolchain.

- `npm run typecheck` passes across every package.
- `npm run lint` passes.
- `vitest` on the suites the review named plus their neighbours: 42 files, 430 tests passing, including `SidebarSurfaceOwnership.vitest.ts` specifically. The `@model/computeModelOptions` mock that spreads `importOriginal` is inert as predicted.

PR #11419, opened 2026-08-25
refactor(agent): collapse ten residual seams in the agent runtime

LionSR:
## Summary

Ten independent collapses in `src/agent/` (plus the one line of `src/tools/ExecutionsTool.ts` this lane owns): pass-through ports whose only implementers were test fakes, two hydration entry points with byte-identical bodies, a mirrored state field, a computed flag nobody read, and ten types exported without a cross-file consumer. Every deletion is behavior-preserving; the deliberate exceptions are listed under "Behavior deltas" below. Five rows leave the knip baseline.

## What landed

**AgentDirectoryService loses three injected ports.** `AgentDirectoryPathStorage`, `AbsoluteDirectoryAccess`, and `AgentDirectoryServiceLogger` had zero production importers and were wired by `createPlatformAgentDirectories` to fixed host-agnostic targets. The service now calls `GlobalStorageFS` and `AbsoluteFS` directly and builds its own `createLog` from a new `channel: string` option. `AgentDirectoryIssueReporter` stays injected: the VS Code extension really does override it. The ENOTDIR-means-absent handling that lived in the factory adapter is not relocated into `src/agent`; it is inherited from `AbsoluteFS.exists`, which already swallows ENOENT and ENOTDIR through `BaseFS.statIfExists`. `AgentDirectoryService.vitest.ts` moves onto `setupPlatform` + `createTempDirPlatform`, the pattern the neighbouring platform-directory suite already uses, keeping `RecordingIssueReporter` and asserting real on-disk directories.

**The fixed template-variable vocabulary has one owner.** `UserVariableValueSchemas` is exported and `USER_VAR_RUNTIME_TOKENS` is now `Object.keys` of it. The reverse guard is not lost: that map already carries `satisfies { [K in keyof Required<UserVars>]-?: z.ZodType<...> }`, whose `-?` requires every key and whose object-literal excess-property check rejects any key that is not in `UserVars`. The hand-written `AssertNever` alias and `_UserVarsStayInRuntimeTokens` are deleted. Token order changes; both consumers build a map or a set from the list.

**Three small collapses.** `ExecutionRegistry.waitForChange` folds into `waitForAnyChange([id], signal)` at its single caller. `isWaitingFlowResult` takes `AgentRuntimeFlowResult` instead of `unknown` and drops its hand-rolled object guard; all seven call sites already hold a typed value, and the two that can see a nullable one (`resumeRun.ts:416`, `childRunLoop.ts:1040`) guard before calling. `textConnection`'s `ConnectionResult` record collapses to a string: `bestConnectionMethod` folds into `agentResponseTextConnector` and `DEFAULT_RESULT` becomes `DEFAULT_CONNECTOR`.

**The Google client resolver returns to its handler.** `resolveGoogleClient`, `ResolveGoogleClientParams` (including its dead `sdkLabel`), and the exported `GoogleClientCache` are gone. The cache is a file-local field type on `ModelHandlerGoogleInteractions`, and `getClient` does the credential comparison, the `retryOptions`-free construction, and `rememberClientCredentialRoute` inline, in the original order. The comment explaining why passing any `retryOptions` value blinds the route gate's 429/5xx classification travels with the code.

**One workspace-snapshot hydration entry point.** `fromCanonicalSnapshot`, the private `fromParsedFields`, and the reflection flow's `workspaceFromSnapshot` collapse into `fromSnapshot`. This is runtime-identical: both public entry points already ran `AgentWorkspaceStateSnapshotSchema.parse` on the same schema. Two comments claiming a legacy migration arm that no longer exists are rewritten.

**The run's storage key stops being mirrored.** `storageKey` is declared on `ReflectionServices` and `OutputDependencies` (the services literal already spreads it from `RunReflectionFlowInput`), so `OutputState.storageKey`, `getStorageKey`, and `setActiveRun`'s idempotence guard all go. `setActiveRun` becomes `startRunWorkspacePreparation`. `checkExpectedOutputs` also drops its now-unused `state` parameter.

**Four smaller items.** `parseToolUseShared` returns `z.ZodSafeParseResult<ToolUseRunShared>` directly, deleting the hand-written result alias, the unread `changed` flag, and the `node:util` `isDeepStrictEqual` import. `processMultipleOutputs` folds into `extractFilesFromXml`, its only caller. `AgentRosterSnapshot.agents` is deleted along with its per-call `byCategory(getVisibleAgents)` resolve, which also lets the interface drop its `Entry` type parameter. Ten types with zero cross-file references lose their `export` keyword.

## What was skipped

- **Inverting `UserVars` to `z.infer` of the schema map** (the second half of the vocabulary item). Refuted and not attempted. `UserVars` stays the documented owner of the vocabulary and the mapped-type `satisfies` pin stays as the two-way guard.
- **The media half of `googleHandlerShared`.** `buildMedia`/`buildLabel` cannot be inlined the way the client half could, and open issue #11297 rewrites those failure branches. Left untouched, so the file keeps its (now media-only) docstring.
- **Inlining `startRunWorkspacePreparation` into `runReflectionFlow`.** It stays a one-statement helper because inlining would force exporting the currently private `collectRunSupportFiles`, trading one export for another.
- **Rewriting `GoogleClientRouteCache.vitest.ts`.** It existed only to drive the deleted `resolveGoogleClient` export, so it is deleted rather than rewritten. The credential-comparison logic moved verbatim into `getClient` and is no longer unit-covered; `GoogleClientRefresh.vitest.ts` overrides `getClient` and does not exercise it.

## Behavior deltas

- Custom-directory creation goes through `AbsoluteFS.ensureDir` instead of `platform().fs.createDirectory`. That is a strict superset: it additionally swallows EEXIST. Both node and VS Code providers use a recursive mkdir that never throws EEXIST for an existing directory, so nothing observable changes.
- Two debug log strings in `textConnection` changed wording because they named the deleted function. No test asserts them.
- `isWaitingFlowResult` no longer tolerates a non-object argument at runtime. Every call site was audited; the two that can hold `undefined` guard first.
- `deps.storageKey` is `undefined` in the two `as unknown as OutputDependencies` test fakes. It reaches only a debug template that those suites never execute. Previously that path threw a "storageKey is unset" error instead.
- The four migrated `extractFilesFromXml` test call sites now also run `withOutputStage`, which is inert: it passes `skip: true`, and `openStage` returns a non-emitting `SkippedStageHandle` that delegates `run` straight through.

## Net elements (R6)

Measured from `git diff --stat origin/main...simplify/L4-agent` (39 files changed, 282 insertions(+), 622 deletions(-)).

Net LoC: **-340**.

| Area | Files | +/- | Net |
| --- | --- | --- | --- |
| Production (`src/agent/**`, `src/tools/ExecutionsTool.ts`) | 31 | +212 / -461 | -249 |
| Centralized tests (`src/test-kernel/**`) | 7 | +70 / -131 | -61 |
| Ratchet (`config/ratchets/knip-baseline.json`) | 1 | +0 / -30 | -30 |

One file deleted (`src/test-kernel/agent/modelHandlers/GoogleClientRouteCache.vitest.ts`), none added.

Export symbols: `^[+-]export` lines are +6 / -24. Six removals are same-symbol re-adds from a rename or a reflow (`processMultipleOutputs`→`extractFilesFromXml`, `setActiveRun`→`startRunWorkspacePreparation`, plus reformatted `parseToolUseShared`, `USER_VAR_RUNTIME_TOKENS`, `agentResponseTextConnector`, `extractFilesFromXml`). Real net exported symbols: **19 removed, 1 added** (`UserVariableValueSchemas`, consumed by `src/agent/prompt/userVars.ts` in the same commit) = **-18**.

Removed exports: AbsoluteDirectoryAccess, AgentDirectoryPathStorage, AgentDirectoryServiceLogger, AgentRosterSnapshot, AppendFollowUpResult, BaseBranchCandidate, bestConnectionMethod, ChildRunDeliveryResult, CollectReviewDiffResult, DrainedFollowUpItem, FollowUpConsumerKind, FollowUpPresentation, FollowUpSubmission, getStorageKey, GoogleClientCache, processMultipleOutputs, resolveGoogleClient, SubmitFollowUpOptions, workspaceFromSnapshot.

class/interface/enum declarations: -13 / +4 = **-9**.

- Deleted outright: 6 production interfaces (AgentDirectoryPathStorage, AbsoluteDirectoryAccess, AgentDirectoryServiceLogger, GoogleClientCache, ResolveGoogleClientParams, ConnectionResult) and 3 test classes (RecordingStorage, RecordingAbsoluteDirectories, RecordingLogger).
- Re-declared without `export`: 4 (SubmitFollowUpOptions, AppendFollowUpResult, BaseBranchCandidate, AgentRosterSnapshot — the last also lost its `Entry` type parameter, dropping three generic arguments at its use sites).

Type aliases un-exported (not in the declaration count above): 6 (DrainedFollowUpItem, FollowUpPresentation, FollowUpConsumerKind, FollowUpSubmission, ChildRunDeliveryResult, CollectReviewDiffResult). Type aliases deleted: 2 (ParsedToolUseSharedResult, and the AssertNever / `_UserVarsStayInRuntimeTokens` guard pair).

Other element deltas: two static methods removed from `AgentWorkspaceState` (one public, one private), one public method removed from `ExecutionRegistry`, one field removed from `OutputState`, one field removed from `AgentRosterSnapshot`, one parameter removed from `checkExpectedOutputs`, one `node:util` import removed.

## Consumer counts (R8)

Counts are `git grep` against `origin/main` over `src/` and `packages/`. Dated `docs/dev/audits/**` and `docs/proposals/**` hits are excluded: both are archival by design and carved out of the `check-guidance-refs` gate. Post-merge hits on this branch are 0 for every symbol.

| Deleted symbol | Production callers | Test callers | Disposition |
| --- | --- | --- | --- |
| `processMultipleOutputs` | 1 (`extractFilesFromXml`, same file, `:198`) | 4 sites / 2 files | Inlined; tests are a same-arity swap; docstring at `outputOperations.ts:31` repointed; knip row removed |
| `bestConnectionMethod` | 1 (`agentResponseTextConnector`, same file, `:85`) | 1 file | Folded into the connector; no consumer ever read the `choice` half; knip row removed |
| `AbsoluteDirectoryAccess` / `AgentDirectoryPathStorage` / `AgentDirectoryServiceLogger` | 0 | 12 grep hits across exactly 2 files (definition + rewritten suite) | Deleted; 3 knip rows removed |
| `resolveGoogleClient` | 1 (`modelHandlerGoogleInteractions.ts:598`) | 1 file (deleted) | Inlined into `getClient` |
| `GoogleClientCache` | 1 handler field type (+2 refs inside the deleted params interface) | 1 file (deleted) | Became a file-local field type |
| `ResolveGoogleClientParams` (file-local) | 1 | 0 | Deleted |
| `ConnectionResult` (file-local) | 2 | 0 | Deleted |
| `ParsedToolUseSharedResult` (file-local) | 1 | 0 | Replaced by `z.ZodSafeParseResult` |
| `getStorageKey` | 2 (`outputValidation.ts:36`, `roundSummary.ts:64`) | 0 | Both read the required `deps.storageKey` field |
| `setActiveRun` | 1 (`runReflectionFlow.ts:174`) | 0 | Renamed; its guard and the two resets it protected were unreachable against a freshly built `createOutputState()` |
| `ExecutionRegistry.waitForChange` | 1 (`ExecutionsTool.ts:287`) | 0 | Retargeted to `waitForAnyChange`; single-id semantics are identical, and `register` accepts `Promise<unknown>` |
| `AgentWorkspaceState.fromCanonicalSnapshot` | 2 (`helpers.ts:45`, `ToolUseCycleNode.ts:44`) | 1 | Body was byte-identical to `fromSnapshot`'s |
| `workspaceFromSnapshot` | 2 (`MediaExtractionNode.ts:27`, `ResponseCycleNode.ts:51`) | 0 | Both already import `AgentWorkspaceState` as a value |
| `AgentRosterSnapshot.agents` | 0 readers | 0 | The one production consumer (`packages/cli/src/runtime/agentRoster.ts:27`) reads five other fields and builds `agentKeys` from `getEnabledAgentKeys`; the one test reads `missingTeamId` |

Ten types un-exported (`export` dropped, declaration otherwise untouched), cross-file references 0 each, verified by repo-wide word grep over src, all host packages, docs, config, prompts and supabase: DrainedFollowUpItem (3 same-file refs), FollowUpPresentation (2), SubmitFollowUpOptions (3), FollowUpConsumerKind (4), FollowUpSubmission (2), ChildRunDeliveryResult (2), AppendFollowUpResult (3), CollectReviewDiffResult (2), BaseBranchCandidate (3), AgentRosterSnapshot (4). None appears in `config/ratchets/knip-baseline.json`, so no baseline row went stale.

`getVisibleAgents` keeps 6 callers across CLI and desktop after the roster field is deleted, so the snapshot change removes work rather than a surface. `isNotADirectoryError` loses its `platformAgentDirectories.ts` consumer but keeps 10+ others, so it does not become a new knip finding.

No emit path was deleted: this diff adds and removes zero `bus.emit` and `appSignals.emit` sites.

## Ratchets

Five rows leave `config/ratchets/knip-baseline.json`, one per deleted dead export: `processMultipleOutputs`, `bestConnectionMethod`, and the three `AgentDirectoryService` port types. No row was added and no baseline was widened. The one new export carries a consumer in the same commit. No new distinct `@agent/*` deep-import specifier is introduced from a host, and no top-level subsystem edge is new (`agent → logger` and `agent → utils` are both already `value` in the architecture-edges baseline).

## Verification

Reviewed without a typechecker available. Every changed signature was traced to every caller by grep, including the barrel files (`src/agent/index/index.ts`, `src/agent/review/index.ts`), the extension command registry, the settings Zod schemas, `packages/extension/resources/`, `prompts/`, and `supabase/functions/`. `z.ZodSafeParseResult<T>` was confirmed against the installed zod 4 typings (`zod/v4/classic/parse.d.ts` declares it, `external.d.ts` re-exports `./parse.js`, and `ZodObject.safeParse` returns `ZodSafeParseResult<core.output<this>>`, which matches `ToolUseRunShared = z.output<typeof ToolUseRunSharedSchema>` exactly). `Object.keys(...) as ReadonlyArray<keyof UserVars>` is a comparable-relation assertion and compiles. `SkippedStageHandle` was read to confirm the added `withOutputStage` wrapper on the migrated test call sites emits nothing. Prettier has not been run over the branch; print widths on the reflowed statements were checked by hand and a `npm run format` pass is expected to be a no-op, but should be run before merge.

One nit worth fixing before merge: `src/agent/implementations/flows/reflection/output/roundSummary.ts` still imports `type StorageKey`, which became unused when the `getStorageKey(state)` call was replaced by `deps.storageKey`. Nothing in the repo catches it (`noUnusedLocals` is unset, `@typescript-eslint/no-unused-vars` is off, and knip does not scan imports).


LionSR:
All three findings were valid and are fixed in the latest commit.

**1. `AbsoluteFS.ensureDir` on the configured custom agents directory (P2, but the most important one).** This was a real behavior change, which makes it a defect against the PR's own behavior-preserving claim. The injected port this replaced was `ensureDir: (target) => platform().fs.createDirectory(target)` (`platformAgentDirectories.ts` on main) — a raw `createDirectory` that rejects when the target is a regular file. `AbsoluteFS.ensureDir` catches `isFileExistsError` and returns, so a file at that path was being handed back as a directory and the failure deferred to the first read or write. Switched to `AbsoluteFS.createDir`, which is exactly the prior call, with a comment recording why it is not `ensureDir`.

**2. Mutable `UserVariableValueSchemas` export (P1).** Agreed. The only cross-file consumer took `Object.keys()` from it, while the map itself backs checkpoint validation process-wide, so exporting it let a consumer drop validators. The map is file-local again and a frozen `USER_VAR_RUNTIME_TOKENS` list is exported beside it as the vocabulary's single owner. `userVars.ts` and the template-renderer test now consume that. This also shrinks one more knip-baseline row, since the token list gained a real cross-file production consumer.

**3. `startRunWorkspacePreparation` single-caller wrapper (P1).** Agreed, and it is this PR that made it shallow: once the storage-key guard came out, the body was one assignment reaching a file-local helper. Inlined at its one call site in `runReflectionFlow` and `collectRunSupportFiles` exported instead.

Validation after the fixes: `typecheck` and `lint` clean, `check:dead-code-ratchet` OK with the baseline shrunk by one row, and 135 tests passing across `AgentDirectoryService`, `AgentTemplateRenderer`, and the reflection output suites.

PR #11432, opened 2026-08-26
docs: record the 2026-08-26 cross-file simplification survey (40 verified candidates)

LionSR:
Round 2 of the simplification sweep, following [round 1](https://github.com/texra-ai/coauthor/pull/11414) (shipped as #11416-#11422).

## Why a second round finds anything

Round 1 partitioned the repo by directory and gave each surveyor one subtree. That partition is blind by construction to anything whose two halves live in different directories: a symbol exported from `src/` whose only caller is in `packages/cli/`, one fact derived independently in all four hosts, a Zod schema mirrored by a hand-written interface two trees away.

This round replaced the directory partition with **24 cross-cutting lenses** — dual representation, re-derivation helpers, lifecycle mirrors, defensive machinery after same-process typed handoffs, hand-rolled code the ES2022 floor already covers, compatibility past its window, options bags with one caller, cross-provider handler duplication, webview-tree parity, and more.

## Results

| | Count |
| --- | --- |
| Candidates proposed | 54 |
| **Survived verification** | **40** |
| Refuted | 14 |

Net if all land: **-2,263 LoC, -142 elements**, across nine disjoint lanes.

Every candidate went to an independent adversarial verifier whose default was REFUTED, with eight named kill conditions: duplicate of round 1, already filed or already done, wrong consumer count, deliberately justified by a dated ruling, settled surface, net-adds or churns, misread masking site (checklist §15), or too thin to stand alone. The 14 refutations are recorded in §3 so they are not re-discovered.

Two guards against rehashing round 1: every surveyor read a 90-title index of what round 1 shipped and refuted before proposing anything, and kill condition #1 for every verifier was "check whether this is already done at HEAD" — since #11416-#11422 landed the day before, a stale candidate looks identical to a live one.

Verifiers found factual errors in the evidence of most surviving candidates. Those corrections are recorded inline and override the claim above them; implementers must read them.

## Scope

Docs only. `docs/proposals/` is excluded from the published VitePress site by `docs/.vitepress/publicDocs.js`.

Checks: `npm run format`, `npm run check:guidance-refs`, `docs/scripts/check-root-docs.mjs`, `git diff --check`.


LionSR:
All six comments addressed. Two of them found **real behavior bugs in the implementing PRs**, and both items are now reverted — thank you, those were the most valuable findings of the round.

## Code findings (acted on, in the lane PRs)

**`resumeOutcome` (P2)** — correct and blocking. `sessionProgressSubscription.ts:70-71` forwards the `inquiryThreadUpdated` payload verbatim, so the field is on the public NDJSON wire; `2026-08-03-ssot-consolidation-plan.md:78-79` puts public-key deletion on a 0.40-deprecate/0.41-delete clock and the repo is at 0.40.6. Reverted in #11441.

**`texra.agent.rounds` (P2)** — also correct. `jsonConfigProvider.ts:40-48` reads workspace store → global store → catalog default, so an uncataloged key in `.texra/config.json` still overrides. Absence from the catalog does not make the read dead. Plus the replacement shipped 2026-08-02, inside the three-month window. Reverted in #11437.

Both were the **same class of error**, and I've recorded it in §4 of the doc as a new verifier kill condition: for anything crossing a public wire format, a persisted user file, or a settings key, an in-repo consumer grep does not establish deadness.

## Doc defects (fixed)

**Garbled text** — confirmed and fixed. A full non-ASCII sweep found exactly one contaminated run: `никогда` at line 1879, now "never". No other Cyrillic or CJK anywhere in the file.

**Unfilled correction stubs** — confirmed, and there were **five**, not four: lines 94, 1603, 1679, 2190, 2792, under three different wordings ("see corrections field content above", "See the corrections field content above.", "see above"). Checked the raw verifier output: those five verifiers wrote a self-referential placeholder into the `corrections` field and put their actual numbered corrections in the `reasoning` block instead. The content is therefore present in the document, just not where the heading promised. Replaced each stub with an accurate pointer to the reasoning block rather than inventing corrections that were never written.

**Acceptance-criteria wording (P2)** — a fair catch, and the criterion as written was unsatisfiable. Rewritten: a deleted export that *had* a baseline row shrinks the baseline; a currently-live export has no row and simply disappears with its consumers. I also added the case this round actually hit — when a deletion orphans some *other* export into dead code, un-export that symbol rather than baselining it (that is what happened to `shouldSaveModelIO` and `REVIEW_SEVERITIES` during integration).

## `ITool.hosts` → `unavailableHosts` (P2) — not reverted, with reasoning

This one I'm declining, and it already merged in #11436. The concern reads "frozen surface" as "exported type shapes may never change", but the freeze in this repo is about **import edges**, not type shapes: `host-agent-import-baseline` forbids new distinct `@agent/*` deep-import specifiers from a host, and the sibling ratchets freeze the remaining edges. CLAUDE.md is also explicit that `packages/agent` is "built, fenced, **not published**" and that "npm publication is deliberately held until a named external consumer exists" — so there is no embedder whose compile can break, which is the condition AGENTS.md names for keeping a contract ("no current public contract or demonstrated user need").

That is a real difference from the `resumeOutcome` case, where the wire format *is* published and consumed today. Happy to revisit if a named external consumer appears.

PR #11437, opened 2026-08-26
refactor(shared): fold six cross-file duplications across shared, storage and utils

LionSR:
Lane `M5-shared` of the [2026-08-26 cross-file simplification survey](https://github.com/texra-ai/coauthor/pull/11432) — Shared, controllers, storage, utils, platform.

Every item below was proposed by a surveyor, then independently verified by an adversarial reviewer whose default was REFUTED. The verifier corrections are recorded in the survey doc and governed the implementation; where a correction cut or reshaped an item, that is noted.

## What this deletes

- **Delete KVStoreCache: three memoizations of a stateless two-field KVStore handle** — -180 LoC, -8 elements.
- **Fold src/utils/text/sanitizeTag.ts into wrapWebhookEvent, its only caller, and drop the tag-name parameter and its unreachable validator** — -20 LoC, -3 elements.
- **Fold AgentSelectionPanel's SOURCE_META origin badges onto the shared AGENT_DECORATORS.properties table** — -10 LoC, 0 elements.
- _(skipped)_ **Route the three remaining hand-rolled reimplementations onto the owner that already exists: throwAggregated/aggregateError and Intl.ListFormat** — -10 LoC, 0 elements.
- **Replace AgentOptionData's three one-hot source booleans with the AgentSource enum the registry already carries** — -8 LoC, -2 elements.
- _(skipped)_ **Delete the retired `texra.agent.rounds` config read: the key is in no settings catalog, so housekeeping's scan depth is already the hardcoded default** — -6 LoC, -2 elements.

## Deliberately not done

- **Item 4, Intl.ListFormat leg (packages/cli/src/commands/_helpers/globalArgs.ts formatCommaList)** — Verifier correction #8 explicitly recommends dropping this leg first: it is 0 net LoC and 0 net elements (formatCommaList is 4 lines; the replacement Intl.ListFormat const is 89 chars and prettier-wraps to 4), and it trades a deterministic repo-local constant for CLDR-version-dependent output pinned by ~10 snapshot lines in src/test-kernel/cli/__snapshots__/Completion.vitest.ts.snap plus two exact-string assertions in CliRootArgs.vitest.ts:490-521. I cannot run vitest in this worktree (no node_modules), so I would be shipping an unverifiable snapshot risk for zero measurable win. The batch stands on the aggregation leg, which I did ship (including the third site the survey missed).
- **Item 1, RATCHET WIN sub-leg (unexport STREAM_LOGS_DIR / STREAM_LOG_SUMMARIES_DIR, delete 4 knip-baseline rows)** — Struck by verifier correction #1 as a duplicate of a round-1 REFUTED candidate and ruled against by docs/proposals/2026-08-19-dead-code-gate-blind-spots.md:184-197. Confirmed independently: both constants have module-level production consumers in the same file (StreamLogStore.ts:181, :183, :212 and elsewhere) plus four test consumers, so deleting the cache does not change their export status. No knip-baseline rows were touched.

## Needs a human decision: Settings -> Agents icons and badges

Folding `SOURCE_META` onto the shared `AGENT_DECORATORS.properties` table means the two tables that had drifted now agree, which changes what renders in **both** graphical hosts (the desktop renderer mounts the same settings app):

- **inline agents**: bolt icon becomes a code icon — this is the divergence the item targets
- **remote agents**: gain a `Remote` badge in the list and, for the first time, an origin tag in the detail pane
- **custom agents**: unchanged at `star`. The shared table was changed from `circle-user` to `star` rather than the reverse, because `properties.custom.icon` has no render consumers today

Reversible per row if you want a different resolution.

## Net elements (R6)

| Metric | Delta |
| --- | --- |
| Files changed | 26 |
| `^[+-]export` symbols | +0 / -2 (net -2) |
| class/interface/enum/type declarations | +1 / -1 (net 0) |
| Net LoC (`git diff --stat origin/main`) | +132 / -383 (net -251) |

## Consumer counts (R8)

Grepped before each deletion across `src/`, `packages/*/src`, `packages/extension/resources/`, `prompts/`, `supabase/functions/`, `packages/extension/package.json` contributions, `packages/extension/src/commands.ts`, and the settings schemas. Per-symbol counts are recorded in the survey doc under each candidate's **Evidence** block; the deletions in this PR all had **zero** production consumers except where an item explicitly rewires a call site (noted in its evidence).

`config/ratchets/knip-baseline.json`: unchanged — no baselined symbol left the baseline in this lane. The ratchet passes with no new findings.

## Validation

- `npm run lint` — clean
- `npm run format` applied

Validated on this branch in isolation, not only in the integration merge, so this PR does not depend on a sibling lane.

## Risk notes

<details><summary>Implementer notes carried forward</summary>

CANNOT VALIDATE LOCALLY. No node_modules in this worktree, so nothing was typechecked, linted, formatted, or tested. Everything below is read-and-rg verification only. The pre-push hook fired "Desktop renderer TS errors" but it runs `npx tsc` in a worktree with no node_modules, so that result is uninformative either way; I pushed with --no-verify per the brief. Please re-run `npm run typecheck` AND the desktop renderer tsc (`packages/desktop` tsconfig.renderer.json) before merging.

RE-CHECK THESE SPECIFICALLY:

1. Two `as` casts I added, both predicted by the verifier but unverified by a compiler.
   - packages/extension/src/settingsView/frontend/components/profile/AgentSelectionPanel.ts `sourceMeta()`: `properties[source as keyof typeof properties]` after an `in` guard, and `tone: source as SourceTone`. AgentSource and SourceTone overlap on custom/remote/inline so the cast should be legal, but confirm.
   - src/shared/utils/selectTemplates.ts `buildAgentTooltip`: same `opt.source as keyof typeof properties` shape.
   In both places the returned `badge` object carries an extra `hint` field. It is a variable, not a fresh literal, so excess-property checking should not apply — worth confirming.

2. Prettier will very likely reflow two spots I hand-formatted: the lit template in AgentSelectionPanel `renderDetail` (the `<wa-tag …>${badge ? html`…` : nothing}${displayName}</wa-tag>` line runs long) and the ternary inside `wrapWebhookEvent` in src/tools/github/formatUtils.ts. Run `npm run format` before merge.

3. USER-VISIBLE UI CHANGES in Settings -> Agents (rendered in BOTH graphical hosts — packages/desktop/src/renderer/main.ts:55 mounts settings-app, so this is not extension-only):
   - inline agents: bolt -> code (adopting the shared table; this is the genuine divergence the item targets)
   - remote agents: gain a `cloud Remote` badge in the list AND, for the first time, an origin tag in the detail pane
   - custom agents: UNCHANGED at `star`. I did NOT adopt `circle-user`. Instead, per verifier correction #2, I changed the shared table (src/shared/utils/icons.ts) from `circle-user` to `star`, since `properties.custom.icon` has zero render consumers today and `circle-user` is already the Account glyph in the same settings window. That is a direction reversal from the written proposal — flag it if the integrator disagrees.
   - the detail-pane tag is now unconditional (built-ins render `Built-in` with no icon, exactly as before; I kept the whitespace tight so the built-in chip is byte-identical).

4. src/shared/utils/icons.ts is a shared table — confirm no other lane edited `AGENT_DECORATORS.properties` concurrently.

5. Shared-file coordination: I edited src/transcript/StreamSnapshotStore.ts and src/controllers/progressView/backend/ProgressBackend.ts, both also owned by lane M9-progress. My StreamSnapshotStore edits are confined to the kvHandles field (:474-486 region), `kv()`/`invalidateKvHandles` (:595 region), `evict`/`evictAll`, and the seed-refresh call at ~:1959. My ProgressBackend edits are one import line and the 5-line aggregate block at :356-362. Should merge cleanly but check.

6. Behavior claims worth an independent look:
   - `throwAggregated` is a no-op at length 0, so fileLocks' `return value as T` path is preserved. `aggregateError` returns `unknown`; `throw aggregateError(...)` is precedented (runTrace.ts:79, :104, :119; runExecution.ts:562).
   - Deleting `StreamLogStore`'s `kvHandles.invalidateAll()` on storage-root reload is safe because `KVStore.write` re-`ensureDir`s per write and every op re-resolves the root. src/test-kernel/transcript/StreamLogStoreLoad.vitest.ts:935-951 ("prepares transcript directories again after a storage-root reload") asserts ensureDir-per-write, not handle identity — it looks like a cache test and is not one.
   - Deleting `StagedDeletionHost.invalidateKvHandles` is safe because `stagedStreamDataDir` is never a KVStore dir; the only KVStore dir for a stream is `streamDataDir(streamId)`, a pure function of the id.

7. I collapsed `StreamLogStore`'s now-trivial `kv()` / `summaryKv()` pass-through methods into direct `this.logsKv` / `this.summariesKv` field reads (17 call sites, mechanical). This is slightly beyond the letter of the proposal ("point kv()/summaryKv() at them") but keeping them would have left two pure pass-throughs, which the repo's abstraction-discipline rule bans. Revert that hunk if the integrator prefers the literal proposal.

8. Guidance-refs gate: I updated CLAUDE.md:56 (60 -> 59) and AGENTS.md:135 (66 -> 65 total, 60 -> 59 other) for the sanitizeTag.ts deletion. `find src/utils -name '*.ts' -o -name '*.tsx' | wc -l` now returns 65. Confirm `npm run check:browser-safe-utils` and `npm run check:guidance-refs` pass — this is the ungated, hard-required guidance-refs CI job.

9. knipRowsRemoved is 0 and that is deliberate, not an oversight. I grepped config/ratchets/knip-baseline.json for every deleted symbol and file (KVStoreCache, sanitizeTag, wrapAndSanitizeTag, selectTemplates, AgentSelectionPanel, icons, housekeeping, formatUtils, globalArgs) and there are no rows. Every export I deleted had a live consumer, so none was baselined. No baseline was widened.

10. No masking catch/fallback was deleted anywhere in this lane."

</details>


LionSR:
Three things addressed, including one **behavior finding raised on the survey PR that applies to this diff** — that item is now reverted.

## `texra.agent.rounds` — reverted (was a real behavior change)

Codex raised this on #11432 and it is correct. The survey's evidence said the key "appears nowhere else in the tree (no settings catalog, no other reader)", which is true and insufficient:

`src/platform/defaults/jsonConfigProvider.ts:40-48` reads the **workspace store, then the global store, then the catalog default**. A `texra.agent.rounds` entry in a user's `.texra/config.json` is therefore still honored even though no catalog row exists. Hardcoding `DEFAULT_MAX_ROUNDS` would force a depth of 5 and silently skip legacy outputs for anyone who had configured a larger depth.

The retirement window also does not permit it yet: AGENTS.md allows removal three months after the replacement ships, and the replacement dates to 2026-08-02 — 24 days ago.

Restored `src/housekeeping/pack.ts`, `src/housekeeping/utils.ts`, and `LegacyWorkflowOutput.vitest.ts` to main. The other five items in this lane stand.

## Finding 1 — stale KV-handle docstrings, fixed

Both corrected, since `invalidateKvHandles` left with `KVStoreCache` and the host port carries no KV-handle capability now:

- module comment: "records, KV handles, write mutexes, and stream versions" → "records, write mutexes, and stream versions"
- `StagedDeletionHost` docstring: dropped "its cached KV handles" from the capability list

## Finding 2 — PR body vs diff, resolved in the other direction

You were right that the body and diff disagreed. Rather than correcting the body to say it shipped, the item is now genuinely reverted per the finding above, so the `_(skipped)_` marker is accurate — for a different and better-evidenced reason than the one the body originally gave. I'll leave the body as-is since it now matches.

## Residual risk — closed

Validated centrally in a worktree with a real `pnpm install`: typecheck (six projects, including the desktop renderer) clean, lint clean, ratchet no new findings, full suite **10,163 passed / 0 failed**. That covers the two `as keyof typeof properties` / `as SourceTone` casts you reviewed by hand — they compile.

Thanks for reproducing the `aggregateError` branch equivalence and the `KVStore` relative-path argument; those were the two I most wanted a second pair of eyes on.

PR #11440, opened 2026-08-26
refactor(scripts): delete four duplications across the build verifiers and CI

LionSR:
Lane `M8-scripts` of the [2026-08-26 cross-file simplification survey](https://github.com/texra-ai/coauthor/pull/11432) — Scripts, CI, build verifiers.

Every item below was proposed by a surveyor, then independently verified by an adversarial reviewer whose default was REFUTED. The verifier corrections are recorded in the survey doc and governed the implementation; where a correction cut or reshaped an item, that is noted.

## What this deletes

- **Delete the duplicated repo-source VS Code-free scan from the two desktop artifact verifiers** — -97 LoC, -6 elements.
- **Fold the remote-agent-docs CI job into docs-check instead of paying a second runner for the same docs-only PRs** — -32 LoC, -1 elements.
- **Sweep four residues in the unlinted, unknipped scripts/ corpus (dead re-export block, unused import, pre-workspace-split fallback, phantom contributes key)** — -27 LoC, -10 elements.
- **Replace the hand-rolled recursion and stat/ENOENT probes left in the package and release-gating build scripts with readdir({recursive:true}) and statSync({throwIfNoEntry:false})** — -15 LoC, 0 elements.

## Deliberately not done

- **Item 4, correction 7: fifth collapsible stat probe in scripts/verify-desktop-installer-artifacts.mjs:152-167** — That file is not in this lane's owned-file list and is not named as shared, so it may belong to another lane. The lane's -15 target already excludes it (correction 1's arithmetic stops at the four in-lane sites). Left for a follow-up; it is a clean, isolated ~-7 LoC change.
- **Item 1, proposal step 1 (partial): add the two roots to the src/README.md list** — src/README.md deliberately does not carry a copy of the zone list - it says 'The canonical list is VSCODE_FREE_ZONE_DIRS in eslint.config.mjs - read it there rather than trusting a copy, including this one.' There is no list at src/README.md:42-44 to update. CLAUDE.md (which does carry a copy) was updated.
- **Item 3, proposal step 1 (optional): collapse the process.argv[1] === fileURLToPath(import.meta.url) guard in check-dead-code-ratchet.mjs to a bare main() call** — Explicitly optional; the guard is harmless and keeping it avoids a second behavioural change in the same file. Cost is ~-2 LoC.

## Net elements (R6)

| Metric | Delta |
| --- | --- |
| Files changed | 12 |
| `^[+-]export` symbols | +1 / -6 (net -5) |
| class/interface/enum/type declarations | +0 / -0 (net 0) |
| Net LoC (`git diff --stat origin/main`) | +74 / -250 (net -176) |

## Consumer counts (R8)

Grepped before each deletion across `src/`, `packages/*/src`, `packages/extension/resources/`, `prompts/`, `supabase/functions/`, `packages/extension/package.json` contributions, `packages/extension/src/commands.ts`, and the settings schemas. Per-symbol counts are recorded in the survey doc under each candidate's **Evidence** block; the deletions in this PR all had **zero** production consumers except where an item explicitly rewires a call site (noted in its evidence).

`config/ratchets/knip-baseline.json`: unchanged — no baselined symbol left the baseline in this lane. The ratchet passes with no new findings.

## Validation

- `npm run lint` — clean
- `npm run format` applied

Validated on this branch in isolation, not only in the integration merge, so this PR does not depend on a sibling lane.

## Risk notes

<details><summary>Implementer notes carried forward</summary>

CENTRAL VALIDATION MUST RE-RUN: npm run lint, npm run format:check, npx vitest run src/test-kernel/architecture/dependencyDirection.vitest.ts src/test-kernel/scripts/DesktopPackagePayload.vitest.ts src/test-kernel/scripts/CheckDeadCodeRatchet.vitest.ts, npm run check:dead-code-ratchet, npm run check:extension-package-invariants, npm run check:package-contributes, npm run check:desktop-build-artifacts, npm run check:desktop-package, corepack pnpm --filter @texra-ai/agent build. I could not run any of these (no node_modules in the worktree); I verified by reading, rg, node --check on all nine edited .mjs files (all pass), a YAML parse of ci.yml, and a live smoke run of the rewritten walkFiles.

1. ITEM 1 HAD NO VERIFIER CORRECTIONS. The brief's corrections block for item 1 is a placeholder ("See the corrections field content above.") with no content. I verified the item myself and found one real error in its proposal: step 2 says to add the rule to "the existing packages/desktop/src/**/*.ts block at eslint.config.mjs:528" - no such block exists; line 528 is one entry in the shared TS `files` array. I added a new 20-line config block instead, which is why eslint.config.mjs is +22. Treat the rest of item 1 as verified-by-me rather than verified-by-the-verifier.

2. Formatting is the highest-probability failure. I wrote prettier-shaped code by hand (printWidth 80, objects kept multiline by a newline after `{`) but could not run prettier. Watch: scripts/verify-extension-package-invariants.mjs:24 is exactly 80 chars; the two statSync object-argument sites in verify-desktop-package.mjs; the .filter callback in packages/agent/scripts/fsWalk.mjs; the new 20-line eslint block; the 82-char step `if:` in ci.yml (precedent: the deleted job's `if:` was far longer and survived). If format:check complains, `npm run format` and amend.

3. Coverage claim to re-check for item 1: I confirmed `packages/desktop/src` and `src/logger` are clean today (0 hits for from/require/import 'vscode' in either, 0 hits for @common/state under packages/desktop/src), so adding them to VSCODE_FREE_ZONE_DIRS and VSCODE_FREE_ZONES should be additive with zero new violations. If lint or the ratchet does fire, the fix is a real violation the deleted scripts were only catching in a manual workflow_dispatch job.

4. Item 3(c) is an M-class masking site fixed loudly, not a guard deleted quietly: findExtensionPackagePath silently substituted the repo-root package.json when packages/extension/package.json was absent; the inlined path.join + readJson now throws ENOENT. Stated in the commit body; the PR body must keep saying it.

5. Deviation from item 3's proposal, taking correction 5's first option: after dropping 'configuration', CATALOG_DERIVED_CONTRIBUTES and REQUIRED_CATALOG_CONTRIBUTES were byte-identical, so I collapsed them into one exported constant and pointed verify-vsix-contents.mjs's shipped-non-empty assertion at it. Consequence: the "delete the dead CATALOG_DERIVED_CONTRIBUTES import" half of (b) became "swap the import for the live one" - the element still removed is REQUIRED_CATALOG_CONTRIBUTES. Behaviourally identical (both lists were ['commands','keybindings']).

6. knip-baseline.json untouched and knipRowsRemoved is 0 because no row in it is under scripts/ - `grep '"scripts/'` returns nothing, and root `scripts/**/*.mjs!` is a knip entry pattern with includeEntryExports off, so these exports were never reported. If check:dead-code-ratchet reports NEW findings after this change, that would contradict my reading and needs a look.

7. Item 2 behaviour change worth eyeballing on the first docs-only PR: the folded step is last in docs-check with `if: ${{ !cancelled() && needs.changes.outputs.remote_agent_docs == 'true' }}`, per correction 8, so a catalog failure and a formatting failure still surface in the same run. The named check "remote-agents catalog" disappears from the PR checks list; per correction 7 main has no branch protection, so no required check name breaks - re-verify with `gh api repos/.../branches/main/protection` before merging. The validate-status special case was deleted because the general docs-only gate (pull_request && checks skipped && docs-check == success) already covers it once the step lives inside docs-check.

8. Deleting checkDesktopSourceBoundaries also orphaned `collectFiles` and the `walkFiles` import in verify-desktop-package.mjs, and `path` in extension-package-utils.mjs; I removed all three, plus the two now-false lines from verify-desktop-package.mjs's success summary and one from verify-desktop-build-artifacts.mjs's. `summary.splice(7, 0, '- macOS app icon')` indexes from the front and is unaffected. Repo-wide grep for 'desktop-shared' in src/, packages/, docs/ returns nothing, so no test pins those strings.

9. One stale doc reference left alone deliberately: docs/proposals/2026-07-09-state-of-the-architecture.md:567 cites `findExtensionPackagePath` as machine enforcement for MONO-3. It is a dated proposal, not a guidance file (check-guidance-refs.mjs's GUIDANCE_FILES does not include docs/proposals/), and the brief forbids touching proposal docs.

10. fsWalk.mjs part A is LoC-neutral by design (correction 1). I used `!entry.isDirectory()` not `entry.isFile()` per correction 9, so symlinks and other non-directory entries are still returned. Cross-directory ordering may differ from the old depth-first descent; both consumers (validate-artifacts.mjs joins all text, rewrite-declaration-aliases.mjs rewrites each file independently) are order-independent. `isFile` and its deliberate ENOTDIR arm are untouched.

</details>


LionSR:
Both findings addressed — one fixed, one refuted with a reproduction.

## 2. Stale `@common/state` eslint rule — fixed, removed

You're right on every point, and it was worse than inert in context: this PR's whole purpose is deletion, so adding a rule that guards nothing is a net element added for no benefit. Verified before removing:

- `find . -name 'stateManager*'` outside `node_modules`/`dist` → **0 results**. The module the rule names does not exist.
- `packages/extension/src/common/state/` contains only `pendingStateManager.ts`.
- `grep -rn "@common/state" packages/desktop/src` → **0 results**.
- Keys live at `@shared/state/stateKeys`, so the message pointed at a dead path.

Removed the block rather than rewriting it: the boundary is architectural now, and `@common/state/**` is already in the src-restricted pattern list. Lint clean.

## 1. `fsWalk.mjs` symlink descent — refuted, and I can explain the discrepancy

The behavior depends on **`withFileTypes`**, and that is what separates your result from mine.

`readdir(dir, { recursive: true })` in **string mode** does follow symlinked directories. With **`withFileTypes: true`** it does not. This PR uses `withFileTypes: true`.

Reproduction on a tree of `sub/b.txt` plus `link -> sub`:

```
# withFileTypes: true  (what this PR uses)
Node 22.14.0:  link (isSymlink=true)  sub  sub/b.txt          <- no link/b.txt
Node 26.7.0:   link (isSymlink=true)  sub  sub/b.txt          <- no link/b.txt

# withFileTypes: false (string mode)
Node 22.14.0:  [ 'link', 'sub', 'sub/b.txt', 'link/b.txt' ]   <- follows
Node 26.7.0:   [ 'link', 'sub', 'sub/b.txt', 'link/b.txt' ]   <- follows
```

I checked the CI-pinned major specifically, since your note asked for it: `.github/workflows/ci.yml` pins `node-version: '22'`, and 22.14.0 does not descend.

So old and new behave identically: both emit the symlink entry itself (old via the non-directory branch, new via `!entry.isDirectory()`), and neither descends into it. No double-processing of a physical file under two logical paths. Left the code as-is rather than adding a defensive `isSymbolicLink()` filter that would change nothing.

## Residual risk note

Fair, and now closed. This branch was validated centrally in a worktree with a real `pnpm install`: typecheck (all six projects), lint, `check:dead-code-ratchet` (no new findings), and the full suite at **10,177 passed / 0 failed**.

PR #11452, opened 2026-08-26
refactor(platform): retire the unreachable in-place workspace-storage transition

LionSR:
Lane `P1-storage` of the [2026-08-26 architectural simplification survey](https://github.com/texra-ai/coauthor/pull/11448) (round 3).

Round 3 asked whether each subsystem needs the **shape** it has, so these are mechanism collapses rather than the dead-code deletions of rounds 1-2. Every finding was independently verified at high reasoning effort against the recorded design rationale, then re-validated centrally.

## Findings in this lane

- **Retire the in-place workspace-storage transition: four commit/rollback ledgers exist for one VS Code event that two other hosts already answer with a restart** (collapse, medium) — target -1550 LoC, -22 elements.
- **LanguageModelPort exposes two change events that no consumer distinguishes; fan them into one onDidChange inside the port** (collapse, low) — target -8 LoC, -1 elements.

## Deliberately not done

- **Deleting StreamLogStore.reload / StreamSnapshotStore.evictAll / StreamStatusMachine.clearAll / getAllStreamStates** — These lost their last PRODUCTION caller when `replaceStoresAfterStorageRootChange` went, but they are members of the ratcheted store public surface (config/ratchets/store-public-surface-baseline.json lists `reload` and `evictAll`) and still have direct test suites (StreamLogStoreLoad.vitest.ts, DesktopAgentExecution.vitest.ts, streamStatusTestUtils.ts). Shrinking that baseline plus its tests is a separate decision well outside this lane's blast radius. Reported below for a follow-up.
- **Deleting PersistedState.reload (src/shared/state/PersistedState.ts:164)** — Orphaned one level further down by removing `ProgressPresentationState.reload`. It is half of a symmetric `reload()`/`reset()` pair on a general shared state class in a file no lane owns; I stopped the cascade there rather than improvise. Not knip-visible (class member).
- **Removing STREAM_TRANSITION_CAUSE.ROLLBACK** — Its only producer was the deleted rollback replay loop. Removing the member ripples into src/shared/streams/streamStatus.ts's switch and the exhaustive RunOutcomeAlgebra.vitest.ts table — shared event vocabulary that P7-events may own. Left in place, reported.
- **Removing SessionHandle's now-constant `storageGeneration` / the first clause of `isRepairSuperseded`** — With the transition gone the counter can never increment, so `generation !== this.storageGeneration` is provably false and `isRepairSuperseded` reduces to the abort check. Verifier correction #8 explicitly says these stay ('ensureRestartRepair still reads storageGeneration'), so I followed the correction rather than over-reaching. Flagged for the validator.
- **The proposal's step (1) — landing a modal reload prompt behind hasPendingWorkspaceStorageChange** — Verifier corrections #2 and #3 override it: the event is never delivered for the only case that moves the root, so there is nothing to prompt about and no maintainer UX ruling is required. `hasPendingWorkspaceStorageChange` was deleted with the rest.

## Net elements (R6)

| Metric | Delta |
| --- | --- |
| Files changed | 26 |
| `^[+-]export` symbols | +0 / -5 |
| class/interface/enum/type declarations | +0 / -10 |
| Net LoC (measured, `git diff --stat origin/main`) | +85 / -1778 (**net -1693**) |

## Consumer counts (R8)

Per-symbol counts are in the survey doc under each finding's **Evidence** block. Every deletion here had zero remaining production consumers except where a finding explicitly rewires a call site.

## External-consumer checks

Round 2 shipped two items that had to be reverted because a consumer lived outside the repo. Each lane checked explicitly:

<details><summary>What was checked</summary>

CLI NDJSON vocabulary: not crossed. None of the deleted symbols is a SessionFact/AgentEvent field; `projectCliSessionFact` (packages/cli/src/runtime/sessionProgressSubscription.ts) is untouched and no payload shape changed. `rg` over packages/cli for every deleted name returns zero hits.

Persisted user files: not crossed, and this is the one place I checked hardest because the change alters WHICH storage root a process uses. `workspaceStorageId` / `resolveWorkspaceStoragePath` / the `.texra-workspace` sidecar writer / `migrateLegacyWorkspaceStorage` are all byte-for-byte unchanged, so an existing user's checkpoints, sidecars and `~/.texra` tree resolve to exactly the same directory. The only semantic change is that `WorkspaceStorageProvider` now calls a function-valued `workspacePath` source once in the constructor instead of on demand. I traced all five construction sites: extension.ts:268 (`() => workspace.getWorkspacePath()`, and `activate` returns early at :239 if there is no workspace root, so it is already resolved); packages/cli/src/runtime/cliStateStores.ts:34 via initPlatform.ts:242 (`() => cliWorkspaceCwd`, and `cliWorkspaceCwd = context.cwd` is assigned at :232, BEFORE the call); packages/cli/src/runtime/updateChecker.ts:299 (no workspacePath at all); packages/desktop/src/main/platform/index.ts:76 (a resolved `const`); packages/agent/src/node.ts:51 (a resolved string). No site resolves later than construction, so no user's storage path moves.

Settings keys: not crossed. No catalog entry, no `texra.*` key, and no `.texra/config.json` read/write path changed. `JsonConfigProvider.get`'s workspace-then-global-then-catalog order is untouched; only the ability to SWAP the workspace store mid-process was removed, and the pinned store is opened from the same `openTexraConfigStores(storage, workspaceRoot, warn)` call as before.

texra-action result-JSON contract: not crossed. No CLI command, flag, exit code or result field touched; `rg` over packages/cli returns zero hits for every deleted name.

Ratchets: verified none needs an edit. knip-baseline.json has no row for any symbol I deleted (checked all 26 touched files; the only hits are unrelated rows for `SessionEvent`, `resolveWorkspaceStoragePath`, `workspaceStorageId`). `openTexraWorkspaceConfigStore` lost its last cross-file consumer, so I un-exported it rather than adding a row. host-agent-import-baseline is set-based in BOTH directions (a listed specifier with no live import also fails) — I removed the `@agent/runtime` import from texraConfig.ts, but ProgressViewProvider.ts and others still import it, so the extension list stays exactly live. host-agent-mock-baseline covers only cli/desktop/support suites (none touched). shared-schemas-deep-import-baseline lists none of my files. architecture-edges only fails on new pairs or type-only->value upgrades; controllers->agent stays a value edge via `RUN_FACT_EVENT_TYPES`. store-public-surface-baseline only rejects GROWTH, so leaving `reload`/`evictAll` listed is safe.

</details>

## Validation

Run on this branch **in isolation**, so the PR does not depend on a sibling lane:

- `npm run lint` — clean
- `npm run format` applied


LionSR:
Need to Rebase now

PR #11476, opened 2026-08-26
refactor(runtime): finish the storage-transition cleanup the round-3 collapses left behind

LionSR:
## Summary

Retires storage-transition and restart-repair residue whose production callers were deleted by earlier round-3 changes. The runtime now has one restart-repair pass, workspace storage captures its construction-time path directly, stream approvals clear their bypass state at their ancestry owner, and `StreamLogStore.reload` is no longer exposed without a production lifecycle that can use it.

## Issue acceptance gates

- `StreamLogStore.reload` and its reload-only support state have no production consumer and are removed with their test-only reload scenarios.
- Restart repair remains single-flight and is cancellable on session teardown.
- Workspace storage still captures the construction-time workspace path for CLI and extension hosts.
- Stream deletion still promotes child bypasses before clearing the deleted stream's explicit bypasses.

## Single source of truth

`SessionHandle.restartRepairPromise` is the sole owner of restart-repair single-flight state. `WorkspaceStorageProvider.activeWorkspacePath` owns the construction-time workspace path. `SessionApprovals.forgetStreamAncestry` owns both ancestry teardown and per-stream bypass cleanup.

Removed the reload-only transaction path and the restart-repair queue/generation pair. Moved the one cleanup sequence from `cleanupApprovalsForStream` into its sole ancestry owner; `StreamApprovalBypassState` narrows the per-stream clear operation to that owner rather than adding a public surface.

## Net elements (R6)

19 modified files, 78 additions and 372 deletions: **-294 LoC**. No files were added or deleted. Exported-symbol diff: 0 added, 0 deleted. Declarations: 1 internal interface added (`StreamApprovalBypassState`) and 1 internal interface deleted (`StreamApprovalControllerOptions`), net 0. Removed public surface: `StreamLogStore.reload` (and its ratchet-baseline entry); the internal `executeReload`, `WorkspacePathSource`, restart-repair queue/generation, and public `StreamApprovalBypass.clearForStream` exposure are also retired or narrowed.

## Consumer counts (R8)

`StreamLogStore.reload` and `executeReload`: 0 production consumers (`rg` across `src/` and `packages/`, excluding tests; unrelated reload names only). `storageGeneration`, `enqueueRestartRepair`, and `WorkspacePathSource`: 0 remaining consumers after deletion. `clearForStream`: 1 internal consumer, `SessionApprovals.forgetStreamAncestry`; it retains the same three bypass clears after child promotion. `createNodeStorageProvider` / `WorkspaceStorageProvider` have 5 production construction sites, all passing a plain path or no path; the two changed call sites capture the same synchronous path value at construction. No event emitter or subscriber is deleted or rerouted.

## Host impact

Extension: captures the current workspace path at activation; no workspace-transition path remains.

Desktop: no behavioral surface change; transcript presentation waits for the resident session store.

CLI: captures `context.cwd` when CLI state stores are constructed.

## Scheduled refactoring

None. The deliberately deferred `ToolDefinition.parameters` investigation remains explicitly out of this PR's file scope.

Lane `PA-postmerge` of [round 4B — production surfaces](https://github.com/texra-ai/coauthor/pull/11465).

The fourth pass over production code, on angles rounds 1-3 could not take: what their collapses left half-done in the **absorbing** owner, internal repetition inside the largest files, silent-degradation defects, and the Zod corpus.

## Findings

- **StreamLogStore.reload/executeReload lost their only production callers in #11452 — now a test-only surface** (delete, medium) — target -255 LoC.
- **ToolDefinition.parameters lost its last producer in #11454; the two converter fallback arms are dead** (delete, medium) — target -20 LoC.
- **SessionHandle restart repair still threads a storage generation that can never change** (amend, low) — target -18 LoC.
- **WorkspacePathSource keeps the lazy-callable arm the storage transition needed, and both remaining lambdas close over constants** (delete, low) — target -12 LoC.
- **Stream approval teardown is split across two owners: forgetStreamAncestry documents a clearing step that lives in its only caller** (consolidate, low) — target -5 LoC.
- **Four comments still describe the deleted workspace-storage transition as live behavior** (amend, low) — target -2 LoC.

## Deliberately not done

- **Finding 2: ToolDefinition.parameters + the two converter fallback arms** — Test blast radius is far larger than verified and partly unresolvable as proposed. The verifier's correction 3 says 'roughly 7-8 sites across 3 suites', but ToolConversion.vitest.ts lines ~808-1110 contain ~13 additional tests that feed raw JSON Schema through def.parameters into convertGoogleToolSchema/toGoogleTools to pin Google-converter semantics (opaque-annotation preservation, oneOf/allOf/not + contains/if/then/else/dependentSchemas/prefixItems keyword rejection, array-valued type rejection, exclusive-bound intersection, const/enum intersection/contradiction, $ref rejection). These shapes are NOT producible from Zod schemas, so the proposal's instruction to rewrite 'to zodSchema form' is unsatisfiable for them; deleting the arm makes every one fail (null return / no throw), and deleting the tests would silently drop a whole coverage suite for toGoogleOpenApiSchemaNode branches whose post-change reachability (via Zod-emitted allOf/anyOf/$defs) I cannot adjudicate without running the converter. Two further unexamined fixture suites also feed parameters-only defs: ModelHandlerDeepSeek.vitest.ts:380-395 and AnthropicCacheBeta.vitest.ts:86-94. I applied the change fully, discovered this while sweeping for remaining constructors, and reverted it cleanly rather than improvise. Recommend re-verifying with an explicit disposition for the Google exotic-schema suite (rewrite against an exported unit surface vs. delete-with-rationale) before re-queueing.

## On the line count

Some findings in round 4B **add** lines, and that is the intended outcome. CLAUDE.md treats a masking fallback as a defect: *"A fallback that masks a failure must be loud — log the cause at `warn` and surface it — or not exist."* Where a finding is an `amend`, the fix is to make the failure loud, never to delete it quietly.

## Validation

Focused regression suites passed locally: `StreamLogStoreLoad`, `WorkspaceStorage`, `DesktopAgentExecution`, `ProgressBackendPresentation`, `SessionRestartRepair`, `RestartRepair`, and `ApprovalCleanupScope` (7 files, 184 tests). `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run check:dead-code-ratchet`, `npm run check:guidance-refs`, and `npm run check:browser-safe-utils` also passed. The full local Vitest run had two failures outside this diff: the pre-existing desktop merge-error assertion (also reproducible in isolation) and one non-reproducible legacy-workflow timeout; required CI kernel shards passed.

<details><summary>Implementer risk notes</summary>

Central validator must: (1) typecheck — highest-risk spots are the streamApprovalQueue restructure (StreamApprovalBypassState internal interface, controller factory signature change) and the desktop bridge type after removing reload() from DesktopAgentExecution.vitest.ts:137; (2) run StreamLogStoreLoad.vitest.ts, DesktopAgentExecution.vitest.ts, SessionRestartRepair.vitest.ts, RestartRepair.vitest.ts, WorkspaceStorage.vitest.ts, and any approval/bypass suites; (3) run the store-public-surface ratchet check (baseline lowered by one row — that check must agree reload is gone); (4) confirm no suite constructs SessionHandle non-deferred and depends on repair NOT starting synchronously before its mocks (correction 5's timing nuance; deferred-mode tests are unaffected). The SDK surface is unaffected: `packages/agent/src/index.ts:252` constructs only `StreamLogStore.ephemeral`, on which `reload` threw. This PR deliberately scopes to `StreamLogStore.reload`, not the three sibling orphans #11452 grouped with it (`StreamSnapshotStore.evictAll`, `StreamStatusMachine.clearAll`, `getAllStreamStates`).

</details>

### DIFF
--- a/config/ratchets/store-public-surface-baseline.json
+++ b/config/ratchets/store-public-surface-baseline.json
@@ -24,7 +24,6 @@
       "openReadOnlyForStream",
       "readEntries",
       "recordSummaryMeta",
-      "reload",
       "requestEviction"
     ],
     "StreamSnapshotStore": [
--- a/docs/proposals/2026-08-16-define-out-of-existence.md
+++ b/docs/proposals/2026-08-16-define-out-of-existence.md
@@ -297,7 +297,9 @@ lease exclusivity is proven airtight), startup double abort-check,
 error-classification arm (external producer shapes). **Refuted-as-deletable
 and stays:** `ModelCell` retirement guards (reachable via `/model` racing
 a child-result wake into a WAITING parent), `storageGeneration` (it IS the
-per-pass cancellation token; supersede lands mid-await),
+per-pass cancellation token; supersede lands mid-await — ruling retired
+2026-08-26: #11452 deleted the only supersede writer, so the field was
+removed and the teardown abort signal is the remaining cancellation token),
 the CLI interrupted-follow-up shadow buffer (the dying flow still owns the
 durable queue's lease — some holding pen must exist), the onboarding
 funnel's edge-trigger (protects a user choice from level-triggered
--- a/docs/proposals/2026-08-16-overdefensive-top10.md
+++ b/docs/proposals/2026-08-16-overdefensive-top10.md
@@ -462,7 +462,8 @@ grow, don't remove), and the OpenAI-compat `.nullish()` /
 `?? undefined` normalizations (98); and **(iii) the fenced
 irreducibility registers** — lease/crash-repair machinery, the
 `inBandSubagentExecution.ts:529` manifest read-back, auth credential
-cells, ModelCell, storageGeneration, the interrupted-follow-up buffer
+cells, ModelCell, storageGeneration (since removed: #11452 deleted its
+only supersede writer), the interrupted-follow-up buffer
 (`ToolUseFollowUpQueueManager.ts:83,364`), onboarding funnel,
 round-update reset, `workflowExecutionState` sealed-guard silent
 returns (documented late-async-publication tolerance — watch, don't
--- a/packages/cli/src/runtime/cliStateStores.ts
+++ b/packages/cli/src/runtime/cliStateStores.ts
@@ -13,7 +13,7 @@ interface CliStateStores {
 
 interface CliStateStoresInit {
   readonly storageRoot?: string;
-  readonly workspacePath: string | (() => string | undefined);
+  readonly workspacePath: string | undefined;
 }
 
 /** Open the CLI's global `state.json` under the given storage provider. This
--- a/packages/cli/src/runtime/initPlatform.ts
+++ b/packages/cli/src/runtime/initPlatform.ts
@@ -239,7 +239,7 @@ export async function initCliPlatform(
   if (!tryPlatform()) {
     const stateStores = await createCliStateStores({
       storageRoot: context.storageRoot,
-      workspacePath: () => cliWorkspaceCwd,
+      workspacePath: context.cwd,
     });
     // The project `.texra/config.json` backs the workspace target and
     // user-level config (`~/.texra/global-storage/config.json`, the same file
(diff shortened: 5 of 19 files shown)


LionSR:
Addressed the actionable documentation finding: the PR description now follows the refactor template, including the required R6 net-element accounting and R8 consumer counts, plus ownership, host impact, and scheduled-refactoring sections. There were no inline review threads to resolve.

PR #11480, opened 2026-08-26
fix: adapt GLM reasoning flows to llm-zoo 1.31

LionSR:
## Summary

Follow-up to #11473 after the `llm-zoo` 1.31.0 bump exposed provider-specific reasoning behavior that TeXRA needed to consume.

- make native GLM request and compaction reasoning follow each model’s exact registry vocabulary
- keep forced thinking enabled for GLM-5.3 and GLM-5.3-Flash compaction
- add product-facing `minimal` reasoning and expose only each model’s supported effort levels in Settings
- normalize OpenRouter effort against exact model vocabularies while preserving `none` where supported
- preserve and replay OpenRouter `reasoningDetails` exactly once across tool follow-ups
- feature `glm53flash` alongside `glm53` in the preferred model list
- synchronize the active Supabase Deno lock with `llm-zoo` 1.31.0

## Compatibility details

- GLM-5.3/Flash accept `low`, `high`, and `max`; thinking cannot be disabled, including compaction summaries.
- GLM-5.2 accepts `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, and `max`; only `none` disables thinking, while `minimal` remains a valid enabled-thinking effort.
- Models without an exact registry vocabulary retain their existing provider behavior.

## Verification

- `pnpm run typecheck` — 721 declarations and 70 external packages validated
- focused GLM/settings/model-list suites — 3 files, 71 tests passed after the final rebase
- independent focused review — 5 files, 105 tests passed; no blocking findings
- ESLint on changed production files
- Prettier on changed supported files
- Deno lock JSON validation
- `git diff --check`

Deno itself is unavailable locally, so the Supabase Deno test was not executed. The lockfile uses the npm 1.31.0 integrity already resolved by pnpm.

### DIFF
--- a/packages/extension/src/settingsView/frontend/components/profile/ModelSelectionList.ts
+++ b/packages/extension/src/settingsView/frontend/components/profile/ModelSelectionList.ts
@@ -139,7 +139,11 @@ export class ModelSelectionList extends LitElement {
         @change=${(e: Event) => this.handleReasoningLevelChange(model.name, e)}
       >
         <wa-option value=""> ${defaultLabel} </wa-option>
-        ${REASONING_LEVEL_OPTIONS.map(
+        ${REASONING_LEVEL_OPTIONS.filter(
+          (option) =>
+            !model.supportedReasoningLevels?.length ||
+            model.supportedReasoningLevels.includes(option.value),
+        ).map(
           (option) => html`
             <wa-option value=${option.value}> ${option.label} </wa-option>
           `,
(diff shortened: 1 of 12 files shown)


LionSR:
Addressed both review findings in 30c3c6d5e4: vocabulary-less GLM and OpenRouter paths now preserve their legacy clamps, and the dead GLM validation override is removed. I also made the nearest-tier tie-break independent of registry array order and corrected GLM  semantics so it remains enabled reasoning while only  disables thinking. Full typecheck, 71 focused tests, changed-file ESLint, and an independent review all pass.

PR #11484, opened 2026-08-27
refactor: simplify data structures flagged by the complexity audit

LionSR:
## Summary

Follow-up to the scheduled data-structure-complexity audit (0 High / 6 Medium / 3 Low findings). This PR lands fixes for 7 of the 9 findings; the remaining 2 turned out, on closer inspection, to conflict with real tested behavior and were left as-is rather than shipped incorrectly (see below).

- **`SessionState.ts`**: merged `_streamStates` and `_sessionState` — the two per-stream maps that always share the same create/clear lifecycle (`claimStreamIdentity`, `clearEphemeralStreamState`) — into one `_ephemeralState` map. `_streamMetadataCache` (a derived read cache), `_streamIncarnations` (identity generation, must outlive an ephemeral clear), and `_removedStreams` (tombstones, must outlive the very state they retire) each have a genuinely different lifetime and stay separate — merging those would have broken real invariants.
- **`runExecution.ts`** (CLI headless shutdown): replaced four of the shutdown/finalization booleans (`shutdownInterrupted`, `workflowOutputPublicationCommitted`, `shutdownArtifactFailure`, `shutdownFinalizationFailureReported`) with one `LaunchVerdict` discriminated union, so "at most one of published/interrupted" is enforced by the type instead of a prose comment. The remaining flags (`shutdownRequested`, `ownedExecutionId`, `shutdownLaunchAborted`, `presentationAttached`, `primaryRunFailure`, `finalizationCompleted`, `runResult`) are orthogonal to the verdict state machine and stay as separate variables.
- **`FileOptions.commit`**: made required like its siblings (`baseFile`/`editedFile`); removed the three duplicated `?? []` guards that compensated for the mismatch.
- **`fileListingRules.ts`**: collapsed `FileListConfig`/`PreparedFileFilters` into one `FileFilterConfig` shape shared verbatim from category selection through normalization, instead of three renamed vocabularies for the same five fields.
- **Auth session refresh**: `SupabaseSessionCoordinator` now uses the same `PQueue(concurrency:1)` single-flight primitive as `SubscriptionOAuthCoordinator`, replacing its own `async-mutex` + manual polling loop. Credential-comparison logic specific to Supabase (`isCurrentStoredSession`, `clearSessionIfCurrent`) was left untouched — it encodes real behavior the generic coordinator doesn't need.
- **`toolAttachmentExtraction.ts`**: `extractToolAttachments` now branches on the `ToolResult` variant and builds each output shape from that variant's own known fields, instead of iterating a generic `Record` against a hand-maintained strip-list. This also fixes the strip-list having drifted to drop `summary` from error results — the field that variant's schema documents as existing specifically for human-facing logs.
- **`executeCommand.ts`**: the wrapped launch input is now validated through one Zod schema at the boundary instead of six ad hoc optional-chained reads.

**Not landed** (audit claim didn't survive contact with the actual code/tests, reverted after investigation):
- A `StatusBarForegroundInput` discriminated union coupling `escapeAction` to `inputActive: true` — the real invariant is more permissive (`escapeAction` is also read on a second, `shortcutsActive === false` branch in `resolveStatusBarBindings`, confirmed by the original doc comment and by existing tests exercising exactly that combination).
- Tightening `extractTouchedFiles`'s optional chaining — an existing test (`tolerates legacy partial state slices`) exercises `stateSlices` values that don't conform to the schema's guarantees, so the extra chaining is load-bearing, not defensive habit.

## Issue acceptance gates

No linked issue — this is a direct follow-up to a scheduled audit report. All 7 landed findings' suggested alternatives from the audit are implemented as described above.

## Single source of truth

- `SessionState`'s `_ephemeralState` map is now the sole owner of per-stream live metadata + execution counters.
- `runExecution.ts`'s `launchVerdict` is the sole owner of the launch's publish/interrupt verdict.
- `fileListingRules.ts`'s `FileFilterConfig` is the sole shape for include/exclude file filters from settings load through to filter matching.
- `SubscriptionOAuthCoordinator`'s generation+`PQueue` pattern is now the shared single-flight-refresh mechanism for both OAuth coordinators.

Removed: two parallel per-stream maps, three renamed vocabularies for one filter shape, two independent mutex mechanisms for the same single-flight-refresh problem, three duplicated `?? []` guards, six ad hoc optional-chained reads. No new abstractions were introduced beyond the discriminated unions/merged types described above, each of which replaces an existing implicit invariant with one the type system checks.

## Net elements (R6)

16 files changed, +294/-209 (diffstat). Exported-symbol delta (`^[+-]export` over the diff): one rename, `FileListConfig` → `FileFilterConfig` (net 0 new exports). No new classes/interfaces beyond type consolidations that replace removed ones (`EphemeralStreamState` replaces `StreamSessionState`; `LaunchVerdict` replaces four flag declarations; `FileFilterConfig` replaces `FileListConfig`).

## Consumer counts (R8)

`FileListConfig` → `FileFilterConfig` rename: all consumers updated in this PR (`desktopFileSelection.ts`, `desktopWorkspaceIpc.ts`, `fileLister.ts`, `listing.ts`, `workspaceFileListing.ts`, plus the test file); grepped repo-wide post-change, zero remaining references to the old name.

## Host impact

Extension: `executeCommand.ts` (Zod boundary), `documentSlice.ts`/`MainApp.ts` (`FileOptions.commit`), `fileLister.ts`/`listing.ts` (filter type rename).

Desktop: `desktopFileSelection.ts`/`desktopWorkspaceIpc.ts` (filter type rename).

CLI: `runExecution.ts` (shutdown state machine).

## Scheduled refactoring

None — the 2 findings left unfixed were determined not to be real bugs, not deferred work.

## Validation

- `npm run typecheck` (all 6 sub-checks: workspace, test-kernel, agent, cli, trace-viewer, desktop) — clean.
- `npm test` — 10,036 passed, 6 skipped, 0 failed.
- `npm run lint` — clean.
- `npm run check:dead-code-ratchet` — no new findings (391 baselined, 391 found).
- `npx prettier --check` on all changed files — clean.

**Update:** addressed independent third-party review feedback — fixed a stale doc comment in `SessionFactApplier.ts` left over from the `SessionState.ts` merge (the review's other two notes were non-blocking/no-change-needed).

---
_Generated by [Claude Code](https://claude.ai/code/session_01PBqL5hjkhCsdQ6gBwWJuLf)_

### DIFF
--- a/packages/cli/src/runtime/runExecution.ts
+++ b/packages/cli/src/runtime/runExecution.ts
@@ -298,18 +298,35 @@ export async function executeCliRequest(
   let ownedExecutionId: ExecutionId | undefined;
   const launchAbortController = new AbortController();
   let shutdownRequested = false;
-  let shutdownInterrupted = false;
-  let workflowOutputPublicationCommitted = false;
-  let shutdownArtifactFailure: unknown;
-  let shutdownFinalizationFailureReported = false;
+  // Workflow-output publication and shutdown-driven interruption race on the
+  // same synchronous tick (see tryCommitWorkflowOutputPublication and the
+  // onShutdown handler below): at most one may own the terminal verdict for
+  // this launch. Modeling that as one variable makes "both committed and
+  // interrupted" unrepresentable instead of relying on two booleans staying
+  // in sync by hand. The artifact-failure and report-dedupe bookkeeping are
+  // only ever meaningful once interrupted, so they live on that variant too.
+  type LaunchVerdict =
+    | { readonly kind: 'undecided' }
+    | { readonly kind: 'published' }
+    | {
+        readonly kind: 'interrupted';
+        artifactFailure: unknown;
+        finalizationFailureReported: boolean;
+      };
+  let launchVerdict: LaunchVerdict = { kind: 'undecided' };
   const reportFinalizationFailure = (error: unknown): void => {
     session.interactions.emit('requestShowError', {
       message: toErrorMessage(error),
     });
   };
   const reportShutdownFinalizationFailure = (error: Error): void => {
-    if (shutdownFinalizationFailureReported) return;
-    shutdownFinalizationFailureReported = true;
+    if (
+      launchVerdict.kind !== 'interrupted' ||
+      launchVerdict.finalizationFailureReported
+    ) {
+      return;
+    }
+    launchVerdict.finalizationFailureReported = true;
     reportFinalizationFailure(error);
   };
   let settleLeaseAcquired: (
@@ -330,12 +347,12 @@ export async function executeCliRequest(
   // Both callbacks enter on this event loop, and neither yields between the
   // check and assignment. Exactly one can therefore own the terminal verdict.
   const tryCommitWorkflowOutputPublication = (): boolean => {
-    if (shutdownInterrupted) return false;
-    workflowOutputPublicationCommitted = true;
+    if (launchVerdict.kind === 'interrupted') return false;
+    launchVerdict = { kind: 'published' };
     return true;
   };
   const finalizeShutdownStatus = (): Promise<boolean> => {
-    if (!shutdownInterrupted) return Promise.resolve(false);
+    if (launchVerdict.kind !== 'interrupted') return Promise.resolve(false);
     shutdownStatusFinalized ??= (async () => {
       const executionId = await leaseAcquired;
       if (!executionId) return false;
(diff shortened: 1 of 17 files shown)


LionSR:
The `track` and `claude-review-anthropic` checks both failed at `Run Claude Code` with quota/rate-limit exhaustion (fallback providers skipped too — see the [track job log](https://github.com/LionSR/TeXRA/actions/runs/33056870348/job/98465737354)), not from anything in this PR's diff. That matches the bot's own comment above. I've added the `re-review` label to retry per its instructions; not re-running further myself since I have no other means to retry these two jobs. Keeping this PR watched until they (and the remaining in-progress checks) go green.

---

PR #11490, opened 2026-08-27
test(progress-view): cover unavailable Clean toolbar

### LINKED ISSUE #11331: Fix Clean button no-op for unavailable chat (ToolUse) streams
## Context

Follow-up from #11330, from #11329 (refactor(runtime): deep clean of the session ownership surface).

The PR collapsed `HELD_STATE_BUTTONS`/`UNCLASSIFIED_STATE_BUTTONS`/`MALFORMED_STATE_BUTTONS` into one `UNAVAILABLE_STATE_BUTTONS` bucket and put `CLEAN_STREAM_BTN` in it unconditionally, with a comment claiming Clean "deletes the run." This was flagged inline by `claude[bot]` against the exact merged head (`daeb815b16`) and never fixed before merge.

## Details

`CLEAN_STREAM_BTN` dispatches `PROGRESS_VIEW_COMMANDS.CLEAN_STREAM` → `ProgressWorkflowActionsController.runFileOperation(stream, 'clean')`, which goes through `withWorkflowConfig` (`src/controllers/progressView/ProgressWorkflowActionsController.ts:96-105`):

```ts
if (!config || config.agentCategory !== AgentCategory.Workflow) return;
```

This is the "Workflow toolbar (diff / pack / clean)" family (see the comment at `src/controllers/progressView/ProgressViewCommandHandlers.ts:519`) — it cleans generated output files for a workflow run and no-ops for anything else. It was never wired to delete a run/history record.

Before #11329, `CLEAN_STREAM_BTN` only appeared in the narrow `MALFORMED_STATE_BUTTONS` bucket. #11329 folds `held`/`unclassified` into the much more common `unavailable` bucket (any stream held by another TeXRA window) and puts Clean in `UNAVAILABLE_STATE_BUTTONS` unconditionally — so a `ToolUse` chat stream held by another window now shows an enabled "Clean" button in the header that silently does nothing when clicked, while the tab-rail's `x` (wired to the real `deleteStream`) is the only control that actually clears the run. This contradicts the PR's own D9 promise ("Delete is the user's only action" for an unavailable run) and the repo's "one home per user action" / "silent degradation is a defect" rules.

Fix by either:
- Gating `CLEAN_STREAM_BTN` out of `UNAVAILABLE_STATE_BUTTONS` for non-workflow (`ToolUse`) streams, or
- Fixing the button/comment so Clean doesn't imply it deletes the run for streams it can't act on.

Relevant file: `packages/extension/src/progressView/frontend/components/StreamHeader.ts:104-135`

## References

- PR: #11329
- Review: https://github.com/LionSR/TeXRA/pull/11329#discussion_r3838250460 (claude[bot], posted against the merged head `daeb815b16`)

LionSR:
Closes #11331

Adds one focused `StreamHeader` regression test for the existing toolbar policy:

- a held ToolUse stream does not render the workflow-only Clean control
- an unreadable Workflow stream keeps Clean visible but disabled
- a completed Workflow stream keeps Clean enabled to remove generated output files

This preserves tab-rail Delete as the run-deletion action and verifies `ProgressWorkflowActionsController` clean semantics through its existing focused suite.

### DIFF
--- a/src/test-kernel/progressView/StreamHeader.vitest.ts
+++ b/src/test-kernel/progressView/StreamHeader.vitest.ts
@@ -15,6 +15,10 @@ import {
 } from '@shared/schemas';
 import { APPROVAL_BYPASS_BADGE } from '@shared/copy/approvalBypass';
 import { DELEGATION_APPROVAL_COPY } from '@shared/copy/delegationApproval';
+import {
+  streamHeldMessage,
+  streamUnreadableMessage,
+} from '@shared/streams/streamStatusDisplay';
 
 // Local file imports
 import {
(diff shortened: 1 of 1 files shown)


LionSR:
Addressed the review feedback in be5c79b102:

- collapsed the duplicate status-detail matrix into one held ToolUse case and one malformed/unreadable Workflow case, preserving both relevant unavailable variants
- asserted that unavailable Workflow Clean is present and not hidden, as well as disabled

Focused suites and test-kernel typecheck pass.

LionSR:
Addressed the follow-up visibility feedback in 66a3a8e4c2. The unavailable Workflow assertion now verifies the production visibility markers directly: it lacks `toolbar-button--hidden`, has `aria-hidden="false"`, and remains disabled. Focused suites and test-kernel typecheck pass.

LionSR:
Addressed the final review suggestion in 859db9cd48 by folding the held ToolUse, unreadable Workflow, and completed Workflow assertions into one focused regression test. It still covers the requested category, unavailable-state, and workflow-clean semantics.

PR #11498, opened 2026-08-27
feat(workflow): separate declared plan labels from issued calls

LionSR:
## Problem

A `delegate_multi_agents` proposal was rendered as one aggregate — `6 tasks · 1 phase`, the first *declared* phase labelled as active, `meta.tasks` labels listed as if they were the calls that will run, run-level files labelled as every call's inputs, and one agent/model as everyone's. `meta.tasks` is `{ id, label, phase? }` and nothing more; a script may declare none and issue everything from runtime data; each call picks its own agent, model, schema, and files. At runtime the cards erased the same distinction: `planned`/`stageBlocked`/`queued` collapsed to one "Planned", and `agent`, `files`, the structured/document contract (never stored anywhere), and the live model were dropped — so real concurrency was invisible.

## What this does (slice 1 of the design)

**Snapshot (one owner).** `WorkflowExecutionCall` gains `issued: true` (stamped by `issueCall`, cleared on hydrate for non-reusable calls), `kind` (`document` | `structured`, from `options.schema`), and the script-declared `model` at issue time. `queueCall` still drops a prior attempt's resolved facts (two persistence tests pin that ruling) but restores the declaration.

**Card projection.** `WorkflowCallProgress` gains `declared` (unissued plan label, whatever stage gate it sits behind) and `queued` statuses, plus `kind`/`agent`/`model`/`files` on any status; a live card re-emits when the host resolves agent/model. Shared copy renders `Document · polish · gemini37f · introduction.tex · 1m 12s · $0.31`; a declared row stays a bare label. Extension/desktop and CLI (dashboard, transcript, headless) consume it unchanged; the CLI dashboard now prefers the card's own model over the stream config. Host-stop recovery settles only `running` rows as failed.

**Proposal card** (extension/desktop, CLI modal, headless summary) via one shared copy module: `2 phases · 3 declared items` or `2 phases · calls issued at runtime` (never `0 tasks`); `Defaults: agent (model) — each call may name its own agent and model.`; declared items grouped under their declared phase with an explicit plan-label note; `Files available to the script`; the fake "active phase" (`phases[0]`) is gone; the CLI modal shows the file groups it used to drop. No proposal wire-schema change.

**Design record:** `docs/proposals/2026-08-28-workflow-plan-vs-issued-calls.md` — findings verified at HEAD, the conceptual model, this slice, and the per-call / per-phase admission gate (slice 2: `admitCall` before `queue.add`, reusing the existing proposal panel per issued call) with the engine invariants it must respect and the rejected alternatives.

## Tests

One regression test at the durable boundary (`WorkflowScriptProgressBridge`: a declared label carries no facts; the issued structured/document calls carry kind/agent/model/files; the never-issued label settles not-reached). Existing pins updated to the new copy/vocabulary; the persistence assertion on requeue was tightened from "model undefined" to "the script's declared model, never the prior attempt's resolved one".

## Verified

- `npm run typecheck` (all per-package tsconfigs) — clean
- `npm run lint` — clean
- `npx vitest run` — 822 files / 10 030 tests passed (1 file / 5 tests skipped, pre-existing)
- `scripts/check-browser-safe-utils.mjs` — 65 total, 6 reachable (unchanged)
- `docs/scripts/check-root-docs.mjs` — boundary OK
- prettier — clean


LionSR:
Addressed all six review findings in e82bf22be0: hydrate also resets `kind`/`agent`; declared rows never borrow the stream model in the CLI; the retry running-latch is gone (a retried call shows `queued` when it really waits); empty declared phases stay in the proposal details; the headless summary labels defaults/files/description as container facts; `/executions` exposes `issued` + `kind`. Affected suites, typecheck, lint, ratchet green locally.

PR #11528, opened 2026-08-28
refactor(schemas): delete the legacy workflow-snapshot reader

LionSR:
## What

Deletes the `z.preprocess` compatibility reader on `WorkflowExecutionCallSchema` that stripped `stageTitle` (removed 2026-08-18), `blockedReason` and `timestamps.queuedAt` (removed 2026-08-28), and mapped a persisted `status: 'starting'` to `running`. The strict call shape is now the schema; a `meta.json` snapshot of an interrupted run written by an older build fails `hydrate()`/`/executions` loudly rather than being translated.

Why now rather than 2026-11-28: these shapes are intermediate-era data from an opt-in prototype feature; the repo's ruling (#9590) is to delete such readers early and loudly, not to age-gate them.

## Net elements (R6)

- `git diff --stat origin/main`: 3 files, +11 / −46 (net −35); production 1 file, −29.
- Exported symbols: 0 added, 0 deleted (`WorkflowExecutionCallShape` was file-local and is renamed to the schema). Declarations: 0. Files: 0.

## Consumer counts (R8)

No emit path or public symbol deleted. The reader had one consumer (`WorkflowExecutionSnapshotSchema.calls`), which now takes the strict shape directly. `rg "stageTitle\b|blockedReason|queuedAt" src packages/*/src` → only `stageTitleFor` and the one test, which now asserts the loud rejection.

## Verified

- `npm run typecheck` — clean · `npm run lint` — clean · `npm run check:dead-code-ratchet` — no new findings · prettier — clean
- vitest over agent/tools/transcript/shared suites — 60 files / 896 tests passed

### DIFF
--- a/docs/proposals/2026-08-28-simplification-survey-multi-agent-dispatch.md
+++ b/docs/proposals/2026-08-28-simplification-survey-multi-agent-dispatch.md
@@ -45,8 +45,11 @@ design, the QuickJS bridge shape, per-call replay.
 - **The multi-agent proposal card's row budget was one row short** (the
   workflow branch of `agentProposalMetadataRows` omitted the outer margin
   row), overflowing by one row in feedback mode.
-- **A persisted `status: 'starting'` snapshot would fail resume loudly** once
-  the status is removed; the existing preprocess strip maps it to `running`.
+- **A persisted `status: 'starting'` snapshot fails resume loudly** once
+  the status is removed. The compatibility preprocess that briefly mapped it
+  (and stripped `stageTitle`/`blockedReason`/`queuedAt`) was deleted the same
+  day — intermediate-era data is disposable, not age-gated; an interrupted
+  run's `meta.json` from an older build is rejected loudly on resume.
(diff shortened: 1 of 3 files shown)


LionSR:
Re the review notes: (1) the R8 grep in the body was mis-quoted — the reproducible form is `rg 'blockedReason|queuedAt|"starting"' src packages/*/src` → 0 production hits, plus `rg '\bstageTitle\b'` → only the derived `stageTitle:` display field in `workflowSummaryView.ts` and a local in `progressBadgeFormatter.ts`, neither the persisted key. (2) Classification: these shapes were written by 0.40.2–0.40.5 builds of the opt-in multi-agent workflow (kill switch off by default) and only into an *interrupted* run's `meta.json`; the maintainer ruling for such intermediate-era state is delete-early-and-loud rather than a dated horizon. A stale run from an older build now fails resume with the strict-schema warning and the model is told to start with a new `meta.name`; no release note beyond the changelog line.

PR #11550, opened 2026-08-28
refactor(cli): stream state reads the shared substrate — status, usage, run config, provenance

LionSR:
"Same state, different rendering; no CLI-specific session state." Four places where the CLI kept a copy of a fact the shared substrate already owns now read the substrate. No new plane, bus, vocabulary, or subscribe surface; every item deletes the layer it replaces.

## What moved

**P1 — status triple → status machine.** `StreamSlice.status/substate/runStartedAt` mirrored `StreamStatusMachine`, which writes its entry *before* publishing the `status` fact (`StreamStatusService.ts` `transition`), and whose `StreamPhaseState.runStartedAt` is the single run-window stamper — it rides the wire (`BackendOwnedFieldsSchema.runStartedAt`) and the progress board already renders elapsed from it. The former private `stateFor` becomes the public per-stream read `getStreamState(stream)`, reached through `SessionState.streamStatus`; renderers call `streamPhaseFor`, a slice-gated read of it (`getAllStreamStates()` stays for the whole-map cases and allocates).

Deleted with it: `PatchableStreamSlice`, `StreamSlicePatch`, `patchStream`'s lifecycle re-spread, `streamSliceWithStatus`, `setStreamStatusInCliState`. The removed/retired refusal that writer enforced survives as `cliStreamAcceptsStatus`, checked by the adapter before it folds the log and releases residency. The adapter's sequencing is byte-for-byte unchanged: fold under the new status → release residency for a stream that just left its active phase → focus-return → applier.

**P2 — headless run config → snapshot store.** `RootRunConfigState`/`applyRunConfig` copied `run.config` into three fields. `StreamSnapshotStore` already retains the whole `AgentConfig` and serves it from `getRunMetadata(stream, { quiet: true })`; `plannedRounds` is `getAgent(config.agent, Workflow)?.rounds`, a registry lookup, not a fact. **Precondition verified:** the store calls `attachSessionEvents(events)` in `SessionHandle`'s constructor, and `SessionEventHub` dispatches subscribers in insertion order, so the store has accumulated the config before the headless renderer's own run-fact callback runs. `rootStreamStatus` → the machine; the run-config repaint stays (it is the live line's subject), it just copies nothing now.

**P3 — latest-usage gauge → cumulative projection.** `StreamSlice.usage` and the `onRunUsageChanged` patch are gone; readers project `StreamArtifactProjection.cumulativeUsage`. `streamPreferredUsage` was a two-source precedence rule with one source left, so it is inlined at its three call sites.

**P4 — store provenance query.** `hydratedArtifactStreams` / `beginLoadedStreamsReconcile` / `markArtifactStreamHydrated` shadowed a fact `StreamSnapshotStore` held privately (`hasDiskProvenance` + the eager-apply overlays). The store publishes it as `hasProvenance(stream)`. **Precondition verified:** `load` calls `evictStreamsExcept` synchronously before its async seed and `evict` drops the whole record, so an evicted stream reports no provenance by construction — that *is* `dropStale`/`reconcile`. The CLI keeps only the revision bump (`bumpStreamArtifactRevision`, promoted from private).

**Deliberately left:** P6, the `StreamSlice.bypass` mirror (−8). It is stream-scoped host permission state with a writer of its own; folding it needs the approval-plane ruling this PR does not ask for.

## Declared behavior changes

- **Headless elapsed origin.** The live line's elapsed now measures the status machine's run window (opening at RUNNING, the same origin the TUI status bar and the progress board use), falling back to renderer attach time only before any window exists. Previously it always ran from attach time.
- **Status-bar token fallback.** Before a stream's first `context.state`, the usage segment shows the store's cumulative input tokens rather than the last call's. After it, the gauge is unchanged (`contextState` owns window/used/utilization).
- **`/status` access route** reads the cumulative projection's `usageRoute` (`sumUsageStats` carries it when unmixed) instead of the last call's.

## Ratchet rider

`config/ratchets/store-public-surface-baseline.json`: `StreamSnapshotStore` **+1 unit** (`hasProvenance`), 22 → 23. Accounting: it converts one caller-visible contract *from* an 86-line host-side shadow of the store's private bookkeeping *into* one honest query on the owner — the surface grows by one method and the CLI loses a set, three functions, and the eviction-reconcile protocol built on top of them. `host-agent-import-baseline` does **not** move: `StreamPhaseState` is re-exported from the `@agent/runtime` barrel the CLI already imports, so no new deep-import specifier appears.

## Net elements (R6)

**Deleted — 18.** Slice fields `status`, `substate`, `runStartedAt`, `usage` (4). Types `PatchableStreamSlice`, `StreamSlicePatch`, `RootRunConfigState` (3). Functions `streamSliceWithStatus`, `setStreamStatusInCliState`, `applyRunConfig`, `beginLoadedStreamsReconcile`, `markArtifactStreamHydrated`, `streamPreferredUsage` (6). State `DefaultRunProgressRenderer.config`, `hydratedArtifactStreams` (2). Parameters/props `workflowTaskMetadata(child)`, `WorkflowTaskRow.child`, `focusedChildFollowUpRoute({ streams })` (3).

**Added — 12.** Functions `sessionStreamPhase`, `streamPhaseFor`, `cliStreamAcceptsStatus`, `StreamSnapshotStore.hasProvenance` (4). Private methods `rootPhaseState`, `rootRunConfig` (2). One type re-export, `StreamPhaseState` from `@agent/runtime` (1). Parameters `statusBarStreamTarget({ phaseOf })`, `focusedChildFollowUpRoute({ phaseOf })`, and the `status` option on the three static-transcript builders (5).

**Renames/visibility, counted as zero:** `stateFor` → public `getStreamState`; `startedAt` → `attachedAt`; `rootStreamStatus` → `rootStreamPhase` (kept, but narrowed: it is now only the ordering companion of `rootPhaseText`, since `StreamStatusMachine.transition` publishes on substate-only changes and a description must survive one — terminal-ness and the run window read the machine); `bumpStreamArtifactRevision` private → exported.

**Net: −6 elements.** LOC: production `23 files, +315 −385` (**−70**). `cliState.ts` 811 → 732, `subscribeStreamArtifacts.ts` 275 → 208, `sessionSignalsAdapter.ts` 276 → 262; `runProgressRenderer.ts` 518 → 533 (**+15**, and honestly: the copied state and its writer are gone, but the two reads that replace them carry the ordering and subscriber-order arguments in comments). Tests/harness `+394 −231`, all of it the mechanical status-driver migration plus one new 44-line test-support file. No new production file.

## Consumer counts (R8)

| symbol | before | after |
| --- | --- | --- |
| `StreamSlice.status` / `.substate` / `.runStartedAt` / `.usage` | 35 read sites in `packages/cli/src`, 6 in `packages/cli/scripts` | 0 — fields deleted |
| `streamPhaseFor` (new) | — | 21 call sites across 12 files |
| `readStreamArtifacts(…)?.cumulativeUsage` | 1 (inside `streamPreferredUsage`) | 5 direct call sites |
| `setStreamStatusInCliState` | 1 production caller (`sessionSignalsAdapter`), 9 harness/validator refs, 54 refs across 9 suites in `src/test-kernel` | 0 — deleted |
| `setCliStreamPhase` (test support, new) | — | 56 refs across the same 9 suites, replacing that writer one-for-one |
| `StreamStatusMachine.getStreamState` | 0 (private `stateFor`, 2 internal callers) | 2 external (`sessionStreamPhase`, `rootPhaseState`) + the same 2 internal |
| `StreamSnapshotStore.hasProvenance` | — | 1 (`readStreamArtifacts`) |
| `bumpStreamArtifactRevision` | 3 (private) | 8 production call sites |
| `markArtifactStreamHydrated` | 5 | 0 — folded into the bump |
| `beginLoadedStreamsReconcile` | 1 | 0 — the store's eviction is the reconcile |
| `streamPreferredUsage` | 3 | 0 — inlined |

## Tests

Zero net new tests. `SubscribeStreamArtifacts.vitest.ts`'s single case is rewritten one-for-one to the new boundary (it drove the deleted bookkeeping; it now drives a real `load` and asserts the store's provenance gate). Everything else is migration: 9 suites drive `setCliStreamPhase` — one test-support helper for the "seed the machine, mint the slice" pairing that appeared 56 times — and the fixtures that carried a `usage` literal on a slice now emit the `usage` run fact the store accumulates. Two suites that had no bound `SessionState` now bind one, because a phase read has to reach the machine.

## Record

`docs/proposals/2026-08-15-single-substrate-hosts-as-renderers.md` §3/§4 status lines and §8 are reconciled: `contextState` and `runStartedAt` **landed**; `thinkingActive` **withdrawn** (fold output, no fact-rail producer, one consumer — the `compactingActive` ruling applies verbatim); root-run stream identity **closed as presentation state** (the CLI's claim precedes stream existence, and `executions.getHandle(id).streamId` already answers the live query); run input files + `plannedRounds` **closed** (the store already carried them). Two new §8 fence rows: the Lit log-rail fold of `contextState` (`LitSessionRenderer.ts:197-201` drops the invalidation on purpose — Lit restores that snapshot from the transcript rail, which also repopulates a reopened stream, and two writers for one footer would be worse), and durable-rail vs ephemeral-rail facts.

**Rebased onto `d2d5c202df` (#11545, transcript plane), then onto `1b94d7f6a1` (#11548, transcript render seams).** Three textual conflicts across the two rebases, all one-for-one merges: `SubagentListDisplay.vitest.ts`'s `workflowAgentSlice` (keep #11545's `workflowPhaseGrouping` derivation *and* this PR's substrate seeding, in that order); `ConversationPane.tsx` (keep #11548's `pendingTranscriptEntries` call shape, apply this PR's `streamPhaseFor` status argument); one import block in `ConversationTranscript.vitest.ts`. Everything else merged cleanly.

**Harness fixtures drive the machine directly.** `packages/cli/scripts/tui-harness.tsx` used to write the slice mirror for display fixtures. It now calls `seedStreamStatusForTest` (the kernel suites' helper) through a local `seedHarnessStreamPhase`: it writes the machine the way a real transition does *before* it publishes, but emits no fact. Two reasons it cannot use a production transition: `StreamStatusMachine` stamps its own `Date.now()` run window and no fact lets a caller supply one, so a scenario like `single-run-liveness` could not show a backdated elapsed; and a published status makes the adapter re-fold the stream from a log that has none of the fixture's synthetic rows, which is what dropped the child description in `subagent-list-focus-full-frame`. Fixtures that exercise the fact pipeline itself (`runChildEventOrderFixture`, the interrupt paths) keep the real transitions.

## Verified

- `npx prettier --check` on every changed file — clean.
- `CI=true npm run typecheck` — 0 errors.
- `npm run lint` — clean.
- `CI=true npm run check:dead-code-ratchet` — `379 combined vs 379 baselined`, no new findings.
- `FORCE_COLOR=0 npx vitest run src/test-kernel/cli src/test-kernel/agent/runtime src/test-kernel/transcript src/test-kernel/controllers src/test-kernel/architecture` — `235 files passed, 3608 passed | 1 skipped`.
- Also ran `src/test-kernel/{desktop,extension,agent,tools}` (the status machine is cross-host): `348 passed | 1 skipped`, one unrelated vitest worker-exit flake.
- `validate:tui` over the 17 scenarios the `build artifacts (linux)` job runs: `validate-tui: all 17 scenario(s) passed`. The two full-suite scenarios that fail locally (`focused-stopped-subagent-tab-focus`, `focused-stopped-subagent-submit`) fail **identically on an untouched `origin/main` build** in a side worktree, and neither is in the CI set.

## Note for the child-elapsed fix

`SubagentList.tsx` rows read elapsed from `streamPhaseFor(streamId)?.runStartedAt` (`cliState.ts`), which is the machine's run window for a stream the CLI holds a slice for. For a roster row whose child stream has no slice yet, the shared read path is the roster row itself: `visibleSubagentRows(parentId, childRosters)` → `ActiveChildInfo.startedAt` (`streamState.ts`), written by the applier. Both are substrate reads; neither needs a new field.

### DIFF
--- a/config/ratchets/store-public-surface-baseline.json
+++ b/config/ratchets/store-public-surface-baseline.json
@@ -42,6 +42,7 @@
       "getRunUsage",
       "getWorkPlan",
       "hasPersistedWorkPlan",
+      "hasProvenance",
       "listPersistedStreams",
       "listStagedDeletions",
       "load",
--- a/docs/proposals/2026-08-15-single-substrate-hosts-as-renderers.md
+++ b/docs/proposals/2026-08-15-single-substrate-hosts-as-renderers.md
@@ -51,6 +51,15 @@
 > (`sessionSignalsAdapter` 372 → 312 against ~90, `runProgressRenderer`
 > 573 → 535 against ~395, `cliState` 929 → 876 against ~490) — read the §3
 > table as a direction, not a budget.
+>
+> **Reconciled again (2026-08-28), after the CLI status / usage / run-config /
+> provenance PR.** The §4 promotion list is closed: `contextState` and
+> `runStartedAt` **landed** (each with one §8 fence row), `thinkingActive` is
+> **withdrawn** (fold output, no producer, one consumer), and root-run stream
+> identity and run input files + `plannedRounds` are **closed as
+> non-promotions** — the first is presentation state with an existing
+> execution-keyed query, the second was already on `StreamSnapshotStore`. The
+> per-row §3 and §4 lines below carry the detail.
 
 ## 0. The directive as a ruling, and what it does to prior rulings
 
@@ -232,7 +241,13 @@ Then, container by container:
   composition: 30 fields → 15, every mirror deleted, but the slice is still a
   standalone interface rather than the `StreamState & CliOnlyFields` shape this
   row named, and the file is 876 L against the ~490 target. The §0.1-item-8
-  supersession is executed.
+  supersession is executed. **Extended by the status/usage PR:** the lifecycle
+  triple (`status`, `substate`, `runStartedAt`) and the latest-usage gauge are
+  gone too — 15 fields → 11 — and with them `PatchableStreamSlice`,
+  `StreamSlicePatch`, `streamSliceWithStatus`, and `setStreamStatusInCliState`.
+  Renderers read `streamPhaseFor`, a slice-gated read of
+  `StreamStatusMachine.getStreamState`. What survives is transcript-fold
+  output plus terminal modality.
 - `sessionSignalsAdapter.ts` — **LANDED #10892/#10895/#10932** for the
   mechanism (the per-field patch forwarders and the dual status path are gone;
   `subscribeStreamStatus.ts` is deleted), **not** for the arithmetic: 372 →
@@ -240,17 +255,33 @@ Then, container by container:
 - `runProgressRenderer.ts` — **LANDED #10932**: a real headless
   `SessionRendererPort` implementation over the shared applier, `RenderState`'s
   six fields down to a three-field `RootRunConfigState`. Net −38, not −175.
+  **`RootRunConfigState` is now gone too:** the agent name, input label, and
+  planned rounds are read from
+  `snapshots.getRunMetadata(root, { quiet: true }).config` at paint, and the
+  root phase and its run window from the status machine. The renderer holds one
+  state field of its own (`rootPhaseText`, whose whole semantic is
+  last-write-wins ordering between a status label and a description) plus an
+  attach-time elapsed origin for the window before any run window exists.
 - `subscribeStreamArtifacts.ts` — **PARTIAL #10718/#10735/#10895.** The
   slice-copying is gone (renderers read the canonical projection), but the file
   survives at 222 L holding the async disk-preload edge and its invalidation.
   The "−135 to zero" line was wrong: there was a real asynchronous concern
-  underneath the copying.
+  underneath the copying. **The hydration set went next:**
+  `hydratedArtifactStreams`, `beginLoadedStreamsReconcile`, and
+  `markArtifactStreamHydrated` shadowed a fact the store already held
+  privately, so the store publishes it as `hasProvenance(stream)` (disk
+  provenance established, or a live write eagerly applied ahead of any seed)
+  and the CLI keeps only the revision bump. The `load`-eviction reconcile
+  disappears with it: `load` evicts synchronously, and an evicted record
+  reports no provenance by construction.
 - `streamViews.ts` — **LANDED #10932**; the CLI calls `buildStreamTabInfo`, so
   the ext/CLI name drift this row named is closed.
-- `statusBarDisplay.ts` context gauge — **OPEN.** Still
-  `MODEL_CONFIGS[model]?.contextWindow`; the correctness bug this row called
-  "a correctness fix wearing a refactor" is live, gated on the `contextState`
-  promotion (§4).
+- `statusBarDisplay.ts` context gauge — **LANDED.** `formatUsage` reads
+  `StreamExecutionState.contextState` (window, used input tokens, and the
+  handler's own `utilizationPercent`); the `MODEL_CONFIGS[model]?.contextWindow`
+  re-derivation and the correctness bug behind it are gone. Its pre-first-
+  `contextState` fallback now shows the store's cumulative input tokens rather
+  than the last call's.
 - `resumeHint.ts` targets — **LANDED #10754** via the `resumeEligible` roster
   field.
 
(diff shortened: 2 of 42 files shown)


LionSR:
Check more surface areas

PR #11590, opened 2026-08-29
feat(workflow): record the declared plan so the dashboard can show unopened phases

LionSR:
## What

The CLI workflow dashboard could not show the next round: a script's `meta.phases` / `meta.tasks` reached hosts only through the approval proposal, and the progress projection opens a phase stage only when the run reaches it (`workflowScriptRun.ts` skips `waiting` stages on purpose).

This adds one run-scoped fact and unions it into the dashboard:

- **`workflow.plan` trace event** (`src/agent/trace/events.ts`), emitted once per attempt from the first execution snapshot in `runPersistedWorkflowScriptWithProgress` — every stage with its order, every call with its phase — before any stage or card. Not added to `RUN_FACT_EVENT_TYPES`: the fact rail, NDJSON output, and the progress view are untouched.
- **Transcript recorder** writes it as an `INTERNAL` `workflowPlan` marker (`WorkflowPlanMarkerSchema` beside `WorkflowAttemptMarker`), the same shape as the existing `workflowAttempt` marker. No new `messageType`; older trace viewers degrade it like any `INTERNAL` row.
- **CLI fold** (`transcriptFold.ts`) parses the marker into a new `StreamSlice.workflowPlan` (newest attempt wins, matching `latestWorkflowAttemptEntries`); a malformed marker is logged at `warn`, never swallowed.
- **`workflowDashboardModel`** unions the plan with task groups and cards: a plan phase with no group becomes a declared group in plan order; a plan task with no card becomes a `declaredTask` under its phase; a card always wins over its plan entry. `WorkflowPhaseGroup` gains `opened` and `declaredTasks`.
- **Rendering**: declared phases use a hollow `◇` (new `STATUS_DIAMOND_OUTLINE`) and tally `N declared`; opened phases append `· N declared` while the plan still holds unissued tasks; declared tasks are disabled rows in the wide task column (nothing to focus, kill, or retry).

Why not open waiting stages: the `◆` divider is projected from `stage.start` (`projectTranscriptRow.ts`), so it would print into scrollback before the phase began — on the extension too.

## Context

Slice 2 of the workflow-dashboard redesign (phase-tab popup); follows #11588. The popup itself is the next slice and consumes `declaredTasks` as-is.

## Verified

- `npm run typecheck` — clean
- vitest: `src/test-kernel/{cli,transcript,shared,progressView}`, `agent/Workflow*`, `tools/WorkflowScript*` — 256 files, 3,853 passed; `src/test-kernel/architecture` — 15 files, 101 passed
- `npm run check:dead-code-ratchet` — no new findings (every new export has a consumer)
- eslint + prettier on changed files — clean
- Two boundary tests: the bridge emits exactly one plan before the first stage and card (`WorkflowScriptProgressBridge.vitest.ts`); the model lists a declared phase and declared tasks without duplicating a card, and keeps declared tasks off the keyboard row list (`WorkflowDashboardModel.vitest.ts`).

### DIFF
--- a/packages/cli/src/chat/tui/App.tsx
+++ b/packages/cli/src/chat/tui/App.tsx
@@ -303,12 +303,23 @@ export function App(props: AppProps): React.JSX.Element {
   // The only derivation: `SubagentList` renders this instance and
   // `ConversationRegion` budgets its rows from it, so rows cannot be grouped,
   // ordered, or deduplicated twice and drift the keyboard off the screen.
+  // A plan-only phase a finished run never reached is nothing to list; only
+  // a known, no-longer-active phase counts as settled (an unknown phase is a
+  // stream still being created, not a finished run).
+  const workflowRootPhase =
+    workflowDashboardRoot === undefined
+      ? undefined
+      : streamPhaseFor(workflowDashboardRoot.streamId)?.phase;
+  const workflowRunSettled =
+    workflowRootPhase !== undefined && !isActivePhase(workflowRootPhase);
   const workflowDashboard = useMemo(
     () =>
       workflowDashboardRoot
-        ? workflowDashboardModel(workflowDashboardRoot, columns)
+        ? workflowDashboardModel(workflowDashboardRoot, columns, {
+            runSettled: workflowRunSettled,
+          })
         : undefined,
-    [columns, workflowDashboardRoot],
+    [columns, workflowDashboardRoot, workflowRunSettled],
   );
   // Stream-less approvals fold onto the root of the visible surface: the
   // scoped child-list root while one replaces the session list, else the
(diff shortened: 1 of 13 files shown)


LionSR:
Review triage for the four P2s on the second push:

- **Phase-less `meta.tasks` dropped** — fixed: they now land under the trailing `Unphased` heading their cards use once issued.
- **Stale plan after a relaunch that fails before recording its plan** — fixed: the fold clears the plan on the `workflowAttempt` marker, so a new attempt has no plan until it records one.
- **Joining phases by redacted title** — declined. The recorder redacts stage labels and plan titles through the same `redactSecrets`, so the join is consistent on both sides; two *distinct* titles that redact to the same text requires secrets in phase titles, and the failure mode is a merged heading, not lost work. Not worth a second identity on the wire.
- **Compacted cards listed as unissued** — declined as unreachable: `compactWorkflowEntries` keeps the newest 2,000 dashboard rows, and the engine's lifetime call cap is 200 per run (`runWorkflowScript.ts`), so no run can produce enough cards for a card to be compacted away.

PR #11594, opened 2026-08-29
feat(cli): show a running workflow as a phase-tab popup, never as a viewport

LionSR:
## What

A running `dispatch_multi_agent` workflow is no longer a viewport you focus. It stays one row in its parent's session list; Enter (or Alt‑N, or Esc from one of its agents) opens a **phase-tab popup** over the parent — the same foreground mechanics as the Ctrl-T reader (row-budgeted, Esc restores the parent untouched, no scrollback repaint). Design D on the board: https://claude.ai/code/artifact/9aa29d74-53e9-45ea-808e-b24f7fbada94#d

Inside the popup:

- **Tab strip**, one tab per phase with its own tally (`◆ Derive (1/2) · 13/75 · 20 running · 2 failed    ◇ Compare (2/2) · 75 declared`); ←/→ switches. Declared phases from the `workflow.plan` fact (#11590) are tabs before they start.
- **One glyph per call** under the tabs (`✗☐☐☐□□□…`), capped at the frame width with `+N` — readable without colour because every state already has its own glyph.
- **Attention first, volume collapsed**: awaiting approval, then failed, then running (transcript order within), then `▸ N queued` / `▸ N done` / `▸ N declared` groups that Enter unfolds in place. Rows scale with *states*, not agents; the `Select` windows the rest with `… N more`.
- `/` type-to-filter, `f` next failed, `s`/`r` skip/retry, `x` kill, Ctrl‑T for the workflow's own log, Esc close (or clear the filter first).
- Enter on a task focuses that agent; Esc from the agent comes back to the parent with the popup reopened where it was.

One rule carries all of it: `presentStream` (`state/childControls.ts`) is the single owner of "show me this stream" — a `multiAgentWorkflow` target lands on its parent with the popup open. `App`'s focus path, the approval-announce path, and the finished-child → owner return all go through it, and it reads the roster row's identity/parent for the interval before `run.start` lands. While the popup (or the workflow's Ctrl‑T log) is in the foreground, that stream is the approval-visible one, so a call-review or plan approval bound to the workflow surfaces over it and hands back when answered. The popup's view is keyed by the workflow stream, so Enter into an agent and Esc back land where the user left it.

## What this deletes

Everything that existed to make the workflow stream a viewport or to draw the two-column dashboard:

- The live status band (`workflowRunStatusSummary`) and its knip baseline entry.
- The two-column `WorkflowDashboard`, `PhaseHeader`, phase/task `ChildListValue` kinds, `workflowDashboardPanelItemCount`, `workflowDashboardSelection`, `WORKFLOW_DASHBOARD_WIDE_MIN_COLUMNS`, and all of `App`'s dashboard selection plumbing.
- Two smoke scenarios that encoded the viewport model (`workflow-timeline`, `workflow-running-composer-hidden`) and the harness seed only they used; `workflow-running` is rewritten for the popup.

Kept on purpose (verified consumers): `WorkflowRunDetails` — it serves ordinary reflection-flow agents focused as viewports (round lifecycle, generated files, compile failures), which the popup does not cover; `compactWorkflowEntries` — with the workflow never active it becomes the *only* producer of the rows the popup renders; `ancestorWorkflowPhaseLabel` (grandchild headers); every helper in `src/shared/copy/workflowCall.ts` (progress view + headless); the session-list `k`/`s`/`r` keys (a task's siblings still list as session rows when a task is focused).

## Net elements (R6)

`git diff --stat origin/main`:

- Production (`packages/cli/src`, `config`, `.github`): 14 files, **+1,065 / −873** (net +192 — this is a `feat:` PR: the popup pane is ~620 LoC new, against ~870 LoC of viewport/dashboard code it deletes)
- Smoke scripts (`packages/cli/scripts`): 2 files, +11 / −175
- Tests (`src/test-kernel`): 5 files, +248 / −1,462
- Files: +1 (`panes/WorkflowPopup.tsx`), −0
- Exported symbols: +12 (`WorkflowPopup`, `workflowPopupRows`, `WorkflowPopupRow`, `WorkflowPopupGroupKind`, `WorkflowPopupView`, `workflowPopupView`, `openWorkflowPopup`, `updateWorkflowPopupView`, `presentStream`, `RowSegment`, `ApprovalSegments` — the last two were module-private and now feed the popup — and `ChildListValue` re-narrowed) / −8 (`workflowRunStatusSummary`, `workflowDashboardPanelItemCount`, `workflowDashboardSelection`, `workflowPhaseListValue`, `workflowTaskListValue`, `isWorkflowTaskListValue`, `WORKFLOW_DASHBOARD_WIDE_MIN_COLUMNS`, old `ChildListValue`)
- Declarations: +5 (`WorkflowPopupView`, `WorkflowPopupRow`, `WorkflowPopupGroupKind`, `WorkflowPopupProps`, `ChildListValue`) / −2 (`WorkflowDashboardSelection`, old `ChildListValue`)

## Consumer counts (R8)

No emit path is deleted or re-routed. Deleted public symbols, grepped at `origin/main` (prod / test):

- `workflowRunStatusSummary` 1 / 12 (its only prod caller was the band JSX, deleted; tests deleted with it)
- `workflowDashboardPanelItemCount` 1 / 7; `workflowDashboardSelection` 2 / 0; `WORKFLOW_DASHBOARD_WIDE_MIN_COLUMNS` 1 / 0
- `workflowPhaseListValue` 4 / 11; `workflowTaskListValue` 12 / 14; `isWorkflowTaskListValue` 3 / 4 — all in the dashboard code or its tests, all gone

## Verified

- `npm run typecheck` — clean (before and after the rebase onto `main`)
- vitest: `src/test-kernel/{cli,transcript,shared,architecture}`, `agent/Workflow*`, `tools/WorkflowScript*` — all passing (199 files / 3,288 tests before the rebase; rerun after)
- `npm run check:dead-code-ratchet` — no new findings (baseline shrinks by one)
- eslint + prettier on every changed file — clean
- Review round 1 (see the triage comment): every finding fixed — `presentStream` as the single owner (covers the approval-announce and finished-child paths and the roster-first interval), popup ids from the workflow roster, view keyed by stream, approval scope for the workflow log, `WorkflowRunDetails` restored, windowed tabs, measured footer, strip reserve, CI smoke set
- `validate-tui`: `workflow-running` (rewritten: Tab → ↓ → Enter opens the popup, approval suffix on the right row, `Esc close`, no `Esc parent`), `subagent-list-focus-full-frame`, `hidden-root-approval-tab-return`, `subagent-followup-summary` — all pass locally
- Tests at the durable boundary: `workflowPopupRows` ordering/grouping/filter (pure), two popup render tests (empty phase; 43-agent phase windowed with failed first), one App-level test (session-list Enter opens the popup over main, a bound approval surfaces and hands back, Enter focuses the agent, Esc returns with the popup restored). Five App tests and ~10 display tests that pinned the two-column dashboard are deleted with it.


LionSR:
Review triage — every finding was real, all addressed in the follow-up commit:

- **Popup controls never fired (P1)** — the execution ids came from the parent conversation's roster. The popup now gets its own map from the workflow stream's roster (`activeSubagentsFor(workflowStreamId, …)`).
- **Other `activeStreamId` writers bypassed the rule** (`announceApproval` → `onActiveStreamChanged`, and the finished-child → owner return) — the rule now has one owner, `presentStream` in `state/childControls.ts`; `App` and both adapter call sites go through it. It also falls back to the roster row's identity/parent for the roster-first interval before `run.start` lands (**P2, "recognize workflow rows before run metadata arrives"**).
- **Agent completion landed on the workflow viewport (P1)** — same fix: the adapter's owner return goes through `presentStream`, so it restores the popup over the parent exactly like the Esc path.
- **View reset on the Enter→Esc round trip (P2 + review note)** — the view is now keyed by the workflow stream, not by the mounted reader; only a *different* workflow starts fresh. The App test now sets view state before the round trip and asserts it survives.
- **Approvals while the workflow's log is open (P1)** — the approval scope is the foreground reader's stream whenever that stream is a `multiAgentWorkflow`, whichever reader kind (popup or its Ctrl-T log) is showing.
- **Ordinary workflow agents lost round/compile details (P1)** — `WorkflowRunDetails` and its rendering in `ConversationPane` are restored (they never depended on the viewport); only the phase band stays deleted (it only ever fired for workflow-script phase groups, which the popup now owns). The PR body's net accounting is updated.
- **Active tab could scroll out of view (P2)** — the tab strip windows around the active phase with a `‹` mark when earlier tabs are hidden.
- **Wrapped footer rows unbudgeted (P2)** — the hint text is measured at the popup width and its wrapped rows subtracted from the list budget.
- **`+N` suffix budget** — reserves the widest count the strip can print (`1 + digits(calls.length)`), so it never exceeds `maxCells` at a digit rollover.
- CI: the retired `workflow-timeline` scenario was still listed in `ci.yml`'s smoke set; removed. The test-fixture union typing that CI's type check caught is fixed.

LionSR:
@codex review

PR #11602, opened 2026-08-29
consolidate(workflow): one shared run model for the workflow display; CLI on it

LionSR:
## What

One model of a workflow-script run for every host. Owner ruling (recorded in `docs/proposals/2026-08-29-simplification-survey-workflow-display-hosts.md`): the TUI and the board must share one workflow-display state structure and differ only in how they paint it; no projector or adapter layers on either side. This PR is the first half — the model, and the CLI on it. The second puts the board on it.

- **`src/shared/streams/workflowRunModel.ts`** — `workflowRunModel({ taskGroups, rows, plan, runSettled })` folds a run's facts once: phase order (plan order, then the order the run opened the rest), newest-attempt scoping of cards *and* phases, the declared-plan union (a card always wins over its plan entry), per-phase and per-run tallies, one status cell per card (the strip), and which card may open which child stream (a stream two cards claim is nobody's). `workflowPhaseRows(phase, view)` is the attention-first row order with counted `queued` / `done` / `declared` groups and the flat filter mode. `workflowMarkerOf(entry)` reads the `workflowAttempt` / `workflowPlan` markers off `INTERNAL` entries so both hosts fold them the same way.
- **`src/shared/copy/workflowCall.ts`** — `WORKFLOW_CALL_STATUS_GLYPH` (`□ ☐ ☑ ✓ ⊘ ✗`, one shape per status so a strip reads without colour), `WORKFLOW_PHASE_GLYPH` (`◆` / `◇`), `formatWorkflowTally`, `formatWorkflowPhaseTally` — the one spelling of a tally.
- **CLI on the model, directly.** `packages/cli/src/chat/tui/state/workflowDashboardModel.ts` is deleted, not adapted: `App` calls `workflowRunModel` on the slice's task groups, rows, plan, and settled state; `WorkflowPopup` paints `WorkflowRunModel` and `workflowPhaseRows`; the fold reads markers through `workflowMarkerOf`; `WORKFLOW_TASK_STATUS_STYLE` takes its markers from the shared glyph table. The popup reads `statusLabel` and `metadataParts` off the row it already has instead of re-deriving them per repaint.

## Dead code cleaned in the same pass (survey C1–C5, C7)

[...]

PR #11604, opened 2026-08-29
feat(progress-view): the board paints the shared workflow run model

LionSR:
## What

Second half of the owner ruling in `docs/proposals/2026-08-29-simplification-survey-workflow-display-hosts.md` (first half: #11602, merged): the TUI and the board share **one** workflow-display model and differ only in how they paint it — no projector or adapter layer on either side.

`TaskGroupList` now folds its task groups, rows, plan, and settled state through `workflowRunModel` — the same call the CLI popup makes — and paints what the popup paints:

- **Phase header** — the tally in the one shared spelling (`14/39 · 6 running · 1 failed · 4 declared`) and the **glyph strip**, one cell per card in issue order from the model's `cells` and `WORKFLOW_CALL_STATUS_GLYPH` (`☑☑☑✗□☐☐□□…`). The strip sits inline after the tally and wraps under the title when a phase is wide; the shapes read without colour, and colour adds failed / running / done / awaiting.
- **Phase body** — `workflowPhaseRows`: cards needing attention first (awaiting approval, failed, running), the rest folded into counted `▸ 18 queued` / `▸ 13 done` / `▸ 4 declared` groups that open in place (a `<button aria-expanded>` per group; open state lives in the existing per-stream `ToggleStateStore`, `false` = open, same convention as the phase groups). The script's own log lines follow the cards in transcript order.
- **Declared phases** — the plan's phases the run has not reached yet render after everything it has, as the hollow `◇ Publish (3/3) · 2 declared` twin of an opened header, holding the plan's tasks as quiet dashed cards. The model drops them once the run settles.
- **Run band** — `formatWorkflowTally(model.tally)`.

The plan reaches the board because `logSlice.applyEntry` reads the `workflowAttempt` / `workflowPlan` markers through the shared `workflowMarkerOf` instead of dropping every `INTERNAL` entry, and `StreamLogs` carries the newest attempt's plan (`workflowPlan`) through `logContext$` → `LogList` → `TaskGroupList` beside the rows and task groups it already carried.

Behavior change worth knowing: the board scopes a phase's cards to the newest attempt exactly as the popup does, so a superseded attempt's cards no longer show as history inside a live phase (a phase whose every card is stale keeps its rows as plain history — the model does not hold it).

Known, unchanged: a card issued outside any phase (the model's trailing `Unphased` phase) has no task-group node on the board, so it still renders as a plain transcript row in the run body rather than through the model's attention order — the pre-PR behaviour, and a follow-up if a script ever relies on unphased calls. Cards bypass the 400-row `groupRowWindows` cap (they carry `content-visibility: auto`); the attention set is bounded by concurrency, the quiet sets by their counted groups.

Both hosts now read `runSettled` off one predicate, `workflowRunSettled(status)` in `@shared/streams/streamStatus` (a known status that is neither running nor waiting); an unknown status is not "ended", so a plan-only phase cannot be dropped before the stream's first status arrives.

### Captured from the real bundle (Electron smoke runner, seeded with a captured run + a synthesized plan marker and a 40-card phase)

Mid-run: `2/6 · 2 running`; `✓ Investigate (1/3) 2/2 ☑☑`; `● Revise (2/3) 0/4 · 2 running □□☐☐` → two running cards, `▸ 1 queued ▸ 1 done`; `◇ Publish (3/3) 2 declared`.
Wide phase (40 cards): `14/39 · 6 running · 1 failed · 4 declared`, strip `☑×12 ✗ □ ☐×6 □×18 ⊘`, then the awaiting-approval card, the failed card, six running, `▾ 18 queued` opened in place.

## Dead code cleaned in the same pass

- `collectWorkflowCalls`, `collectLatestAttemptWorkflowCalls`, `renderWorkflowCallTally`, `renderGroupProgress`, `GROUP_DOT_LIMIT`, and the `.group-dot*` / `.group-progress--failed` styles — the board's private fold of the cards.
- The popup's private `GROUP_LABEL` table — `formatWorkflowRowGroup` on the model is the one spelling of `12 queued`.

## Net elements (R6)

`git diff --stat` against `origin/main`:

- Production (`src/shared`, `packages/cli/src`, `packages/extension/src`): 10 files, **+344 / −125** (net +219)
- Tests: 1 file, +100 / −5
- Files: ±0
- Exported symbols: +3 (`formatWorkflowRowGroup` — two consumers; `formatWorkflowDeclaredTaskTemplate` — one; `workflowRunSettled` — two) / −0
- Declarations: ±0

Net-positive, stated reason: this is the feature half of the ruling — the strip, the counted groups, the declared phases and tasks, and the plan plumbing did not exist on the board — layered on the deletion of the board's own fold. Nothing is duplicated: every fold the board paints from lives in `src/shared/streams/workflowRunModel.ts`.

## Consumer counts (R8)

No emit path is deleted or re-routed. Deleted symbols, grepped at `origin/main` (prod / test): `collectWorkflowCalls` 4 sites / 0, `collectLatestAttemptWorkflowCalls` 3 / 0, `renderWorkflowCallTally` 3 / 0, `renderGroupProgress` 2 / 0, `GROUP_DOT_LIMIT` 4 / 0, `.group-dot` 13 CSS sites / 0, `GROUP_LABEL` 2 / 0. `logSlice` still drops every other `INTERNAL` record; only the two workflow markers are read.

## Verified

- `npm run typecheck` — clean
- `npm run lint` (repo-wide) + prettier — clean
- vitest `src/test-kernel/{progressView,cli,shared,architecture}` — 251 files, 3,413 tests passing
- Real-bundle capture (`npm run vite:webviews` + `scripts/smoke-webviews-electron-runner.cjs`) of a mid-run, a settled run, and a 40-card phase with every status; the counted group opened by a real click, the queued cards then in the DOM; no schema or console errors.
- Tests kept minimal: the two existing phase assertions updated (shared tally spelling; done cards sit in a group), one new case for attention order, the group toggle, and a declared phase.

### DIFF
--- a/packages/cli/src/chat/tui/App.tsx
+++ b/packages/cli/src/chat/tui/App.tsx
@@ -20,7 +20,10 @@ import {
   rewriteKittyEnterInput,
 } from '@cli/tui/inputKeys';
 import { type StreamTabId, type WorkflowControlAction } from '@shared/schemas';
-import { isActivePhase } from '@shared/streams/streamStatus';
+import {
+  isActivePhase,
+  workflowRunSettled,
+} from '@shared/streams/streamStatus';
 import { SESSION_LIST } from '@shared/copy/nestedRuns';
 
 // Local imports - TUI surfaces and state
@@ -321,8 +324,7 @@ export function App(props: AppProps): React.JSX.Element {
     workflowPopupStreamId === undefined
       ? undefined
       : streamPhaseFor(workflowPopupStreamId)?.phase;
-  const workflowRunSettled =
-    workflowRootPhase !== undefined && !isActivePhase(workflowRootPhase);
+  const workflowPopupRunSettled = workflowRunSettled(workflowRootPhase);
   const workflowPopupModel = useMemo(
     () =>
       workflowPopupRoot
@@ -330,10 +332,10 @@ export function App(props: AppProps): React.JSX.Element {
             taskGroups: workflowPopupRoot.taskGroups,
             rows: workflowPopupRoot.entries,
             plan: workflowPopupRoot.workflowPlan,
-            runSettled: workflowRunSettled,
+            runSettled: workflowPopupRunSettled,
           })
         : undefined,
-    [workflowPopupRoot, workflowRunSettled],
+    [workflowPopupRoot, workflowPopupRunSettled],
   );
   const workflowPopup = useSignal(workflowPopupViewSignal);
   // The popup controls its own grandchildren: their execution ids live on the
--- a/packages/cli/src/chat/tui/panes/WorkflowPopup.tsx
+++ b/packages/cli/src/chat/tui/panes/WorkflowPopup.tsx
@@ -48,6 +48,7 @@ import {
   formatWorkflowTally,
 } from '@shared/copy/workflowCall';
 import {
+  formatWorkflowRowGroup,
   workflowPhaseRows,
   type WorkflowPhaseModel,
   type WorkflowPhaseRow,
@@ -81,12 +82,6 @@ import { ApprovalSegments, RowSegment } from './SubagentList';
 import { pendingApprovalRowDisplay } from './SubagentListDisplay';
 import { WORKFLOW_TASK_STATUS_STYLE } from './transcriptEntryLayout';
 import type { PendingApprovalKind } from '../state/approvalQueue';
-
-const GROUP_LABEL = {
-  queued: 'queued',
-  done: 'done',
-  declared: 'declared',
-} as const satisfies Record<WorkflowRowGroup, string>;
 
 /** Rows of chrome inside the panel beyond what the shared budget already
  *  counts: the tab strip and the per-call status strip. The filter line adds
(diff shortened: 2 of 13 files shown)


LionSR:
Addressed in e5214d0666:

- **`runSettled` divergence (both reviewers):** one predicate now, `workflowRunSettled(status)` in `@shared/streams/streamStatus` — a known status that is neither running nor waiting — called by `App.tsx` and `TaskGroupList`. That also answers the ordering question: an unknown/null status is *not* settled, so a plan-only phase cannot be dropped before the stream's first status arrives, regardless of how `activeStreamState$` and `activeStreamLogs$` interleave.
- **Unphased cards:** acknowledged in the body — a card issued outside any phase has no task-group node on the board and keeps rendering as a plain row, the pre-PR behaviour; noted as a follow-up.
- **Chronology / unbounded cards:** documented in the body. The 40-card capture (six running + failed + awaiting, 18 queued opened by click) rendered without lag; cards carry `content-visibility: auto`, and the attention set is bounded by concurrency.

PR #11606, opened 2026-08-29
feat(workflow): join each child run's live progress to its card in both hosts

LionSR:
## What

A card for a running tool-use agent said nothing about what the agent was doing — twenty `Structured · researcher · gpt56 · Running` rows with no way to tell them apart. The facts already exist exactly once: each child's own stream state (elapsed origin, `conversationProgress.toolCallCount`) and, where a host projects usage per stream, tokens and spend. This PR **joins** them to the card instead of copying them:

- **`workflowRunModel`** takes `childProgress: ReadonlyMap<StreamTabId, ChildRunProgress>` beside `taskGroups / rows / plan / runSettled` and exposes `liveOf: ReadonlyMap<rowId, ChildRunProgress>` — the join runs through the existing `childStreamOf`, so a stream two cards claim is nobody's here too.
- **`formatWorkflowCallLiveParts(call, live, nowMs?)`** is the one spelling of the in-flight segments: `12s · ↓12k · $0.030 · 7 tools`. Elapsed needs a clock (`nowMs`); elapsed and spend show only while the call is live (a settled card carries its own duration and cost); tokens and tool calls stay, since nothing else records them.
- **CLI:** `App` reads each child's progress once off the session's own records (`streamPhaseFor`, `streamStateFor`, `readStreamArtifacts`) and hands the map to the model; the popup paints `model.liveOf` and its **private `liveTaskMetadata` read is deleted** — the popup no longer touches a stream.
- **Board:** a value-compared leaf selector `activeChildProgress$` (elapsed origin + tool count for the active stream's roster children) rides `logContext$ → LogList → TaskGroupList`; the card template takes the live parts. The board has no clock, so it shows tokens/spend/tools without elapsed — and today's board holds no per-stream usage, so in practice `· 5 tools`.
- `TOKENS_GENERATED` (`↓`) moves from the CLI glyph table to shared copy; both hosts read it there.

Not in scope, as noted in the design: a *current tool name* is not a fact anywhere yet (only a count) and would need a new emitted field on `conversation.progress`.

### Captured from the real bundle

Wide phase, live cards: `Structured · researcher · gpt56 · 2 tools` (awaiting approval), `Structured · lean-search · gpt56 · 5 tools`, `Document · polish · gemini37f · figures.tex · 1 context · 8 tools`, … — page: https://claude.ai/code/artifact/ee281655-b576-4504-8511-94a0886998da

## Net elements (R6)

`git diff --stat origin/main`:

- Production (`src/shared`, `packages/cli/src`, `packages/extension/src`): 12 files, **+224 / −59** (net +165)
- Tests: 3 files, +54 / −4
- Files: ±0
- Exported symbols: +2 (`ChildRunProgress`, `formatWorkflowCallLiveParts`) / −0; `TOKENS_GENERATED` re-homed (CLI glyphs → shared copy)
- Declarations: +1 (`ChildRunProgress`)

Net-positive, stated reason: a feature (live child progress on both hosts) built on one shared join; the only pre-existing host path for it — the popup's `liveTaskMetadata` — is deleted rather than kept beside it.

## Consumer counts (R8)

No emit path is deleted or re-routed. `liveTaskMetadata` (deleted): 2 sites in 1 file / 0 tests. `TOKENS_GENERATED` (moved): 5 sites in 3 files, all updated.

## Verified

- `npm run typecheck` — clean
- `npm run lint` (repo-wide) + prettier — clean
- vitest `src/test-kernel/{progressView,cli,shared}` — 236 files, 3,312 tests passing
- `validate-tui workflow-running`, `subagent-list-focus-full-frame` — pass
- Real-bundle capture: the running and awaiting cards in a 40-card phase carry their child's tool count; no schema or console errors
- Tests: one assertion block added to the shared model suite (the join follows the claimant rule; the copy with and without a clock) and one to the board's phase test (`3 tools` on the live card)

### DIFF
--- a/packages/cli/src/chat/tui/App.tsx
+++ b/packages/cli/src/chat/tui/App.tsx
@@ -94,8 +94,8 @@ import {
   childRosters as childRostersSignal,
   parentStream as parentStreamSignal,
   sessionStateRevision,
-  sessionStreamExecution,
   streamMetadataFor,
+  streamStateFor,
   subagentExecutionLabels as subagentExecutionLabelsSignal,
   visibleSubagentRows,
 } from './state/childExecutions';
@@ -352,7 +352,7 @@ export function App(props: AppProps): React.JSX.Element {
       childProgress.set(child.childStreamId, {
         ...(runStartedAt !== undefined ? { runStartedAt } : {}),
         toolCallCount:
-          sessionStreamExecution(child.childStreamId)?.conversationProgress
+          streamStateFor(child.childStreamId)?.conversationProgress
             .toolCallCount ?? 0,
         ...(usage ? { outputTokens: usage.outputTokens } : {}),
         ...(usage?.cost !== undefined ? { costUsd: usage.cost } : {}),
--- a/packages/cli/src/chat/tui/App.tsx
+++ b/packages/cli/src/chat/tui/App.tsx
@@ -27,7 +27,10 @@ import {
 import { SESSION_LIST } from '@shared/copy/nestedRuns';
 
 // Local imports - TUI surfaces and state
-import { workflowRunModel } from '@shared/streams/workflowRunModel';
+import {
+  workflowRunModel,
+  type ChildRunProgress,
+} from '@shared/streams/workflowRunModel';
 import {
   appDraftDiscardActive,
   approvalVisibleForActiveStream,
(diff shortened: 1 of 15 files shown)


LionSR:
Addressed in ccd8e17db4:

- **Duplicate helper (both reviewers):** `sessionStreamExecution` deleted; `App` calls the existing `streamStateFor`.
- **Dead imports** left by the deleted `liveTaskMetadata` (`isTerminalWorkflowCallProgress`, `formatCompactTokenCount`) and the **orphaned glyph comment** removed.
- **Render-error fallback on board task cards:** left as is — the card's inputs are the zod-validated row plus strings from `formatWorkflowCallLiveParts`, and the phase body's other paths (declared cards, group buttons) never went through `formatLogEntry` either. If a formatter throw ever surfaces here, the fix is a live-parts-aware `formatLogEntry`, not a second try/catch.

PR #11613, opened 2026-08-29
fix(workflow): list finished calls as ticked rows instead of folding them

LionSR:
## What

Owner feedback on a live run: finished calls folded into `▸ 7 done` should stay listed — tick the box, do not hide the row.

`workflowPhaseRows` (the shared model, so the popup and the board change together) now orders a phase as: attention rows (awaiting approval, failed, running), then every finished call as its own row in transcript order (`☑`/`✓`/`⊘`/`✗`), then the counted `▸ N queued` / `▸ N declared` groups for what has not started. `WorkflowRowGroup` loses its `done` member; the board's toggle-key list follows.

## Net elements (R6)

`git diff --stat origin/main`: production 2 files, +7 / −7; tests 3 files, +8 / −10. Exports ±0; `WorkflowRowGroup` narrowed by one member.

## Consumer counts (R8)

No emit path deleted. `'done'` group consumers at `origin/main` (prod / test): the model's own fold, the board's `WORKFLOW_ROW_GROUPS` list, three test expectations — all updated here.

## Verified

- CLI + extension typecheck clean; eslint + prettier clean on changed files
- vitest `src/test-kernel/{shared,cli,progressView}` — 236 files, 3,312 tests passing
- `validate-tui workflow-running` — pass (real TUI, ttyd frames)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QESykrKh1yzvF2bB4pjDrS

PR #11612, opened 2026-08-29
simplify(workflow): drop the per-call and per-phase review scope

LionSR:
## What

Owner decision: the Approve menu of a multi-agent workflow proposal no longer offers **"Approve, review the first call of each phase"** and **"Approve, review each call"**. Plain approve — plus the existing "approve all delegated work" bypass — is the whole choice.

Build implies delete, so the entire admission path behind those two rows goes, not just the rows:

- **Wire and schema:** `WORKFLOW_CALL_REVIEW_SCOPE` / `WorkflowCallReviewScope` (`prompts.ts`), `callReview` on the approve action (`progressView/inbound.ts`, `HostInteractions`, the CLI approval queue/settlement, the board's `events.ts`/`eventHandlers.ts`, `ProgressAgentProposalController`), and the reviewed-call proposal field `workflowCall` (`proposalFields.ts`) with its `admitsPhase` flag.
- **Engine:** `WorkflowScriptRunOptions.admitCall`, `WorkflowCallAdmissionRequest`/`WorkflowCallAdmission`, the admission wait in `runWorkflowScript` (44 lines), `WorkflowExecutionState.awaitAdmission`, and the `awaitingApproval` call status it alone produced — from the snapshot enum, the card union, the label/glyph tables, the model's attention ranks, both hosts' status styles/CSS/icons, the headless log classifier, and `deriveWorkflowCounts`.
- **Tool/runner:** `createWorkflowCallAdmission` (its only reason to build a proposal per call) and the `admitCall` plumbing through `WorkflowScriptTool` → `workflowScriptStrategy`.
- **Copy and UI:** `WORKFLOW_CALL_REVIEW_COPY`, `workflowCallCardLine`, the CLI card's explanation line, `p`/`c` extra actions and their footer-compact labels, the "Run workflow call …?" title and "Workflow call" category; the board's split-button items, `canReviewWorkflowCalls`/`reviewWorkflowCallsHandler`, the review-card context block, and the "no Setup on a review card" rule (every proposal offers Setup again).
- **Docs:** the guide paragraph; the engine README bullet; a status line on `docs/proposals/2026-08-28-workflow-plan-vs-issued-calls.md` recording the retirement (the declared-vs-issued model, plan marker, and card projection stand).

Persisted cards in the retired `awaitingApproval` status are one-day-old intermediate data: they now fail the card schema loudly rather than mapping to a default, per the intermediate-data ruling.

## Net elements (R6)

`git diff --stat origin/main`:

[...]

PR #11614, opened 2026-08-29
docs: simplification survey — the shared workflow run model, one day in

LionSR:
## What

`/find-simplification` over everything that landed for the workflow display on 2026-08-29 (#11588 #11591 #11590 #11594 #11602 #11604 #11606; #11612 and #11613 in flight). Four read-only scouts by domain, each with its own refutation section, then two adversarial verifiers re-deriving every claim at `fa365c163f`.

**Filed as this proposal:** 21 verified candidates in seven bundles, ≈ −285 production LoC, −8 exports, −4 declarations, −1 trace event, −1 transcript marker kind, +1 shared schema. Zero new tests.

Headline finds:
- **S5 — the `workflowAttempt` marker is a second representation of a fact already on every plan and card** (found independently by two scouts); ≈ −40, one trace event and one INTERNAL kind gone.
- **S6 — one `WorkflowDeclaredPlan` schema**; the proposal panel folds through `workflowRunModel` and the plan-grouping fold that #11608 just relocated into the panel (closing #11605) is deleted outright.
- **S7 — the projection keeps the card it last wrote**, which also fixes the backstop sweep dropping `kind/agent/model/files` (the define-out-of-existence form of #11574's hand fix).
- **S18 — compute the board's model in a selector** (four edit sites per model input today → one), verified net −26 with a strict memo improvement.
- **S19 — the board opens a child via the model's `childStreamOf`**, closing the last host divergence the owner ruling reclassified as a defect.

**Consolidated:** A-C4 (relocate the proposal fold) into S6; two duplicate tally finds into S2; two attempt-marker finds into S5.
**Rejected with evidence:** 13 (tree-vs-model grouping, plan as one *event*, status tables, snapshot counts, stage `total`, concurrency owners, root elapsed/cost, …).
**Excluded:** the settled surfaces, `WorkflowRunDetails`, the cost tracker, the child-activity emitter.
**Not filed as issues:** the classifier gates issue creation without an explicit ask; each bundle is written to be an issue or PR body.

## Checks
`npm run format`, `npm run check:guidance-refs`, `git diff --check` — clean.

### DIFF
--- a/docs/proposals/2026-08-29-simplification-survey-shared-workflow-model.md
+++ b/docs/proposals/2026-08-29-simplification-survey-shared-workflow-model.md
@@ -102,7 +102,7 @@ alternative does not give a full reset either (cards still scope by their
 own `attemptId`), so it is not offered. Net ≈ −40. Risk: low; `events.ts` is
 the unpublished `@texra-ai/agent` contract with no external consumer.
 
-### Bundle 3 — one declared-plan shape (≈ −15; supersedes #11605)
+### Bundle 3 — one declared-plan shape (≈ −20; deletes the fold #11608 relocated)
 
 **S6. `WorkflowDeclaredPlan` once; the proposal panel folds through the
 model.** The approval proposal carries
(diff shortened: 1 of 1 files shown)


LionSR:
Both staleness notes are addressed in 4ba4e41b9d (pushed right after #11608 merged): Bundle 3 / S6 now reads as the next step after #11608's relocation — the panel-local fold is deleted and the panel folds through `workflowRunModel` — and the Tracking line records that #11605 closed with #11608.

LionSR:
All three addressed in 2ccc76a11b: Bundle 3 header and the grand total now match S6's −15 (≈ −285 overall); the concurrency refutation cites the real numbers (83 `runWorkflowScript(...)` test calls rely on the default, 15 pass `concurrency:`); S6 names the helper's shared origin and its post-#11608 home without line numbers, since that file moved twice today.

PR #11635, opened 2026-08-29
simplify(workflow): the projection keeps the card it last emitted

LionSR:
## What

Bundle 4 (S7) of [the 2026-08-29 survey](docs/proposals/2026-08-29-simplification-survey-shared-workflow-model.md): the card projection keeps the card it last emitted.

`ProjectedWorkflowCall` stored six fields hand-copied out of each emitted card, and the finally sweep rebuilt a settled card from that subset — omitting `kind`, `agent`, `model`, and `files` from a card the reader had just seen, since `StreamLog` replaces `data` wholesale (#11574 hand-added `attemptNumber` to the same subset). The record is now `{ logId, card }`: `emitCall` stores the card it emits, the fold's diff reads it (`status`, `childStreamId`, `agent`, `model`), and the sweep spreads it with `status: 'failed'` — so a sweep-settled card keeps every issued-call fact it showed.

The phase re-derivation goes with it: the engine pins `stageId` when the call is issued and throws if a declared task is issued elsewhere, so the card's own `phase` is the one fact (the verifier and a bridge test confirm a phase-less declared task issued inside a phase stays under the parent). `openPhaseStage`/`markPhaseFailed` lose the `index`/`total` parameters no caller passed.

**Kept on purpose** (the verifier's two corrections): `error ?? WORKFLOW_CALL_UNFINISHED_NOTE` in `cardFor` — the persisted snapshot is not a discriminated union, so the default is a real boundary; and `call.model ?? attempts.at(-1)?.model` — reachable on a retry re-queue followed by a cancel, so deleting it would be a behaviour change, not a cleanup.

The stale "two channels" comment on `WORKFLOW_CALL_NOT_REACHED_NOTE` now says which note has which writers.

## Net elements (R6)

`git diff --stat` against the base: production 2 files, **+44 / −90** (net −46); tests 1 file, +4 / −2 (one bridge pin required an explicit `phase: undefined` key the old emitter always wrote; it now asserts the fact). Declarations: `ProjectedWorkflowCall` 6 fields → 2; −4 dead parameters.

## Consumer counts (R8)

No emit path deleted. `ProjectedWorkflowCall` fields at the base: read only inside `workflowScriptRun.ts` (fold diff, sweep). `openPhaseStage`/`markPhaseFailed` extra parameters: 0 callers pass them (4 call sites, one argument each).

## Verified

- `npm run typecheck` — clean; eslint + prettier — clean on changed files
- vitest `src/test-kernel/{agent,shared,tools}` — 192 files, 2,996 tests passing (incl. `WorkflowScriptProgressProjectionFailure`, the backstop-sweep suite from #11574)

### DIFF
--- a/src/shared/copy/workflowCall.ts
+++ b/src/shared/copy/workflowCall.ts
@@ -113,10 +113,11 @@ export function formatWorkflowPhaseHeading(
 }
 
 /**
- * What a run that ended first says about a call, written once here because two
- * channels report it: the execution snapshot the engine terminalizes and the
- * trace cards the run's progress projection settles. Two spellings of one
- * sentence is drift, not two facts.
+ * What a run that ended first says about a call. `NOT_REACHED` is the note
+ * on a declared call the run never issued (`workflowCallDetail` below);
+ * `UNFINISHED` has two writers — the engine's terminal sweep and the card
+ * projection's own backstop for a transition the writer never landed — so it
+ * is spelled once here. Two spellings of one sentence is drift, not two facts.
  */
 const WORKFLOW_CALL_NOT_REACHED_NOTE =
   'The workflow ended before this call was reached.';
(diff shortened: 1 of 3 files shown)


LionSR:
Can we simplify to remove projection layer

PR #11669, opened 2026-08-31
fix: bound long-lived memory retention

LionSR:
## Summary

- replace permanently retained snapshot version counters with revocable generation tokens while preserving stale async-work rejection
- time out and dispose pending desktop file reads/listings, keep uncancellable writes authoritative, and make late responses no-ops
- clear write-failure diagnostics at stream lifecycle boundaries and bound CLI broken-entry diagnostics
- remove whole-store reset APIs that existed only for tests; tests now use fresh stores or the public reload boundary

## Issue acceptance gates

- Snapshot generation tokens are revoked on eviction and recreated with fresh identity, so stale continuations remain rejected without retaining historical stream IDs.
- Desktop reads/listings time out after 60 seconds; unload disposal rejects all pending requests; late responses are ignored. Writes remain pending because the main process cannot cancel them safely.
- `writeFailureWarned` is cleared on per-stream and whole-store cleanup, and broken transcript-entry diagnostics use a 1,000-ID LRU bound.
- Test-only `StreamSnapshotStore.evictAll`, registry-wide eviction, and staged-deletion reset surfaces are removed.
- Pending snapshot writes and desktop writes remain authoritative.

## Single source of truth

`ResidentStreamRegistry` owns revocable per-stream generation identity. `fileRequests.ts` owns renderer-document request registration, timeout, settlement, and disposal.

The shared desktop request registration path replaces three duplicated map/send sequences and owns the timeout/disposal invariant. No pass-through layer was added. Test-only whole-store reset surfaces were deleted.

## Net elements (R6)

- Files: 1 added, 12 modified
- Exported symbols: +1 (`disposePendingFileRequests`), removed public method surface: `StreamSnapshotStore.evictAll`
- Classes/interfaces/enums: +1 internal interface (`PendingFileRequestEntry`); no new classes or enums
- LoC: +241 / -122, net +119

## Consumer counts (R8)

- `disposePendingFileRequests`: 1 production consumer (`packages/desktop/src/renderer/main.ts` unload lifecycle), plus its regression suite
- removed `StreamSnapshotStore.evictAll`: 0 production consumers; all test callers replaced
- removed staged-deletion reset and registry-wide eviction: 0 remaining consumers

## Host impact

Extension: Snapshot lifecycle and transcript write-diagnostic cleanup apply through shared storage; no extension-specific wiring changes.

Desktop: Pending reads/listings have a 60-second backstop, all request kinds reject on renderer unload, late responses do nothing, and uncancellable writes do not time out.

CLI: Broken transcript-entry diagnostic IDs are LRU-bounded at 1,000; tests no longer use whole-store reset hooks.

## Scheduled refactoring

None.

## Validation

- Focused transcript/desktop/CLI/storage tests: 393 passed
- Full Vitest suite: 823 files passed, 1 skipped; 9,987 tests passed, 6 skipped
- Architecture suites: 15 files / 101 tests passed
- `corepack pnpm run format:check`
- `corepack pnpm run lint`
- `corepack pnpm run check:dead-code-ratchet`
- `corepack pnpm run check:guidance-refs`
- `corepack pnpm run check:browser-safe-utils`
- Raw `check:dead-code` still exits 1 on the same 6 baselined unused exports; the ratchet reports 295 combined findings versus 295 baselined and no new findings.

### DIFF
--- a/config/ratchets/store-public-surface-baseline.json
+++ b/config/ratchets/store-public-surface-baseline.json
@@ -29,7 +29,6 @@
     "StreamSnapshotStore": [
       "attachSessionEvents",
       "deleteStream",
-      "evictAll",
       "flush",
       "getCompileFailures",
       "getExecutionIdMap",
--- a/packages/cli/src/chat/tui/panes/transcriptViewport.ts
+++ b/packages/cli/src/chat/tui/panes/transcriptViewport.ts
@@ -3,6 +3,7 @@
 import { createLog } from '@logger/logUtils';
 import type { TranscriptRow } from '@shared/transcript';
 import type { ExecutionLabels } from '@shared/tools/executionsDisplay';
+import { createBoundedIdSet } from '@utils/core/boundedIdSet';
 import { toErrorMessage } from '@utils/errors/errorMessage';
 import {
   transcriptEntryLayout,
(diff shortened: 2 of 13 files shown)


LionSR:
@codex review exact head 0fb19233a5e8cc8961e013f02f07bb8fb2d44c3b

PR #11673, opened 2026-08-31
fix(workflow): preserve unresolved compile rejection

### LINKED ISSUE #11672: Preserve unresolved workflow compile rejection across repair rounds
## Problem

A workflow can consume `compileFailureContext` while preparing a repair round, then end without a successful compile because the round is cancelled, crashes, or produces no compile result. The rejection evidence is lost and the run can incorrectly complete.

Related downstream behavior also treats rejected artifacts as publishable or resumable:

- failed CLI workflows can copy generated artifacts to `--output` / `--output-dir`;
- plain text output can look successful;
- history can advertise a terminal rejected workflow as resumable;
- final outputs can be auto-opened despite a failed outcome.

## Expected behavior

- Keep a durable unresolved compile-rejection marker separate from one-shot prompt feedback.
- Clear it only after explicit compile success or when rejection policy is disabled.
- At the configured round cap, unresolved rejection yields `FAILED` without fabricating provider/runtime error metadata.
- Only `COMPLETED` workflows publish or auto-open final outputs.
- Do not advertise terminal compile rejection as resumable, while preserving its checkpoint for explicit resume after configuration changes.
- Migrate persisted state produced before the durable marker safely.

LionSR:
## Summary

- separate durable unresolved compile rejection from one-shot repair prompt feedback
- keep rejection through cancellation, crashes, and absent compile results until explicit compile success or policy disablement
- migrate legacy persisted rejection state and remove only the exact synthetic final-compile error
- publish and auto-open workflow artifacts only for `COMPLETED` outcomes
- label failed CLI output and suppress terminal rejection from resume advertising without deleting its checkpoint
- define compile rejection as an outcome-only domain failure rather than fabricated provider/runtime error metadata

## Tests

- 10 focused test files: 237 passed, 1 skipped
- workspace, test-kernel, and CLI typechecks passed
- changed-file ESLint and Prettier passed
- `git diff --check` passed
- independent scoped review found no remaining issues

Closes #11672

### DIFF
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -6,7 +6,10 @@ import {
   persistChildRunResultMeta,
   type ResumabilityDecision,
 } from '@agent/storage';
-import type { AgentConfigPayload } from '@agent/runtime';
+import {
+  isTerminalPersistedCompileRejection,
+  type AgentConfigPayload,
+} from '@agent/runtime';
 import { RUN_OUTCOME, type ExecutionId, AgentCategory } from '@shared/schemas';
 import { toErrorMessage } from '@utils/errors/errorMessage';
 
@@ -198,7 +201,7 @@ export async function executeCliWorkflowConfig(
       typeof shared === 'object' && shared !== null
         ? (shared as Record<string, unknown>).lastError
         : undefined;
-    return lastError == null;
+    return lastError == null && !isTerminalPersistedCompileRejection(shared);
   };
   const writeResumeHint = (
     executionId: ExecutionId,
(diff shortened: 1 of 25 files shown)


LionSR:
Rebased onto current `main` and removed two unrelated concurrent hunks that entered the original commit while the shared workspace was dirty:

- restored `CompiledPdfArtifact[]` and `artifact.latestPdf` in `OutputNode`;
- restored the existing `generateSessionDescription(..., runSession, signal)` call.

The cleaned commit is `86766cba88`. On that exact commit, full workspace typecheck passes and the 10 focused suites pass with 237 tests (1 skipped). The inline findings are now outdated.

PR #11687, opened 2026-08-31
fix(setup): continue after provider key read failures

LionSR:
## Summary

- isolate API-key lookup failures to the provider being checked
- continue scanning later providers after a keychain or environment lookup error
- keep reporting each failed probe through the existing setup warning path
- add a regression proving a failed OpenAI lookup does not hide a valid Anthropic key

## Why

The previous outer probe wrapped the entire provider loop. If one provider lookup threw, setup treated the whole provider-key scan as failed and never checked later providers, so a usable credential could be reported as missing.

## Validation

Reconstructed on current `main` (`6e3dda7f05`) and validated at `1230c5fbfe`:

- focused setup/model/provider suites: 7 files, 65 tests passed
- `corepack pnpm run compile:safe`
- `corepack pnpm run lint`
- `corepack pnpm run format:check`
- `corepack pnpm --filter @texra-ai/cli run check:architecture`
- `corepack pnpm run check:dead-code-ratchet`
- `corepack pnpm test` (822 files passed, 1 skipped; 9,985 tests passed, 6 skipped)
- `git diff --check`

### DIFF
--- a/src/model/setupCredentialAccess.ts
+++ b/src/model/setupCredentialAccess.ts
@@ -14,10 +14,16 @@ import { toErrorMessage } from '@utils/errors/errorMessage';
 /** True when any provider has a usable API key in secret storage or the environment. */
 async function hasAnyUsableProviderApiKey(
   secrets: PlatformSecrets,
+  onProbeFailure: (message: string) => void,
 ): Promise<boolean> {
   for (const provider of API_PROVIDERS) {
     // Keep the scan sequential so the first usable key ends the lookup.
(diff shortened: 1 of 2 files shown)


LionSR:
@codex review

Please review exact reconstructed head `1230c5fbfee705320299dd853253c8b95d35c889`. Confirm per-provider failure isolation, later-provider continuation, first-valid-key short-circuiting, warning reporting, provider order, and unchanged secret-to-environment lookup semantics.

PR #11704, opened 2026-08-31
fix(cli): preserve compact workflow fallback

LionSR:
Follow-up to #11702 for the two final Codex findings that arrived immediately before that PR was merged externally.

- Preserve current-plan issued-task knowledge across transcript eviction so compact popup fallback does not revive cutoff tasks as `Declared`.
- Keep the original declared phase array through `workflowRunModel`, then apply the CLI retention projection to the model so future phases retain their original index and total.
- Extend the popup/full-transcript lifecycle regression with the eviction window and original phase numbering.

Issue context: #11552.
Review findings: https://github.com/LionSR/TeXRA/pull/11702#discussion_r3892114289 and https://github.com/LionSR/TeXRA/pull/11702#discussion_r3892114291.

### DIFF
--- a/packages/cli/src/chat/tui/App.tsx
+++ b/packages/cli/src/chat/tui/App.tsx
@@ -89,7 +89,10 @@ import {
   streamPhaseFor,
 } from './state/cliState';
 import { appendLocalAssistantTranscript } from './state/transcript';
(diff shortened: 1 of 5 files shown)


LionSR:
@codex review

PR #11715, opened 2026-08-31
fix: enforce Google Interactions transport deadlines

### LINKED ISSUE #11713: Enforce long-running timeout policy across the Google Interactions SDK boundary
## Context

Investigation of #11378 found that the direct Google Interactions handler does not actually receive TeXRA's long-running Undici dispatcher policy.

`ModelHandlerGoogleInteractions` passes `fetchOptions.dispatcher = composeLongRunningModelDispatcher()` to pinned `@google/genai` 2.18.0. The SDK flattens those options into a native `Request`, then calls `fetch(request)` without an init object. The non-standard `dispatcher` property is discarded during `Request` construction. The abort signal survives.

As a result, the intended response-header and stream-inactivity policy in `src/platform/defaults/longRunningModelTransport.ts` is not authoritative for Google Interactions requests. Those requests fall back to the host/native-fetch transport plus Google or intermediary deadlines.

This is separate from #11378's user-facing documentation remedy. It affects extension, desktop, and CLI because all hosts share the direct Google handler.

## Acceptance gates

- Stop relying on `fetchOptions.dispatcher` as the Google Interactions timeout boundary.
- Keep timeout durations and policy in `src/platform/defaults/longRunningModelTransport.ts` as the single source of truth.
- Enforce the response-header deadline through a primitive preserved by `@google/genai`, such as a composed `AbortSignal`.
- For SSE responses, enforce/reset the existing inactivity budget while consuming stream events without imposing a total-generation wall clock.
- Preserve user cancellation and host shutdown abort propagation.
- Do not replace the process-global dispatcher or extend unrelated auth, telemetry, or ordinary HTTP requests.
- Add a regression against the pinned SDK boundary with mocked global fetch that proves the final native `Request` receives and reacts to the timeout signal. Do not assert the unsupported `dispatcher` option.
- Cover unary, streaming, and background submit/poll requests as appropriate, including cleanup of timers/listeners.

## Relevant paths

- `src/agent/modelHandlers/google/modelHandlerGoogleInteractions.ts`
- `src/platform/defaults/longRunningModelTransport.ts`
- `src/test-kernel/agent/modelHandlers/GoogleInteractions*.vitest.ts`
- `src/test-kernel/agent/modelHandlers/LongRunningModelFetch.vitest.ts`

## Evidence

A runtime probe against the pinned SDK observed the final `fetch` receiving a native `Request` with no dispatcher while retaining a dependent abort signal. Static inspection of the SDK confirms the Request reconstruction and `fetch(request)` call shape.

LionSR:
## Summary

- stop relying on `fetchOptions.dispatcher` for Google Interactions, because `@google/genai` 2.18.0 rebuilds a native `Request` and calls `fetch(request)`, dropping that non-standard field
- install a request-local HTTP client on the pinned SDK's generated Interactions subclient, after the SDK has built its final `Request`
- enforce the shared 10-minute deadline until native fetch resolves with response headers, then enforce a resettable 30-minute inactivity deadline around response-body reads
- apply the policy to foreground unary and streaming calls, background submit/get/cancel, and compaction while preserving caller abort reasons and the existing server-side cancellation flow

Closes #11713.

## Issue acceptance gates

- [x] No Google Interactions request relies on `fetchOptions.dispatcher`.
- [x] Timeout durations remain defined only in `src/platform/defaults/longRunningModelTransport.ts`.
- [x] The 10-minute timer is cleared when native fetch resolves, not when the outer SDK promise finishes parsing the body.
- [x] Every response-body read gets the shared 30-minute inactivity budget; activity resets it and there is no total generation wall-clock timer.
- [x] The final SDK `Request` signal composes with user/RunScope/host aborts and forwards their reason.
- [x] Timers and abort listeners are removed on fetch failure, body completion, stream cancellation, timeout, or caller abort.
- [x] No process-global dispatcher or global fetch replacement is installed. Unrelated HTTP retains its existing transport.
- [x] The request-local seam covers foreground unary, streaming, background submit/poll/cancel, and compaction.
- [x] The regression invokes the real pinned SDK with mocked global fetch. It verifies the final native `Request`, dropped `dispatcher`, propagated caller abort, header timeout, resettable body inactivity timeout, and a unary body active for more than 10 minutes.

No live-key test was run, so Google background cancellation state transitions retain their existing smoke-test caveat.

## Single source of truth

`src/platform/defaults/longRunningModelTransport.ts` owns both timeout durations and both request phases. The Google-specific adapter consumes those values rather than defining provider-local policy.

One transport function implements the native fetch header/body boundary. One narrow installer applies it to every API-version client produced by the pinned Interactions runtime. No provider-wide transport facade or duplicate constants were added.

## Net elements (R6)

3 files changed, +303/-18 lines. Added one transport helper export and two file-local structural types. No files, classes, or enums added.

## Consumer counts (R8)

The new transport export has one production consumer, the Google Interactions client installer. The installer applies it to every generated Interactions client, so all `create`, `get`, and `cancel` call sites share the policy without per-call wrappers.

## Host impact

Extension, desktop, and CLI share the same direct Google Interactions handler and receive the same request-local behavior. OpenAI, OpenRouter, uploads, and unrelated native fetch calls are unchanged.

## Validation

- Direct reproduction against `@google/genai` 2.18.0: final fetch input was a native `Request`; `fetch` received no init; `dispatcher` was absent; the cloned signal reacted to source abort.
- Focused Google + transport suites: 7 files passed, 1 live suite skipped; 100 tests passed, 4 live tests skipped.
- `corepack pnpm run typecheck`: all six typecheck lanes passed.
- `corepack pnpm run lint`: passed.
- `corepack pnpm run format:check`: passed.
- `corepack pnpm run compile:fast`: passed.
- `corepack pnpm run check:dead-code-ratchet`: passed, 295 findings matched 295 baselined.
- `corepack pnpm test`: 822 files passed, 1 skipped; 9,990 tests passed, 6 skipped. One unrelated desktop timing test failed in the full parallel run and passed immediately in isolation: `DesktopAgentExecution.vitest.ts > presents a merge failure that occurs before lifecycle startup`.

### DIFF
--- a/src/agent/modelHandlers/google/modelHandlerGoogleInteractions.ts
+++ b/src/agent/modelHandlers/google/modelHandlerGoogleInteractions.ts
@@ -49,7 +49,7 @@ import {
   takeTail,
 } from '@common/errors/sdkError/errorPatterns';
 import { handleStreamingFailure } from '@common/errors/sdkError/streamFailure';
-import { composeLongRunningModelDispatcher } from '@platform/defaults/longRunningModelTransport';
+import { longRunningGoogleInteractionsFetch } from '@platform/defaults/longRunningModelTransport';
 import {
   type FileLocation,
   type MediaAttachmentKind,
@@ -376,6 +376,25 @@ type InteractionsRequestOptions = {
   fetchOptions: RequestInit;
 };
 
+type GoogleInteractionsSdkClient = {
+  _httpClient: { request(request: Request): Promise<Response> };
+};
+type GoogleInteractionsClientInternals = {
+  getClient(apiVersion?: string): GoogleInteractionsSdkClient;
+};
+
+/** Install the pinned SDK's request-local HTTP seam on every API-version client. */
+function installLongRunningInteractionsTransport(client: GoogleGenAI): void {
+  const interactions =
+    client.interactions as unknown as GoogleInteractionsClientInternals;
+  const getClient = interactions.getClient.bind(interactions);
+  interactions.getClient = (apiVersion?: string) => {
+    const sdk = getClient(apiVersion);
+    sdk._httpClient = { request: longRunningGoogleInteractionsFetch };
+    return sdk;
+  };
+}
+
 export class ModelHandlerGoogleInteractions extends ModelHandler<
   Step,
   Usage | null,
@@ -665,13 +684,15 @@ export class ModelHandlerGoogleInteractions extends ModelHandler<
     // `{ attempts: 1 }` — switches apiCall into a pRetry wrapper that converts
     // non-ok responses into bare Errors with no status field, blinding the
     // route gate's 429/5xx classification for Google.
+    const googleClient = new GoogleGenAI({
+      apiKey: credential.apiKey,
+      httpOptions: {
+        baseUrl: credential.baseUrl ?? undefined,
+      },
+    });
+    installLongRunningInteractionsTransport(googleClient);
     const client = this.rememberClientCredentialRoute(
-      new GoogleGenAI({
-        apiKey: credential.apiKey,
-        httpOptions: {
-          baseUrl: credential.baseUrl ?? undefined,
-        },
-      }),
+      googleClient,
       credential.route,
       credential.apiKey,
     );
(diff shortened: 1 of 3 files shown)


LionSR:
do the cleanest cia the latest api

LionSR:
Reworked the implementation at exact head `4772a60258` after the independent review found that the first version kept the 10-minute timer armed while the unary SDK parsed `response.text()`.

The new design uses the pinned SDK’s request-local generated-client HTTP seam. It sees native `fetch(Request)` resolution directly, clears the 10-minute timer at response headers, then wraps body reads with the resettable 30-minute inactivity budget. A regression now proves an active unary body can run beyond 10 minutes, while also covering header timeout, body inactivity, dropped `dispatcher`, and caller abort propagation.

@codex review

PR #11774, opened 2026-09-02
fix(agent): re-check stream currency after a failed lease-lock claim

### LINKED ISSUE #11765: markUnavailable overwrites a live stream reservation; restartRepair's lease-lock catch lacks the post-await isCurrent re-check
## Summary

A latent defect introduced by #11757 (`03fa583`). Verified first-hand at `e599027`; every line below was re-opened.

#11757 made a hold the third `StreamEntry` arm — a good collapse, and the survey that found this records it as healthy structure. The side effect is that `markUnavailable` now **replaces whatever entry is present, a live `reserved` one included**:

```ts
// src/agent/runtime/StreamStatusService.ts:293-301
markUnavailable(stream: StreamTabId, detail: string): void {
  const entry = this.streams.get(stream);
  const state = entry ? effectiveState(entry) : undefined;
  this.streams.set(stream, { kind: 'hold', detail, ...(state ? { state } : {}) });
}
```

`effectiveState` synthesizes `{ phase: RUNNING, substate: STARTING, runStartedAt }` for a `reserved` entry (`:76-81`), so the hold carries a running-looking state forward. After that overwrite:

- `releaseIfReserved` bails, because it requires `entry?.kind === 'reserved'` (`:159`) — so **a failed launch's rollback never runs**;
- the stream reads in-flight *and* unavailable, i.e. stuck.

## Why it is reachable

`restartRepair` re-checks staleness through the single authority (`isRepairCandidateCurrent`, via the local `isCurrent` at `:165-166`) at the top of each phase-2 iteration (`:247`) **and again after its inner await** (`:266`) — precisely because a stream can be reused across an await.

The lease-lock catch at `:325` runs after its own await (`withLeaseLock` / `settle`) and has **no such re-check**:

```ts
// src/agent/runtime/restartRepair.ts:324-331
} catch (error) {
  if (settleStarted) throw error;
  applyHold(streamId, executionId, {
    kind: 'unclassified',
    cause: `lease lock unavailable (${toErrorMessage(error)})`,
  });
  continue;
}
```

So a stream that a fresh run reserved while the lock acquisition was in flight gets its reservation overwritten by a hold.

## Proposed fix (3 lines)

Add the missing re-check at `restartRepair.ts:325`, matching its sibling at `:266` and reusing that site's debug wording so it routes through the one staleness authority:

```ts
if (!isCurrent(streamId, executionId)) { options.logger?.debug(...); continue; }
```

**Optional hardening, separate call:** make `markUnavailable` refuse to overwrite a `reserved` entry. That is a behavior decision about which owner wins, so it should not ride along silently.

## Evidence (re-opened at `e599027`)

- `src/agent/runtime/StreamStatusService.ts:293-301` — the unconditional overwrite
- `src/agent/runtime/StreamStatusService.ts:72-85` — `effectiveState` synthesizing RUNNING/STARTING for `reserved`
- `src/agent/runtime/StreamStatusService.ts:155-160` — `releaseIfReserved` bails on a non-`reserved` entry
- `src/agent/runtime/restartRepair.ts:324-331` — the unfenced catch
- `src/agent/runtime/restartRepair.ts:247`, `:266` — the two fenced siblings; `:165-166` — the `isCurrent` helper

## Risk and accounting

Risk medium (status-plane semantics feed every host's liveness display). Net **+3 production lines** — this is the one item in the survey that adds rather than deletes, and it should not wait on any cleanup.

Found by the runtime/host/session/lifecycle ownership survey; full context in `docs/proposals/2026-09-02-runtime-host-session-lifecycle-ownership-survey.md` §7.

LionSR:
## Summary

Fixes #11765, filed by the 2026-09-02 runtime/host/session/lifecycle ownership survey (`docs/proposals/2026-09-02-runtime-host-session-lifecycle-ownership-survey.md` §7) and re-verified against current `main` before implementing.

`restartRepair`'s phase-2 settle loop re-validates that a candidate stream is still current (not reused by a fresh run) at two points: the top of each iteration (`restartRepair.ts:247`) and again right before settlement (`:266`), because a stream can be reused across an `await`. The lease-lock `catch` block — reached when `runWithInactiveExecutionLease`'s own `claimLease` throws, before the settlement callback ever runs — had no such re-check, and applied `markUnavailable` unconditionally.

Since #11757 made a reservation a first-class `StreamEntry` kind, `markUnavailable`'s unconditional overwrite replaces a live `reserved` entry with a hold. `releaseIfReserved` only fires on a `kind === 'reserved'` entry, so once overwritten a failed launch's rollback can never run again, and the stream reads in-flight-and-unavailable — stuck.

## Issue acceptance gates

#11765's proposed fix is the 3-line addition: re-check `isCurrent(streamId, executionId)` in the lease-lock catch, matching its sibling at `:266` (same debug wording pattern). Implemented as described; the optional hardening (`markUnavailable` refusing to overwrite a `reserved` entry) is explicitly called out in the issue as a separate, later decision and is not included here.

## Single source of truth

No new vocabulary or module — the fix reuses the same `isCurrent` closure the two sibling checks in the same function already call.

N/a — this is a bug fix, not a consolidation. No abstraction added or removed.

## Net elements (R6)

2 files changed, 47 insertions(+). Production: `src/agent/runtime/restartRepair.ts` +6 lines (one `if` guard). Test: `src/test-kernel/agent/runtime/RestartRepair.vitest.ts` +41 lines (one regression test plus the module-mock scaffolding it needs). No exported symbols added or removed.

## Consumer counts (R8)

N/a — no export or emit path was deleted or re-routed.

## Host impact

Extension: none (shared runtime code, not host-specific).

Desktop: none directly; benefits all hosts equally since `restartRepair` runs in the shared agent runtime.

CLI: none directly; same shared-runtime benefit.

## Scheduled refactoring

None. The issue's own "optional hardening" (making `markUnavailable` refuse to overwrite a `reserved` entry) is a separate behavior decision per the issue text and is left for a future PR if the maintainer wants it.

## Validation

- Added a regression test (`skips a candidate reused while the lease lock was unavailable, instead of marking it unavailable`) that mocks `runWithInactiveExecutionLease` (via `vi.mock` with `importActual` passthrough, so the other 15 tests in the file keep exercising the real lease implementation) to reject once, with `isRepairCandidateCurrent` reporting current at loop-entry and stale by the time the lock rejects.
- Verified the test is a real regression check: reverted the production fix locally, confirmed the new test fails with the exact bug described in #11765 (a hold's `detail` message where none should be), then re-applied the fix and confirmed it passes.
- `npm run typecheck` (workspace + test-kernel) — clean.
- `npx eslint` on both touched files — clean.
- `npx prettier --check` on both touched files — clean.
- `npx vitest run` on `RestartRepair.vitest.ts` (16/16) and the sibling `SessionRestartRepair.vitest.ts` (12/12) — all pass.

https://claude.ai/code/session_01NQCfK7TeMFDGvueTFw4ytt

### DIFF
--- a/src/agent/runtime/restartRepair.ts
+++ b/src/agent/runtime/restartRepair.ts
@@ -322,6 +322,12 @@ export async function repairRestartedStreams(
       // A settlement that already mutated must surface; a lock that could
       // not even be taken proves nothing, so the stream stays unavailable.
       if (settleStarted) throw error;
+      if (!isCurrent(streamId, executionId)) {
+        options.logger?.debug(
+          `Skipped restart repair for stream ${streamId}: it was reused while the lease lock was unavailable`,
+        );
+        continue;
+      }
       applyHold(streamId, executionId, {
         kind: 'unclassified',
         cause: `lease lock unavailable (${toErrorMessage(error)})`,
--- a/src/test-kernel/agent/runtime/RestartRepair.vitest.ts
+++ b/src/test-kernel/agent/runtime/RestartRepair.vitest.ts
@@ -1,5 +1,21 @@
 import { describe, expect, it, vi } from 'vitest';
 
+const mocks = vi.hoisted(() => ({
+  runWithInactiveExecutionLease: vi.fn(),
+}));
+
+vi.mock('@agent/storage/executionLease', async (importActual) => {
+  const actual =
+    await importActual<typeof import('@agent/storage/executionLease')>();
+  mocks.runWithInactiveExecutionLease.mockImplementation(
+    actual.runWithInactiveExecutionLease,
+  );
+  return {
+    ...actual,
+    runWithInactiveExecutionLease: mocks.runWithInactiveExecutionLease,
+  };
+});
+
 import { getExecutionStore } from '@agent/storage';
 import { flowKey, type FlowRecord } from '@agent/node/persistedFlow';
 import type {
@@ -417,6 +433,31 @@ describe('repairRestartedStreams', () => {
     });
 
     expect(setup.streamStatus.get(setup.streamId)).toBeUndefined();
+    expect(closeRunningGroups).not.toHaveBeenCalled();
+  });
+
+  it('skips a candidate reused while the lease lock was unavailable, instead of marking it unavailable', async () => {
+    const setup = setupStream('reused-during-lock');
+    const closeRunningGroups = vi.fn(closeAllGroups);
+    mocks.runWithInactiveExecutionLease.mockRejectedValueOnce(
+      new Error('lock contention'),
+    );
+    let currentCalls = 0;
+
+    await runRepair(setup, {
+      closeRunningGroups,
+      finalizeRun: createDurableFinalizer(),
+      classifyRun: classifyAs({ kind: 'resumable' }),
+      isRepairCandidateCurrent: () => {
+        currentCalls += 1;
+        // Current at the top-of-loop check (so the run reaches the lease
+        // attempt), reused by the time the lock rejects.
+        return currentCalls === 1;
+      },
+    });
+
+    expect(setup.streamStatus.get(setup.streamId)).toBeUndefined();
+    expect(setup.streamStatus.holdState(setup.streamId)).toBeUndefined();
     expect(closeRunningGroups).not.toHaveBeenCalled();
   });
 


LionSR:
Heads-up before this gets reviewed: **the production change here is already on `main`.**

#11765 was also implemented in #11785, which merged at `3422b5f` ("fix(agent): keep a live reservation when restart repair holds a stream"). Two Claude sessions picked up the same issue independently — this one at 22:48Z from `34adb91`, the other about half an hour later. No fault on either side; just flagging so a reviewer doesn't spend time on a diff `main` already carries.

Concretely, `restartRepair.ts`'s lease-lock catch on current `main` reads:

```ts
if (settleStarted) throw error;
if (!isCurrent(streamId, executionId)) {
  options.logger?.debug(
    `Skipped restart repair for stream ${streamId}: it was reused before settlement`,
  );
  continue;
}
```

— the same guard, with the same sibling-matching debug wording this PR proposes.

**What merged additionally**, which this PR deliberately scoped out: the issue's "optional hardening" (`markUnavailable` returning `false` rather than overwriting a live `reserved` entry) and a second re-check closing the narrower window after `await currentLeaseOwner()`. The reasoning for taking the hardening was that the re-check alone is insufficient — `applyHold` has four call sites, and `git show 03fa583^` shows holds lived in a separate map before #11757, so refusing the overwrite restores a prior invariant rather than inventing a policy.

**One thing here that did *not* merge and may be worth keeping:** your regression test. #11785's approach was a `StreamStatusService`-level test for the reservation-versus-hold invariant, and it explicitly declined a `restartRepair`-level test on the grounds that `runWithInactiveExecutionLease` is a direct module import rather than an injectable option. This PR solves that with `vi.mock` + `importActual` passthrough, and reports revert-verifying it — which is stronger evidence than what merged has. If the maintainer wants that coverage, the test is portable to `main` on its own.

Otherwise this can probably be closed as superseded. Not my call, and not my PR to close.

---

PR #11789, opened 2026-09-02
fix: stop dropping the extension's resume-failure notice

### LINKED ISSUE #11766: The extension silently drops its resume-failure notice: showInfoMessage has no extension implementor
## Summary

`HostInteractions.showInfoMessage` is an optional port member with its own replay branch, implemented by **only** the desktop adapter. The VS Code extension never attaches it — yet the extension's resume path calls it, so a user whose resume is refused sees **nothing at all**.

```ts
// packages/extension/src/commands/agent/resumeFromResumeData.ts:50-54
session.interactions.showInfoMessage(describeFollowUpFailure(result.failed), {
  replayWhenAttached: true,
});
```

That dispatches to `active.interactions.showInfoMessage?.(message)` (`HostInteractions.ts:526-537`), which is `undefined` on the extension, so the call is a silent no-op. A queued replay drops it too. Per the repo's silent-degradation rule, a fallback that masks a failure must be loud or not exist; this one is neither.

## Consumers (grepped)

- **Implementors:** 1 — `packages/desktop/src/main/desktopAgentExecution.ts:381`.
- **Extension adapter:** `createExtensionHostInteractions` (`packages/extension/src/progressView/extensionHostInteractions.ts:17-41`) — does not attach it.
- **Producers:** 2 — the extension resume path above, and desktop's resume, which already routes its *sibling* failure through `emit('requestShowError', …, { replayWhenAttached: true })` (`desktopAgentResume.ts:93-110`).

## Two fixes — this needs your decision, not code

1. **Keep the port, make it loud.** Add a three-line implementor to `createExtensionHostInteractions`:
   `showInfoMessage: (m) => void vscode.window.showInformationMessage(m)`. Zero surface change.
2. **Delete the one-message port** (`HostInteractions.ts:355-356`, `:526-537`, and the desktop implementor) and route both producers through `emit('requestShowError', …)`, the shape desktop already uses, so the message reaches all three hosts through the compile-checked presentation map. ≈ **−15 production lines**.

**Option 2 carries a behavior change that must be stated in its PR:** there is no info-level presentation event, so a warning-class message would start rendering as an **error** toast/dialog on both GUI hosts.

## Evidence

`src/agent/runtime/HostInteractions.ts:355-356`, `:526-537`; `packages/extension/src/progressView/extensionHostInteractions.ts:17-41`; `packages/extension/src/commands/agent/resumeFromResumeData.ts:50-54`; `packages/desktop/src/main/desktopAgentExecution.ts:381`; `packages/desktop/src/main/desktopAgentResume.ts:93-110`.

Test retargets are test-only (`SessionInteractions.vitest.ts`, `DesktopAgentExecution.vitest.ts:1956-1996`, `ResumeFromResumeData.vitest.ts:58-69`).

Found by the ownership survey (finding D1); context in `docs/proposals/2026-09-02-runtime-host-session-lifecycle-ownership-survey.md`.

LionSR:
## Summary

`HostInteractions.showInfoMessage` was an optional port member with its own bespoke replay branch, implemented by **only** the desktop adapter. The VS Code extension never attached it — yet the extension's resume path called it, so `active.interactions.showInfoMessage?.(message)` resolved to `undefined` and a user whose resume was refused saw **nothing at all**. A queued replay dropped it too.

This takes **Option 2** from the issue — delete the one-message port, route both producers through the compile-checked presentation map — but with a correction that removes the option's only stated cost.

### The issue's behavior-change caveat does not apply, and shouldn't be paid

The issue says Option 2 must promote the message to an **error** toast because "there is no info-level presentation event". There is one: **`requestShowInstruction`** (`src/agent/runtime/runtimePresentationEvents.ts:21`).

- Desktop renders it via `showInstructionDialog`, which is literally `dialog.showMessageBox({ type: 'info' })` — its own comment says "an instruction is actionable guidance, not a failure, so it uses the info-style dialog" (`desktopAgentExecution.ts:254`, `index.ts`).
- The extension renders it via `handleRequestShowInstruction` → `vscode.window.showInformationMessage` (`agentEventListeners.ts:79`).
- The CLI logs it at `info`.
- `showSuppress: false` with no actions is an established production shape (`AgentLaunchContext.ts:208-213`, `OutputNode.ts:263`).

Routing through `requestShowError` would also have been **internally inconsistent**: the sibling progress-view follow-up path already presents these exact `describeFollowUpFailure` strings through `showInfo` on both GUI hosts (`progressFollowUpSubmit.ts:82-84` → `showInfoMessage` on desktop, `showInformationMessage` on the extension). The same message would have been error-styled in one code path and info-styled in the other, in the same process.

So this is a **pure fix with no severity promotion on any host**.

`showSuppress: false` is deliberate: the extension's `showInstructionWithSuppress` only consults its persisted "never remind again" key when `showSuppress` is true, and a refusal notice must never become permanently suppressible.

## Issue acceptance gates

Closes #11766. The issue asked for a decision, not code; the decision taken is Option 2 with `requestShowInstruction` instead of `requestShowError`, which dominates both stated options — it removes the optional-member trap (Option 1 keeps it, so the next unimplemented host silently drops again) *and* costs no severity change (Option 2's only downside). Say the word if you'd rather have the three-line implementor instead and I'll swap it.

## Single source of truth

The runtime's presentation-event map (`RuntimePresentationEventPayloads`) is now the only channel from the agent runtime to a host's notification surface. There is no second, optional side channel a host can leave unimplemented.

Note the port deleted here is `HostInteractions.showInfoMessage`. The unrelated `UiHost`/`MessageHost.showInfoMessage` notification surface (`src/hosts/uiHosts.ts:34`, and ~60 desktop/settings call sites) is a different member and is untouched.

Removes a port member, its class method, its bespoke replay branch, and its only implementor — four constructs collapsed onto one already-shared channel. No abstraction added. `queuePresentationReplay` and `AgentRuntimeEmitOptions` stay: `emit` still uses both.

## Net elements (R6)

Files: 4 production, 1 doc, 3 test. Exported symbols: 0 added, 0 removed. Interfaces: 1 member removed (`HostInteractions.showInfoMessage?`); 1 class method + its replay branch removed. Net LoC: **≈ −15 production** (the desktop spread collapses to a bare call as a side effect), with test churn roughly neutral.

## Consumer counts (R8)

`HostInteractions.showInfoMessage` — grepped at HEAD: **1 implementor** (`desktopAgentExecution.ts:381`), **2 producers** (`packages/extension/src/commands/agent/resumeFromResumeData.ts:52`, `packages/desktop/src/main/desktopAgentResume.ts:94`), **0 in `packages/cli`**, **0 in `packages/agent`**. Both producers are re-routed; the implementor is deleted. `requestShowInstruction` gains one producer and has handlers on all three hosts already.

## Host impact

Extension: **the fix** — a refused resume now shows an information message where it previously showed nothing.

Desktop: unchanged dialog class. `showInstructionDialog` is a `type: 'info'` message box, the same class `showInfoMessage` used; the only delta is a "Dismiss" button label instead of the default OK.

CLI: unchanged — it never implemented the port; the instruction event logs at `info` as before.

## Scheduled refactoring

None.

## Validation

- `npm run typecheck:workspace`, `typecheck:test-kernel`, `typecheck:desktop` — all clean.
- `npx eslint` on all four production files — clean. `npx prettier --check` on every changed file — clean.
- `node scripts/check-guidance-refs.mjs` — OK (58 files). `docs/architecture/2026-07-26-embedding-the-agent-runtime.md:476` named `showInfoMessage` in the port's typed shape and is updated with it; that directory is inside `GUIDANCE_DIRS`.
- `npx vitest run` over `SessionInteractions.vitest.ts`, `ResumeFromResumeData.vitest.ts`, `DesktopAgentResume.vitest.ts` — 45 passed, 0 failed.
- `npx vitest run src/test-kernel/desktop/DesktopAgentExecution.vitest.ts` — 75 passed, **2 failed**. See below.

### ⚠️ Two pre-existing failures on `main`, not from this PR

`DesktopAgentExecution.vitest.ts` fails these two on the current `main` as well:

- `presents a merge failure that occurs before lifecycle startup`
- `presents a terminal merge failure exactly once from its result`

Verified by stashing this branch's changes and running the same file at `origin/main` (`34adb91`) twice: **2 failed | 75 passed** both times, the same two tests, with identical `showErrorMessage … Number of calls: 0` assertions. They are in the merge-failure presentation path, which this PR does not touch. Flagging rather than fixing, since it is unrelated to the diff — happy to open a separate issue.

Test retargets are mechanical and test-only, as the issue predicted: four `SessionInteractions.vitest.ts` cases lose their now-duplicated `showInfoMessage` arms or move to `requestShowError`, `ResumeFromResumeData.vitest.ts:58-69` asserts on the emit, and the desktop replay test moves to `showInstructionDialog` (asserted with the trailing `undefined`, since the handler passes `instruction.actions` and the payload has none).

---

### DIFF
--- a/docs/architecture/2026-07-26-embedding-the-agent-runtime.md
+++ b/docs/architecture/2026-07-26-embedding-the-agent-runtime.md
@@ -473,8 +473,8 @@ the runtime already knows the parking behaviour exists.
 
 `cancel` is the one **required** member of `HostInteractions`
 (`src/agent/runtime/HostInteractions.ts:336`); every other member — the seven
-request methods plus `emit`, `dispose`, `showInfoMessage`, the diagnostics
-readers, and `setApprovalBypassState` — is optional
+request methods plus `emit`, `dispose`, the diagnostics readers, and
+`setApprovalBypassState` — is optional
 (`src/agent/runtime/HostInteractions.ts:293-338`). So the compiler already
 forces you to write `{ cancel: … }` — the trap is not a badly-typed object, it
 is **never calling `interactions.use` at all**, which no type can catch.
--- a/packages/desktop/src/main/desktopAgentExecution.ts
+++ b/packages/desktop/src/main/desktopAgentExecution.ts
@@ -367,19 +367,17 @@ export class DesktopProgressBridge {
         this.backend.approvalHandlers.toolEdit.dismiss(requestId),
       detachCause: SESSION_DISPOSED_CAUSE,
     });
-    this.hostInteractions = {
-      // The shared progress-view host port plus the Electron shell's
-      // notification surface, which the runtime reaches through
-      // `session.interactions`.
-      ...createProgressHostInteractions({
-        interactions: presentationHost,
-        session: this.session,
-        getApprovalHandlers: () => this.backend.approvalHandlers,
-        getToolEditApprovals: () => this.toolEditApprovals!,
-        setApprovalBypassState: this.backend.setApprovalBypassState,
-      }),
-      showInfoMessage: (message) => this.options.host.showInfoMessage(message),
-    };
+    // The shared progress-view host port. Every notice the runtime raises now
+    // reaches this window through `presentationHost`'s presentation-event
+    // map, so there is no second, optional channel a host can leave
+    // unimplemented.
+    this.hostInteractions = createProgressHostInteractions({
+      interactions: presentationHost,
+      session: this.session,
+      getApprovalHandlers: () => this.backend.approvalHandlers,
+      getToolEditApprovals: () => this.toolEditApprovals!,
+      setApprovalBypassState: this.backend.setApprovalBypassState,
+    });
     this.fileActions = new DesktopProgressFileActions(this.options.host, {
       startExecution: (request) => {
         const logger = this.logger;
--- a/packages/desktop/src/main/desktopAgentResume.ts
+++ b/packages/desktop/src/main/desktopAgentResume.ts
@@ -91,8 +91,15 @@ async function resumeDesktopStream(
     });
     if ('started' in result) return result.delivered;
     if (!isCancellationRequested()) {
-      await session.interactions.showInfoMessage(
-        describeFollowUpFailure(result.failed),
+      // Same info-styled channel the sibling progress-view follow-up path
+      // uses for this wording; desktop renders it as an 'info' message box.
+      await session.interactions.emit(
+        'requestShowInstruction',
+        {
+          key: 'resumeRefused',
+          message: describeFollowUpFailure(result.failed),
+          showSuppress: false,
+        },
         { replayWhenAttached: true },
       );
     }
--- a/packages/extension/src/commands/agent/resumeFromResumeData.ts
+++ b/packages/extension/src/commands/agent/resumeFromResumeData.ts
@@ -49,8 +49,21 @@ export async function tryResumeFromResumeData(
     if ('started' in result) return result.delivered;
     if (!isCancellationRequested()) {
       logger.warn(`Stream ${streamId} was not resumed: ${result.failed}`);
-      await session.interactions.showInfoMessage(
-        describeFollowUpFailure(result.failed),
+      // A refused resume is actionable guidance ("start a new agent task",
+      // "send the message there"), not a failure: the shared progress-view
+      // path presents these same `describeFollowUpFailure` strings through
+      // `showInfo` on both GUI hosts (progressFollowUpSubmit.ts), so this
+      // rides the info-styled `requestShowInstruction` event -- the one
+      // presentation channel every host's compile-checked handler map must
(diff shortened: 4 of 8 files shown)


LionSR:
Correction to the "⚠️ Two pre-existing failures on `main`" note in the description: **CI is green here**, both kernel-test shards included, so that note overstates the case.

What I observed was real but environment-specific. In this container, `DesktopAgentExecution.vitest.ts` fails `presents a merge failure that occurs before lifecycle startup` and `presents a terminal merge failure exactly once from its result` — identically on this branch and on a stashed-clean `origin/main` (34adb91), twice each. So they are still **not caused by this diff**, which is the part that mattered for reviewing it. But they are not broken on `main` either: CI runs both to green, so the right reading is a local/container flake in those two `vi.waitFor` assertions, not a repository-level break.

No action needed on this PR. Flagging in case the same two ever show up flaky in CI.

---

PR #11792, opened 2026-09-02
refactor: delete five dead surfaces in the runtime and CLI approval layers

### LINKED ISSUE #11772: Batch: five high-confidence dead-surface deletions in the runtime and CLI approval layers
## Summary

Five independent deletions from the ownership survey, each cleared by two adversarial lenses at high confidence. Batched because each is small and they touch nearby code; they can also be taken separately. Combined ≈ **−140 production lines**, no new elements.

---

**1. The run-subscriber assertion is a tautology** (≈ −50 prod, ≈ −87 test)

`SessionEventHub` keeps two counters updated on *every* subscribe/unsubscribe so `assertRunSubscribersAttachedBeforeActivation` (`:207-225`) can throw in dev and warn in production. But `SessionHandle`'s constructor calls `snapshots.attachSessionEvents(events, …)` (`:188-192`), which registers a `{ scope: 'run', types: SNAPSHOT_RUN_FACT_TYPES }` subscription with no streamId, detached only at teardown (`StreamSnapshotStore.ts:687`). Every session therefore has a run-scope subscriber from construction and the guard's early return is **always** taken. Both production call sites (`AgentLaunchContext.ts:408`, `childStream.ts:124`) reach the hub through a `SessionHandle`.

Making it meaningful would need subscriber tagging — new vocabulary, banned by R4. Delete the method, both counters, the bookkeeping, the `isDevAssertionMode` import, the two call lines, and the two tests that observe it (one uses a bare `new SessionEventHub()`, the other spies on it as an ordering probe).

**2. The live wire union accepts a retired seven-value status** (≈ −10)

`StreamLifecycleStatusSchema` (`src/shared/schemas/stream.ts:195-200`) carries a fourth member, `StreamStatusSchema.transform(streamStatusToLifecycleStatus)` (`:199`), that re-maps `error|stopped|resuming|initializing`. Its one consumer is `BackendOwnedFieldsSchema.status` (`streamState.ts:127`), whose producers are typed `StreamPhase | 'ready' | 'unavailable'` and ship in the same bundle. The real trace-import boundary is `StreamSnapshotSchema.status`, whose own `LegacyStreamStatusAsPhaseSchema` (`streamSnapshot.ts:95-100`) normalizes archived `trace.json` **before** anything reaches this schema.

Delete the member and fix the comment above `STREAM_STATUS`, which currently points at the wrong boundary. `z.output` is unchanged, so no consumer retypes. `StreamStatus` keeps one in-file consumer and can only lose its `export`.

**3. A three-arm pseudo-event vocabulary switched on once, in the file that builds it** (≈ −50)

`CliDecisionApprovalRequest` (`packages/cli/src/runtime/approval/approvalPrompts.ts:24-49`) copies the progress-view wire arm names that nothing emits — its own docblock concedes this — and its only consumer, `approvalAdapter.ts`, constructs it at three sites and immediately destructures it in a switch and an `isRetry` branch.

Replace `decideApprovalEvent(request, …)` with a `decideGated(context, hooks, immediate, content, options)` that each call site parameterizes directly; delete the union, `summarizeApprovalEvent`, and the file's only `assertNever`. Prompt text, stderr lines and the `prompted` flag unchanged.

*Sequence against the open #11461*, which would relocate the same `immediate` policy settlement — land this first or fold it in, but do not reshape `decideApprovalEvent` concurrently.

**4. A test-only injection seam the design doc never specified** (≈ −15 prod, −100…−173 test)

`createLatexExecutionDiscovery(dependencies)` (`src/agent/storage/executionListing.ts:288-310`) declares a dependency interface, a frozen default, and a defaulted parameter whose doc says it exists so the contract "can be unit tested without scanning real execution storage". Both production callers pass nothing (`latexdiffCommands.ts:433`, `desktopProgressFileActions.ts:222`); the only passer is its own suite. The 2026-08-15 latex-agent-port design specifies the function with **no parameter**.

Delete the seam; seed the suite from real storage as `ExecutionListing.vitest.ts` does, or keep one real-storage case.

**5. A dead relative-path fallback** (≈ −13)

`ToolEditApprovalController.relativeDisplayPath` (`:44-55`) wraps `WorkspaceFS.relativePath` in `try/catch → basename` "for a host without an initialized platform". That throw requires an uninstalled platform; all three hosts install one before any run, the TUI calls the same function unguarded for the same request (`subscribeApprovals.ts:162`), and vitest installs a fake platform globally. Delete it and call `WorkspaceFS.relativePath` directly, so a missing platform fails loudly instead of rendering a basename.

---

Per R6/R8 the implementing PR should state net elements and the grepped consumer counts; the survey record carries both. Context: `docs/proposals/2026-09-02-runtime-host-session-lifecycle-ownership-survey.md` §2.2 (findings S1, S3, S5, S9, S11).

LionSR:
## Summary

The five deletions from #11772, taken as one batch. Each was independently re-verified at `34adb91` before implementing; all five held, with only cosmetic line drift.

**1. The run-subscriber assertion is a tautology.** `SessionHandle`'s constructor calls `snapshots.attachSessionEvents`, which registers a `{ scope: 'run', types: SNAPSHOT_RUN_FACT_TYPES }` subscription with no `streamId`, detached only at teardown. Every session therefore has a run-scope subscriber from construction, both production call sites (`AgentLaunchContext.ts`, `childStream.ts`) reach the hub through a `SessionHandle`, and the guard's early return is always taken. I looked specifically for a production path constructing `SessionEventHub` *without* the snapshot attach and found none. Making it meaningful would need subscriber tagging — new vocabulary, banned by R4.

**2. The live wire union accepted a retired seven-value status.** `StreamLifecycleStatusSchema`'s `StreamStatusSchema.transform(...)` member re-mapped `error|stopped|resuming|initializing`. Its one consumer is `BackendOwnedFieldsSchema.status`, whose producers are typed `StreamPhase | 'ready' | 'unavailable'` and ship in the same bundle. The real trace-import boundary is `StreamSnapshotSchema.status`, whose `LegacyStreamStatusAsPhaseSchema` normalizes archived `trace.json` *before* anything reaches this schema.

**3. A three-arm pseudo-event vocabulary switched on once, in the file that built it.** `CliDecisionApprovalRequest` copied progress-view wire arm names nothing emits — its own docblock conceded this.

**4. A test-only injection seam the design doc never specified.** Both production callers of `createLatexExecutionDiscovery` passed nothing; the only passer was its own suite, and `docs/proposals/2026-08-15-latex-agent-port-design.md:166-181` specifies the function with no parameter.

**5. A dead relative-path fallback.** `relativeDisplayPath`'s `try/catch → basename` requires an uninstalled platform to fire. All three hosts install one at their composition root before any run, the TUI calls `WorkspaceFS.relativePath` unguarded for the same request (`subscribeApprovals.ts:162`), and vitest installs a fake platform globally.

## Issue acceptance gates

Closes #11772. All five items implemented. One deliberate scope addition inside item 1: **`withSessionEventRecording` is deleted, not kept with a rewritten docblock.** Its only reason to exist was the deleted assertion (its own docblock said so), and `recordSessionEvents` merely subscribes a recorder nothing asserts — so after this change `withSessionEventRecording(() => X)` is provably equivalent to `X`. Keeping it would enshrine a no-op wrapper, which the abstraction-discipline rule forbids.

Sequencing against **#11461** (item 3's rider): #11461 would relocate the same `immediate` policy settlement. This lands first and does not reshape it concurrently — `settleRetry`/`settleExecutable` keep their current call sites and semantics, only moving from inside `decideApprovalEvent` to the three callers.

## Single source of truth

- Item 1: `SessionHandle`'s constructor is the single place a run-scope subscriber is guaranteed; the hub no longer keeps a shadow count of that fact.
- Item 2: `LegacyStreamStatusAsPhaseSchema` inside `StreamSnapshotSchema.status` is now the *only* place the retired vocabulary is normalized. The corrected comment above `STREAM_STATUS` says so.
- Item 3: each call site owns its own policy settlement and its own content; `decideGated` owns only "settled already, or prompt".
- Item 4: real execution storage is the only source for the discovery projection.

Removes: one guard method + two counters + their bookkeeping, one no-op test wrapper, one dead schema union member, one 3-arm discriminated union + its summarizer + an `assertNever`, one dependency interface + frozen default + defaulted parameter, one try/catch wrapper. **No new abstraction.** `decideGated` is not a new layer — it is `decideApprovalEvent` with the union-dispatch removed, keeping its single behavior (settle-or-prompt).

## Net elements (R6)

Files: 8 production, 4 test, **1 test file deleted**. Diffstat: 13 files changed. Exported symbols: **−2** (`CliDecisionApprovalRequest`; `StreamStatus` loses `export` and stays as a module-local type). Interfaces/types deleted: 2 (`CliDecisionApprovalRequest`, `LatexExecutionDiscoveryDependencies`). Functions deleted: 5 (`assertRunSubscribersAttachedBeforeActivation`, `withSessionEventRecording`, `summarizeApprovalEvent`, `relativeDisplayPath`, plus `decideApprovalEvent` replaced by `decideGated`). Consts deleted: 1 (`DEFAULT_LATEX_EXECUTION_DISCOVERY_DEPENDENCIES`). Net LoC: **≈ −140 production**, **≈ −250 test** (including the 173-line deleted suite), no new elements. `z.output` of `StreamLifecycleStatusSchema` is unchanged, so no consumer retypes.

## Consumer counts (R8)

- `assertRunSubscribersAttachedBeforeActivation`: 2 production call sites (both deleted), 2 test observers (both deleted). Post-change grep across `src`/`packages` returns zero.
- `CliDecisionApprovalRequest` / `summarizeApprovalEvent` / `decideApprovalEvent`: 1 consumer file (`approvalAdapter.ts`), 3 construction sites. Post-change grep returns zero. Note `src/test-kernel/agent/progressTestUtils.ts` keeps a *parallel* `{event:'showPlanApproval'|…}` fixture vocabulary — it imports none of these symbols and is deliberately untouched; unifying it would widen the blast radius.
- `createLatexExecutionDiscovery`: 2 production callers (`latexdiffCommands.ts:433`, `desktopProgressFileActions.ts:222`), both already passing nothing, both unchanged.
- `relativeDisplayPath`: 1 consumer, module-local, never exported.
- `StreamStatus`: re-exported only by the `src/shared/schemas/index.ts` star barrel, which nothing imports it from; `packages/agent/src/schemas.ts` does not re-export it, so the fenced SDK surface is untouched.

## Host impact

Extension: item 5 only — a missing platform in the tool-edit approval path now throws instead of silently rendering a basename. No host installs the controller without a platform.

Desktop: same as extension for item 5; item 4's `desktopProgressFileActions.ts` caller is unchanged.

CLI: item 3 — prompt text, stderr lines, the `prompted` flag and the retry pre-prompt double-call are all byte-identical. Verified by mapping each of the three call sites: `request.plan` is the same field as `request.payload.plan` (`HostPlanApprovalRequest = Readonly<PlanApprovalPermission>`), `buildAgentProposalApprovalContent` takes exactly `HostAgentProposalRequest`, and left-to-right argument evaluation keeps `settleExecutable(context)` running before the summary is built — which matters because `settleExecutable` can itself emit the `warnApprovalDenied` stderr line.

## Scheduled refactoring

Not coupled to survey item S2 (the hub's `dispose()`). The survey suggested landing them together "so no stale counter survives on a dead hub", but once the counters are deleted there is no stale counter to survive. S2 remains separate.

## Validation

- `npm run typecheck` — **all six projects clean** (workspace, test-kernel, agent, cli, trace-viewer, desktop).
- `npm run lint` over every changed file — clean. `npm run format` / `prettier --check` — clean.
- `npm run check:dead-code-ratchet` — OK, no new findings (294 vs 294 baselined). No baseline or ratchet file needed editing: no new export, no host-boundary import, no schema deep-import, no snapshot.
- `npx vitest run` — `SessionEventHub`, `ChildStreamProgressEvents`, `ExecutionListing`, `ToolEditApprovalController`, `VscodeToolEditApproval`, `DesktopToolEditApproval`, `SchemaMigrations`, `SharedStreams`: **101 passed**. `ApprovalAdapter`: **40 passed**. `ApprovalDeniedWarn`, `ApprovalQueue`, `BashApproval`, `EditApproval`, `PlanApproval`, `TuiApprovalRetry`: **89 passed**. 0 failed across all of them.

Note on gates for items 1 and 3: nothing in the repo would have caught a leftover unused import — `@typescript-eslint/no-unused-vars` is `off`, no tsconfig sets `noUnusedLocals`, and knip analyzes files/exports/types/duplicates, not imports. The `isDevAssertionMode`, `AgentProposalPermission`/`PlanApprovalPermission`, `ExecutionMeta` and `assertNever` imports were removed deliberately and verified by grep, not by a gate.

Test accounting: one 173-line suite deleted (three of its four cases asserted a single `?.streamId` optional chain through the mock being deleted; the fourth tested projection against a stubbed `listExecutions`, which never proved the default wiring worked), replaced by **one** real-storage case extending `ExecutionListing.vitest.ts`. The two tests observing the deleted assertion are removed with it. `SchemaMigrations.vitest.ts:125-153` already covers item 2's surviving boundary in both directions, so no test was added there.

---


packages/cli/src/runtime/approvalAdapter.ts
```diff
@@ -30,7 +29,6 @@ import {
   type CliApprovalContent,
   type CliApprovalDecision,
   type CliApprovalPromptHooks,
```
LionSR: All these also are suspicious and might be possible to collapse too

Issue #11771, opened 2026-09-02
Historical ownership decisions implemented; obsolete compatibility superseded


LionSR:
## Current status — 2026-09-15

Reviewed on 2026-09-15 against main `781d6630848878cf53aba9c6c92f00778370fbcd`. This status supersedes historical paths, counts and unfinished-PR claims below. Source review is not a new live-service or interactive verification run.

## Disposition

The three recorded decision PRs #11794, #11795 and #11796 merged. The legacy trace permanence direction was subsequently superseded by the accepted 1.0 no-legacy export policy (#12359); do not revive its old dated-bookkeeping requirement. The narrow quit-policy implementation landed. Any current interactive exit regression belongs to the relevant CLI/lifetime issue, not another vote on this completed decision record.

Closing this bounded historical issue does not claim completion of unrelated runtime, release or operator requirements.

<details>
<summary>Historical issue description (superseded status; retained for evidence)</summary>

## Why this is one issue

Three questions surfaced by the runtime/host/session/lifecycle ownership survey that are **decisions, not work**. Each is cheap to answer and two of them shape other items. None should be implemented before it is ruled on.

---

## 1. The `taskRuns` absolute-path arm has no ledger row anywhere

`runStorageLocationFromAnyAbsolutePath` still loops over a second root when classifying absolute paths from persisted transcripts and resume records (`src/utils/files/runStorageFs.ts:166-181`, the sole production reader of `WORKSPACE_STORAGE_LAYOUT.legacyRuns`, `src/common/storage/storageLayout.ts:12`). Its comment cites a **#6981 retention policy that exists as a row nowhere**: #6981 closed 2026-08-09, and neither #9422 nor #9627 carries a `taskRuns` row.

History: #9430 item 5 ruled "keep the pure legacy absolute-path parser … track that condition in #6981"; #9623 executed that; #9620 (the same deletion proposed here) was closed unmerged as superseded; #10887 R8 then kept it citing the now-nonexistent retention.

Two complications worth your judgement:

- The directory probe this arm served is **already gone** (`runStorageFs.ts:44-54` resolve only `executions/`).
- The arm matches only paths under the **current** storage root, and the extension root moved to `~/.texra` — so the population it was kept for is largely unreachable anyway.

[...]

Issue #9422, opened 2026-07-30
Tracking: retire the four remaining persisted-format readers


LionSR:
## Current scope

Four compatibility readers remain on current main. Each stays only until three months after its replacement became the sole writer:

- `AgentWorkspaceLegacySnapshotSchema` in `src/agent/core/state/AgentWorkspaceState.ts`: retire after **2026-08-14**.
- `LegacyRunUsageAccumulatorSchema` in `src/agent/core/usage/RunUsageAccumulator.ts`: retire after **2026-09-22**.
- persisted disk `meta.taskState` fallback in `src/transcript/StreamSnapshotStore.ts` and its schema allowance in `src/shared/schemas/streamData.ts`: retire after **2026-10-05**. The separate public CLI NDJSON task-state vocabulary is not part of this deletion.
- pre-`modelId` recovery in `migrateSharedState` at `src/agent/implementations/flows/tooluse/nodes/types.ts`: retire after **2026-11-01**.

These dates are based on replacement-writer history, not on the age of the readers.

## Completed arms

- #9593 removed the old result-metadata parser and its compatibility-only tests.
- #9605 removed pre-May nested round sidecars, archived instruction backfill, old typed stream-log rows, historical delegation names, and obsolete checkpoint documentation.
- #9606 removed the January nested shared-state wrapper and the February `conversation` alias.

## Retirement rule

For each remaining date:

1. Reconfirm the old writer, current writer, replacement date, all readers, and protected public surfaces against then-current main.
2. Remove the complete obsolete cluster: schema arm, transform or fallback, comments, fixtures, and compatibility-only tests.
3. Preserve only current behavioral tests and any released public protocol explicitly excluded above.
4. Make malformed unsupported data fail or warn at its real boundary; do not silently turn authored data into an ordinary default.
5. Update #6981 and close this issue only after all four arms have been removed or moved to a precise external blocker.

LionSR:
Related: #9431 (legacy-storage probes) and #9424 (history migration) are the other two survivors of the same pre-shared-storage generation. Item #9433's resultMeta precedence flip and #9432's dated v0.41 removal also touch legacy read arms; sequence the sunset decisions together.

LionSR:
Umbrella: #9434 proposes the boundary-only migration system and the authored-vs-derived data tiering that would stop these legacy arms from accumulating. Worth agreeing the tiering there before deciding this sunset, since derived-tier data can be discarded-and-rebuilt rather than migrated.

[...]

Issue #8622, opened 2026-07-16
Extension: share the ~/.texra storage root so CLI/desktop conversations show up in VS Code


LionSR:
## Problem

Conversations run in the \`texra\` CLI (and desktop) are invisible in the VS Code extension. All three hosts use byte-identical storage code and on-disk formats (\`ExecutionKVStore\`, \`@transcript\` stores), but the extension roots \`platform().storage\` at VS Code's per-extension \`context.storageUri\`, while CLI and desktop share \`~/.texra/workspace-storage/<basename>-<hash8>/\` (#7987). The extension is the lone outlier; \`src/transcript/streamDataPaths.ts\` explicitly anticipates this "cross-host shared-root flip" as a configuration change.

## Plan

- Wire the extension's \`storage\` port to \`WorkspaceStorageProvider(DEFAULT_NODE_STORAGE_ROOT, ...)\` (the desktop construct at \`packages/desktop/src/main/platform/index.ts:124\`). Only the storage port moves — VS Code Mementos, secrets, and \`.vscode/settings.json\` config stay put.
- One-time best-effort migration of existing extension data from \`context.storageUri\` / \`globalStorageUri\` into the shared root (desktop's \`migrateLegacyDesktopDataRoot\` pattern: per-directory move, never clobber; merge id-keyed dirs like \`executions/\` and \`streamData/\` per child).
- Cross-platform: normalize the Windows drive letter — VS Code's \`uri.fsPath\` lowercases it while the CLI hashes \`process.cwd()\`, which would land in a different workspace bucket.
- Freshness: invalidate the execution-listing cache on history refresh so runs written by another host appear without reloading the window.

## Concurrency notes (why this is safe)

All JSON writes go through write-file-atomic (temp+fsync+rename), so cross-host readers never see torn files. Each execution has a unique id and is only ever written by the host that started it (single-writer-per-execution, the model desktop↔CLI already rely on). \`listExecutions()\` already tolerates half-created execution dirs (missing meta → skipped, per-execution try/catch).

Remaining cross-write risk (delete/housekeeping while another host is mid-run) and live file-watching refresh are deferred to a follow-up issue.

PR #11794, opened 2026-09-03
refactor(storage): delete the taskRuns absolute-path arm

LionSR:
## Summary

Implements **#11771 ruling 1**, as decided: delete the arm, create no tracker row.

`runStorageLocationFromAnyAbsolutePath` looped over a second root when classifying absolute paths from persisted transcripts and resume records, citing a #6981 retention policy that exists as a row nowhere.

**The "stamp it" option was not available.** #6981 (completed 2026-08-09), #9422 and #9627 (both `not_planned`, same day) are all closed, and no open tracking-labelled retirement ledger exists. Stamping would have meant first creating an issue whose only purpose is to hold a row for this code.

**The arm is unreachable, not merely stale.** Both roots are built with `StorageFS.fullPath(...)`, which joins against the *current* storage base, and matching is `isPathWithin` — so it can only match `<current storage root>/taskRuns/<executionId>/…`. The last build that wrote `taskRuns/` was **2026-02-12** (`bedbf19e64`, pinned by #9623); the extension's storage port moved to the shared `~/.texra` root on **2026-07-16** (#8622 / PR #8624). Five months apart, so no build ever wrote that combination. Absolute `taskRuns` paths in released transcripts sit under the *old* VS Code root and fail `isPathWithin` against the current one. The one route that could have produced such a directory was #9623's migration, which *consumed* it into `executions/` — and #10887 deleted that migration outright.

**It was not load-bearing even on a hypothetical match.** `childRunOutput.ts:18` uses the result purely as a lookup token and immediately re-resolves through `inspectRunStorageEntry(reference.executionId, reference.relativePath)` — under `executions/`, never `taskRuns/`. The arm relabels a path; it cannot restore access to anything.

**#10887's R8 keep rested on a factual error.** It reasoned `runStorageFs.ts:180` "still reads it under a *different* retention (#6981…) — a live reader outside this ruling's scope." #6981 had closed nine days earlier. Deleting it applies that PR's own recorded ruling that dated compat horizons for intermediate-era data are void.

## Issue acceptance gates

Part of #11771 (ruling 1); rulings 2 and 3 ship separately.

- Arm deleted, `legacyRuns` layout key deleted with its only reader, two compat-only test assertions deleted — **done**.
- SQLite PRD edited — **done, in all five places**, not the one the issue named: `docs/prds/2026-08-16-sqlite-workspace-state.md` lines 133-134 (§3.1 target-layout diagram), 137-140 (the grandfathering paragraph), 714 (§7 ratchet allowlist), 742-746 (the strictness condition), 769-770 (§9 acceptance criterion). §7 and §9 still enumerate the same set, as that criterion requires. Line 126 stays — it is the "today" column and describes a directory that can still sit on disk.
- **No tracker row created**, as ruled.

## Single source of truth

`resolveRunStoragePath()` → `executions/` is now the only run-storage root anywhere. This was the last surviving two-root site: `findExistingRunStoragePath`, `findRunDir`, `ensureRunDir`, `getRunStorageAbsolutePath` and `inspectRunStorageEntry` already resolved exclusively through it.

Removes a two-root loop and the layout key that fed it. No abstraction added. The single-root form also fixes a latent quirk: today a path under the primary root with a bad execution id returns `undefined` immediately rather than falling through to the second root, so behavior for the surviving root is unchanged.

## Net elements (R6)

Files: 2 production, 1 test, 1 doc. Exported symbols: 0 added, 0 removed (`WORKSPACE_STORAGE_LAYOUT` stays exported and used; it loses one key). Net LoC: **≈ −40** including doc edits.

## Consumer counts (R8)

`WORKSPACE_STORAGE_LAYOUT.legacyRuns` / `taskRuns`: **1 production reader** (`runStorageFs.ts:180`, deleted). Remaining hits before this PR were two compat-only test assertions (`RunStorageEntryInspection.vitest.ts:177`, `:210`, both deleted), the PRD, and historical proposal docs. Post-change grep across `src`/`packages` returns nothing.

`runStorageLocationFromAnyAbsolutePath` keeps its three callers (`childRunOutput.ts:18`, `taskRunStorage.ts:122` and `:145`); only the second root is gone.

## Host impact

Extension: a hypothetical `~/.texra/workspace-storage/<id>/taskRuns/…` path in a persisted transcript now classifies as workspace/external rather than `runStorage`. No shipped writer produces that path.

Desktop: same, and it never wrote `taskRuns/` under the shared root at all (it was on `~/.texra` from #7987).

CLI: same.

### Accepted behavioral outcome, stated loudly

If such a directory somehow exists and a transcript references it, `TaskRunFileService.locateSource` classifies the path as workspace/external and `resolveChildRunOutput` throws **"Workflow output is not inside task-run storage."** That is a loud failure on data no shipped writer produces — not silent degradation.

## Scheduled refactoring

None. No ledger row created, no other tracker needs updating.

Noted while here, **not acted on**: two other code comments point at "a row on #9627", a closed tracker — `src/shared/deliveryTags.ts:55` and `src/agent/storage/executionLease.ts:104`. That is a separate, wider hygiene problem outside this ruling.

## Validation

- `npm run typecheck:workspace` and `npm run typecheck:test-kernel` — clean.
- `npx vitest run src/test-kernel/utils/files` — 36 passed, 0 failed.
- `npx eslint` on both changed source files — clean. `npx prettier --check` on all changed files — clean.
- `node docs/scripts/check-root-docs.mjs` — OK (the edit is under `docs/prds/`, already excluded from publication, so the texra.ai deploy is untouched).

No new tests: the remaining cases in `RunStorageEntryInspection.vitest.ts` already cover the surviving single-root behavior, and the one test whose title claimed "current and legacy" is renamed to match what it now asserts.

---

### DIFF
--- a/docs/prds/2026-08-16-sqlite-workspace-state.md
+++ b/docs/prds/2026-08-16-sqlite-workspace-state.md
@@ -130,14 +130,12 @@ workspace storage today                     workspace storage after
                                             ├── pasted/  recordings/
                                             │                 (user-media caches: pasted
                                             │                  images, audio — documents)
-                                            └── taskRuns/     (legacy, until #6981
-                                                               retention drains it)
 ```
 
-Legacy `taskRuns/` (`WORKSPACE_STORAGE_LAYOUT.legacyRuns`) is grandfathered:
-it stays read-only on its existing #6981 retention path, is never migrated
-into rows, and its layout key retires when that policy drains it. It is the
-one deliberate exception to the §2 rule, time-bounded by retention.
+Legacy `taskRuns/` is not part of the target layout. Its last writer was
+retired on 2026-02-12, its absolute-path reader and the
+`WORKSPACE_STORAGE_LAYOUT.legacyRuns` key were deleted (#11771 ruling 1), and
+any residual directory is inert on disk. There is no §2 exception.
 
 Engine: `node:sqlite` (fallback: `better-sqlite3`, accepting the packaging
 cost, if Stage 0 finds the built-in flagged or unfit on any host). Schema and
@@ -711,7 +709,7 @@ public surfaces (`store-public-surface-baseline`); hosts-as-renderers plane
 rules untouched. **Added:** the §2 rule, pinned by a new architecture test.
 The test lands at Stage 1 as a **shrinking ratchet, not a strict gate**: it
 starts with an explicit allowlist of the not-yet-migrated directories
-(`streamLogSummaries/`, `streamData/`, `streamData.deleting/`, `taskRuns/` — the staged-
+(`streamLogSummaries/`, `streamData/`, `streamData.deleting/` — the staged-
 deletion namespace `StagedDeletionCoordinator` renames into, retained in
 the baseline through the **Stage-7** deletion-protocol cutover that retires
 the coordinator — `streamLogs/`, `executions/` KV records, lease/lock
@@ -739,11 +737,7 @@ After Stage 6 one **importer-owned temporary** entry remains alongside the perma
 `*.pre-sqlite-backup`, which the §5 retirement window requires for 90 days
 and two releases — the importer-removal PR deletes the backups and this
 ratchet entry together. Strict (permanent entries only) is that removal
-PR's exit criterion, not Stage 6's — and strictness is additionally
-conditional on legacy `taskRuns/` having drained under #6981's own
-retirement condition, which is independent of the importer window: if the
-importer retires first, `taskRuns/` remains a shrinking temporary entry
-until #6981 removes it, never a re-widen.
+PR's exit criterion, not Stage 6's.
 
 ## 8. Non-goals
 
@@ -766,8 +760,7 @@ normalization of entry payload internals; no publication of this doc
   `{ texra.db, texra.db-wal, texra.db-shm, original, memories, state.json,
 config.json, _workspace.json, pasted, recordings }` plus the per-execution
   artifacts
-  area (§5 Stage 5) — and legacy `taskRuns/` until #6981 retention drains
-  it (§3.1 exception). The §7 `exports` allowlist entry is a category for
+  area (§5 Stage 5). The §7 `exports` allowlist entry is a category for
   caller-chosen destinations outside the workspace-storage bucket, not a
   `WORKSPACE_STORAGE_LAYOUT` key. §7 and this criterion therefore enumerate
   the same bucket set by construction; a divergence between the two lists is
--- a/src/common/storage/storageLayout.ts
+++ b/src/common/storage/storageLayout.ts
@@ -9,7 +9,6 @@ export const WORKSPACE_STORAGE_LAYOUT = Object.freeze({
   memory: 'memories',
   runs: 'executions',
   executionLeases: 'executionLeases',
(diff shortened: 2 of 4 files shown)


LionSR:
## On the P2 review comment — the mechanism is right, my "fails loudly" claim was wrong, and the right fix is broader than this PR

Codex flagged that a `<storage>/taskRuns/<executionId>/…` path no longer reaches `resolveChildRunOutput`'s gates. **I verified it, and codex is correct on the mechanism.** `resolveInvocationFileList` (`src/tools/delegation/workflowScriptAgentRunner.ts:44-58`) partitions inputs by exactly this predicate:

```ts
runStorage: runStorageLocationFromAnyAbsolutePath(file) !== undefined
```

`runStorage: false` files go to `assertWorkflowFilesExist` and are returned as-is; only `runStorage: true` files go through `resolveChildRunOutput`'s direct-child, declared-output and completed-run checks. So on this path a `taskRuns` file that exists on disk moves from "rejected" to "accepted as an ordinary input" — **not** the loud failure my PR body describes. That sentence in the description is wrong for this code path, and I'm correcting it rather than defending it. The loud-failure claim holds only where `resolveChildRunOutput` is called unconditionally.

**But this is not a hole this PR opens.** `runStorageLocationFromAnyAbsolutePath` recognizes exactly one root — `resolveRunStoragePath()` → `executions/`. Every other absolute path inside the workspace storage bucket already takes the `runStorage: false` branch today: `streamLogs/`, `streamLogSummaries/`, `streamData/`, `memories/`, `original/`, `executionLeases/`, `pasted/`, `recordings/`. Deleting the `taskRuns` arm does not create a new class of bypass; it moves `taskRuns` into the same bucket every other non-`executions` storage directory has always been in.

So the real finding is a general one: **`resolveInvocationFileList` treats "not parseable as run storage" as "is a workspace file", and absolute paths pass `RelativeFS.resolvePath` through unchanged.** The proportionate fix is a check in that partition — reject any absolute path inside the storage bucket that does not resolve to a valid run-storage location — not a `taskRuns`-specific reader kept alive for the rejection. Resurrecting `WORKSPACE_STORAGE_LAYOUT.legacyRuns` to reject one of eight directories would leave the other seven open while re-adding the exact undated compat branch ruling 1 was about.

Reachability of the specific `taskRuns` case remains what the PR argues: it needs a `~/.texra/workspace-storage/<id>/taskRuns/<executionId>/…` file to exist, and no shipped build ever wrote that combination (writer retired 2026-02-12; the extension moved to the shared root 2026-07-16; #9623's migration consumed any such directory into `executions/` and #10887 deleted that migration). It requires a hand-created directory.

**Proposed follow-up, not folded in here** (per "don't widen the PR on your own"): add the storage-bucket guard to `resolveInvocationFileList` so *any* absolute path under the storage root that fails to resolve as run storage is rejected rather than treated as a workspace input. That closes all eight directories at once and is independent of this deletion. Happy to open it as its own issue + PR — say the word, or tell me to fold it in here instead.

The PR body's "Accepted behavioral outcome" paragraph is amended accordingly: on the `resolveChildRunOutput` callers the failure is loud; on the workflow-script invocation path the file is treated as a workspace input, which is the pre-existing behavior for every other storage subdirectory.

---

PR #11796, opened 2026-09-03
fix(cli): converge /exit on the teardown's quit policy

LionSR:
## Summary

Implements **#11771 ruling 3**, as decided: converge, in the narrow shape, and cut the four redundant `stopRequested` writes in the same PR.

`/exit` unconditionally set `stopRequested`, called `interruptActive()`, then requested exit. Ctrl-C's preserve arm exits **without** interrupting a resumable-idle root, and the graceful teardown it reaches already owns that decision — so on an idle WAITING conversation `/exit` marked the run CANCELLED and applied the detach/approval-clearing side effects where Ctrl-C left it WAITING.

`/exit` was not *skipping* the policy, it was *pre-empting* it. `beginTeardown`'s graceful arm computes `resumableIdle` and interrupts only when `chatTuiRunPending(session) && !resumableIdle`, and `/exit` already ran that exact code immediately after its own interrupt. Deleting the call lets the interrupt happen at the layer that already decides whether it should, so the three quit gestures agree by construction rather than by three copies of a predicate staying in sync.

## Issue acceptance gates

Part of #11771 (ruling 3); rulings 1 and 2 ship separately.

**`stopRequested` stays**, as the issue warned — and the deadlock behind that warning is concrete. `/exit` → `requestInputExit()` → `ink.unmount()` → `gracefulTeardown()` → `beginTeardown` awaits `ctx.followUpQueue.onIdle()` at `sessionExitController.ts:329`. The queued task at `chatSubmitDriver.ts:285-292` spins `while (!followUpTarget && !session.stopRequested && !session.runCompleted)`. The teardown's own `stopRequested = true` sits at `:336` — *after* that await. With a follow-up queued and no stream id yet, dropping the `/exit` write hangs the quit forever. The signal paths never reach `:329`, which is why only `/exit` exposes it. It is now the **sole writer** on that path, and the handler says so.

**No `requestExit()`**, per the ruling. Its three would-be callers need three different exit causes and two different process outcomes — Ctrl-C is a four-way machine exiting three ways (`requestInputExit()`, `teardown({signal, Interrupted})`, `teardown({signal, runExitCode})`); SIGTERM/SIGHUP must `process.exit` 143/129; `/exit` must return its code normally. It could only be `teardown(cause)` with a predicate bolted on — and `teardown(cause)` already exists, is memoized, and is already the single funnel. It would add a name, not a decision, and would leave `SlashCommandContext.interruptActive` alive for one caller.

## Single source of truth

`beginTeardown`'s graceful arm is now the only place that decides whether quitting interrupts. `/exit` was the one gesture reaching past it.

Removes a duplicated policy decision (the pre-empting interrupt), a context field with one consumer, and four dead assignments. No abstraction added — explicitly rejecting the `requestExit()` wrapper.

## Net elements (R6)

Diffstat: **6 files changed, 25 insertions(+), 20 deletions(-)**. Interface members removed: 1 (`SlashCommandContext.interruptActive`). Exported symbols: 0 added, 0 removed. Net code LoC: **≈ −12** (the insertions are largely the explanatory comments the deleted call's absence now needs).

## Consumer counts (R8)

`SlashCommandContext.interruptActive` (`slashContext.ts:16`): **exactly one consumer**, `registerBuiltins.tsx:733`, now deleted; its producer at `runChatTui.tsx:269` goes with it. The **local** `interruptActive` const at `runChatTui.tsx:407-409` stays — it is still passed to the exit controller at `:561`.

### The five `stopRequested` writes, separated

`stop()` calls `requestStop()` first and synchronously (`chatSessionController.ts:888-891`), and `requestStop` sets the flag unconditionally (`:295-301`) with no `await` in between — `blockRecoveryUntilInterruptedRunSettles()` and the `activeAutoResumeCancellation` cancel are both synchronous. So every write immediately preceding an `interruptActive()` is a strict duplicate with no observable gap:

| Site | Verdict |
|---|---|
| `sessionExitController.ts:213` (`clean-exit`) | redundant — **deleted** |
| `sessionExitController.ts:234` (`interrupt-and-arm-exit`) | redundant — **deleted** |
| `sessionExitController.ts:246` (`handleTermSignal`) | redundant — **deleted** |
| `sessionExitController.ts:336` (graceful `beginTeardown`) | redundant — **deleted** |
| `registerBuiltins.tsx:732` (`/exit`) | **load-bearing — kept** |

The fifth was redundant *only* because of the `interruptActive()` call this PR deletes; it flips to load-bearing under the ruling.

### Two stale comments, not one

The issue said "one stale comment to reword"; there were two, and both were provably false since #11304/#11315 — `waitingTermination.ts:175-179` passes `flowRecord: retainFlowRecordUnlessCompleted(RUN_OUTCOME.CANCELLED)`, `runToolUseFlow`'s `finally` preserves on both `signal.aborted` and `outcome !== COMPLETED`, and `deriveResumability` states the terminal outcome "never blocks":

- `sessionRunState.ts:205-208` — "interrupting clears its per-execution flow record in `runToolUseFlow`'s finally, destroying the only resumable state."
- `sessionExitController.ts:330-332` — "its flow record survives only if we DON'T interrupt the flow (interrupt clears it)."

Both now say what is actually preserved: the run's persisted status and its side effects, not the checkpoint.

## Host impact

Extension: unaffected. Desktop: unaffected.

CLI: on an **actively-running** turn nothing observable changes — the teardown's `runPending && !resumableIdle` arm fires, interrupts, and awaits `session.runPromise` exactly as today. Only the **idle-WAITING** case changes, and it changes toward Ctrl-C: the run stays WAITING instead of being persisted CANCELLED, active subagents are not swept by `detachSubagentsOnStop()`, and the terminal exit code stops reporting a cancellation the user never asked for.

## Scheduled refactoring

None.

## Validation

- `npm run typecheck:cli` and `npm run typecheck:test-kernel` — clean.
- `npx eslint` on every changed file — clean. `npm run format` / `prettier --check` — clean.
- `npx vitest run src/test-kernel/cli` — **143 files, 2441 passed, 1 skipped, 0 failed** (the whole CLI kernel, not just the exit path).

One test changed, mechanically: `SlashCommandDispatch.vitest.ts`'s `/quit` case drops the `interruptActive` stub and its assertion, keeps `expect(session.stopRequested).toBe(true)` — now the more meaningful assertion, since that write is the sole one on the path — and gains a comment saying why the interrupt is deliberately absent. The fixture at `:168` loses the same field.

### One risk to hold

With an active run, `/exit` moves the interrupt from before `ink.unmount()` to after `ctx.disposables.dispose()` and `followUpQueue.onIdle()`. Both are bounded — `stopRequested` is already true, so the poll loop at `chatSubmitDriver.ts:285-292` bails immediately at `:293`, and `submitFollowUp` resolves at admission rather than at delivery. I could not exercise this end-to-end in a headless container: **worth one manual pass — `/exit` during a live turn with a queued follow-up should still exit promptly.**

---

### DIFF
--- a/packages/cli/src/chat/tui/commands/handlers/slashContext.ts
+++ b/packages/cli/src/chat/tui/commands/handlers/slashContext.ts
@@ -13,7 +13,6 @@ export interface SlashCommandContext {
   readonly processCwd?: CliContext['cwd'];
   readonly initialAgent: string;
   readonly initialModel: string;
-  readonly interruptActive: () => void;
   readonly requestInputExit: () => void;
   readonly getApprovalPolicy: () => TexraApprovalPolicy;
   readonly setApprovalPolicy: (policy: TexraApprovalPolicy) => void;
--- a/packages/cli/src/chat/tui/commands/registerBuiltins.tsx
+++ b/packages/cli/src/chat/tui/commands/registerBuiltins.tsx
@@ -729,8 +729,15 @@ export function registerBuiltinSlashCommands(options?: {
     category: 'session',
     echo: 'never',
     handler: (_remainder, context) => {
+      // Deliberately does NOT interrupt: the graceful teardown owns that
+      // policy and skips the interrupt for a resumable-idle root, so `/exit`
+      // agrees with Ctrl-C by construction instead of pre-empting it.
+      //
+      // `stopRequested` stays and is the sole writer on this path. The
+      // teardown awaits `followUpQueue.onIdle()` BEFORE setting the flag
+      // itself, and the queued task polls this flag — dropping it would hang
+      // `/exit` forever with a follow-up queued and no stream id yet.
       context.session.stopRequested = true;
-      context.interruptActive();
       context.requestInputExit();
     },
   });
--- a/packages/cli/src/chat/tui/runChatTui.tsx
+++ b/packages/cli/src/chat/tui/runChatTui.tsx
@@ -266,7 +266,6 @@ export async function runChat(
     processCwd: process.cwd(),
     initialAgent: agent,
     initialModel: model,
-    interruptActive,
     requestInputExit: exitController.requestInputExit,
     getApprovalPolicy,
     setApprovalPolicy,
--- a/packages/cli/src/chat/tui/runChatTui.tsx
+++ b/packages/cli/src/chat/tui/runChatTui.tsx
@@ -258,7 +258,7 @@ export async function runChat(
     patchSessionMeta({ approvalPolicy: policy });
   };
   // The slash-command context is identical at every call site; build it once
-  // lazily so the closures it captures (interruptActive, resetSessionForClear,
+  // lazily so the closures it captures (resetSessionForClear,
   // chatController.resume) are all defined before the first use.
   const slashCommandContext = (): SlashCommandContext => ({
     cliContext: context,
--- a/packages/cli/src/chat/tui/sessionExitController.ts
+++ b/packages/cli/src/chat/tui/sessionExitController.ts
@@ -326,10 +326,24 @@ export function createSessionExitController(
     // A suspended (idle/WAITING) root session is resumable, so it is left
     // uninterrupted: the checkpoint survives either way since #11304/#11315,
     // but interrupting would persist a CANCELLED outcome, clear approvals and
-    // sweep active children. This is the one owner of that policy — `/exit`
-    // and both signal paths all land here rather than deciding it themselves.
-    // See chatTuiIsResumableIdleOnExit for the live-flow check that
-    // distinguishes this state from a resume slot that is still rehydrating.
+    // sweep active children. See chatTuiIsResumableIdleOnExit for the live-flow
+    // check that distinguishes this state from a resume slot that is still
+    // rehydrating.
+    //
+    // Scope: this owns the policy for the GRACEFUL path only — `/exit` and
+    // Ctrl-C's `clean-exit`, both via `requestInputExit`. Signal quits
+    // (`handleTermSignal`, and `handleSigint`'s force/preserve arms) return
+    // above at the `cause.kind === 'signal'` branch and decide for themselves
+    // through `chatTuiSigintAction`/`canStopActiveRun`.
+    //
+    // One residual divergence, deliberately left alone: `clean-exit` calls
+    // `interruptActive()` itself before `requestInputExit()`, outside this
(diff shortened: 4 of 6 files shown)


LionSR:
All four review findings addressed across `8d03ea6` and `bd8e9f5`. Both reviewers were right on every point; three of the four were mine to fix.

**Codex P1 — `/exit` could hang. Real, fixed in `8d03ea6`.** Verified the mechanism: `submitFollowUp` reaches `await dispatch.resume` (`src/agent/followUp/ToolUseFollowUp.ts:245`), which resolves only when the resumed turn finishes and never reads `stopRequested`. My PR body claimed the poll loop bails on the flag — true only *before* the resume attaches, which is exactly the window codex named. `beginTeardown` now runs the interrupt decision before `followUpQueue.onIdle()` as well as after it, so an actively-running turn is interrupted first and the queue drains promptly; the post-drain re-check preserves the original behavior for a run the drain itself starts. Your ruling's exception is untouched — a resumable-idle root is still never interrupted on any gesture.

**"both signal paths all land here" — factually wrong, fixed in `bd8e9f5`.** Confirmed: `handleTermSignal` and `handleSigint`'s force/preserve arms return at the `cause.kind === 'signal'` branch (`:305-316`) before `resumableIdle` is computed. Only the graceful path — `/exit` and Ctrl-C's `clean-exit`, both via `requestInputExit` — reaches that arm. Reworded to say exactly that.

**"single source of truth" overstated — narrowed, not silently dropped.** `clean-exit` does call `interruptActive()` outside the predicate (`:212-215`), and since `chatTuiRunPending` is false on a completed root while `stopAgentStream`'s child sweep fires without a root handle, Ctrl-C on a finished turn still detaches background children where `/exit` no longer does.

I took your **option (b)** — narrow the claim and document the divergence in the comment — rather than (a). Removing `clean-exit`'s interrupt would change **Ctrl-C's** behavior, and #11771 ruling 3 authorized converging `/exit` onto the existing policy, not altering the gesture it converges toward. Silently sweeping children is a side effect someone may be relying on for Ctrl-C; that deserves its own ruling rather than riding along here. Say the word and I'll open it as a follow-up.

**Finding 2 (stale `interruptActive` in the `slashCommandContext` comment)** was already fixed in `8d03ea6`, which landed after the commit you reviewed (`1e08e26`).

Validation on the current head: `npm run typecheck:cli` clean, `npx eslint` and `prettier --check` clean, `npx vitest run src/test-kernel/cli` — 143 files, 2441 passed, 1 skipped, 0 failed.

---

PR #11798, opened 2026-09-03
ci: close three gate gaps found by the docs/audit sweep

LionSR:
## Summary

First batch from a sweep of `docs/proposals`, `docs/prds`, `docs/dev/audits`, `docs/design` and `docs/architecture` for findings still unimplemented at HEAD. These three are CI-gate gaps: each is a check that reads green while not actually checking the thing it names.

**1. The docs-root boundary gate never ran in CI.** `check-root-docs.mjs` is wired only as `docs/package.json`'s `check:content`, and no workflow invokes it — the `docs build` job runs `format:check` and `docs:build`, and it is gated off entirely on code-only PRs. So an unclassified root-level doc could freeze the texra.ai deploy with CI fully green, which is precisely the failure mode CLAUDE.md warns about ("A doc landing at the `docs/` root can silently freeze the texra.ai deploy"). It now runs in `guidance-refs` — the deliberately ungated job that already covers both sides of the docs/code split, and which needs no build step (the script is plain Node).

**2. `claude-code-review` reported green after silently skipping.** With no provider token the job sets `skip=true`, every later step is gated off it, and the job ends **successfully** — the check reads "reviewed" when nothing was reviewed. It now emits a `::warning::` annotation and a job-summary block naming the provider.

Deliberately **not** a failure: the workflow's own comment (`:54-55`) notes fork PR tokens are read-only and their review jobs already skip for exactly this reason. Failing would make every fork PR unmergeable. Loud-but-passing is the correct shape here.

**3. `HaaLeo/publish-vscode-extension` was referenced by mutable tag at both credential-receiving call sites.** `release.yml:84` and `:92` pass `secrets.VSCE_PAT` and `secrets.OVSX_PAT` to `@v2`. A moved tag hands those credentials to unreviewed code. Pinned to the commit `v2` currently resolves to, with the tag kept as a trailing comment so the next bump moves both together.

## Issue acceptance gates

No issue — these come from the sweep of existing survey/audit documents (the open-source-readiness audit and the dead-code-gate blind-spots proposal among them). Each was **re-verified against the workflows at `d418d45`** rather than taken from its source document; three other items from the same sweep were discarded on verification because merged PRs had already fixed them.

## Single source of truth

`guidance-refs` becomes the single ungated home for repo-hygiene gates that must run on both docs-only and code-only PRs. It already held the guidance-reference and browser-safe-utils checks; the root-docs boundary belongs in the same place for the same reason.

No abstraction added. One script invocation moved into a job that already exists, one annotation added, one ref pinned.

## Net elements (R6)

3 files, **+26 / −2**. No exported symbols, no source files, no classes/interfaces. Runtime code untouched.

## Consumer counts (R8)

n/a — no emit path or export changed. `check-root-docs.mjs` keeps its `docs/package.json` `check:content` binding; this adds a CI caller rather than moving it.

## Host impact

Extension: none at runtime. The publish step that ships the VSIX is now pinned.

Desktop: none.

CLI: none.

## Scheduled refactoring

`pnpm/action-setup@v6`, `denoland/setup-deno@v2.0.5`, `EndBug/label-sync@v2` and the first-party `LionSR/agent-ci-actions@v1` are also tag-referenced. I pinned **only** the two that receive publish credentials — that is where a moved tag is a credential-disclosure risk rather than a build-reproducibility one, and pinning all of them is a larger policy call (plus ongoing bump burden) that belongs to you, not to this PR.

## Validation

- All three workflow files parse as YAML (`yaml.safe_load`).
- The rewritten `if`/`else` shell fragment in `claude-code-review.yml` parses under `bash -n`.
- `node docs/scripts/check-root-docs.mjs` runs clean from the repo root with no install: *"docs boundary OK: every root-level entry is classified (6 public docs, 1 public dirs); 5 internal doc areas are excluded and timestamped."* — confirming it works in `guidance-refs` context.
- `npx prettier --check` on all three files — clean.
- The pinned SHA was resolved from the upstream tag directly (`git ls-remote … refs/tags/v2` → `4004fc217b5ac892b15a1c3cbe6458de97242dde`), not copied from a document.

No tests: there is no runtime code in this diff, and CI configuration is verified by CI running it.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,10 +256,12 @@ jobs:
         run: node scripts/check-guidance-refs.mjs
 
       # The docs-root boundary decides what the texra.ai site publishes. It
-      # only ran as `docs/package.json`'s `check:content`, which no CI job
-      # invokes — `docs build` runs `format:check` + `docs:build` and is gated
-      # off on code-only PRs. An unclassified root doc could therefore freeze
-      # the deploy with CI green. This job is the ungated one, so it runs here.
+      # runs inside `docs:build` (via `check:content`), but that job is gated
+      # off on code-only PRs and on pushes to main — exactly the paths where a
+      # root-level doc gets added alongside code. An unclassified root doc
+      # could therefore freeze the deploy with CI green. This job is the
+      # ungated one, so it also runs here; the overlap on docs-only PRs is
+      # harmless.
       - name: Check docs root boundary is classified
         run: node docs/scripts/check-root-docs.mjs
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,6 +255,14 @@ jobs:
       - name: Check guidance file references resolve
         run: node scripts/check-guidance-refs.mjs
 
+      # The docs-root boundary decides what the texra.ai site publishes. It
+      # only ran as `docs/package.json`'s `check:content`, which no CI job
+      # invokes — `docs build` runs `format:check` + `docs:build` and is gated
+      # off on code-only PRs. An unclassified root doc could therefore freeze
+      # the deploy with CI green. This job is the ungated one, so it runs here.
+      - name: Check docs root boundary is classified
+        run: node docs/scripts/check-root-docs.mjs
+
       - name: Check browser-safe utils counts resolve
         run: node scripts/check-browser-safe-utils.mjs
 
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -104,6 +104,16 @@ jobs:
           else
             echo "skip=true" >> "$GITHUB_OUTPUT"
             echo "reason=$REVIEW_PROVIDER token is not configured" >> "$GITHUB_OUTPUT"
+            # Say so loudly. Every later step is gated on this flag, so the job
+            # otherwise ends green and the check reads "reviewed" when nothing
+            # was reviewed. Not a failure: fork PRs legitimately have no
+            # provider credentials and must still be mergeable.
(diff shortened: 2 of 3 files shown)


LionSR:
Both findings were correct and both were mine. Fixed in `c2f7671`.

**The pinned SHA was the annotated tag object, not a commit — and it would have broken the release.** Verified directly:

```
refs/tags/v2     -> 4004fc217b5ac892b15a1c3cbe6458de97242dde   (tag object)
refs/tags/v2^{}  -> ca5561daa085dee804bf9f37fe0165785a9b14db   (commit)
```

Actions requires a commit's full SHA in `uses`, so the pin would not have resolved — and it would have failed *only* at release time, taking the Marketplace and Open VSX publish down with it. Worse, my PR body advertised the wrong lookup (`git ls-remote … refs/tags/v2`) as the method, which is precisely the footgun. Now pinned to `ca5561da…`, with the comment stating the `^{}` form and why the plain ref is wrong, so the next bump doesn't repeat it. A hardening change that breaks the publish is worse than the exposure it closes; thanks for catching it before it shipped.

**"no CI job invokes `check:content`" was false.** `docs/package.json` defines `docs:build` as `npm run check:content && npm run sync-content && vitepress build .`, and the `docs build` job runs `docs:build` — so the boundary check already covers docs-only PRs. Confirmed by reading the script.

The gap is narrower than I described but still real: `docs-check` is gated off on **code-only PRs and pushes to main**, which is exactly where a root-level doc gets added alongside code and where nothing would catch it. The comment now says that, and notes the double-run on docs-only PRs as harmless redundancy rather than pretending it isn't there.

Validation on `c2f7671`: both workflow files parse as YAML, `prettier --check` clean, and both `uses:` lines carry the commit SHA.

---

PR #11800, opened 2026-09-03
legal: attribute Font Awesome Free, and generate the NOTICE copies from one source

LionSR:
## Summary

Two things, the second added in response to review.

**1. License-compliance gap in a shipped product.** `@fortawesome/free-solid-svg-icons` is a **production dependency** (`^7.3.1`), and the installed package declares its license as `(CC-BY-4.0 AND MIT)`. `src/shared/wa/webAwesomeIcons.ts` imports ~118 named icons straight from it, and they are bundled into the shipped extension, desktop and CLI artifacts — so TeXRA **redistributes** the icons. CC BY 4.0 requires attribution on redistribution, and upstream's `LICENSE.txt` is explicit that it *"applies to all icons packaged as SVG and JS file types"*, exactly the form imported here. All five NOTICE files listed only PocketFlow-TypeScript.

**2. There was no single source of truth for the NOTICE files** — the review question on this PR, and a fair hit. They were byte-identical *by convention*: five hand-maintained copies with nothing enforcing agreement. This PR was itself the demonstration, since I edited all five by hand and checked `md5sum`, which works right up until someone doesn't.

`scripts/sync-notice-files.mjs` now generates the four package `NOTICE.txt` files from the root `NOTICE`, in the shape of `sync-tsconfig-paths.mjs` and `sync-package-contributes.mjs`: `sync:notice` writes, `check:notice` gates in CI. The root `NOTICE` is now the only file to edit.

The duplication itself stays because it must: the VSIX, the Electron installers and the two npm packages are distributed separately, and the obligation attaches to each artifact. What is gone is maintaining the copies by hand.

## Issue acceptance gates

No issue — from the docs sweep (an open-source-readiness audit row). Verified against the installed package rather than the document: the `dependencies` entry, `node_modules/@fortawesome/free-solid-svg-icons/package.json`'s `"license": "(CC-BY-4.0 AND MIT)"`, the upstream `LICENSE.txt` wording, and the import site.

## Single source of truth

The root `NOTICE`. `check:notice` fails CI when a package copy drifts, naming the file and the fix command.

Five hand-maintained copies become one source plus a generator, following an existing repo pattern rather than inventing a mechanism. No new abstraction: the script mirrors two that already exist.

## Net elements (R6)

10 files, **+265 / −2**. New: `scripts/sync-notice-files.mjs`, two npm scripts (`sync:notice`, `check:notice`), one CI step. Exported symbols: 0. No runtime code.

## Consumer counts (R8)

n/a — no emit path or export changed. `packages/cli` and `packages/agent` gained `NOTICE.txt` in their `files` whitelists (see below).

## Host impact

Extension: `NOTICE.txt` ships in the VSIX (already a required packaged path in `verify-extension-package-invariants.mjs`).

Desktop: ships in the installers.

CLI / SDK: **previously did not ship at all** — `packages/cli`'s `files` whitelisted only `["dist", "!dist/bin/tui-harness.js", "LICENSE.txt"]`, so `npm pack` omitted the NOTICE while `dist/resources` carried the built trace-viewer bundle containing the icons. `packages/agent` had the same gap. Both now list it; verified with `npm pack --dry-run --ignore-scripts --json ./packages/cli` → `NOTICE.txt included: true`.

## Scheduled refactoring

Deliberately **not** done here: a full third-party license audit walking every production dependency and reconciling it against the NOTICE files. This PR fixes the one dependency I could verify end-to-end. Also untouched: `docs/guide/acknowledgments.md` and other user-facing credits surfaces — the NOTICE files are what the redistribution obligation attaches to.

## Validation

- `npm run format:check` — clean. `npx eslint scripts/sync-notice-files.mjs` — clean.
- **The gate was proven to fail on drift, not assumed to**: appending a line to `packages/cli/NOTICE.txt` makes `check:notice` exit 1 naming the file and the fix command; `sync:notice` restores it and the check exits 0. (My first attempt at that check read a pipeline's `tail` exit and reported a misleading 0 — caught and redone against the real exit code.)
- All five NOTICE files confirmed byte-identical after the change.
- `npm pack --dry-run` confirms the CLI now ships its NOTICE.
- Full CI green on `076d2e7`, `validate` included.

Three review findings were fixed on this branch: the unshipped CLI/SDK notices, a `Code: MIT License` line that over-attributed (upstream puts the imported `fa*` JS modules under CC BY 4.0, not MIT — the section now says which license covers what and reproduces the MIT permission notice as the PocketFlow section does), and a garbled closing sentence.

**Not legal advice.** I verified what the package declares and what its terms say; whether this wording satisfies your obligations is a judgment for you or counsel.


LionSR:
All three findings correct; fixed in `7047bcf`.

**The CLI's attribution was never published — the P1 is right, and it defeated the PR's purpose.** `packages/cli/package.json`'s `files` whitelisted only `["dist", "!dist/bin/tui-harness.js", "LICENSE.txt"]`, so `npm pack` omitted `NOTICE.txt` while the packed `dist/resources` carries the built trace-viewer bundle containing the icon artwork. Consumers got the icons *without* the notice. `packages/agent` had the identical gap (`["dist", "LICENSE.txt"]`). Both now list `NOTICE.txt`, verified with `npm pack --dry-run --ignore-scripts --json ./packages/cli` → `NOTICE.txt included: true`.

**The `Code: MIT License` line over-attributed.** You're right that Font Awesome's LICENSE.txt splits the download three ways and the imported `fa*` modules are JS files carrying icon SVG path data — CC BY 4.0, not MIT. Rather than drop the line, I took the other option you offered: the section now says explicitly which license covers which material (CC BY 4.0 for the icons "packaged as SVG and JS file types, which is the form imported here"; MIT for the package's non-icon, non-font files) **and reproduces the MIT permission notice in full**, matching how the PocketFlow section above handles its MIT text. Naming a license without its notice was the inconsistency; this removes it either way, whether or not non-icon code ends up bundled.

**The closing sentence was garbled** — it read as though the whole download were CC BY 4.0. Reworded so the license applies to the icons, as upstream states.

**On the host-impact overstatement:** correct that neither `packages/cli/src` nor `packages/agent/src` imports `@fortawesome/*` or `@shared/wa/*` — I re-grepped and confirmed zero hits. Direct source imports are extension and desktop only. The CLI still redistributes the artwork via the built trace-viewer assets in `dist/resources`, which is what makes the `files` fix load-bearing rather than conservative over-attribution; the SDK package is the genuinely conservative one. The PR body's Host impact section overstated the mechanism for both, and I'm not going to restate it as though I'd had it right.

All five NOTICE files remain byte-identical. `npm run format:check` clean.

---

LionSR:
single source of truth?

LionSR:
Fair hit — there wasn't one. Fixed in `076d2e7`.

My SSOT section said the five NOTICE files are byte-identical "by convention." That's not a source of truth, it's a coincidence with a good track record: nothing enforced agreement, so the next attribution would have been added to some of them and not others. This PR was itself an example — I edited all five by hand and verified with `md5sum`, which works exactly until someone doesn't.

**The duplication is genuinely required** — the VSIX, the Electron installers and the two npm packages are distributed separately, and the attribution obligation attaches to each artifact, so each needs its own copy. What isn't required is maintaining those copies by hand.

`scripts/sync-notice-files.mjs` now generates the four package `NOTICE.txt` files from the root `NOTICE`, following the shape of `sync-tsconfig-paths.mjs` and `sync-package-contributes.mjs`:

- `npm run sync:notice` writes the copies
- `npm run check:notice` is the diff gate, wired into `static checks` beside `check:tsconfig-paths` and `check:package-contributes`

**Verified the gate actually fails on drift** rather than assuming it — appending a line to `packages/cli/NOTICE.txt` makes `check:notice` exit 1 naming the file and the fix command; `sync:notice` restores it and the check exits 0. (I checked the exit code directly; my first attempt read the exit of a pipeline's `tail` and reported a misleading 0.)

The root `NOTICE` is now the only file to edit when attribution changes.

One thing I did **not** do: generate the copies at pack/build time instead of committing them. That would remove the files from the tree entirely, but `packages/extension/NOTICE.txt` is a required packaged path in `scripts/verify-extension-package-invariants.mjs`, and the CLI's is now in its `files` whitelist — both expect a real file. The sync-and-gate shape matches what the repo already does for generated content, so I kept to it rather than inventing a second mechanism.

---

Issue #11822, opened 2026-09-03
Startup: remove restart repair and the orphan sweeps from the ready path (S1 to S3)


LionSR:
## Problem, measured (2026-09-03)

`texra chat --agent X --model Y` takes 48 to 112 s to show the input box in a workspace with ~4,100 execution directories (3,125 still holding a `flow_<id>.json` checkpoint, 1.8 GB). The pickers are not the cost:

| Step after the pickers | Time |
| --- | --- |
| `StreamLogStore.open` (summary cache) | 1.5 to 2 s |
| `SessionHandle` restart repair (`waitUntilReady`) | 43 to 108 s |
| `sweepLeftoverStreams` | 2 to 15 s |
| everything else | < 0.5 s |

Inside repair: scan 1.8 s; phase 1 classifies ~1,900 checkpointed streams (~4 s, each a full checkpoint parse plus a lease inspection); phase 2 settles the same ~1,900 streams sequentially under `runWithInactiveExecutionLease`, re-parsing each checkpoint, 37 s and up, writing nothing for runs that already carry a persisted outcome. Per launch the executions directory is scanned 3 to 4 times and every checkpoint is parsed up to 3 times (launcher history row, repair phase 1, repair phase 2). The extension and desktop run the same pass.

Root cause: completed tool-use runs keep their checkpoint so Resume can continue them, so "has a checkpoint" is the whole history and only grows. Repair is O(history) instead of O(unsettled runs). None of Claude Code, Codex, OpenCode, or Gemini CLI has a startup pass over prior sessions; all derive interrupted state at read time.

## Scope of this issue (lifecycle only; engine-independent)

Per `docs/proposals/2026-09-03-startup-repair-is-the-wrong-shape.md` (landing with the docs PRs) and the single-owner-sessions rulings D3 and D7 ("classify once, mutate nothing"; "status is derived, never remembered"):

- **S1.** Delete restart repair and the orphan sweep from `waitUntilReady`. A stream's phase becomes a pure function over the execution tuple (id, stream edge, outcome, resumable): live flow context here gives RUNNING or WAITING; otherwise persisted outcome; otherwise interrupted, checkpoint present. No write. The history list and resume picker advertise "checkpoint present" from one `stat`, never `classifyRun`; the resume path stays the only checkpoint parser and refuses loudly on a malformed record.
- **S2.** Terminal facts are written where the run ends (host exit paths; on resume for a hard crash), never at boot.
[...]

PR #11823, opened 2026-09-04
docs: record why startup restart repair is the wrong shape

LionSR:
Records the 2026-09-03 startup measurement (48 to 112 s to the first prompt in a workspace with ~4,100 executions, of which restart repair is 43 to 108 s and the leftover sweep 2 to 15 s), the peer comparison (Claude Code, Codex, OpenCode, Gemini CLI: none runs a boot pass over prior sessions), and the lifecycle target S1 to S3 that #11822 tracks. Storage-engine steps S4 to S7 defer to the companion `2026-09-03-persistence-substrate-decision.md`, which lands separately.

Docs only; `docs/proposals/**` is excluded from the published site and the root-docs gate passes.

Refs #11822

### DIFF
--- a/docs/proposals/2026-09-03-startup-repair-is-the-wrong-shape.md
+++ b/docs/proposals/2026-09-03-startup-repair-is-the-wrong-shape.md
@@ -37,21 +37,26 @@ Inside repair (`src/agent/runtime/SessionHandle.ts` `runRestartRepair`,
 
 Raw read plus `JSON.parse` of all 3,125 checkpoints is 7 s; the Zod envelope
 is shallow (`shared: z.unknown()`). The cost is per-run overhead multiplied
-by duplication, not parsing per se. Per launch the executions directory is
-scanned three or four times (repair scan, `readExecutionStreamIndex` in the
-orphan sweep, `listExecutionStreamReferences` again in
-`sweepOrphanedExecutions`, `listExecutions` in the launcher) and every
-checkpoint is parsed up to three times (launcher history row `classifyRun`,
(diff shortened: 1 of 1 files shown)


LionSR:
Corrected in the follow-up commit: completed runs delete their checkpoint (D8 as shipped in #11304) and it is the cancelled, failed, and crashed runs that keep one, which is what keeps repair's candidate set growing; the orphan sweep is named as a host-level startup step rather than part of `waitUntilReady`; the scan and parse counts are attributed to the launch mode that incurs them. On the Codex threads: the stream-less legacy cohort is excluded from Resume by the listing's stream-id gate (now stated in §6, and implemented in #11835); the cleanup survives S1 by moving off the interactive path (#11825) rather than being deleted; S2 now names the transcript close beside the outcome write (implemented at exit in #11830 and at row open after #11837); the importer's retirement condition belongs to the companion substrate proposal, whose Stage 7 sets it (90 days and two releases after the cutover).

LionSR:
Round four at d15a7a80d9, all five taken: the launcher accounting now counts both listing parses (`classifyRun` and `readCliSessionResumeData`); S1's listing rule keeps the workflow compile-rejection exclusion and states exactly when it may read (FAILED or missing outcome only), matching #11835; the startup-time clauses of D3/D7 are superseded explicitly rather than described as already ruling this shape; S6 retention is bounded by eligibility (ephemeral, orphaned, stream-less legacy), never a TTL over a resumable checkpoint; and the Stage 5 cutover is stated as file-level per the storage PRD, with generated artifacts staying at the execution path.

PR #11824, opened 2026-09-04
feat(session): derive a non-live stream's phase at read time

LionSR:
Part of #11822 (S1). Behaviour-preserving while restart repair still runs: the new rule's live arms win wherever repair wrote a phase, so this lands green and can be verified against the pass it replaces before that pass is deleted.

## What

- `SessionState.resolveStreamPhase(stream)` is the one read-time rule for a stream's phase, pure and synchronous: hold, then live machine state; for an unhydrated stream the always-resident transcript summary (`StreamLogStore.hasUnfinishedOutput`, refactored out of `getUnfinishedStreamIds`) says "interrupted" or "pending"; for a hydrated stream the run tuple its own sidecar hydration captured decides: authority-read failure and a foreign or self-held lease render the read-only `unavailable` sentinel with the reason (D3, D9), a persisted outcome is the phase, and a surviving checkpoint or unclosed transcript with no outcome is CANCELLED. CANCELLED is the value repair writes today, so no wire schema, button table, or label table changes.
- `StreamSnapshotStore` records `outcome` (from the `meta.json` hydration already parsed and discarded), `checkpointPresent` (one `exists` stat) and `lease` (one read-only `inspectExecutionLease`) per hydrated stream, in memory only, cleared on execution handoff. The checkpoint and lease probe start beside the metadata read and are awaited after the overlays replay, so the existing accumulator-reset window does not widen. The hydration reads the row strictly, so a corrupt `meta.json` renders as unreadable rather than as "no outcome".
- Consumers moved onto the rule: `LitSessionRenderer.metadataFor` (both `getAllStreamStates()` calls and `SessionFactApplier.buildStreamStatesForRefresh` deleted; the status being applied travels as one `phaseOverride`), the CLI's `sessionStreamPhase` and `streamUnavailableDetailFor` (a foreign owner now locks the composer before a refused send), the Ctrl-C whole-map status-bar scan (absent phase is not live), and CLI transcript eviction (an absent phase evicts instead of retaining forever). `sessionRunState`'s undefined arm is deliberately untouched: it covers the launch-to-reservation gap.
- `ProgressBackend.load()` starts an un-awaited, chunked (8-wide), yielding, dispose-cancelled sidecar hydration pass through the storage queue, re-checking membership per chunk, so restored tabs converge off the ready path. Interim until the persistence substrate's `executions` rows carry outcome and resumability; the comment says so.

## Why

Restart repair re-proves every checkpointed run at every launch (43 to 108 s in a 4,100-execution workspace) only to assign in-memory phases that a read-time function over already-resident facts can answer. This PR installs that function so the next PR can delete the pass without changing what any host renders.

## Verification

Root typecheck (all projects), `check:dead-code-ratchet`, eslint and prettier on the changed files, and the full vitest suite (823 files, 10,008 tests) green on the final commit. One new test file, `DerivedStreamPhase.vitest.ts`, pins the four arms that matter (persisted outcome plus checkpoint, unhydrated open transcript, held foreign lease, unreadable authority); two existing CLI tests that pinned the inverted defaults were rewritten, one with a sibling case asserting a pending run without a machine entry still reads "stop". `config/ratchets/store-public-surface-baseline.json` records the widened `getRunMetadata` and the new `hasUnfinishedOutput` read in the same commit.

Adversarially reviewed before opening; four confirmed findings fixed in the second commit (hydration pass escaping the storage queue, hold read after phase, unhydrated arm parked by re-seed, tolerant meta read on a corrupt row).


LionSR:
Addressed in 1a22485a44 and f05cfedaf8:

- **Bounded residency (Codex P1).** The run tuple no longer lives on the heavy record: `StreamSnapshotStore` keeps a small always-resident `runFacts` map (outcome, checkpoint present, lease, authority failure), written once per hydration, cleared on execution handoff and on real disappearance, kept across eviction. After the background pass publishes a chunk it releases the warmed records through the existing sidecar-retirement rule, which now keys off the derived phase and releases every unowned child (held elsewhere, unreadable, lease without a run, nothing durable left), not only finished ones. Roots stay resident because `lookupStreamExecutionId` reads their record synchronously; that residual is stated in the code.
- **Ratchet baseline (Codex P1).** `getRunMetadata` is back to its five fields; the facts come through one accessor, `getRunPhaseFacts`, registered in `aggregateContracts` with its four fields so the count is exact (the ratchet's field extractor now unwraps a `T | undefined` return). `StreamLogStore.getUnfinishedStreamIds` is deleted (callers filter `keys()` through `hasUnfinishedOutput`), so that store is net zero. `StreamSnapshotStore` is +4 caller-visible units in this PR, recorded as interim in the baseline's semantics note; the offset I named earlier does not hold (`getExecutionIdMap` has a second consumer in `deleteAllOnce`), so I am not claiming it. The accessor goes with the substrate cutover's `executions` rows.
- **Failed chunk (Codex P2).** Per-stream preloads settle under `Promise.allSettled` before the chunk publishes or yields.
- **Timer (Codex P1).** `setTimeout` from `node:timers/promises`.
- **Foreground activation (Codex P2).** `hydrateAndCommitSelection` pushes the stream's metadata right after its preload succeeds.
- **Stop hint (Codex P2).** The status bar derives "stop" from the same run-pending predicate the SIGINT path uses, so a pending run with a stream id but no phase reads "stop"; the existing test row now is that case.
- **Stale comments (texra-ai).** Both rewritten.
- **Config handoff.** `setRunConfig`'s guard reads the summary mirror's execution id like `setRunStart` does, so a record minted after a release does not carry the previous run's facts into a new execution.

Deliberately unchanged, with reasons:

- Corrupt checkpoint reads as CANCELLED with Resume (texra-ai): per the lifecycle proposal, listings advertise checkpoint presence from one stat; the resume path is the only parser and refuses loudly on a malformed record.
- "Interrupted arm narrower than repair" (texra-ai): repair's candidate set is the unfinished-transcript seed plus checkpointed executions, so a hydrated stream with no checkpoint and a closed transcript was never a candidate and renders `ready` today too.
- Probe on every seed (claude): intended; one stat and one lease read for the one stream being seeded.
- `readMetaStrict` discarding `config` on a corrupt `meta.json` (claude): intended fail-closed; a row whose authority is unreadable is shown unavailable rather than partially.

Full validation on f05cfedaf8: typecheck, dead-code ratchet, lint, full suite (823 files, 10,008 tests) green.

LionSR:
On the three new Codex threads:

- **Validate checkpoints before marking streams cancelled**: deliberately not done. The lifecycle proposal fixes the contract that listings and the rail advertise checkpoint presence from one stat, and the resume path is the only parser, refusing loudly on a malformed or spent record. Parsing every checkpoint to decide a label is the O(history) cost this series removes; a run whose checkpoint delete failed after completion is a storage defect the resume refusal surfaces rather than hides.
- **Preserve valid outcomes when optional metadata is malformed**: agreed, fixing. The phase facts will parse only the core metadata schema strictly, so a malformed optional projection keeps the outcome and only a malformed core row is unreadable.
- **Order phase reads behind the lease snapshot**: agreed, fixing. The probe will read the lease, then the checkpoint, then the outcome, so a free lease is always followed by the outcome a completed finalize has already written.

Fix commit to follow on this branch.

LionSR:
Fixed in a1697a899d:

- The phase facts now parse only the core metadata schema strictly (`readMetaCore`, the same shape `deriveResumability` uses): a malformed core row is unreadable, a malformed optional `workflow` projection keeps the outcome. The identity and description hydration shares that read. One new case pins a valid outcome beside a malformed optional projection resolving to Completed.
- The probe is ordered lease, then checkpoint, then outcome, with the invariant and the foreign finalize order (outcome write, checkpoint delete, lease release, verified against `finalizeRun`) stated in a comment. It is still awaited after the overlay replay. The outcome now has one producer, the probe; if the probe fails the tab reads unavailable with the cause rather than a terminal phase from an unordered read.

Validated alone: typecheck, lint, dead-code ratchet, and the agent, transcript, progress-view, and desktop suites (318 files, 4,397 tests).

PR #11825, opened 2026-09-04
perf(session): take the leftover-stream sweep off every host's ready path

LionSR:
Part of #11822 (S1). The first of the startup fixes: no host awaits `sweepLeftoverStreams()` before its UI is up.

## What

- New `scheduleLeftoverStreamSweep(session)` in `src/controllers/session/` runs the sweep once per session on an unref'd ~1.5 s timer, awaits nothing on the caller's side, logs failure at `warn` with the cause, and returns a cancel. The CLI schedules it from `initializePersistentSession` so headless `texra run` keeps an owner (a short run exits before the timer fires and the next launch sweeps); the extension after `registerCommands`, cancel on `context.subscriptions`; the desktop after `reopenMainWindow()`, cancel on `processResources`. The three awaited boot calls are deleted.
- The sweep can now overlap live runs, which the boot-only version could not, so `sweepLeftoverStreams` takes `runningStreams` (tracked handles plus in-flight phases, computed by the scheduler from the session): those streams are never offered to the ephemeral half, whose sequential `deleteStream` loop would otherwise await a live stream's deletion behind its execution lane for the run's lifetime, and they are retained by the orphan half. The orphan half re-reads liveness at deletion time and again after staging through `shouldDelete`, so a stream registered after the sweep's snapshot (a new chat tab, a fresh shell) is never taken for an orphan.
- One executions walk instead of two: `sweepOrphanedStreams` lists the directory once and hands the references to `sweepOrphanedExecutions`. A listing failure now skips both halves loudly instead of treating every orphan as provably unowned.
- `deleteStream` reconciles its own stream's staged residue first (the private `reconcileStagedDeletions` widened with an optional scope, precedent `adjacentStreamCleanup.ts`), so a user deletion that lands before the deferred sweep does not hit the permanent `stage()` throw.
- Swept shells are published to presentations as the existing `removeStream` session fact, since a rail may already be showing them and no host repaints for a deletion the store made on its own.

## Why

Measured 2 to 15 s in front of every host's first prompt in a 4,100-execution workspace (details in #11822). The sweep owes the prompt nothing that cannot wait a beat. On the extension this also reverts the activation-blocking placement introduced by #11808.

## Verification

Root typecheck, `check:dead-code-ratchet`, eslint on the changed files, and the storage, progress-view, CLI, desktop, architecture, and agent-runtime suites (320 files, 4,222 tests) green on the final commit run alone. Existing sweep tests keep their assertions and call the sweep directly; the CLI reclaim test awaits the scheduled sweep's effect; one case added for the running-stream exclusion. Adversarially reviewed before opening; the two confirmed findings (stale live-set snapshot, swept shells not published) are fixed in the second commit.

### DIFF
--- a/packages/cli/src/runtime/transcriptSession.ts
+++ b/packages/cli/src/runtime/transcriptSession.ts
@@ -4,7 +4,7 @@ import {
   tryDefaultSession,
   type SessionHandle,
 } from '@agent/runtime';
-import { createSessionStores } from '@controllers/session/createSessionStores';
+import { scheduleLeftoverStreamSweep } from '@controllers/session/scheduleLeftoverStreamSweep';
 import { createTexraResponseTextProcessing } from '@latex/texraResponseTextProcessing';
 import { ephemeralTranscriptWarning, StreamLogStore } from '@transcript';
 
@@ -52,7 +52,11 @@ async function initializePersistentSession(
       responseTextProcessing,
     }),
   );
-  await createSessionStores(result.session).sweepLeftoverStreams();
+  // Off the ready path: the sweep reads the whole storage root, and no prompt
+  // waits for it. Both CLI shapes schedule it here so headless `texra run`
+  // keeps an owner for the shells it leaves behind; a short headless run
+  // exits before the unref'd timer fires and the next launch sweeps them.
+  scheduleLeftoverStreamSweep(result.session);
   return result;
 }
 
--- a/packages/cli/src/runtime/transcriptSession.ts
+++ b/packages/cli/src/runtime/transcriptSession.ts
@@ -45,6 +45,7 @@ async function persistentSession(
 
 async function initializePersistentSession(
   transcripts: StreamLogStore,
+  sweep: { readonly delayMs?: number } = {},
 ): Promise<CliTranscriptSession> {
   const result = await persistentSession(
     initializeDefaultSession({
@@ -53,10 +54,13 @@ async function initializePersistentSession(
     }),
   );
   // Off the ready path: the sweep reads the whole storage root, and no prompt
-  // waits for it. Both CLI shapes schedule it here so headless `texra run`
-  // keeps an owner for the shells it leaves behind; a short headless run
-  // exits before the unref'd timer fires and the next launch sweeps them.
-  scheduleLeftoverStreamSweep(result.session);
+  // waits for it. The TUI takes the default delay, which keeps the read off
+  // its first paint. A headless `texra run` has no paint to protect and can
+  // finish inside that delay — the unref'd timer would never fire, leaving
+  // its shells for a launch that may not come — so it schedules with none.
+  // Overlapping the run is safe: the sweep excludes the streams this process
+  // is running, and it is idempotent if the process exits first.
+  scheduleLeftoverStreamSweep(result.session, sweep);
   return result;
 }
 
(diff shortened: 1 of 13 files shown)


LionSR:
Addressed in 1cea20e9a1:

- `deleteAdjacentStreamState` now performs the same scoped `reconcileStagedDeletions(..., new Set([stream]))` as `deleteStream` before staging, so a `history delete` that lands before the deferred sweep no longer hits the permanent `stage()` throw on a stream with staged residue (both reviews).
- The CI failure in `TranscriptSessionPolicy` ("reclaims an orphaned stream sidecar after a headless session opens") was a race in the test: it awaited the empty sidecar listing and asserted the goal-store cleanup outside the wait, but the goal is cleared after the directory goes. Both facts are now awaited together.

On the residual risks: the CLI dropping the cancel is deliberate for the headless shape and documented at the call site; a session torn down before the timer fires leaves the sweep for the next launch, and a failed sweep is not retried within the session by design (one pass per launch).

LionSR:
Follow-ups in f567e636cb and 0adeb5caba:

- Orphan-stream admission now also consults the shared authoritative store (`hasAuthoritativeStream`), so a stream another host registered after this process opened its index is retained; the existing orphan-sweep test gained that case (Codex P1).
- A `removeStream` fact for a stream whose durable deletion already committed no longer restarts deletion: `deleteStream` short-circuits before the scoped reconcile and the executions-directory walk when nothing resident remains (goal entries are still forgotten, which the cleanup test pins). No new fact or payload field (Codex P2).
- Both reconcilers are now serialized by the deletion queue: `deleteStream`'s scoped reconcile moved inside the enqueued step, and `sweepOrphanedStreams`'s unscoped reconcile is enqueued too (texra-ai residual risk).
- The swept-shell publication was inert for the GUI hosts: `deleteStreamNow` returned before posting `DELETE_STREAM` when no durable data remained. The backend now remembers the last projected roster and runs the presentation removal for any stream it showed; the leftover-shells test asserts the `DELETE_STREAM` post.

The stale `pendingStreamDeletions` sentence in the body is corrected.

LionSR:
Round 3, in e2efa936ca:

- **Cross-instance serialization.** The right owner is the session's snapshot store, which every `SessionStores` instance shares: `StagedDeletionCoordinator.reconcile` now runs through a concurrency-1 queue, so the sweep's unscoped pass and a user delete's scoped pass on another instance cannot both rename the same staged directory. The two comments that credited the per-instance deletion queue with this are corrected; that queue orders work within one instance only.
- **Authoritative admission at the commit.** `deleteStreamSidecars` takes a `confirmDeletable` probe, checked after staging and before the irreversible commit; the orphan body passes `!hasAuthoritativeStream(stream)` for both admission and confirmation. The residual window (a host registering the same id between that answer and the commit) is documented; closing it needs a cross-host lock no deletion path holds.
- **Executions the stream half already deleted** are filtered out before the execution half.
- **Module-level memo deleted.** Each host schedules once at bring-up; a cancelled schedule can be scheduled again.
- **Rendered roster lives in `LitSessionRenderer`**, updated by both `UPDATE_STREAMS` and the incremental `UPDATE_STREAM_METADATA`, so a tab first shown incrementally still gets its removal.
- **Headless CLI sweeps with no delay** (`delayMs: 0` on the `'fail'` policy path); the TUI keeps the 1.5 s delay. The sweep excludes streams this process runs and is idempotent if the process exits first.

Validated alone: typecheck, lint, dead-code ratchet, and the storage, transcript, progress-view, CLI, desktop, and controllers suites (305 files, 4,187 tests).

LionSR:
Both residuals fixed in 8b55ea9cee: the orphan half's rejection is logged and the swept shells are still returned, so their `removeStream` facts publish regardless; a `confirmDeletable` read that rejects after staging is treated as not deletable, rolls the staging back, and reports the deletion superseded. Storage, progress-view, and transcript suites green alone (74 files, 911 tests); typecheck and lint clean.

PR #11830, opened 2026-09-04
feat(session): close a run's transcript at its own exit, and render an unclosed one as interrupted

LionSR:
Part of #11822 (S1/S2). Stacked on #11824; the base retargets to main when that merges.

## What

- **Write side.** The exit drain in `SessionHandle.settleLiveSessionExecutions` now closes the run's transcript groups as CANCELLED inside the same lease-fenced `afterArtifactsDrained` callback that writes the outcome, resolving the stream from the handle, re-checking the abort signal between the two writes, and logging a skipped close at `warn`. The drain's doc block and deadline warning no longer promise that "a later launch" recovers the run; one settlement is an outcome write plus a transcript flush, which the deadline budgets for.
- **Read side (hard crash or deadline skip).** One pure helper, `taskGroupDisplayStatus(group, runDurablyFinal)`, beside `upsertTaskGroupFromStreamLog`: a group still running on a stream whose resolved phase is a terminal outcome with no producer left renders as CANCELLED. `normalizeGroupStatusEntry` is untouched (its output is written back to disk). Consumed by the progress view's task-group list (icon, label, row class, and the completion-chime map so painted and remembered statuses agree), the terminal's workflow run details, and the shared workflow run model, where the interrupted repaint happens once on the row so the status strip, tallies, and cards cannot disagree.
- `runDurablyFinal` comes from `SessionState.resolveStreamPhase` (origin `derived` with a terminal outcome), carried to the CLI through `childExecutions` and to the webview as `statusDurablyFinal` on the backend-owned stream metadata (prefault `false`); a user stop that turns the phase terminal while the flow still unwinds does not qualify. `workflowRunModel` takes the stream phase instead of a pre-digested `runSettled` boolean, so a foreign-held live run's cards are never repainted as cancelled.

## Deliberately not done

The CLI settled-prefix promotion for typed-terminal rows was implemented and reverted: a run settled in this process reaches its terminal phase before the workflow bridge's cleanup rewrites its rows (pinned by `WorkflowScriptStreamTranscript.vitest.ts`), and a stream with no phase yet can resolve to a derived CANCELLED during the launch-to-reservation gap, so promoting those rows would print stale text into append-only scrollback. A crashed run's promoted-but-unsettled workflow-call rows still render as running in the CLI live pane; only task groups and the phase status strip are corrected here.

## Verification

Typecheck (all projects), dead-code ratchet, lint, and the shared, progress-view, CLI, agent-runtime, transcript, desktop, and architecture suites green on the rebased commit run alone. Adversarially reviewed; the two confirmed findings (user-stop repaint, rows disagreeing with the strip) are fixed in the second commit.


LionSR:
Review feedback applied in 456d639b7f and b4b5a7bdf2 (rebased onto main after #11824 merged), plus a test fix in 3e65f2b2b3:

- One vocabulary for an interrupted card: the producer's settlement is one function, `interruptedWorkflowCall`, called by both `endRunningGroupsForStreams` and the read-side repaint, so a launched call reads `failed` with the same message and an unlaunched one `skipped` (not reached) on both paths; tallies follow. The comment that claimed the repaint "cannot drift" now states the real guarantee.
- The exit drain closes the run's groups with the outcome `finalizeRun` actually left standing (its ok result now carries it), so a driver that won the terminal write keeps COMPLETED or FAILED in the transcript; `taskGroupDisplayStatus` paints an open group with that same outcome instead of a constant.
- One durably-final rule for both hosts, `SessionState.streamDurablyFinal`: a terminal outcome with origin derived, or live with no hold and no tracked execution handle (the untracked-terminal-with-open-group case). The CLI and the renderer both call it.
- The completion chime re-checks on a durability-only change.
- Two `finalizeRun` result assertions updated for the new shape.

Validated alone on the final commit: typecheck, dead-code ratchet, lint, and the shared, agent, progress-view, CLI, controllers, transcript, desktop, architecture, and tools suites. The only failures were the three desktop merge-failure presentation cases, which fail on clean main too (three filtered runs there: 3, 2, and 0 failures); that is #11790, not this PR.

LionSR:
Round two applied in ed591aa030 (then rebased onto main after #11832 merged):

- The read-side repaint now derives the card's `detail` the same way the projector does (`workflowCallDetail`), so a hard-crash repaint shows the same error or note line as a drained transcript; the settlement test asserts both arms.
- `statusDurablyFinal`'s doc names `SessionState.streamDurableOutcome` as the producer and states both cases (origin derived; this process's own live entry with no hold and no tracked execution).
- The drain's doc block records the ordering assumption for the reverse race (drain writes CANCELLED first, driver overwrites the header afterwards).

Typecheck, lint, and the shared, progress-view, and agent-runtime suites green alone.

LionSR:
Round three in 3edc2d0acd:

- A terminal status transition now also pushes the stream's metadata (once per terminal transition, no phase override), so `statusDurablyFinal` reaches the webview on the incremental path and an open group repaints as interrupted without waiting for an unrelated refresh.
- The pre-settlement deadline warning says the run's open groups render as interrupted from the derived outcome until a later settlement persists the close.

On "serialize transcript closure with the driver's finalization": the reverse race (drain writes CANCELLED first, driver overwrites the header afterwards) is acknowledged in the doc block and left as is; making the driver preserve an existing CANCELLED would let the drain's outcome override the driver's real result, which is the worse trade.

Typecheck, lint, and the progress-view, controllers, agent-runtime, desktop, CLI, and transcript suites green alone.

LionSR:
Declined: a live-execution guard on the derived arm of `streamDurableOutcome`. The scenario needs `resolveStreamPhase` to answer `derived` at the moment `handleRunConfig` clears the run facts, and it cannot: `run.config` is emitted from the lifecycle over a launch context that `AgentLaunchContext` builds only after `acquireStreamOrThrow` has reserved the stream (`AgentLaunchContext.ts:622`, `AgentRunLifecycle.ts:689-701`), and a reserved entry answers `live` RUNNING. The tracked-but-not-yet-reserved window before that exists, but nothing clears the facts inside it, so the row keeps showing the previous run's outcome it already showed. The resume path goes through the same launch context.

PR #11831, opened 2026-09-04
fix(tools): stop inferring a run's liveness from a missing outcome

LionSR:
Part of #11822 (S2/S3). Stacked on #11824.

## What

One rule, `resolveExecutionLiveness(executionId, outcome)` in `src/tools/executions/executionLiveness.ts`, answers "is this run still going" for both surfaces that used to infer it from process-local facts: a live handle in this process wins; otherwise `classifyRun` decides, and a run held by a live foreign owner, or one whose facts are unreadable, is reported as unsettled with the held-elsewhere copy, never as `cancelled`; a free lease with no outcome and a valid checkpoint reads as interrupted; everything else keeps today's reading. `getExecutionStatusInfo` and the listing and child formatters are async over it (the listing page fans out at 16-wide through `p-map`); `turnAttributionNote` says "still running" only when something alive owns the run; `showOutput`'s footer derives from the same arm as its header, so one tool result cannot contradict itself. Three stale comments that named restart repair as a recovery path are corrected.

## Why

An orchestrator polling a sibling that another process is executing was told `unknown` or "still running" before, and would have been told `cancelled` by a naive checkpoint stat. A terminal claim to a model must rest on the lease and the checkpoint's validity, not on the absence of a handle here.

## Verification

Typecheck, dead-code ratchet, lint, and the tools, agent, progress-view, CLI, controllers, shared, transcript, desktop, and architecture suites green on the rebased commit (the five CLI files that timed out under concurrent load pass alone). Adversarially reviewed across three rounds; the final ladder is a stat plus at most one lease read per row, with `classifyRun` kept only for the single-run paths that parse a record.

### DIFF
--- a/src/agent/runtime/ExecutionHandle.ts
+++ b/src/agent/runtime/ExecutionHandle.ts
@@ -22,6 +22,12 @@ import { runIdentityName } from '@shared/schemas';
 export interface ExecutionStatusInfo {
   status: StreamPhase | 'unknown';
   elapsed: string | null;
+  /**
+   * Why the status reads the way it does, when the phase alone would mislead:
+   * a run another process holds, or one interrupted with a checkpoint still
+   * on disk. Rendered after the status by `formatStatusInfo`.
+   */
+  detail?: string;
 }
 
 /**
--- a/src/agent/runtime/detachSubagentsOnStop.ts
+++ b/src/agent/runtime/detachSubagentsOnStop.ts
@@ -24,9 +24,12 @@ import { readPlatformSetting } from '@utils/config/platformSettings';
  *   exiting process, so honoring the toggle there would strand children
  *   without finalization (`packages/cli/src/runtime/runExecution.ts`).
  *
- * Host quit is a separate axis this toggle does not govern. The extension and
- * desktop abandon native children to the restart-repair path; the CLI kills or
- * interrupts them, because the process that owns them is the one going away.
+ * Host quit is a separate axis this toggle does not govern, and detaching does
+ * not opt a child out of it: `detachActiveChildren` detaches a child from its
+ * parent without untracking its handle, so the shared exit drain
+ * (`settleLiveSessionExecutions`, #11355) still settles every tracked child on
+ * the way out, on every host. The CLI additionally kills or interrupts them,
+ * because the process that owns them is the one going away.
  *
  * The platform must be initialized before a run can be stopped or killed.
  */
--- a/src/agent/runtime/executionRegistry.ts
+++ b/src/agent/runtime/executionRegistry.ts
@@ -477,8 +477,9 @@ export class ExecutionRegistry {
 
   /**
    * Kill only background OS processes (bash, codex) without touching agent
-   * stream status. Agent executions are left in RUNNING so restart recovery can
-   * restore them to WAITING (resumable) if a flow record exists.
+   * stream status. Agent executions are left in RUNNING: whether one is
+   * resumable afterwards is decided from its durable facts (a flow record on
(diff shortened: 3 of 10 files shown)


LionSR:
Review feedback applied in e68929d73e and 4591ffc2be, then the listing path was reshaped in bacd442f99 after the architecture audit:

- `resolveExecutionLiveness` no longer calls `classifyRun` or parses a checkpoint on the listing or child-line paths. Order, cheapest first: a live handle in this process; a persisted outcome (the listing row the caller already holds, or one core-metadata read); one `exists` stat for the checkpoint; and only when a checkpoint exists, one `inspectExecutionLease` read: held by a live foreign owner reads unsettled with the held-elsewhere copy, owned by this process with no handle reads unsettled and logs the registry/lease disagreement, free reads interrupted. A malformed checkpoint therefore reads as interrupted on the listing, the series' accepted contract; the single-run open path refuses loudly.
- The settled arm carries the persisted outcome, so a run that finished while this process lags releasing its lease still reports its outcome (regression case added).
- `formatChildren` fans out at 16-wide like the listing page; the output footer states only what is established ("no TeXRA process owns this execution and no result was recorded"); the turn-attribution docstring matches the held wording; any read failure reads unsettled with the cause at `warn`.

Validated alone on bacd442f99: typecheck, dead-code ratchet, lint, and the tools, agent, CLI, progress-view, and architecture suites (503 files, 6,904 tests). Cost per listing row is one stat plus at most one lease read; rows with a persisted outcome cost nothing beyond the listing read.

LionSR:
Round two applied in befbf18fbf (then rebased onto main after #11832 merged):

- The ladder reads the lease before the checkpoint stat for any row without a persisted outcome, so a background shell or workflow-script container another live TeXRA process owns reads held elsewhere instead of unknown (they hold a lease for their lifetime and never write a checkpoint); a free lease then lets one stat decide interrupted or settled. Rows with an outcome still pay nothing; one case pins the held shell.
- `showOutput` resolves liveness without the snapshot it read before awaiting the transcript, closing the stale-metadata window on the single-run path; listing and child fan-outs keep the caller's row.

Declined, as the series' contract: parsing the checkpoint on the interrupted arm (a malformed record reads as interrupted on listings; the single-run open path refuses loudly), and re-reading each listing row's outcome (a millisecond window on a bounded page).

Typecheck, lint, dead-code ratchet, and the tools and agent-runtime suites green alone.

LionSR:
Round three in dee853d4b6:

- The interrupted arm's detail is presence-only ("a flow record remains (not validated here)"); status stays CANCELLED and the listing still never parses. Only the single-run paths that parse a record may promise a resume.
- `turnFate`'s live arm mirrors `showOutput`'s footer: "still running" only while the handle's phase is in flight, otherwise "ended with its run (<phase>)"; one case pins a tracked handle whose stream already completed.
- The listing and `formatChildren` comments describe the lease-first order.

Typecheck, lint, and the tools and agent-runtime suites green alone. The PR body's verification line will be refreshed at merge; the head uses the stat ladder, not `classifyRun`.

LionSR:
Round four at bf6835bca5: the local import block in `executionFormatters.ts` carries its header.

Declined: re-reading the lease after the checkpoint stat. The listing is a point-in-time reading and every ordering of two reads has a window in which the other read goes stale; the earlier post-facts recheck narrowed that window, it never closed it. The acquire on the run the user actually opens is the authority, and the next listing corrects a line that a resume overtook mid-read. Adding a third read per row to shrink a window that cannot be closed is the kind of guard this series is deleting.

PR #11832, opened 2026-09-04
feat(runtime): record the read-only hold when a run is opened for write

LionSR:
Part of #11822 (S3). Stacked on #11824.

## What

`StreamStatusMachine.markUnavailable` gains production writers at open-for-write, so the "held elsewhere" display fact survives the deletion of restart repair's two boot-time producers: `resumeRun`'s refusal on `ExecutionLeaseActiveError` (both the workflow launch and the tool-use resume paths) records `streamHeldMessage(owner)` on the stream, a stale hold is dropped once at the point the run is opened for write, and `ToolUseFollowUp.classifyRefusal`'s held-elsewhere arm records the same while its successful classifications clear it. A hold now publishes as a `streamHoldChanged` session fact (a hold carries no phase, so it cannot ride the status arm), hosts re-read the resolved phase on it, and the unavailable toolbar keeps a Resume affordance so a stale hold has an in-tab action that re-reads the run. The `unclassified` arm deliberately records nothing: `classifyRun` reports it for transient read failures, and a sticky hold would turn a blip into a permanent read-only tab; the unreadable display fact keeps its producer in PR2's run tuple.

Only the one run the user acted on is inspected; nothing is added to any all-runs path, and nothing is written to disk.

## Design note

S1/S3 supersede D3/D9's "shown as held at load" clause from the single-owner-sessions proposal: after restart repair is deleted, a foreign live run is shown held once its sidecar hydrates (PR2's per-stream lease probe) or when the user acts on it (this PR), never by a boot pass.

## Verification

Typecheck, dead-code ratchet, lint, and the agent, progress-view, CLI, controllers, shared, transcript, desktop, and architecture suites green on the rebased commit. The moved test ("holds a foreign-owned run read-only while its owner is live") now drives the follow-up refusal in `DerivedStreamPhase.vitest.ts`. Adversarially reviewed; the unpublished hold, the sticky transient hold, and the missing clear on successful classification were fixed in the second commit.

### DIFF
--- a/packages/cli/src/runtime/sessionProgressSubscription.ts
+++ b/packages/cli/src/runtime/sessionProgressSubscription.ts
@@ -59,7 +59,7 @@ type CliRunFactEvent = Extract<
 
 /**
  * Project session facts onto the frozen NDJSON progress-event vocabulary.
- * `followUpSent` is intentionally session-local.
+ * `followUpSent` and `streamHoldChanged` are intentionally session-local.
  */
 function projectCliSessionFact(
   fact: SessionFact,
@@ -72,6 +72,10 @@ function projectCliSessionFact(
     case 'updateQueuedFollowUps':
       return { event: 'updateQueuedFollowUps', payload: fact.payload };
     case 'followUpSent':
+      return undefined;
+    // A read-only hold is a session-local display fact; the frozen NDJSON
+    // vocabulary carries no record for it.
+    case 'streamHoldChanged':
       return undefined;
     case 'setActiveStream':
       return { event: 'setActiveStream', payload: fact.payload };
--- a/packages/extension/src/progressView/frontend/components/StreamHeader.ts
+++ b/packages/extension/src/progressView/frontend/components/StreamHeader.ts
@@ -102,11 +102,19 @@ const TERMINAL_STATE_BUTTONS = [
 
 /**
  * Buttons enabled while the run is unavailable here (held by another TeXRA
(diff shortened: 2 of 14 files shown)


LionSR:
Review feedback applied in 9a4225ad00 and da86dcadcf (rebased onto main after #11824 merged):

- The stale hold is dropped before the `finished` refusal, and that refusal is now decided from the lease through the same `recordRunRefusal` the follow-up path uses, so a run another process just claimed keeps its fresh held-elsewhere hold instead of being reported finished; a readable checkpoint this retrieval cannot prepare refuses as `not_resumable` rather than `finished`.
- `clearHold` gained `discardRetainedPhase`: the `finished` and `resumable` classifications delete the entry rather than resurrecting a retained WAITING phase; `owned_here` keeps a hold with the owned-here diagnostic instead of clearing.
- The delayed status push re-reads the hold after `ensureLoaded` and drops its override when a hold now exists; `transition()`'s early return no longer fires on a hold entry, so a hold always converts through the publish path.
- One `markUnavailableOrLog` replaces the three copies of the mark-or-log pattern.
- The unavailable sentinel no longer carries the retained phase's substate or `runStartedAt`.
- The moved test's comment now says the case is moved ahead of PR6's deletion of the boot-time producer, which is still live at this commit.

Validated alone on da86dcadcf: typecheck, dead-code ratchet, lint, and the shared, agent, progress-view, CLI, controllers, transcript, desktop, architecture, and tools suites (638 files, 8,328 tests).

PR #11835, opened 2026-09-04
perf(cli): advertise a resumable history row from one stat, not a checkpoint parse

LionSR:
Part of #11822 (S1, second half). Independent of the PR3 to PR5 stack.

## What

- `listExecutions` carries `checkpointPresent` (one `exists` stat inside the per-execution reads it already does) and the `streamId` it already read. `texra history list` and the `/resume` picker derive `resumable` and `status` from those facts through the unchanged `resolveHistoryRunStatus`, so the per-row lease inspection and the two or three full checkpoint parses per row are gone. One rule, `isCliRunResumable`, serves both the listing and `history show`, so the two surfaces cannot disagree; the detail path keeps a single parse only for the resumed model.
- Two free gates stay: a row without a stamped stream id is not resumable (the ~1,100 legacy rows in a large workspace no longer fill the picker), and a workflow row is checked for a terminal compile rejection only when its outcome is FAILED (a terminal rejection always finalizes FAILED), so cancelled and interrupted workflow rows cost one stat.
- The open path is honest about what the listing admits. A new `unusable_checkpoint` refusal ("This run's saved state could not be loaded, so it cannot be continued. Delete it from history and start a new agent task.") is returned by `resumeRun` when retrieval yields nothing while the checkpoint file exists, and by `texra resume` for an unclassifiable record; a run predating stream stamping says so instead of "finished". The chat TUI now probes the resume before it clears the transcript or switches streams, so a refused resume leaves the conversation as it was (one regression test).

## Accepted consequences

A run another process is executing right now lists as resumable and is refused when opened (PR5's hold at open-for-write). The `model` column shows the startup model; the resumed-model override lives only inside the checkpoint and `history show` still reports it.

## Verification

Typecheck, dead-code ratchet, lint, and the CLI, agent, tools, storage, architecture, and progress-view suites green on the rebased commit run alone. Adversarially reviewed; the reviewer's blocker (unusable rows advertised and then refused with misleading wording) is what the open-path changes address, without re-adding any parse to the listing.

### DIFF
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -581,6 +581,17 @@ export function createChatSessionController(
         session.streamId = streamId;
         session.executionId = id;
         rootStreamId.set(streamId);
+        // The session held no stream until the line above: `markRunPending`,
+        // inside the synchronous slot claim at the top of `resume`, dropped
+        // the pre-resume one, so a Ctrl-C in the window before adoption could
+        // not fabricate an interrupted marker on a stream this resume is
+        // leaving behind. It also found nothing to interrupt, so re-read the
+        // request here — the way `startRootRun`'s `onStreamResolved` does —
+        // and let it land on the run the user asked to continue. `resumeRun`
+        // re-reads `isCancellationRequested` once this hook returns, so the
+        // stop still refuses the launch; this only decides which stream it
+        // marks recoverable.
+        if (session.stopRequested) interruptActiveRun();
 
         await runtimeSession.transcripts.ensureLoaded(streamId);
         // `load` evicts every other record synchronously before its async
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -12,6 +12,7 @@ import {
   describeFollowUpFailure,
   detachSubagentsOnStop,
   lookupStreamExecutionId,
+  probeResumeRefusal,
   resumeRun,
   runAgent,
   type AgentConfig,
@@ -556,6 +557,22 @@ export function createChatSessionController(
       if (!recovery) {
         restoreInterruptedRecovery(supersededRecovery);
         appendLocalErrorTranscript(describeFollowUpFailure('not_resumable'));
+        session.markRunCompleted();
+        resolveRunPromise();
+        return;
+      }
+
+      // Validate before mutating anything. A history row is advertised from
+      // its checkpoint file alone (one `stat`, no parse), so a run whose saved
+      // state cannot be loaded reaches here; refusing it after the transcript
+      // was cleared and the session switched to its dead stream would take the
+      // user's chat with it and answer in the wrong window.
+      const refusal = await probeResumeRefusal(id, config, streamId);
+      if (refusal) {
+        runtimeSession.followUps.release(recovery, 'recoverable');
+        recovery = undefined;
+        restoreInterruptedRecovery(supersededRecovery);
+        appendLocalErrorTranscript(describeFollowUpFailure(refusal));
         session.markRunCompleted();
         resolveRunPromise();
         return;
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -12,7 +12,6 @@ import {
   describeFollowUpFailure,
   detachSubagentsOnStop,
   lookupStreamExecutionId,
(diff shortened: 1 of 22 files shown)


LionSR:
Follow-up in cec522dbcb: the resume path now parses a checkpoint once. `probeResumeRefusal` is deleted; `resumeRun` takes an `onResumeResolved` hook fired after its own retrieval succeeds and before dispatch, and the chat TUI performs its transcript-clear, stream-switch, and focus inside that hook, so a refusal (including the new `unusable_checkpoint`) returns before any mutation without a second parse. The refusal now also flows through the run chain's shared handling (exit code set as for the other refusals). Typecheck, lint, dead-code ratchet, the chat-session-controller suite, and the agent-runtime suites green.

LionSR:
Round two applied in 0906a220bb:

- **Ownership before mutation.** `resumeRunWithRecoveryProvenance` inspects the execution lease once before firing the adoption hook; a live foreign owner refuses `owned_elsewhere` with the hold recorded, before the chat clears anything. The later acquire still settles the race window. One lease read per open.
- **`unusable_checkpoint` is typed, not guessed.** `deriveResumability`'s unreadable decision now carries a `ResumabilityFault` (metadata unreadable or malformed, checkpoint unreadable or malformed); retrieval throws `PersistedFlowStateError` only for a malformed checkpoint, and `resumeRun` returns `unusable_checkpoint` only when the cause chain names one. A transient read (`KV timeout`), a snapshot failure, or a lease failure propagates as the operational error it was ("Could not load session …", exit code 1); the test that had pinned the opposite is restored and a positive malformed-record case added.
- **`texra resume`'s unclassified arm** words only a malformed checkpoint as unusable state; lease and metadata read failures get the neutral "Could not read the state of execution" message.
- **A failing checkpoint stat keeps the row** (`checkpointExists` warns and reads false), in both the listing and `history show`.
- **A config-less run is not resumable** in `history show`.
- **The `flowKey` barrel re-export is gone**; the CLI reaches `checkpointExists` through the existing barrel, no new deep import.

Validated alone on 0906a220bb: typecheck, lint, dead-code ratchet, and the CLI, agent, architecture, and tools suites (448 files, 6,369 tests).

LionSR:
Rebased onto main after #11832 merged (0ad327671c). The two PRs rewrote the same refusal arm, so the merged rule is: the lease-aware classification runs first and records or clears the hold exactly as #11832 does; `held_elsewhere` and `owned_here` return that reason with no checkpoint read; `unclassified` reads `unusable_checkpoint` only for a `checkpoint-malformed` fault; `finished` and `resumable` let one `checkpointExists` decide between `unusable_checkpoint` and the classified reason. The pre-hook lease refusal reuses `markUnavailableOrLog`. Every test either PR added is present; the one shared arrangement (workflow config, readable checkpoint, empty retrieval) now expects `unusable_checkpoint`, which is what the merged rule mandates. Typecheck, lint, dead-code ratchet, and the CLI, agent, architecture, and tools suites green alone.

LionSR:
Round three in 8eb65ae014:

- The terminal-rejection gate skips the parse only for CANCELLED and COMPLETED, so FAILED and outcome-less workflow rows (the crashed-workflow cohort; legacy outcome-less rows fail the stream-id gate first) are read before being advertised. The doc states the cost honestly.
- A parse failure in that gate now advertises and defers to open (logged at warn) instead of hiding the row, so the "refuse at open" contract holds for this branch too.
- Ctrl-C before adoption: the premise that `session.streamId` still holds the previous stream is not right on this path, since claiming the root-run slot clears it before any await, so no stale marker could be fabricated; the real defect was that the stop was dropped. `adoptResumedStream` now re-checks `stopRequested` right after it sets the stream id, the way `startRootRun`'s `onStreamResolved` does; one case pins that a stop in the gap targets the resumed stream and never the previous one.

On "hold the execution lease before adopting the stream": the pre-hook read is a snapshot by design; acquiring in `resumeRun` would split lease ownership away from `resumeToolUseTurn`. The launch's own acquire settles the race the snapshot cannot see; the residual window is two processes resuming the same row at the same instant, accepted.

Typecheck, lint, dead-code ratchet, and the CLI and agent-runtime suites green alone.

LionSR:
Round four at 5b335b7a4e: the lease preflight runs before the earlier refusal's hold is touched. A read that throws leaves the hold standing (the attempt learned nothing that disproves it); once the inspection answers, the hold is cleared and a held lease overwrites it with the current owner. Typecheck and the agent-runtime plus CLI suites pass alone.

PR #11836, opened 2026-09-04
refactor(session): read run facts at row open; delete the background hydration pass

LionSR:
Part of #11822 (S1/S3), correcting two pieces of #11824 after the architecture audit: the post-load background hydration was an O(history) pass moved off the ready path, and the run facts widened a frozen store surface.

## What

- **No boot pass over rows.** `ProgressBackend`'s background hydration (chunked walk over every selectable stream with a checkpoint stat and a lease read each) is deleted outright. Rows the user has not opened render from the always-resident summary tier; the open-transcript flags give the interrupted arm at zero I/O; everything else reads as pending until opened.
- **Run facts belong to the controller.** `StreamSnapshotStore` loses `probeRunPhase`, the `runFacts` map, `getRunPhaseFacts`, and the hydration-time lease, checkpoint, and outcome reads. `SessionState` keeps an in-memory per-stream tuple (outcome, checkpoint present, lease, authority failure; never persisted) filled by one `hydrateRunFacts(stream)` at row open, reading the lease, then the checkpoint stat, then the core metadata, in that order so a free lease is always followed by the outcome a completed finalize already wrote. Cleared on execution handoff and stream deletion. `resolveStreamPhase`'s arms are unchanged; they read the tuple instead of the store.
- **Row open is the only caller**: the progress backend's selection hydration and the CLI's focused-stream hydration. No all-rows path.
- **Ratchet reverted**: `config/ratchets/store-public-surface-baseline.json` returns `StreamSnapshotStore` to its pre-#11824 five-field `getRunMetadata` with no new method; `StreamLogStore` keeps `hasUnfinishedOutput`.

Net: 9 files, +200 / -344.

## Why

The startup proposal forbids any boot-time walk over history, bounded or not, and the repo freezes the shared store surfaces. A liveness probe belongs at row open (S3), not at hydration or at boot.

## Verification

Typecheck clean; lint and prettier clean; progress-view, transcript, architecture (including the store-surface ratchet against the lowered baseline), desktop, and agent-runtime suites green; `DerivedStreamPhase` cases re-seeded through `hydrateRunFacts`. Central dead-code ratchet and CLI suites run before merge.


LionSR:
Both findings fixed in 4bdef04e39:

- `readRunFacts` now reads the execution config in the same try (after the lease, checkpoint, and outcome reads, so the ordering invariant is unchanged), so a malformed or unreadable `config.json` lands in the tuple's `authorityFailure` and the row renders unavailable with the cause instead of resolving to its outcome; one case pins it and fails without the read.
- The strict core-metadata read has one home: `readExecutionMetaCore(store)` in `ExecutionKVStore.ts`, used by both `StreamSnapshotStore` and `SessionState`, reached through the existing barrel both already import.

Typecheck, lint, dead-code ratchet, and the agent-runtime, transcript, and architecture suites green.

LionSR:
Applied the incarnation re-check in 1279680574: `hydrateRunFacts` captures the stream's incarnation before its reads and drops the tuple if it moved, which covers a same-execution-id re-run and a tombstone-dropping re-claim that the execution-id comparison alone missed. The eviction note is as intended: facts belong to the session, not the record. Typecheck, lint, the derived-phase and progress-view suites green.

PR #11837, opened 2026-09-04
perf(session): delete the startup restart-repair pass and the readiness gate

LionSR:
Part of #11822 (S1). The deletion the series exists for.

## Measured

`texra chat --agent research --model deepseekproT` in the TNLean workspace (4,245 execution directories, 3,222 still holding a checkpoint), wall clock from launch to the input prompt, tmux recipe, same machine, back to back:

| Build | Time to prompt |
| --- | --- |
| main at c0870c7fbf (PR1, PR2, PR5 merged) | 31.4 s |
| this branch, run 1 | 1.6 s |
| this branch, run 2 | 1.6 s |

The deleted pass's read-only floor on a clone of that bucket was a 0.1 to 0.2 s scan plus about 5 s of checkpoint parsing over 1.8 GB before the sequential lease-locked settle that dominated the earlier 43 to 108 s measurements.

## What

- `src/agent/runtime/restartRepair.ts` deleted. In `SessionHandle`: `waitUntilReady`, `ensureRestartRepair`, `repairStoresAfterRestart`, `runRestartRepair`, `computeStartupSeedSet`, `isRepairSuperseded`, the repair promise and abort controller with its teardown entry, and the `restartRepair: 'deferred'` init option with its constructor branch.
- The readiness awaits are gone from the extension entry, the desktop entry (and its `deferred` option), the CLI's transcript session (whose persistent-session helpers became synchronous), and `ProgressBackend.load`. The desktop bridge's own `waitUntilReady` is a different object and is untouched.
- Machinery that lost its last caller went with it rather than into a baseline: `StreamStatusMachine.getGeneration`, `STREAM_TRANSITION_CAUSE.RESTART_REPAIR` and its transition-table arm, the `classifyRunFacts` export. `config/ratchets/host-agent-mock-baseline.json` drops its two stale repair mocks; the store-surface baseline keeps its counts (only prose naming deleted callers changed).
- Tests pinned to the boot pass are deleted (`RestartRepair`, `SessionRestartRepair`, `ProgressBackendPresentation`) and the desktop cases that asserted its input ordering; two desktop cases that had relied on the readiness await as an incidental settle point now wait on the fact they read.
- Prose in five production files that asserted restart repair exists is corrected; dated proposals under `docs/` stay as history.

What replaces it is already on main: the read-time phase rule (#11824, refined by #11836) and the hold written at open-for-write (#11832).

## Settling a crashed run's transcript

The review pass found that deleting repair left `endRunningGroupsForStreams` with no production caller, so a transcript a dead host left open was never closed on disk again. An interim commit ran that settle on the post-UI timer the leftover-stream sweep uses; that timer pass is gone (1d60213430). The write now happens at row open: `SessionState.hydrateRunFacts`, for the one stream a user just opened, calls `settleInterruptedRun` when the tuple it read says the lease is free, no outcome was ever recorded, and the transcript still has unfinished output. Under `runWithInactiveExecutionLease` it re-reads the outcome, records CANCELLED with the checkpoint preserved, closes the open groups, flushes, and re-reads the tuple so the row renders what was written. A refusal (a live owner) writes nothing. No timer, no roster walk, no background pass. A run whose outcome is already persisted is not rewritten: #11830's read-time rule paints its open groups from that outcome. #11830 adds the exit-time close for graceful shutdowns.

Net: 34 files, +400 / −2,287.

## Verification

Typecheck across all projects (the agent SDK build validates 731 declarations), lint, dead-code ratchet, the 15 architecture ratchet suites, and the full vitest suite (820 files, 9,980 tests) green on the current head run alone. Adversarially reviewed; one finding rejected because PR5 on main gives the hold machinery its writers.


LionSR:
Declined: settling a transcript that already has a persisted outcome. That row is exactly what #11830 answers at read time: `streamDurableOutcome` returns the persisted outcome and both renderers paint the open group or card as interrupted with it (`taskGroupDisplayStatus`, `sessionStreamDurableOutcome`), so nothing renders as in progress. The settle here exists to record an outcome where none was ever written; rewriting a transcript whose outcome is already durable would put a second writer on state the read-time rule already covers. Merge order is #11830 before this PR.

LionSR:
Rebased onto main after #11830 merged (bfdd3b6b3e): one delete/modify conflict on `RestartRepair.vitest.ts`, resolved by keeping the deletion. Typecheck, lint, dead-code ratchet, and the full suite (820 files, 9,982 tests) green on the rebased head run alone. The PR body's settle section now describes the row-open settle in place of the timer.

LionSR:
Rebased again after #11831, #11835, and #11823 merged (0588f511c2): two prose conflicts in `detachSubagentsOnStop.ts` and `ExecutionKVStore.ts`, resolved to main's wording, which is already repair-free. Typecheck, lint, dead-code ratchet, and the full suite (820 files, 10,005 tests) green on this head run alone; no `restartRepair` reference remains outside dated docs.

PR #11854, opened 2026-09-04
Skip restart repair for streams reused during lease lock contention

LionSR:
## Summary

When a stream is reused (becomes current again) while the restart repair process is waiting to acquire an execution lease lock, the repair should be skipped rather than marking the stream as unavailable. This prevents incorrectly holding a stream in an "unclassified" state when the stream has already been reclaimed by active execution.

The fix adds a currentness check after a lease lock acquisition failure. If the stream is no longer current at that point, the repair is skipped with a debug log rather than applying a hold.

## Issue acceptance gates

n/a — no issue referenced

## Single source of truth

`repairRestartedStreams()` in `src/agent/runtime/restartRepair.ts` owns the decision of whether to apply a hold or skip repair based on stream currentness and lock contention.

n/a — no duplication removed or abstraction added

## Net elements (R6)

- Modified files: 2
- Net LoC: +37 (test setup +16, test case +11, production logic +6, test assertions +4)
- No new exports or public symbols

## Consumer counts (R8)

n/a — internal logic change with no new exports

## Host impact

Extension: None
Desktop: None
CLI: None

## Scheduled refactoring

None

## Validation

Added test case `skips a candidate reused while the lease lock was unavailable, instead of marking it unavailable` that:
- Mocks `runWithInactiveExecutionLease` to reject with lock contention
- Verifies the stream is current on the first check but reused (not current) when the lock fails
- Asserts the stream is not marked unavailable and no groups are closed

The test uses the existing `vi.hoisted()` mock setup pattern to intercept the lease lock module.

https://claude.ai/code/session_01NQCfK7TeMFDGvueTFw4ytt


LionSR:
Closing note for context: this PR's change (`restartRepair.ts`'s lease-lock catch missing an `isCurrent` re-check) was already fixed independently in #11785, which merged before this branch was rebased. Then #11837 deleted `restartRepair.ts` and the whole startup restart-repair pass entirely (superseded by `SessionState.resolveStreamPhase`), so the file this PR touched no longer exists on `main`.

The branch (`claude/modest-lovelace-1mzg7q`) has been reset to `main` — that's why the diff here collapsed to empty and the PR auto-closed. No action needed.

---

PR #11930, opened 2026-09-06
refactor(cli): fold viewSignal into the shared toSignal helper

LionSR:
## Summary

- `packages/cli/src/chat/tui/state/sessionView.ts`'s `viewSignal` (added in the one-fold-three-renderers lane, #11881) reimplemented `src/shared/signals.ts`'s `toSignal` line for line: same `signal(initial)` + `runtime.runFork(Stream.runForEachArray(...))` + `Object.assign(s, { dispose })` shape, in a package that already imports three other helpers from that exact module (`subscribeToSignalChanges`, used in `terminalTitle.ts`, `runChatTui.tsx`, `subscribeApprovals.ts`, `useSignal.ts`).
- `viewSignal` now delegates to `toSignal`, and the undefined-array guard the CLI copy had (`if (last !== undefined) s.set(last)`) is folded into `toSignal` itself, so `toSignal`'s two pre-existing call sites (`packages/extension/src/progressView/frontend/sessionTransport.ts`'s `view$`/`host$`) also drop the unsound `arr.at(-1) as A` cast that could silently write `undefined` into a typed signal if a drained array were ever empty.
- This is a scheduled routine's daily consolidation/native-methods survey pass; the write-up is `docs/proposals/2026-09-06-consolidation-and-native-methods-survey.md`, following the existing series (`docs/proposals/2026-0{8-29,8-30,9-02,9-03,9-04,9-05}-consolidation-and-native-methods-survey.md`) — this is the first entry in that series since 2026-08-29 to find and fix a real candidate; every prior round came up empty.

## Issue acceptance gates

No tracked issue; this is a scheduled routine's own consolidation sweep, not a response to a filed issue.

## Single source of truth

`src/shared/signals.ts`'s `toSignal` is now the one implementation of "a signal fed by an Effect stream, with the fiber's interrupt," used by both `packages/cli/src/chat/tui/state/sessionView.ts` and `packages/extension/src/progressView/frontend/sessionTransport.ts`.

## Duplication and abstraction budget

Removed: one duplicate reimplementation of `toSignal` (`viewSignal`'s body), and the three now-unused `Effect`/`Fiber`/`Stream` imports it required. No new abstraction was added — `viewSignal` stays as a one-line adapter naming `toSignal`'s CLI-specific call site, not a pass-through worth inlining further since it's still exported indirectly via `bindSessionView`.

## Net elements (R6)

[...]

PR #11946, opened 2026-09-06
style(docs): format the consolidation survey so main's format:check passes

LionSR:
## Summary

`main` is red. #11930 landed `docs/proposals/2026-09-06-consolidation-and-native-methods-survey.md` without running prettier, so `pnpm run format:check` fails in the **static checks (linux)** job, which fails the **validate** gate. CI run [34027329517](https://github.com/LionSR/TeXRA/actions/runs/34027329517) on `fd32b8b` reports exactly one file:

```
$ prettier --check .
[warn] docs/proposals/2026-09-06-consolidation-and-native-methods-survey.md
[warn] Code style issues found in the above file.
```

Every open PR inherits the failure, since `format:check` runs over the whole tree.

This is prettier 3.9.6's output, unedited — `prettier --write` on that one file, nothing else touched:

- Two fenced TypeScript examples get statement-per-line bodies instead of single-line arrow bodies.
- The second line of the status blockquote loses its `> ` prefix. That prefix sat *inside* a multi-line inline code span (`` `chore(deps-dev): bump …` ``) and was being rendered literally as part of the code. Markdown's lazy continuation keeps the line in the blockquote without the marker, so the rendered document is unchanged and the stray `>` disappears from the code span.

## Issue acceptance gates

No issue is closed. This restores the base branch to green; the substance of #11930's survey is untouched.

## Single source of truth

n/a: formatting only.

n/a: documentation only.

## Net elements (R6)

Files: 1 modified, 0 added, 0 deleted. No exported symbol, class, interface, or dependency changes. Net LoC +10, all of it prettier's line wrapping.

## Consumer counts (R8)

n/a: no emit path or code change.

## Host impact

## Scheduled refactoring

None. Worth noting for the next docs-only PR: `format:check` covers `docs/**`, so a doc added without prettier turns the whole base branch red, which is how this one slipped through.

## Validation

- `prettier --version` → 3.9.6, the pinned devDependency CI installs.
- `prettier --check docs/proposals/2026-09-06-consolidation-and-native-methods-survey.md` fails before this change and passes after.
- `node docs/scripts/check-root-docs.mjs`: unaffected — the file is under `docs/proposals/`, an internal area excluded from the published site.
- No code changed, so no typecheck, lint, or test run applies.

https://claude.ai/code/session_01UyJjcnpYcFopcGbYdWayxp

---
_Generated by [Claude Code](https://claude.ai/code/session_01UyJjcnpYcFopcGbYdWayxp)_

### DIFF
--- a/docs/proposals/2026-09-06-consolidation-and-native-methods-survey.md
+++ b/docs/proposals/2026-09-06-consolidation-and-native-methods-survey.md
@@ -1,7 +1,7 @@
 # Survey: code consolidation and native-method opportunities (2026-09-06)
 
 > **Status:** Written 2026-09-06 against branch HEAD `03d2cfd` (`chore(deps-dev):
-> bump the development-dependencies group across 1 directory with 6 updates`,
+bump the development-dependencies group across 1 directory with 6 updates`,
 > #11842). Scheduled routine re-ran the standing question — "find
 > duplicate/similar logic to consolidate, and hand-rolled code that a native
 > method or the standard library already covers" — three days after
@@ -56,10 +56,16 @@ were the same function:
 const s = signal(initial);
 const fiber = runtime.runFork(
   Stream.runForEachArray(changes, (arr) =>
-    Effect.sync(() => { s.set(arr.at(-1) as A); }),
+    Effect.sync(() => {
+      s.set(arr.at(-1) as A);
+    }),
   ),
 );
-return Object.assign(s, { dispose: () => { runtime.runFork(Fiber.interrupt(fiber)); } });
+return Object.assign(s, {
+  dispose: () => {
+    runtime.runFork(Fiber.interrupt(fiber));
+  },
+});
 
 // packages/cli/.../sessionView.ts — viewSignal (new in this window, #11881)
 const s = signal(SubscriptionRef.getUnsafe(view));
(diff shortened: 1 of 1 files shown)


LionSR:
Heads-up on what this unblocks, and what it uncovers.

The **static checks (linux)** job runs `format:check` at step 1 and `check:effect-migration-ratchet` several steps later, so the prettier failure has been masking the ratchet — it never got to run. With this merged, it will, and it fails:

```
@adapter-until check failed: 1 marker(s) present; there are no temporary adapters.
  - src/tools/github/PollingSourceBase.ts:413
```

That marker came in with #11923 and the check that rejects it came in with #11920; the two are individually fine and only conflict once both are on `main`. So expect `main` to go from red-on-prettier to red-on-ratchet, and this PR is still the right first step — it is what lets every open PR get past step 1.

The marker is honest, not lazy: it covers one `setInterval` + `timer.unref()` in `ensureTimer`, kept off Effect's clock because rc.112 ships no unref-capable scheduler (there is no `unref` anywhere in `node_modules/effect/dist`), so a `Schedule`-driven fiber would pin the `texra` process alive for as long as anything is subscribed.

The fix is real and already written in the w5 rebuild lane: an unref'd sleep expressed as `Effect.callback` over `setTimeout(...).unref()`, which `Schedule.spaced` then drives — so the poll cadence moves onto Effect's clock (R8) without the process-lifetime regression, and `ensureTimer`, `stopTimer`, and the marker all go away. That lane is mid-conversion in this session; I will open it as its own PR as soon as it commits and validates, and it should land close behind this one so `main` is not red on the ratchet for long.

If you would rather not have that window at all, say so and I will fold the `PollingSourceBase` conversion into this PR instead — I kept it out because a formatting fix and a runtime conversion do not belong in one diff.

---

PR #11954, opened 2026-09-06
refactor: post-refactor simplification sweep since v0.40.8 (Effect-native + declarative)

LionSR:
A 54-agent code-simplifier sweep over everything `main` changed since the `v0.40.8` tag — the session-fold PRD lanes, the Effect v4 runtime migration, the progress-view and desktop-shell lanes, the session ledger — minus every file touched by a then-open PR.

**Scope:** 414 changed non-test `.ts`/`.tsx` files, minus the 163 touched by the 12 then-open PRs (#11946 #11945 #11943 #11942 #11936 #11928 #11926 #11922 #11921 #11919 #11893 #11818), minus `src/test-kernel/` → 395 files / 108.5k lines, bin-packed into 54 disjoint directory-coherent batches. Charter priorities on top of the standard rules: **Effect-native** (for files already importing `effect`) and **declarative**.

**27 of 54 batches reported zero edits.** That is the honest result for code this recently reviewed, and it extends the 2026-08-20 whole-tree finding that per-file simplification is largely exhausted here. Agents were explicitly told zero edits is a valid outcome.

## What landed — 37 files, +267/−423, 61 named elements removed

- **Dead imports left by the fold refactor** — 7 in `packages/cli/src/chat/tui/App.tsx` alone (`AgentCategory`, `isPlainAgentIdentity`, `USER_FOLLOW_UP_SUPPORT`, `isActivePhase`, `isInFlightPhase`, `StreamView`, `streamPhaseOf`).
- **Single-caller pass-through collapses** — clearest is `withOpenTurnUpdate` in `src/tools/inquiry/externalInquiryStorage.ts`, whose own docstring called it "the shared guard behind `recordAnswerForOpenTurn`" while having exactly that one caller. Mutex/guard/write ordering preserved.
- **Effect-idiom fixes** — `Stream.unwrap(Effect.sync(() => …))` → `Stream.suspend(…)` in `SessionEvents.ts`'s `readAll`/`readListing`. The third `Stream.unwrap` in that file is deliberately left alone: its body does real Effect work before producing the stream, so `Stream.unwrap(Effect.gen(…))` is correct there.
- **Declarative hardening** — `SessionStores.ts`'s execution-deletion if-ladder is now an exhaustive `switch` with a `never` guard, so a new `ExecutionDeletionOutcome` variant compile-fails instead of silently falling into the `retained` repair path. Net +LOC, justified under dispatcher-as-contract (#7159).

## Gates

Workers ran no repo-wide gates (54 concurrent editors would read each other's half-written files); everything was serialized centrally afterwards.

| Gate | Result |
|---|---|
| `format:check` | pass |
| `lint` | pass |
| `typecheck` (all 6 tsconfig projects) | pass |
| `test` | **764 files / 9070 tests, 0 failures** |
| `check:dead-code-ratchet` | pass (286 vs 286 baselined) |

**One regression, caught by the gate and fixed here:** the `hostDraftRequests.ts` `tryPromise` collapse dropped an `async` that was load-bearing — `runInSession` returns `Promise<X> | X` and the original `try: async () => runInSession(…)` was coercing that union. The `async` is restored; the collapse stands. An agent verifying by reading could not have caught it, which is the argument for the central-gate topology.

**Ratchet baselines are untouched — this widens nothing.**

## ⚠️ CI will be red on a break this branch inherits from `main`

`static checks (linux) :: Check Effect migration ratchet` fails on `main` today (run 34030234404, `.github/workflows/ci.yml:214`), and every branch cut from `main` inherits it. Verified against a pristine `origin/main` worktree.

Cause is two-PRs-combine-badly, not a mistake in either: #11923 introduced an `@adapter-until 2026-11-05` marker in `src/tools/github/PollingSourceBase.ts` while markers were legal; #11920, merged **after**, added the check that forbids all markers. Each was green on its own base.

Worth noting the marked code is correct: the `setInterval(...).unref()` cadence is deliberately off Effect's clock because rc.112 has no unref-capable scheduler, so a `Schedule`-driven fiber would pin the CLI process. Resolving this is an owner call on the R1 ruling, and out of scope here.

Relatedly, `--update` on the Effect baseline currently refuses (11 files hold `Effect.run*` below the sanctioned boundary), so accumulated shrinks — including a verified-stale `import:p-timeout` row for `UsageLogService.ts` — cannot be locked in yet.

## Also in this PR

`docs/proposals/2026-09-06-post-refactor-simplify-sweep.md` records the sweep and, more usefully, the **78 Effect-port opportunities** the agents recorded rather than took (they were barred from porting non-Effect files, since the migration PRD owns those lanes). 24 name a hand-rolled primitive with a direct Effect equivalent. Four were spot-verified — three held (`approvalQueue.ts`'s raw `new Promise` latch in an Effect-importing file; `goalStore.ts` running `async-mutex` alongside Effect; the stale `p-timeout` baseline row) and **one was refuted** (`perKeyQueue.ts` already uses Effect `Deferred` by design). Verify before scheduling.

https://claude.ai/code/session_01QKbhoXfmZmZDgfFFZwrpqA

### DIFF
--- a/docs/proposals/2026-09-06-post-refactor-simplify-sweep.md
+++ b/docs/proposals/2026-09-06-post-refactor-simplify-sweep.md
@@ -69,23 +69,57 @@ on open PRs — but to record the opportunity instead. **78 of the 261 skips are
 Effect-port opportunities**, and 24 name a hand-rolled primitive with a direct
 Effect equivalent.
 
-These are agent claims. Four were spot-verified; three held and one did not.
-**Verify before scheduling.**
+These are agent claims, and they do not survive inspection at anything like
+face value. Five were spot-verified: **two held, two were refuted, one is
+already being fixed elsewhere.** Treat the list as leads, never as a queue.
 
 ### Verified
 
-| Site                                               | Finding                                                                                                                                                                        |
-| -------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `packages/cli/src/chat/tui/state/approvalQueue.ts` | Imports `Effect` (line 15) yet settles `HostRequest`/`HostReservation` through a raw `new Promise<ApprovalDecision>((resolve) => …)` latch (line 510) instead of a `Deferred`. |
-| `src/tools/goal/goalStore.ts`                      | Imports `Mutex` from `async-mutex` (line 1) _and_ `Effect, Fiber, Stream` (line 2) — two concurrency systems in one file; `indexMutex` wants an Effect Semaphore.              |
-| `config/ratchets/effect-migration-baseline.json`   | Carries a stale `import:p-timeout` row for `src/telemetry/UsageLogService.ts`, which no longer imports it. A free shrink — see the baseline note below.                        |
+| Site                                            | Finding                                                                                                                                                           |
+| ----------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `src/tools/goal/goalStore.ts`                   | Imports `Mutex` from `async-mutex` (line 1) _and_ `Effect, Fiber, Stream` (line 2) — two concurrency systems in one file; `indexMutex` wants an Effect Semaphore. |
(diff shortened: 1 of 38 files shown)


LionSR:
Do these opportunities if nobody is

PR #11986, opened 2026-09-06
refactor(tools): keep request programs native to Effect

LionSR:
## Summary

Request timeout and retry operations now accept Effect programs. All six callers in web fetch, web search, Loogle and Zotero use that contract; the Promise-only callback and its arity guard are removed together.

Web fetch reads response bodies with Effect Stream, preserving byte limits, charset fallback and the final decoder flush. Each attempt owns the complete response lifetime. Web search combines headers and JSON decoding at one foreign HTTP boundary; Zotero keeps body-only error recovery separate from request failure classification. The existing tool `execute()` Promise boundary remains; converting the full tool dispatcher is outside this request-interface change.

## Issue acceptance gates

The request-program portion of R1, R5 and R6 is complete: every production caller is converted, interruption reaches foreign I/O, and each attempt owns its resources. Expected failures, defects and interruption remain distinct. Retry policy and output behavior are preserved. Zotero writes retain the incoming cancellation fence and deadline; this does not promise a durably known remote outcome after interruption.

## Single source of truth

`src/tools/timeouts.ts` owns deadlines, transient retries and request-error classification. Callers own the foreign HTTP operations and response validation.

The callback contract and manual `readBodyWithLimit` loop are deleted. No overload, adapter, service or new runtime entry is added. Stream reading and release use the existing Effect implementation.

## Net elements (R6)

- Files: 9 modified; none added or deleted. Total +224 / −234 lines, net −10.
- Production: +178 / −170 lines, net +8. Tests: +46 / −62 lines, net −16. Ratchet: 2 lines deleted.
- Exported symbols, classes, interfaces and enums: none added or deleted. The two existing request functions accept Effects in place of callbacks.
- Private helpers: `readBodyWithLimit` deleted. No dependency changes or baseline widening.

## Consumer counts (R8)

Six production request sites in four modules are converted: one each in WebFetchTool, WebSearchTool and LoogleTool, and three in bbtClient. The shared retry operation also calls the timeout operation internally. The deleted body reader had one caller. No event emission or subscription interface changes.

## Host impact

Extension, desktop and CLI retain the existing shared tool execution boundary and cancellation input. No host-specific behavior changes.

## Scheduled refactoring

None for this request-interface conversion. The larger native runtime and tool migration remains incomplete and is not claimed by this PR.

## Validation

Current head `8579a90003983f982d39d4fa59d6a7552b42456a` includes current main `7ee6c20382`, preserving both sets of Effect ratchet decreases.

- Formatting, all six typechecks, safe build, full lint, and Effect/browser/guidance/dead-code guards pass.
- Full suite on this source: **9,089 tests passed, 5 skipped; 763 files passed, 1 skipped**, with four workers.
- The existing cancellation cases prove both the independent deadline and deferred caller interruption. The retained assertion distinguishes an abort before completion from scope cleanup after completion.
- All three review threads are resolved after a fresh check. Current-head CI passes every applicable job (run 34096433772), including static checks, both kernel shards, build artifacts and macOS smoke. The PR is mergeable against current main.

Review verification: pinned Effect 4.0.0-rc.112 `Stream.runCollect` returns a mutable `Array<A>` (`node_modules/effect/src/Stream.ts:18584`), so appending the decoder flush is valid. Zotero intentionally preserves separate status/body handling so only body parse failures recover to a status message; transport failures retain their existing classification.

### DIFF
--- a/config/ratchets/effect-migration-baseline.json
+++ b/config/ratchets/effect-migration-baseline.json
@@ -197,8 +197,6 @@
       "src/tools/arxiv/ArxivSearchTool.ts": 1,
       "src/tools/github/PollingSourceBase.ts": 1,
       "src/tools/lean/direct/leanServer.ts": 1,
-      "src/tools/web/WebFetchTool.ts": 2,
-      "src/tools/zotero/bbtClient.ts": 1,
       "src/utils/files/relativeFS.ts": 2
     }
   },
--- a/src/test-kernel/tools/Cancellation.vitest.ts
+++ b/src/test-kernel/tools/Cancellation.vitest.ts
@@ -82,7 +82,7 @@ describe('rateLimitedApiCall cancellation', () => {
 /**
  * `callZoteroConnector` runs its non-idempotent `saveItems`/`saveSnapshot`
  * write under `Effect.uninterruptible` so cancelling a run cannot tear it
- * mid-request and lose the outcome. That rests on one non-obvious property of
+ * mid-request. That rests on one non-obvious property of
  * the effect runtime: the region defers the *caller's* interrupt but not the
  * deadline, because `timeoutOrElse` races through `raceAllFirst`, which forks
  * both racers interruptible regardless of the enclosing region. An effect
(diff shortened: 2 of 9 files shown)


LionSR:
Can Tool become an effect program

Issue #11868, opened 2026-09-04
Tracking: finish Effect runtime ownership and direct composition


LionSR:
## Current status — 2026-09-15

Checked against main `2c9898445d809b98b038ca9b133b16b57ef7794a` on 2026-09-15. Current status below supersedes old counts, branch reservations, and implementation-stage descriptions.

Parent: #12025. This is the canonical runtime implementation tracker.

## Landed

- [x] Durable RunLedger/foldRunState foundation (#12287; #12135).
- [x] Both ledger loops and old flow-engine retirement (#12314; #11964).
- [x] Native model-handler replacement (#12320) and Effect tool execution (#12337).
- [x] Durable request/child vocabulary (#12329), follow-up rows/seeded Queue (#12446), and run-file lease deletion (#12494).
- [x] Runs session service (#12451).

## Remaining

- [x] #12425 — Requests coordination, SDK composition and independent session/request registry retirement.
- [x] #12421/#12433 — explicit project/configuration/filesystem/session ownership across fiber suspension and host callbacks.
- [ ] #12070 — internal collaborators still reached through Promise wrappers, including prompt/model configuration, reflection output and usage recording; convert actual I/O directly and keep pure work synchronous.
- [x] #12253 — concrete runtime defects; #12442 covers child admission racing parent stop.
- [x] #12427 — reconcile shipped recovery with the design and establish consequential reopen/stop/performance evidence.

The implementation is the existing ordinary Effect loops and ledger. The rejected workflow-engine proposal #12081 is not another pending implementation. No legacy importer, graph cursor, second run engine or intermediate compatibility interface belongs in this work.

<details>
<summary>Historical issue description (superseded status; retained for evidence)</summary>

Tracking record for moving the agent runtime onto Effect as a ledger loop. Source: `docs/proposals/2026-09-04-agent-runtime-on-effect.md`, landing with the decision-doc PR #11843. This program is stage 5 and lane D of the persistence substrate cutover (#11867); it is not a separate release. Related to the one-fold program #11861 (not a parent): the session fold reads the same `event` table, and the `flow.transition` arm in the session fold is this program's PR 3.

## Shape (proposal sections 0 and 2)

[...]

PR #11989, opened 2026-09-06
feat: connect native Effect diagnostics to host logging

LionSR:
## Summary

Native Effect logs and named spans now reach the existing host logger. Debug mode shows span duration, parent identity and bounded scalar attributes; ordinary logs retain their levels and secret redaction. No remote exporter or second execution history is introduced.

Session operations provide only their owned session service, preserving a caller-supplied tracer, logger and other services.

## Issue acceptance gates

Advances the diagnostic-context work in #11868. The process installs the diagnostic layer, and the existing logger suite verifies native logging, nested spans, redaction and debug-mode behaviour. This does not complete run-context propagation or the broader runtime migration.

## Single source of truth

`src/logger/effectDiagnostics.ts` defines the native logger and local tracer. `logUtils.ts` remains the output and redaction owner; `sessionLayer.ts` installs the layer at process construction. Durable `AgentTrace` events retain their existing owner.

Adds one diagnostic layer and one bounded span implementation. It retains no span registry, span-event history or links, and introduces no new runtime entry or Promise facade.

## Net elements (R6)

Three files changed, including one new production file: **+142 / −3 lines (net +139)**. Exported declaration lines: +1 / −0 (`effectDiagnosticsLayer`). Classes: +1 / −0 (`DiagnosticSpan`); interfaces and enums: no change. One test is added to the existing logger suite to protect diagnostic output and secret redaction.

## Consumer counts (R8)

One production consumer installs `effectDiagnosticsLayer`. Four process initializers reach that composition: extension, desktop, CLI and agent SDK. No application event publisher or subscriber is replaced.

## Host impact

Extension: native logs use the existing output channel; span summaries require debug mode.

Desktop: the same diagnostic policy uses its existing host sink.

CLI: the same policy uses its existing logger. Effect SDK callers retain their supplied services.

## Scheduled refactoring

Broader run identity and lifetime conversion remains tracked in #11868; no temporary compatibility path is added.

## Validation

- `npm run format`
- `npm run typecheck` (all workspace packages)
- `npm run lint`
- `npm run compile:fast`
- `node scripts/check-effect-migration-ratchet.mjs`
- `git diff --check`
- Full suite on the combined diagnostics/HTTP tree: **9,089 passed, six skipped**.
- Diagnostics-only full suite: **9,087 passed, two failed, six skipped**. The two failures are `DesktopDiffHost` assertions at lines 341 and 495 comparing the shared temporary-directory list while concurrent suites added or removed matching directories. All **16 tests in that file passed** when rerun separately. No other failures occurred.

### DIFF
--- a/src/auth/authProgram.ts
+++ b/src/auth/authProgram.ts
@@ -7,6 +7,7 @@
  * sanctioned boundary, not here.
  */
 import { Cause, Data, Deferred, Effect, Exit, Option, Semaphore } from 'effect';
+import type { HttpClient } from 'effect/unstable/http';
 
 /**
  * A host port (secret storage), an SDK call, or a provider policy rejected.
@@ -111,7 +112,7 @@ function rethrowPortCause(error: unknown): never {
  * `Effect.run*` call itself in boundary code (PRD R1).
  */
 export type AuthProgramEdge = <A, E>(
-  program: Effect.Effect<A, E>,
+  program: Effect.Effect<A, E, HttpClient.HttpClient>,
 ) => Promise<Exit.Exit<A, E>>;
 
 let authProgramEdge: AuthProgramEdge | null = null;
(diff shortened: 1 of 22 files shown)


LionSR:
we should use native Effect logging as much as possible

PR #12005, opened 2026-09-07
feat: bound conversation readers and scope SDK transcript residency

LionSR:
A paused conversation view now stops its own source read while other views continue. Each reader has acknowledged delivery, bounded frame/replay storage, and scoped transcript interests. Reconnect uses the last published view; oversized histories produce an explicit error without changing saved content. Late SDK readers wait for transcript hydration before terminal completion, and automatic interests end when the run finishes.

The existing bridge, framer, and fold remain the delivery path. Frames target 256 KiB with a separate 16 MiB frame limit; replay is capped at 128 MiB / 1 million rows, and SDK trace delivery at 512 events / 8 MiB. A retained row above the batching target gets its own frame. Each attachment owns a unique interest key, so an old close finalizer cannot clear a reopened port’s interest.

This now composes directly with main’s SQLite event plane. Public SQL reads apply the optional display budget before decoding or retaining each row. SQL selects the requested aggregate suffix and compacts listing facts before accounting. `readInputBatch` keeps the event prefix, high-water cursor, and existence/claim state in one transaction. Intermediate frames are acknowledged, but the decoder publishes only complete replay/drained batches. The canonical owner reads remain unrestricted. Unselected live text is excluded before traversal and retention; selected text charges its full envelope and row count.

Main’s retired in-memory log stays deleted. The obsolete reader-only file parser, spill budgets, and their fixtures are removed. There is no second event store or fold authority.

Validation:

- At `c3c21b3f9f4105b4afc71c132b33a823a2030a3b`: formatting, all six typechecks, safe build, full lint, and dead-code/Effect/browser/guidance guards pass. **8,888 tests pass, 5 skip; 761 files pass, 1 skip.** The four conflict resolutions also pass 59 focused cases and independent source review. No baseline is widened.
- The existing SQL read regression checks both budget dimensions on all five public read paths, verifies that excess rows are not decoded, and confirms unchanged complete owner reads afterward.
- Existing scenarios cover independent paused/healthy readers, close/reopen cleanup, SDK trace overflow and late hydration, incomplete replay recovery with pending requests, complete SQL batch publication, selected live-text bounds, and oversized retained rows without loss.
- The documented September 7 SQLite-composed measurement (`node scripts/measure-session-readers.mjs`) retained all 100,000 synthetic rows in 391 frames. During a paused-reader interval, source pulls remain at 257 while the healthy reader delivers 100,000 rows; one frame stays outstanding and settled heap changes by 316,920 bytes. Source hashes, full measurements, and limits are recorded in `docs/architecture/2026-09-07-bounded-session-readers.md`. These observations include canonical/source memory and are not a process-memory ceiling.
- The final SQL integration received an independent source review. Current-head remote CI and reviews are tracked separately below.

Head `c3c21b3f9f4105b4afc71c132b33a823a2030a3b` preserves the external `cb1240f38b` head and main `13218da867` through a merge commit. Main’s native SDK launch and deletion/ownership changes are retained. All 13 review threads are resolved in the final current-head check. All applicable CI jobs passed in run [34177818236](https://github.com/LionSR/TeXRA/actions/runs/34177818236); the PR is mergeable and ready for review. No PR merge is requested or performed.


LionSR:
Is this a bit over engineered

PR #12008, opened 2026-09-07
feat(editor): apply VS Code LM token counts to max_tokens

LionSR:
## Summary

Keep the editor-supplied language-model token-counting path and actually use it.

`ModelHandlerVscodeLm` now runs `applyTokenCountLimit` before `sendRequest`, using the existing `LanguageModelPort.countTokens` estimator. Reduced output budgets are passed as `maxTokens` on the host-neutral request options and mapped to Copilot's `modelOptions.max_tokens`.

This replaces the previous approach of deleting the unused counting API and advertising `supportsTokenCounting: false`. The VS Code `countTokens` API remains, and Copilot requests now get a real max-token cap instead of an advertised-but-unwired capability.

## Review follow-up

Addresses the request to turn counting on rather than retire it. Counting failures still follow the shared soft-failure envelope (proceed with the unreduced budget unless the prompt already exceeds the context window).

## Test plan

- [x] Existing VS Code LM handler and port suites
- [x] New handler case: counted prompt leaves no output room → `sendRequest` gets reduced `maxTokens`
- [x] Port adapter maps `maxTokens` to `modelOptions.max_tokens`

### DIFF
--- a/packages/extension/src/frontend/lm/createLanguageModelPort.ts
+++ b/packages/extension/src/frontend/lm/createLanguageModelPort.ts
@@ -143,6 +143,9 @@ function toVscodeOptions(
       inputSchema: tool.inputSchema,
     })),
     toolMode: toVscodeToolMode(options.toolMode),
+    ...(options.maxTokens == null
+      ? {}
+      : { modelOptions: { max_tokens: options.maxTokens } }),
   };
 }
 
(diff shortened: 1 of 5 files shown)


LionSR:
Shouldn't we turn that on in stead

LionSR:
Turned counting on instead of retiring it.

`ModelHandlerVscodeLm` now calls `applyTokenCountLimit` before `sendRequest`, and the reduced budget is forwarded as `maxTokens` → Copilot `modelOptions.max_tokens`. The port `countTokens` path stays; `supportsTokenCounting` remains true.

PR #12049, opened 2026-09-07
fix(cli): remove the sixth hostPort copy and restore the auth port error contract

LionSR:
## What was broken

Four defects that shipped with the recent CLI typed-recovery work.

**1. A sixth copy of `hostPort`, propping up a tightened baseline.** `packages/cli/src/chat/chatSessionController.ts` defined its own `tryPromise` with the identity catch mapper. `src/common/hostPort.ts` already is that helper, and its docstring records that five copies had been consolidated into it. Worse than the duplication was what the copy was used for: it wrapped five whole Promise chains of this file's own code in a single boundary crossing, and the file's `catch:effect-importer` ratchet row was lowered from 7 to 1 on the strength of that wrapping. Since baselines never widen, the file could no longer take an honest raw catch while still holding the chains the row was supposed to have retired.

**2. `rethrowPortCause` broke the contract its own class documents.** `AuthPortError`'s docstring promises that the Promise edge re-throws the port's own error unchanged, "so every `instanceof` and message check a host makes still holds". The function had been changed to re-throw `ensureError(error.cause)`, which turns a non-`Error` cause such as `{status: 401}` into `new Error('[object Object]')`. This is shared code: 33 `runAuthProgram` call sites take that default rethrow, spread across `packages/extension/src/frontend/auth/SupabaseAuthProvider.ts` (27 port/program sites), `src/auth/SupabaseSession.ts`, `src/auth/SupabaseClient.ts`, `src/auth/oauth/SubscriptionOAuthCoordinator.ts`, `packages/desktop/src/main/desktopSupabaseAuth.ts` and the CLI. The blast radius reaches all three hosts, not just the CLI.

**3. An architecture-test exemption admitted on a false premise.** `BARE_EFFECT_RUN_SITES` in `src/test-kernel/architecture/dependencyDirection.vitest.ts` allowed one bare `Effect.runPromise` in `cliConfig.ts` because "`buildCliContext` resolves `.texra/config.json` and the user approval policy BEFORE `initCliPlatform` ... so no process runtime exists to borrow". That is true of one caller. `packages/cli/src/runtime/agentRoster.ts` calls `effectRuntime().runPromise(loadAgents(...))` and then `loadWorkspaceCliConfig(cwd)` on the next line, and `resolveChatDefaults` is reached only from `runChatTui` (after `effectRuntime().runPromise(loadAgents())`) and from `orchestrate`'s `canLaunchWithDefaultModel` (after `initInteractiveCliPlatform`). Both had a runtime to borrow and did not use it. This allowlist is the only artifact gating new bare runs, so a false justification in it is worse than no entry.

**4. A ratchet-invisible silent swallow.** `notifyCliUpdate` used `installCliProcessRuntime().then(() => true, () => false)` and returned on false. Reading `scripts/check-effect-migration-ratchet.mjs` confirms the counter only sees `ts.isCatchClause` nodes and `.catch(` property calls, so a two-argument `.then` is invisible to it: the file's row could be deleted while an error-discarding handler survived.

## What changed

**The controller.** The local `tryPromise` is gone. Each of its five call sites was judged against the line `src/common/hostPort.ts` draws between a foreign Promise edge (where the identity catch belongs) and our own code (where wrapping is a facade):

- `startRootRun`, `resume`'s `attemptResume`, `tryResumeStream`'s `attempt` and `startSession` were our own composition and are now `Effect.gen` programs. `hostPort` appears only at the genuine Promise leaves: `runAgent`, the execution-store reads, `lookupStreamExecutionId`, `setCliHelperModel`, `selectCliRunnableModel`. `snapshotStore.preload` and `resumeRun` are already Effects, so they are yielded directly instead of being run on a nested `runPromise`.
- The best-effort settle-wait on `session.runPromise` is a foreign promise handed to us, so it keeps the identity catch and simply calls `hostPort`.

One shared `recoverRun` combinator replaces the four `Effect.catch` tails. `Effect.catch` recovers typed failures only, but these bodies used to be `try`/`catch` blocks that also caught a throw from the imperative code around the awaits. `recoverRun` uses `Effect.catchCause` and hands the recovery the original value through `Cause.squash`, so a failure and a defect are treated exactly as before, while interruption keeps propagating rather than being reported as a run failure.

Converting `startRootRun` surfaced a real ordering bug. A fiber calls `runAgent` synchronously, so the run could resolve and publish its stream before `session.markRunPending` ran in the caller's synchronous tail, and `markRunPending` resets `session.streamId`. The old `Promise.resolve().then(...)` hop hid that by accident. `startRootRun` now claims the root-run slot through a `pDefer` before the program starts, which is the pattern `resume` and `tryResumeStream` already document and use, and forwards the chain's settlement to the claimed promise so exit-drain still awaits the real run.

**The auth edge.** `rethrowPortCause` throws `error.cause` again, with a comment saying why it must not go through the unwrap helper. The docstring is untouched, because the contract is the thing worth keeping. `unwrapAuthPortCause` stays exported: it still has three consumers, all CLI `Effect.mapError` sites in `supabaseAuth.ts` and `supabaseAuthCallbackServer.ts`.

**The config readers.** `loadWorkspaceCliConfig` and `loadUserApprovalPolicy` are Effect programs now. `readCliAgentRoster` and `resolveChatDefaults` settle them on `effectRuntime()` like any other program, so they inherit the process runtime's layers instead of spinning up a bare one. The single remaining bare run is `loadCliStartupConfig`, whose only caller is the genuinely pre-runtime `buildCliContext`, and the allowlist comment now says exactly that. `loadUserApprovalPolicy` is no longer exported, since `buildCliContext` was its only external caller and exports are contracts.

**The update check.** The two-argument `.then` is gone; the install is plainly awaited. Both callers of `notifyCliUpdate` reach `initCliPlatform` moments later and await the same install with no handler at all, so absorbing the rejection here only moved the identical crash a few statements down while hiding the cause. The check's own best-effort silence is unchanged: it is the `Effect.ignoreCause` around the part that actually runs on the runtime.

## What was verified

- `npm run typecheck`: exit 0 across workspace, test-kernel, agent, cli, trace-viewer and desktop.
- `npx vitest run src/test-kernel/cli/ src/test-kernel/architecture/`: 150 files, 2282 passed, 1 skipped.
- `npx vitest run` (whole suite): 761 files passed, 1 skipped; 8945 tests passed, 6 skipped.
- `node scripts/check-effect-migration-ratchet.mjs`: every one of the twelve rows equals the baseline, "@adapter-until markers OK: none present", "Effect migration ratchet OK: no count grew, no baseline headroom". No baseline JSON was edited.
- `npm run lint`: exit 0.
- `npm run check:dead-code-ratchet`: "Dead-code ratchet OK: no new findings".

One post-commit run of the two suite directories showed five 10-second timeouts in `ResumeCommand`, `StaticBandResize`, `TranscriptSessionPolicy` and `WorkflowRunCommand` while the machine load average was 63. Those four files pass in 3.5 seconds on their own, the same directory command passed 150/150 both before and after, and the full suite is green, so they are load-induced timeouts rather than a regression.

## What was deliberately left alone

- The `AuthPortError` docstring, and `unwrapAuthPortCause` itself, which still earns its keep at three CLI mapError sites.
- The raw catch in the controller's synchronous dispose callback, which is the file's one allowed entry and cannot return an Effect.
- Tests: no new ones. The three suites that touch the config readers follow their signature change, and `InitConfig.vitest.ts` gets shorter because it can yield the reader directly instead of wrapping it in `Effect.promise`.

https://claude.ai/code/session_012ha5vnpy67jRyC369MtxZU

### DIFF
--- a/packages/cli/src/chat/chatSessionController.ts
+++ b/packages/cli/src/chat/chatSessionController.ts
@@ -3,7 +3,7 @@
 // Host-neutral (no Ink/TUI rendering dependencies): the Ink component
 // consumes narrow commands exposed here.
 
-import { Effect, Option, Stream, SubscriptionRef } from 'effect';
+import { Cause, Effect, Option, Stream, SubscriptionRef } from 'effect';
 import pDefer from 'p-defer';
 import PQueue from 'p-queue';
 
@@ -42,6 +42,7 @@ import {
   runOutcomeExitCode,
   type TurnOutcome,
 } from '@cli/runtime/terminalStatus';
+import { hostPort } from '@common/hostPort';
 import { hasErrorPresentationClaimed } from '@common/errors/sdkError/errorMetadata';
 import type { RunModelDecisionReason } from '@model/runModelDecision';
 import type { DisposableStore } from '@platform/disposable';
@@ -132,10 +133,25 @@ interface AutoResumeOptions {
 // Controller
 // ---------------------------------------------------------------------------
 
-/** `Effect.tryPromise` with the identity catch every Promise boundary below
- *  wants: the rejection value flows through unchanged as the error. */
-const tryPromise = <A>(run: () => Promise<A>): Effect.Effect<A, unknown> =>
-  Effect.tryPromise({ try: run, catch: (error) => error });
+/**
+ * The recovery tail every run/resume program below shares. A typed failure
+ * from a `hostPort` leaf and a throw from the imperative body alike reach
+ * `recover` with the original value, which is exactly what the `try`/`catch`
+ * around these bodies did before they became programs. Interruption is not
+ * folded in: a fiber the runtime is tearing down is not a run failure to
+ * report, and the caller's own settlement still sees it.
+ */
+const recoverRun = <A, E, R>(
+  program: Effect.Effect<A, E, R>,
+  recover: (error: unknown) => A,
+): Effect.Effect<A, E, R> =>
+  program.pipe(
+    Effect.catchCause((cause) =>
+      Cause.hasInterrupts(cause)
+        ? Effect.failCause(cause)
+        : Effect.sync(() => recover(Cause.squash(cause))),
+    ),
+  );
 
 /**
  * Narrow commands the chat-session controller exposes to the Ink component.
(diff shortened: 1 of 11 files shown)


LionSR:
Took the review suggestion: added one regression test extending `src/test-kernel/auth/SupabaseSessionLifecycle.vitest.ts` (no new file), asserting that a port rejecting with a non-Error value re-throws that exact value.

Verified it pins the regression rather than passing vacuously: restoring the broken `unwrapAuthPortCause(error)` rethrow fails it with a reference-equality assertion error; the fix passes it. 26 tests in the suite pass, root `tsc -p tsconfig.json` clean.

Left the ordering-fix comment as informational, since it describes the change rather than asking for one.

https://claude.ai/code/session_012ha5vnpy67jRyC369MtxZU

Issue #12076, opened 2026-09-07
Telemetry on Effect: finish the usage pipeline, and add the Metric/Tracer instrumentation the completion gates require (tracking)


LionSR:
Lane tracker under #12025 and #12074. Two distinct pieces of work that both sit under "telemetry", and conflating them has hidden the second one.

## Piece 1 — the usage pipeline is nearly Effect-native already

`src/telemetry/UsageLogService.ts` (549 lines) already imports `Clock, Data, Duration, Effect, Exit, Fiber, Queue, Scope, Semaphore` and models its failures as `Data.TaggedError('UsageBatchUndelivered')`. This is not a Promise subsystem awaiting conversion — it is one of the more Effect-idiomatic files in the repository.

What is left is small and specific:

- [x] `ky` for delivery → `HttpClient` (`effect/unstable/http`) or `Undici` (`@effect/platform-node`). It is one of the 28 A1 foreign edges in W7 (#12073). Done in #12086.
- [ ] `randomUUID` from `node:crypto` → `Crypto`/`Random`, only if determinism under test is wanted; otherwise leave it.
- [ ] `export const UsageLogService = new UsageLogServiceImpl()` is a **module-level singleton**, not a `Layer`. Its lifetime is the process, its dependencies (`SupabaseClient`, `ConfigProvider`, `workspaceRoots`) are ambient reads. That makes it the same shape #12079 is collapsing elsewhere, and it is why `platform()` and `workspaceRoots()` appear inside it.
- [ ] `unrefSleepClock` — a hand-rolled unref'd timer — against `Clock`/`Effect.sleep` semantics.

## Piece 2 — there is no instrumentation at all, and a completion gate depends on it

Measured on `1cf1a84`: **zero** `Metric.` uses in production, **zero** OpenTelemetry, and `Tracer` appears in exactly two files (`effectDiagnostics.ts`, which routes spans into the string logger, and `leanServer.ts`).

Meanwhile #12025's **Performance** completion gate requires: cold open, idle memory, two-paper concurrency, event-loop delay under SQLite contention, p95 commit/stop latency, replay memory, and bytes written as history grows. **Nothing in the repository produces any of those numbers today.** #12068 records a pre-cutover baseline by measurement harness, which is the right way to get a one-off number and the wrong way to get a p95 under contention in a shipped build.

Every Effect program in the repo already emits spans; they are simply discarded. `Metric` is free-standing and does not require an exporter to be useful in tests.

[...]

Issue #12073, opened 2026-09-07
Rulings requested: five decisions blocking the Effect migration tail


LionSR:
Consolidated decision issue under #12025. Five questions no lane can answer for itself, each blocking a named number of sites. The long-form briefs live in #12025's comments; this issue is the list, so that "what is blocked on a ruling" is one lookup rather than a comment scroll.

Measured at `1cf1a84`: the migration has 210 remaining sites. **Rulings below gate at least 59 of them**, plus the terminal form of two rows.

> **Correction, 2026-09-09 (after #12108 and the accepted [TeXRA 1.0 direction](https://github.com/LionSR/TeXRA/blob/main/AGENTS.md#texra-10-direction)).** The fresh-state cutover landed between this brief and now: #12108 merged execution metadata into SQLite session records on `main`, the official Effect SQL client is adopted (`@effect/sql-sqlite-node` and `@effect/platform-node` at `4.0.0-rc.112` are dependencies on `main` — the R-1 evidence below saying the pin "is not installed" is stale), and `src/platform/defaults/jsonStore.ts` is Effect-native there. The 1.0 [implementation plan](https://github.com/LionSR/TeXRA/blob/main/.agents/docs/proposed/architecture/2026-09-09-texra-1-0-implementation-plan.md) §5 also bounds every ruling below: do not fund another round of Effect conversions inside the retiring JSON store (its application-state uses), the file lease implementation, the persisted graph, or the old model handlers merely to reduce a count of Promise calls — convert durable consumers to their final services; and no dual systems, time-boxed bases, or temporary replacement systems. Row counts have moved with the cutover churn (`platform()` is 111 sites in 52 files at `dff2602689`); re-measure at HEAD before implementing any ruling.

## 1. Terminal row form — what is a row's end state when it has a permanent resident?

The completion criterion in #12025 is "the migration is done when every row is zero and deleted". Two rows can never satisfy it.

[...]

PR #12086, opened 2026-09-08
refactor(telemetry): send usage batches through Effect's HttpClient

LionSR:
## Summary

`UsageLogService` was already one of the most Effect-native files in the repository — it uses `Clock`, `Data`, `Duration`, `Effect`, `Exit`, `Fiber`, `Queue`, `Scope` and `Semaphore`, and models failure as `Data.TaggedError('UsageBatchUndelivered')`. `ky` was the last foreign edge in it. This replaces that one edge with `effect/unstable/http`.

This is billing data, so the delivery contract — not the syntax — was the thing under review. Behaviour is preserved deliberately, **with one wire-level exception noted below**:

- **The body read stays inside the timed effect.** `ky`'s `timeout` only guarded until response headers arrived (it clears the timer once fetch settles), so a server that answered headers and then stalled mid-body would hang the `.json()` read indefinitely, wedging the flush lane and `dispose()`. `HttpClient.withScope` creates the `AbortController` and registers `abort` as a scope finalizer, so `Effect.timeoutOrElse` + `Effect.scoped` aborts the underlying fetch exactly as the hand-passed signal did. `raceAllFirst` awaits the loser's unwinding, so the abort is prompt.
- **A non-2xx status is still not a failure.** The endpoint answers a rejection with a body that the acknowledgement checks read, so status handling stayed below the error channel.
- **`Effect.catchTag('UsageBatchUndelivered')` still matches**, because `mapError` collapses before the timeout wrapper.
- **No hidden retry layer was removed** — `ky`'s default retry does not cover POST.
- **The opt-out path is untouched.** `TELEMETRY_ENABLED_KEY` / `usageLoggingOptOut()` gate the pipeline as before; all six consent tests pass.
- **No new event-loop holder** — setting `scopedController` skips both `responseRegistry.register` branches.

One deliberate improvement: a transport failure now reports the wrapper's message *and* its cause. The wrapper names the request that failed; only the cause names the refused connection or reset socket, and a billing warning is worth both.

The scope finalizer runs with no context of its own, so the shutdown drain is explicitly given the client this initialization received.

### The one behaviour change reviewers should decide on

`HttpClient.TracerPropagationEnabled` is a `Context.Reference` whose default is `constTrue` (`unstable/http/HttpClient.js:678`), and TeXRA installs a real `Tracer` via `effectDiagnosticsLayer` (`sessionLayer.ts:727`). **The POST therefore now carries trace-propagation headers to the Supabase usage endpoint that the `ky` version did not send.**

**Correction after review:** an earlier version of this description said "`traceparent` headers". `HttpTraceContext.toHeaders` emits **both `b3` and `traceparent`** (`HttpTraceContext.js:26-32`) — two headers, not one.

Functionally harmless — the server ignores unknown headers — but it exports internal trace/span ids to that endpoint. I did not suppress it, because doing so means adding a line that nothing in this change's validation covers. If exact prior wire parity is wanted, it is one service on the send effect: `Effect.provideService(HttpClient.TracerPropagationEnabled, false)`.

**The bearer token is not exposed.** `HttpClient.make` copies request headers into span attributes, but `Authorization` lands as `<redacted>` because `Headers.CurrentRedactedNames` defaults to `["authorization", "cookie", "set-cookie", "x-api-key"]` (`Headers.js:329`). The exported ids are not credentials. This was independently confirmed in review; an override of that Reference would nonetheless be security-relevant.

## Issue acceptance gates

**#12076, piece 1.** Three of its four items are satisfied; two are answered as negative results and one is deliberately deferred.

- [x] `ky` → `HttpClient`.
- [x] `ky` may **not** be removed from `package.json` — nine other importers.
- [ ] **`unrefSleepClock` must stay — negative result, do not attempt.** Effect's `sleepMillis` uses a ref'd `setTimeout`, and there are zero `unref` calls in `effect/dist/internal/effect.js`. The repo already pins this with a `handle.hasRef() === false` assertion. Converting it would keep the Node event loop alive, and a CLI that will not exit is a severe regression.
- [ ] **`randomUUID` → `Crypto`/`Random` — not done, deliberately.** `Crypto` is a no-default `Context.Service`; converting would widen `ProcessRuntime`'s requirements to buy test determinism the repo already gets by injection.
- [ ] **Singleton → `Layer` — deliberately out of scope.** It reaches `packages/extension/src/extension.ts`, which `cutover/persistence-substrate` and `stage/6-execution-registration` are both rewriting. Recorded as follow-up on #12076.
- [ ] **Piece 2 (`Metric`/`Tracer` instrumentation) — untouched.** Separate lane.

## Single source of truth

`HttpClient.HttpClient`, taken from context, is now the single owner of outbound HTTP for this file. It replaces a module-level `ky` import that resolved `globalThis.fetch` per call. The delivery decision (delivered / undelivered-retry / undelivered-give-up) remains owned by `UsageLogServiceImpl`'s acknowledgement checks, unchanged.

No abstraction added. One dependency edge removed from this file. The `Effect.tryPromise` wrapper that adapted `ky`'s promise into the error channel is deleted outright rather than replaced — the client's typed `HttpClientError` reaches `mapError` directly. No pass-through layer.

## Net elements (R6)

- Files: 2 changed, +59 / −29 across two commits (net **+30** lines; the growth is comment, explaining the timeout/abort contract that was previously implicit in `ky`'s options, plus the empty-reason guard below).
- Exported symbols: `^[+-]export` over the diff = **0 added, 0 removed**.
- Classes / interfaces / enums: **0 / 0 / 0**.
- Dependencies: none added. `ky` retained (nine other importers).

## Consumer counts (R8)

`dispose()`'s signature widens from `Effect<void>` to `Effect<void, never, HttpClient.HttpClient>`. Grepped consumers of `UsageLogService.dispose()` / `.initialize()`:

- **6 production call sites**, all of which already run through `effectRuntime()`: `packages/cli/src/runtime/initPlatform.ts:386,401`, `packages/desktop/src/main/platform/index.ts:177,185`, `packages/extension/src/extension.ts:514,635`.
- **Zero host edits were required**, because `ProcessRuntime` is already `ManagedRuntime<HttpClient.HttpClient, never>`. This is the first production datapoint for open ruling **R-2** (do Platform capabilities get Effect `Context.Service` keys) and is recorded on #12073.
- No emit path or subscription key changed.

## Host impact

**Extension:** none. `extension.ts` calls `initialize`/`dispose` through `effectRuntime()`, which already provides the client. Not edited.

**Desktop:** none. Same, via `platform/index.ts`. Not edited.

**CLI:** none. Same, via `initPlatform.ts`. Not edited. The shutdown ordering comment at `initPlatform.ts:103` (lifecycle shutdown before the NDJSON stdout flush) still holds — the drain now runs against a client captured at initialize time, which is inert for the stateless `FetchHttpClient` but would matter if that layer were ever swapped for a pooled client.

## Scheduled refactoring

- #12076 — piece 1's singleton→`Layer` conversion, deferred until `extension.ts` is free.
- #12076 — piece 2, the `Metric`/`Tracer` instrumentation that #12025's Performance completion gate requires. No instrumentation currently produces those numbers in a shipped build.
- #12073 — the `TracerPropagationEnabled` and `Headers.CurrentRedactedNames` observations are recorded there as R-2 evidence.

## Validation

Run by me on a clean tree, at `1cf1a84` for the first commit and re-run in full after the second:

- `npm run typecheck` — **exit 0**, all legs (root, agent, cli, trace-viewer, desktop, extension).
- `npx vitest run src/test-kernel/telemetry` — **35 passed (35)**.
- `npx eslint src/telemetry/UsageLogService.ts` — clean. `prettier --check` — clean.
- `node scripts/check-effect-migration-ratchet.mjs` — **green**; no count grew, no baseline headroom, no `@adapter-until` marker. Baseline byte-identical.
- Pre-empting CI: ran the **12 other suites that reference `UsageLogService`, `testHttpClientLayer` or `FetchHttpClient`** — 26 files, **252 tests, all passing**.

**The two files in the first commit must land together.** `tsc` exits 0 with the production change and the *unmodified* test stub, so typecheck alone does not catch the mismatch — splitting them merges a red build (22 of 35 failing). They are one commit for that reason.

Not run: full `npm test`, `npm run lint` on the whole tree.

Provenance: implemented in an isolated worktree, reviewed by two independent adversarial passes (scope/build and semantics), then re-verified by hand before commit. Those reviews corrected three claims from the original implementation — a wrong failing-test count, an overstated typecheck result, and a "land the patch" verdict that would have shipped the production file without its test. The `review` job then found two further items, both verified and one fixed in `995966d` (see below).

### `995966d` — empty-reason guard

Review found that `sendBatch`'s non-`HttpClientError` branch could log a blank reason. `HttpClientRequest.bodyJson` fails with `HttpBodyError`, which `Data.TaggedError` builds from `{ reason, cause }` with **no message** — but it still extends `Error`, so `toErrorMessage` takes the `err instanceof Error` path and returns `err.message`, the empty string. An undelivered billing batch would have been logged with no reason at all.

Verified empirically against the pinned package: such an error gives `message === ""` and `String(err) === "HttpBodyError"`. `extractErrorMessage` returns `undefined` rather than `""`, so the fallback now yields the tag and the reason can never be blank.

Unreachable today — every `UsageLogEntry` field is a JSON-safe primitive — so this is defence for a failure mode that would otherwise be undiagnosable if the wire shape ever changed.

---

https://claude.ai/code/session_01VRKNXQf68EFbEiMrsNboLo

### DIFF
--- a/src/telemetry/UsageLogService.ts
+++ b/src/telemetry/UsageLogService.ts
@@ -28,7 +28,10 @@ import {
   TELEMETRY_ENABLED_KEY,
 } from '@shared/schemas';
 import { CODING_PLAN_SUBSCRIPTIONS } from '@shared/codingPlanSubscriptions';
-import { toErrorMessage } from '@utils/errors/errorMessage';
+import {
+  extractErrorMessage,
+  toErrorMessage,
+} from '@utils/errors/errorMessage';
 import { isEnvFlagEnabled } from '@utils/system/envFlags';
 import { unrefSleepClock } from '@utils/system/unrefSleepClock';
 
@@ -483,7 +486,11 @@ class UsageLogServiceImpl {
         Effect.scoped,
         Effect.mapError((error) => {
           if (!HttpClientError.isHttpClientError(error)) {
-            return undelivered(toErrorMessage(error));
+            // `bodyJson` fails with `HttpBodyError`, which carries `reason`
+            // and `cause` but no message, so `toErrorMessage` renders it as
+            // the empty string — an undelivered billing batch with a blank
+            // reason. Fall back to the tag so this can never be silent.
+            return undelivered(extractErrorMessage(error) ?? String(error));
           }
           // The wrapper's message names the request that failed; only its
           // cause names the transport failure (a refused connection, a reset
--- a/src/telemetry/UsageLogService.ts
+++ b/src/telemetry/UsageLogService.ts
@@ -11,7 +11,11 @@ import {
   Scope,
   Semaphore,
 } from 'effect';
-import ky from 'ky';
+import {
+  HttpClient,
+  HttpClientError,
+  HttpClientRequest,
+} from 'effect/unstable/http';
 
 import { SupabaseClient } from '@auth/SupabaseClient';
 import { SUPABASE_CUSTOM_DOMAIN } from '@auth/config';
@@ -254,7 +258,15 @@ class UsageLogServiceImpl {
       // Run before the sender's interruption finalizer. Closing this child on
       // dispose also removes its parent finalizer, so reinitialization neither
       // retains old shutdown callbacks nor changes the drain-before-stop order.
-      yield* Scope.addFinalizer(lifetime, this.shutdown());
+      // A scope finalizer runs with no context of its own, so the drain it
+      // performs carries the client this initialization was given.
+      const client = yield* HttpClient.HttpClient;
+      yield* Scope.addFinalizer(
+        lifetime,
+        this.shutdown().pipe(
+          Effect.provideService(HttpClient.HttpClient, client),
+        ),
+      );
     }
 
     if (isTelemetryDisabledByEnv()) {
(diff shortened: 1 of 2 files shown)


LionSR:
Both findings verified against the pinned `effect@4.0.0-rc.112`. One is fixed in `995966d`; the other corrects this PR's description.

### Finding 2 — the empty reason. Real, and fixed.

This one is a genuine defect in code this PR introduced, not just a note for completeness. Confirmed empirically rather than by reading:

```
HttpBodyError-shaped Data.TaggedError({ reason, cause })
  → message: ""
  → String(): "HttpBodyError"
```

`HttpBodyError` extends `Error`, so `toErrorMessage` takes its `err instanceof Error` branch and returns `err.message` — the empty string. An undelivered **billing** batch would have been logged with no reason at all. That is precisely the silent-degradation shape CLAUDE.md names: a failure quietly becoming a wrong-but-quiet answer.

Fixed by using `extractErrorMessage`, which returns `undefined` rather than `''`, with `String(error)` behind it — so the reason is now never blank. Your reachability assessment is right (every `UsageLogEntry` field is a JSON-safe primitive), and that is exactly why it was worth fixing: an unreachable path that logs nothing is undiagnosable if the wire shape ever changes.

### Finding 1 — I had the header count wrong

Confirmed: `HttpTraceContext.toHeaders` emits **both `b3` and `traceparent`** (`HttpTraceContext.js:26-32`), not just `traceparent` as this PR's description claimed. The description is corrected, and now says "trace-propagation headers … two headers, not one" with the correction called out rather than edited away.

Your redaction verification is also stronger than what I wrote. I had said the token is safe "only because `Headers.CurrentRedactedNames` defaults to include `authorization`" — you confirmed the actual mechanism, that it lands in span attributes as `<redacted>` with the full default list at `Headers.js:329`. The description now states that, while keeping the note that overriding that Reference would be security-relevant.

**The decision itself remains open and is yours**, which is why I did not suppress it: `Effect.provideService(HttpClient.TracerPropagationEnabled, false)` on the send effect restores exact prior wire parity. I have deliberately not added that line, because nothing in this change's validation covers it and the change is arguably desirable.

### Validation after `995966d`

Full `npm run typecheck` **exit 0** (all legs), `vitest run src/test-kernel/telemetry` **35/35**, eslint and prettier clean on the changed file, migration ratchet green with a byte-identical baseline.

Thank you for the `Scope.addFinalizer` vs `Effect.addFinalizer` check at `internal/effect.js:1627`/`:1702` — that confirms the re-provision in the finalizer is necessary rather than defensive, which is more than I had established.


---

PR #12144, opened 2026-09-09
refactor(storage): drop the v1/v2 execution-lease compatibility shim

LionSR:
The v3 per-token claim protocol shipped in 0.40.4 carried a reader for the
two retired single-file protocols plus a writer that mirrored every claim
as a v2 record, so a 0.40.3 process sharing ~/.texra would back off instead
of claiming beside an owner it could not see. V1 accepts the break, so the
whole shim goes: LegacyLeaseSchema, legacyShadowRecord, legacyLeasePath,
readLegacyRecord, the shadow merge in readClaims and the legacy unlink in
unlinkOwnClaim.

A claim is now exactly one file, so StoredClaim.files collapses to a single
`file` and deleteFiles becomes deleteFile.

Claude-Session: https://claude.ai/code/session_015ypYRFv4NjpGgA3Xtg9toN


LionSR:
Reviewed this while surveying what the 1.0 direction actually authorizes deleting. Three things, in order of consequence.

## 1. The CI failure and its cause are the same finding

`static checks` fails on two `TS2339`s in the Deno edge-function typecheck:

```
Property 'usedRelay' does not exist on type '{ usageRoute: ...; timestamp: string; model: string; ... }'
```

The diff drops `usedRelay` from `supabase/functions/log-usage/usageValidation.ts:37` while `index.ts:151` still reads `entry.usedRelay ?? false` into the `used_relay` column.

That is not just a missed call site. **The edge function is deployed infrastructure serving released clients**, not part of the 1.0 storage transition. Retiring its wire tolerance is owned by #10921, which dates it after 2026-11 and gates it on #10920, the ops drain that is still open and still labelled `priority:p0`. The client-side `usageRoute: 'relay'` removals in `src/shared/schemas/usage.ts`, `src/shared/copy/modelAccess.ts` and `packages/cli/src/runtime/modelAccessRoute.ts` are the same item.

## 2. The dated shims split into two classes, and only one is authorized

The 1.0 direction disapplies the three-month window **for the 1.0 storage transition**: legacy JSON stores, histories and execution checkpoints are simply not read, so their readers can go now. That covers most of this diff, including the persisted checkpoint compatibility readers and the archive-boundary and trace-export shims, and those deletions are welcome.

It does not cover compatibility with software that is still running. Several shims here carry their own `delete after 2026-11-29 / 2026-11-30` comments and serve either the deployed relay endpoint or `release/0.40` clients. Splitting the branch along that line would leave a large, uncontroversial deletion that can merge today.

## 3. The title and body no longer describe the diff

Both still describe dropping the v1/v2 execution-lease shim. `executionLease.ts` is not in the file list any more, and `LegacyLeaseSchema`, `legacyShadowRecord`, `legacyLeasePath` and `readLegacyRecord` have zero occurrences on main. Commit `257b4cd8f2` removed the legacy lease paths at 15:29 today, and `StoredClaim` already carries a singular `file` at `executionLease.ts:155-158`.

What the branch actually contains is eight commits retiring compatibility readers across schemas, roster, settings, archive boundary and trace export: 83 files, +456/-1531. That is the largest concrete tech-debt deletion currently in flight and deserves a title that says so.

PR #12185, opened 2026-09-10
docs: plan Effect-native injection, context, and pipelines

LionSR:
TeXRA already runs Effect in 443 files over a real LayerMap and
ManagedRuntime. It also runs seven other dependency-carrying mechanisms
beside it. This proposal supplies the two things no existing doc has: a
complete carrier inventory with a per-carrier disposition (tag, lifetime,
composition point, ratchet row), and a retirement order in which every
step deletes a competing authority.

Adds, beyond what the PRD, the delivery plan, and the promise-boundary
audit already own:

- A bounded service tag manifest, plus an explicit list of what must NOT
  become a service (plain arguments, loop state, generator locals).
- Disposition for the two AsyncLocalStorage carriers no document owned
  (ToolFileInteractionContext, executionLease's maintenance set).
- Context.Reference as the native ambient-with-default mechanism. Effect 4
  has no FiberRef; the repo uses Reference zero times against five ALS
  carriers.
- Reconciliation of the provide topology, specified three incompatible
  ways across three docs: process layer, session LayerMap, run layer, and
  call-local provideService.
- Per-domain ordering for platform() retirement across its 51 baseline
  files.
- Stream composition, previously unowned.
- Per-file conversion mode (rewrite / edit / delete) with named exemplars,
  and the hard rule that a rewrite-mode file's guaranteed behaviors are
  enumerated before the old file is deleted.

Records one correction to the record: the largest injector is not
platform() (109 sites, ratcheted) but effectRuntime() (~330 sites across
~105 files, unmeasured).

Resolves the recorded contradiction between PRD Phase 2 and the audit's
W1-before-W6 ordering by splitting the RunContext carrier from
workspaceRoots, which were only one item because RunContext nests them.

Four open questions are left to an owner ruling rather than assumed; the
filesystem ruling gates the context half. The provenance section records
that adversarial verification did not run, so the per-step ratchet
arithmetic is a planning estimate.

Claude-Session: https://claude.ai/code/session_015M7qNHuWejkA64mPPgUWzb

### DIFF
--- a/.agents/docs/proposed/architecture/2026-09-10-effect-native-injection-context-pipelines.md
+++ b/.agents/docs/proposed/architecture/2026-09-10-effect-native-injection-context-pipelines.md
@@ -1,0 +1,393 @@
+---
+created: 2026-09-10
+status: proposed
+---
+
+# Effect-native injection, context, and pipelines: the carrier manifest and its retirement order
+
+**Recommendation:** stop describing the Effect work as "adopting Effect." TeXRA already
+runs Effect in 443 files, pins `effect@4.0.0-rc.112`, and operates a real `LayerMap` over a
+real `ManagedRuntime`. What it also runs, beside that, is **seven other dependency-carrying
+mechanisms**, none of which is Effect's context. The 1.0 work is to route those seven into
+the composition points that already exist and delete them — not to introduce a pattern.
+
+This document supplies the two things no existing proposal contains: a **complete carrier
+inventory with a per-carrier disposition** (tag, lifetime, composition point, ratchet row),
+and a **retirement order** in which every step deletes a competing authority. It also
+records the conversion mode — rewrite, edit, or delete — for every file the order touches,
+under the owner directive of 2026-09-09 that full-file rewrites are preferred and are to be
+done fast where the file is a good candidate.
+
+## 1. What this document adds
+
+Nearly every _rule_ the ask implies is already ratified and this document does not restate
+it. [The Effect 4 runtime migration PRD](./2026-08-26-effect-4-runtime-migration.md) owns
+R1 (boundary kinds), R2 (service granularity), R3 (four lifetimes), R4 (plain-Effect
+loops), R5 (interruption), R6 (`Scope`), R7 (typed errors), R8 (`Schedule`), R9 (traces),
+R10 (delete-on-replace), the §8.1 one-carrier-per-lifetime table, the §8.7 collapse ledger,
+and §15 decision 8's idiom ruling (`Context.Service` + static layers, `Effect.fn`,
+`Data.TaggedError`, Zod not Effect Schema). [The delivery
+plan](./2026-09-06-effect-runtime-delivery-plan.md) (the tree's only `accepted-direction`
+doc) owns the work order and two binding negatives: **no second runtime/session registry**
+and **no tag per existing class**. [The agent runtime
+proposal](./2026-09-04-agent-runtime-on-effect.md) §2.4 owns the only concrete tag roster.
+[The promise-boundary audit](./2026-09-07-promise-boundary-audit.md) §6 forbids a new lint
+rule, a `tryPromise` ratchet row, and any adapter. [The observability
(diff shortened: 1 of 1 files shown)


LionSR:
Checked this note against `origin/main` at `2b4e9ffcba` (13 commits past its pin `cf88d2d`). The pinned numbers were accurate at the pin; main has moved under four of them, and one move proves the note's own thesis.

**What changed under it**

| Claim (§2) | At `cf88d2d` | On main | Consequence |
| --- | --- | --- | --- |
| `effectRuntime()` "~330 sites / ~105 files, unmeasured" | 344 / 112 | **404 / 147** | Grew 17% in one day while unwatched. Step 1 (seal the row) is the most urgent step in the note, not a preliminary. It is also not an injector: every one of those sites is an Effect-to-Promise boundary that the R1 ratchet's `Effect.run*` row (22 sites) does not see. The real boundary count is ~400, and that is the migration's honest progress metric. |
| ALS consumer files "~94" | 65 by the grep in the note's §6 sense | 72 | Also growing. Same conclusion. |
| "The flow layer is a Promise island, zero `effect` imports" | 0 | 2 (#12183 converted the LaTeX toolchain leaf-to-root and added four `Effect.run*` allowlist rows inside `flows/reflection`) | Step 8's "row deleted" must also delete those four rows; the baseline comment already says the rewrite deletes their hosts. |
| Carrier 9 / step 11: `executionLanes` `PQueue` + three `pDefer`s | present | **gone** (#12184: `Deferred` hand-off chain, `ExecutionBusy` typed error) | Step 11's `p-queue 13 → ~5` arithmetic and the `AbortController −3` in step 6 are stale; the lanes half of carrier 9 is done, the lease ALS half is not. |

Unchanged and re-verified on main: both `AsyncLocalStorage.bind` sites (`executeAgent.ts:439`, `:618`), the Promise-typed `runner` seam (`AgentRunLifecycle.ts:478-481`), all five ALS carriers at the cited lines, `Context.Reference` at zero uses, 18 `Context.Service` tags in 14 files (the two `packages/agent/src/effect/` tags the TeXRA review notes are the missing pair), tool `execute()` still `Promise<ToolResult>` at `src/tools/core/base.ts:89` after #12191 (which runs programs at the `execute()` boundary, so R1 kind (b) is intact and Q4 still open), and every cross-document link.

**Review triage.** The TeXRA review's four findings are count accuracy; the 13-ports one matters for a manifest that claims completeness, the rest are edits. Codex's eight are mostly substantive, and three change the plan:

- **Leases (L253) is the important one.** Carrier 9's disposition builds an `ExecutionLeases` service over `SynchronizedRef` + `Semaphore`. The accepted 1.0 plan §3 deletes the file lease with the KV cutover, and this note's own §8 cites plan §5 against converting retiring subsystems. The 2026-09-10 execution-ownership note (D1) reaches the same place: the sequence-row claim is the one durable authority. Carrier 9 should read "deleted with the KV cutover; the maintenance ALS goes with it", and step 11 shrinks to per-key concurrency only.
- **Ordering (L252, L287).** Step 7 deletes the `ToolCall` ALS while tools are still Promise-typed, and step 6 flips the `runner` seam while both flow runners are still Promise. Neither compiles without the adapter the note forbids. Both belong inside step 8's atomic rewrite, or step 8 moves ahead of them.
- **`Layer.scoped` not `Layer.effect` (L119)** for `AgentRun`, or run-owned resources outlive an interrupted run.

The rest (StageStack around every stage, LogSink ordering vs the observability plan, the widened roots tag contradicting §8, `AppState` vs `GlobalState`, the 18-globals catch-all row) are real but local.

**One structural remark.** Three of the five ALS carriers exist to carry the run identity or something keyed by it (`RunContext`, `ToolFileInteractionContext` joining run and stage, the lease's maintenance set keyed by execution id), and the fourth nests inside the first. Under the observability plane (identity from the fiber) and the one-run-model note (#12203, one run id), they are one carrier: a run layer that provides one `Run` service, with the stage as a log annotation. The manifest reads as five conversions where one deletion would do. Naming follows: this note says `AgentRun`, the runtime note says `RunContext`, the code says `RunScope`; #12203 proposes one word for all of them.

Issue #12181, opened 2026-09-09
Four hosts each walk stream topology themselves, a fact the fold already states on every row


LionSR:
Surveyed on `main` @ `98afef5a7a`.

The one-fold PRD's rule is that hosts carry only functions of width, locale,
color depth and the clock; every other derivation belongs to the fold. Stream
topology is not one of those. `StreamView` states both halves of it on every
row — `ancestors` and `childIds` (`src/shared/session/sessionView.ts:141-146`),
computed once in `sessionFold.ts` and reference-stable.

Four hosts walk it themselves anyway, in four different ways.

| Site | Shape | Root |
| --- | --- | --- |
| `packages/cli/src/chat/tui/state/sessionView.ts:117-133` (`descendantStreamIds`) | BFS over `childIds` with a `seen` set | included |
| `packages/cli/src/runtime/runProgressRenderer.ts:489-495` | inline `ancestors.some` filter, no name | included |
| `packages/agent/src/effect/sessions.ts:236-247` (`runStreamIds`) | `ancestors.some` filter | included |
| `packages/extension/src/progressView/frontend/components/BackgroundTasksPanel.ts:243-259` (`childrenOf` + `descendantsOf`) | recursion over `childIds` | excluded |

Grepped consumer counts: `rg "ancestors\.some"` over `src` and `packages`
excluding tests returns exactly three sites, two of them above plus
`appInteractionPolicy.ts:138` (a single-ancestor membership test, not a walk).
`descendantStreamIds` has four external production callers
(`panes/StatusBar.tsx:90`, `runChatTui.tsx:442`, `state/resumeHint.ts:75, 115`).
`runStreamIds` has one (`sessions.ts:469`); `descendantsOf` has one (`:279`).

## Proposal

One exported `descendantStreams(view, rootId, { includeRoot })` beside the fold
in `src/shared/session/sessionView.ts`, which already exports `emptySessionView`.
Delete `descendantStreamIds`, `runStreamIds`, `descendantsOf`/`childrenOf`, and
point the `runProgressRenderer` filter at it — its workflow-identity predicate
stays at the call site.

The one behavioral subtlety to preserve: `BackgroundTasksPanel` wants the root
excluded and the other three want it included. That is the `includeRoot` flag,
not a second function.

## Ratchet check

`packages/agent` would gain one `@shared/session/*` import. It already imports
`@shared/session/sessionView`, `runtimeRequest`, `sessionEvents` and
[...]

PR #12188, opened 2026-09-10
refactor: consolidate four independent stream-descendant walks into one

LionSR:
## Summary

Four hosts each independently re-derived a session's descendant-stream set
from facts `SessionView` already states on every row (`childIds`,
`ancestors`), instead of composing the walk once beside the fold that
computes those facts:

- `packages/cli/src/chat/tui/state/sessionView.ts` — `descendantStreamIds`,
  a BFS over `childIds`.
- `packages/agent/src/effect/sessions.ts` — `runStreamIds`, an
  `ancestors.some` scan over every stream in the view.
- `packages/cli/src/runtime/runProgressRenderer.ts` — an inline,
  unnamed `ancestors.some` filter.
- `packages/extension/src/progressView/frontend/components/BackgroundTasksPanel.ts`
  — `descendantsOf`, a recursion over `childrenOf`.

Same conceptual walk, four independent authors, four different
implementations (and, for the root-included/root-excluded question, two
different answers depending on the caller) — exactly the kind of drift a
canonical module exists to prevent, and a second author changing the
topology shape would have had to find and update all four by hand.

## Issue acceptance gates

Addresses #12181 ("Four hosts each walk stream topology themselves, a fact
the fold already states on every row"):

- [x] One exported `descendantStreams(view, rootId, { includeRoot })` beside
  the fold in `src/shared/session/sessionView.ts`.
- [x] `descendantStreamIds`, `runStreamIds`, and the inline
  `ancestors.some` filter are deleted; their call sites point at the shared
  function.
- [x] `BackgroundTasksPanel`'s `descendantsOf` now composes from the shared
  walk instead of its own recursion. `childrenOf` (a direct-children
  accessor used independently for row rendering, not a topology walk) is
  unchanged.
- [x] The root-included-vs-excluded difference across call sites is the
  `includeRoot` flag, not a second function.

## Single source of truth

`descendantStreams` in `src/shared/session/sessionView.ts`, beside
`emptySessionView`, is now the one place the four **production hosts**
(CLI TUI state, CLI plain-output renderer, the `@texra-ai/agent` package,
and the extension's progress-view webview) walk `SessionView` topology.

Not covered: `packages/cli/scripts/tui-harness.tsx`'s own `descendantsOf`
(a `.shift()`-based BFS, same shape as the ones this PR removes) is a
manual dev harness excluded from the published CLI bin
(`!dist/bin/tui-harness.js` in `packages/cli/package.json`), out of
#12181's scope (its consumer-count grep excluded `scripts/`) and out of
this PR's. Flagging it rather than implying the consolidation reaches it.

Removed: two duplicate walk implementations (`descendantStreamIds` BFS,
`runStreamIds` linear scan) plus one duplicate inline filter and one
duplicate recursion, all computing the same descendant-id set from the same
two fold-provided fields.

Added: one function plus a narrow structural parameter type
(`StreamTopology`) so a deep-readonly projection of `SessionView` (the
`@texra-ai/agent` package's own `ReadonlyDeep<SessionView>`) satisfies the
same signature without importing the agent package's type into `@shared` or
duplicating the function. No new pass-through layer — every call site now
calls straight through to the one implementation.

Side effect worth naming: `packages/agent/src/effect/sessions.ts`'s
`@shared/session/sessionView` import turns from type-only into a value
import (it now calls `descendantStreams`), so that module's runtime code —
not just its types — enters the published `@texra-ai/agent` bundle. The
package already value-imports `@shared/schemas` elsewhere, so this isn't a
new category of dependency, and `npm run typecheck:agent` (which builds
the package) shows no unexpected size jump, but it's a load-surface change
worth calling out explicitly rather than leaving implicit.

## Net elements (R6)

- Files changed: 9 for the consolidation itself (8 in the original commit,
  1 test fake fixed in the post-review round below), no new files. A 10th
  file on this branch, `src/tools/delegation/childStream.ts`, is an
  unrelated ported CI fix — see the standing-down comment on this PR — and
  is excluded from the counts below.
- Exported symbols: `^[+-]export` over the diff shows exactly one removed
  (`descendantStreamIds`) and one added (`descendantStreams`) — net 0, but
  consolidates four independent realizations (one exported function, one
  file-private function, one inline filter, one private-method recursion)
  into one.
- Class/interface/enum declarations: 0 net (one new `type` alias,
  `StreamTopology`, not in this bucket).
- Net LoC: +46 (`git diff --stat origin/main...HEAD` over the 9
  consolidation files, current head: 114 insertions, 68 deletions).

**Why net-positive LoC for a `refactor:`:** the prior audit (#12181)
estimated roughly −45 LoC assuming a leaner drop-in. Implementing it, two
things added lines the estimate didn't carry: (1) doc comments on the new
export and its structural parameter type, matching this file's existing
density (`emptySessionView` carries the same); (2) the `StreamTopology`
type needed so the `@texra-ai/agent` package's `ReadonlyDeep<SessionView>`
type-checks against the same function without a second implementation or a
new `@shared`→agent dependency. A post-review round (below) added a test
fixture fix and an index-cursor rewrite of the walk itself. The deletion is
real (four call sites lose their own walk logic) — the added lines buy one
documented, type-safe, linear-time owner instead of a leaner but
undocumented and quadratic one.

## Consumer counts (R8)

Grepped before this change (`rg` over `src` and `packages`, excluding
tests):

- `descendantStreamIds`: 4 external call sites
  (`panes/StatusBar.tsx:90`, `runChatTui.tsx:442`,
  `state/resumeHint.ts:75`, `state/resumeHint.ts:115`) plus 1 internal
  (`anyStreamRunning` in the same file) — all 5 migrated to
  `descendantStreams`.
- `runStreamIds`: 1 call site (`sessions.ts:469`) — migrated.
- The inline `ancestors.some` filter in `runProgressRenderer.ts`: 1 site —
  migrated to a `Set` built from `descendantStreams`.
- `descendantsOf`: 1 call site (`:280`, `renderChildren`'s `since` calc) —
  migrated. `childrenOf` keeps its 2 unrelated call sites (`:271`, `:369`)
  unchanged; it is a direct-children accessor, not a topology walk, so it
  is out of scope for this consolidation.

## Host impact

Extension: `BackgroundTasksPanel`'s dispatch-card rollup math (running
badge, "since" timestamp) is unchanged — same descendant set, sourced from
the shared walk instead of local recursion.

Desktop: none (no desktop-specific code touched).

CLI: `StatusBar`, `runChatTui`'s scrollback-clear eviction, and
`resumeHint`'s usage/target collection all read the same descendant set as
before; `runProgressRenderer`'s plain-output workflow filter is unchanged
in behavior (same membership test, now via a `Set` instead of a per-stream
`ancestors.some` scan) — see the coverage caveat under Validation below.

## Scheduled refactoring

None. This closes out #12181; no further follow-up needed for this
consolidation.

## Validation

- `npm run typecheck:workspace` — pass.
- `npm run typecheck:agent` — pass (package build + declaration validation).
- `npm run typecheck:cli` — pass.
- `npm run typecheck:test-kernel` — pass.
- `npx eslint <all changed files>` — pass.
- `npx prettier --check <all changed files>` — pass.
- `npm run check:dead-code-ratchet` — pass, no new findings.
- Targeted `vitest run`:
  `src/test-kernel/cli/StatusBar.vitest.ts`,
  `src/test-kernel/cli/ResumeHint.vitest.ts`,
  `src/test-kernel/cli/RunProgressRenderer.vitest.ts`,
  `src/test-kernel/frontend/StatusBarSessionEvents.vitest.ts`,
  `src/test-kernel/frontend/StatusBarUsageTracker.vitest.ts`,
  `src/test-kernel/agent/AgentPackage.vitest.ts` — 182 tests passed.

**Coverage caveat, called out by review rather than left implicit:**
`RunProgressRenderer.vitest.ts` exercises only the stderr status-line
renderer, not `attachWorkflowPlainOutput` (the function this PR's
`runProgressRenderer.ts` hunk actually changes); `RunExecution.vitest.ts`
mocks `attachWorkflowPlainOutput` rather than exercising its internals.
Behavior-equivalence for that one site rests on typecheck plus manual
reading (the `Set`-membership rewrite matching the old
`stream.id === root || stream.ancestors.some(...)` filter), not on the
listed suite. `BackgroundTasksPanel.ts` has no dedicated test file either;
same caveat applies there. Every other touched site (`sessions.ts`'s
subscription drain, `StatusBar`, `resumeHint`, `anyStreamRunning`) is
exercised by the suites above — `AgentPackage.vitest.ts` specifically
covers `sessions.ts`'s `descendantStreams` call via its "descendant joins
the run's subscription" case, added to the fake's coverage in the
post-review round below.

**Post-review round:** two findings from automated review, both verified
and fixed:
- `descendantStreams` used `Array.shift()` to pop its BFS queue, making a
  large fan-out's walk quadratic; switched to an index cursor over the
  append-only queue.
- `AgentPackage.vitest.ts`'s `FakeStreamView`/`enterStream` never carried
  `childIds`, so every faked stream crashed `descendantStreams`'
  `pending.push(...stream.childIds)` on `undefined` — this would have
  broken every run-lifecycle test in that file the moment the full suite
  ran it. Gave the fake a `childIds` field and had `enterStream` link a
  stream onto its immediate parent's `childIds` when `ancestors` are
  given, mirroring how the real fold keeps the two in sync.

Full repo-wide `vitest run` was not completed within this session's time
budget; the targeted suites above cover every host touched by this change,
with the one caveat noted.

https://claude.ai/code/session_015aATDS8WgAbwbSyRqJ7kve

### DIFF
--- a/packages/agent/src/effect/sessions.ts
+++ b/packages/agent/src/effect/sessions.ts
@@ -65,10 +65,11 @@ import {
 import type { RequestError } from '@shared/session/requestErrors';
 import type { Outcome, RuntimeRequest } from '@shared/session/runtimeRequest';
 import type { SessionEventsShape } from '@shared/session/sessionEvents';
-import type {
-  SessionView as RuntimeSessionView,
-  StreamView as RuntimeStreamView,
-  TranscriptView as RuntimeTranscriptView,
+import {
+  descendantStreams,
+  type SessionView as RuntimeSessionView,
+  type StreamView as RuntimeStreamView,
+  type TranscriptView as RuntimeTranscriptView,
 } from '@shared/session/sessionView';
 import { generateExecutionId } from '@utils/core';
 import { toErrorMessage } from '@utils/errors/errorMessage';
@@ -231,20 +232,6 @@ const log = createLog('agentPackage');
  * nothing here discards what it has yet to read.
  */
 const TRACE_HANDOVER_EVENTS = 512;
-
-/** The run's stream and every descendant the view holds. */
-function runStreamIds(view: SessionView, streamId: StreamTabId): StreamTabId[] {
-  const ids: StreamTabId[] = [];
-  for (const stream of view.streams.values()) {
-    if (
-      stream.id === streamId ||
-      stream.ancestors.some((ancestor) => ancestor.id === streamId)
-    ) {
-      ids.push(stream.id);
-    }
-  }
-  return ids;
-}
 
 /**
  * The refusal the package states from the caller's own input, before
@@ -466,7 +453,11 @@ function start(
         const drain = yield* Effect.forkDetach(
           Stream.runDrain(
             view.pipe(
-              Stream.tap((level) => interest(runStreamIds(level, streamId))),
+              Stream.tap((level) =>
+                interest(
+                  descendantStreams(level, streamId, { includeRoot: true }),
+                ),
+              ),
             ),
           ),
           { startImmediately: true },
--- a/packages/cli/src/chat/tui/panes/StatusBar.tsx
+++ b/packages/cli/src/chat/tui/panes/StatusBar.tsx
@@ -14,6 +14,7 @@ import type {
   SubscriptionUsageSnapshot,
   UsageRoute,
 } from '@shared/schemas';
+import { descendantStreams } from '@shared/session/sessionView';
 import { isActivePhase } from '@shared/streams/streamStatus';
 import { toErrorMessage } from '@utils/errors/errorMessage';
 
@@ -30,7 +31,6 @@ import {
 import {
   ancestorPhaseLabel,
   cumulativeUsageOf,
-  descendantStreamIds,
   sessionView,
   streamPhaseOf,
   streamViewOf,
@@ -87,7 +87,7 @@ export function StatusBar(props: StatusBarProps): React.JSX.Element {
     status: streamPhaseOf(streamViewOf(view, rootRunStreamId)),
   };
   const ownedStreamIds = useMemo(
-    () => descendantStreamIds(view, rootStreamId),
+    () => descendantStreams(view, rootStreamId, { includeRoot: true }),
     [view, rootStreamId],
   );
   const target = statusBarStreamTarget({
--- a/packages/cli/src/chat/tui/runChatTui.tsx
+++ b/packages/cli/src/chat/tui/runChatTui.tsx
@@ -53,6 +53,7 @@ import type {
 } from '@shared/schemas';
 import { AgentCategory, STREAM_PHASE } from '@shared/schemas';
 import { subscribeToSignalChanges } from '@shared/signals';
+import { descendantStreams } from '@shared/session/sessionView';
 import { getFirstRunDone } from '@shared/state/onboardingState';
 import { isActivePhase } from '@shared/streams/streamStatus';
 import { platformSettingsStores } from '@utils/config/platformSettings';
@@ -87,7 +88,6 @@ import {
 import {
   bindSessionView,
   currentView,
-  descendantStreamIds,
   sessionView,
   streamPhaseOf,
   streamViewOf,
@@ -439,9 +439,10 @@ export async function runChat(
     // Release this conversation's resident transcripts when their remaining
     // readers and writers leave. Clearing the terminal does not delete history.
     const store = runtimeSession.transcripts;
-    for (const streamId of descendantStreamIds(
+    for (const streamId of descendantStreams(
       currentView(),
       rootStreamIdSignal.get(),
+      { includeRoot: true },
     )) {
       store.requestEviction(streamId);
     }
--- a/packages/cli/src/chat/tui/state/resumeHint.ts
+++ b/packages/cli/src/chat/tui/state/resumeHint.ts
@@ -9,13 +9,12 @@ import {
   type TokenUsageStats,
 } from '@shared/schemas';
 import { usageCostLabel } from '@shared/copy/modelAccess';
-import type { SessionView } from '@shared/session/sessionView';
-
 import {
-  cumulativeUsageOf,
-  descendantStreamIds,
-  streamViewOf,
-} from './sessionView';
+  descendantStreams,
+  type SessionView,
+} from '@shared/session/sessionView';
+
+import { cumulativeUsageOf, streamViewOf } from './sessionView';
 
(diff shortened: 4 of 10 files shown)


LionSR:
The `static checks (linux)` failure is not this PR's: `prettier --check` fails on `src/tools/delegation/childStream.ts`, a file this PR's diff never touches. Confirmed by running `prettier --check` against that file's content at the unmodified base commit (`cf88d2d`) directly — it already fails there.

Opened #12190 with the standalone fix and ported the same one-line reformat into this branch (`caf4ac3`), so this PR's CI now includes it too; it will no-op once `main` picks up #12190.

---

LionSR:
Both Codex findings verified and fixed in `4681780`:

- **P1** (`AgentPackage.vitest.ts`'s fake missing `childIds`): confirmed — it would have crashed every run-lifecycle test that drives the session view, since `descendantStreams` spreads `stream.childIds`. Added `childIds` to `FakeStreamView`, defaulted to `[]` in `enterStream`, and had `enterStream` link a stream onto its immediate parent's `childIds` when `ancestors` are given (mirroring how the real fold keeps the two in sync). The "descendant joins the run's subscription" test now actually exercises the walk rather than vacuously passing. `AgentPackage.vitest.ts`: 16/16 pass.
- **P2** (quadratic `Array.shift()` BFS): confirmed and fixed — swapped to an index cursor over the append-only `pending` array, so the walk stays linear.

---

LionSR:
Addressed the residual (non-blocking) points from the TeXRA review bot's pass:

1. **Overstated test coverage for `runProgressRenderer.ts`** — confirmed: `RunProgressRenderer.vitest.ts` doesn't reach `attachWorkflowPlainOutput`, and `RunExecution.vitest.ts` only mocks it. Corrected the PR body's Validation section to say so plainly rather than implying coverage that isn't there.
2. **"One place" claim missed `tui-harness.tsx`** — confirmed: that dev harness script has its own `.shift()`-based `descendantsOf`. It's excluded from the published CLI bin and out of #12181's scope (its grep excluded `scripts/`), so I left it alone, but reworded the PR body to scope the claim to the four production hosts and flag the harness explicitly instead of implying full coverage.
3. **Runtime-dependency note** (type-only → value import in `packages/agent`) — added a paragraph under "Duplication and abstraction budget" naming it explicitly.
4. Naming nit — acknowledged, left as-is (cosmetic).

---

PR #12203, opened 2026-09-10
docs: one run model for 1.0, superseding the duplicate-concept families

LionSR:
Takes #12186 further. The census counted eight families of duplicated vocabulary; this note treats them as one cause (a run declared once per layer and once per era) and states the 1.0 run model once, with its deletion ledger and its order against the runtime lane.

**The design** is four rules (one fact log and one fold per question; commands in, facts out; one protocol at every boundary; identity minted once and never encoded) applied to seven facts about a run: identity, parent edge, phase and outcome, tool-call status, result, events, and waiting on a human. Each fact gets one declaration; every current spelling becomes a `z.infer` or a `.pick()` of it or is deleted. The trace union and the CLI NDJSON output become projections of `sessionEvent.ts` instead of parallel vocabularies.

**Bold items, listed for owner ruling in section 7:**
- One aggregate per run instead of `stream` + `execution` sharing a logical id. Deletes the second sequence counter, the second ownership claim, and the `RunAggregates` two-key signature that the PR1 ledger note records as a deviation it was forced into.
- The CLI NDJSON contract is versioned at 1.0 (`version: 2`, same field names as the session vocabulary) and the texra-action pin moves with it. The current rename table is frozen by a comment, not by a consumer this repo can see.
- Inquiry joins the request protocol. Today approval facts are durable but the prompt is an in-memory promise, so a restart loses every pending approval, retry, and question, while inquiry survives only because it was built a second time with its own database. One `request.opened` / `request.decided` pair over all seven kinds; pending is the fold; the run parks in its scope (lane D3).

**Sequencing.** The identity and vocabulary changes are schema work that must land before PR1 of the runtime lane and before the Tier-1 SDK manifest, both of which otherwise freeze the duplicate.

**Verification.** Every census family was re-investigated against live `main` by eight independent read-only passes with `path:line` evidence before the note was written. Section 8 records what was refuted so it is not re-derived: `parentStreamId`/`parentExecutionId` are one edge (the only site using both guards a state the seq-1 rule forbids); `StreamTabInfo` and the per-host grouping tables were already deleted today by #12199 and #12189; the `AgentEvent` duplication is hand-written interfaces mirroring Zod schemas, identical in all but two arms; `ExecutionStatus`/`HistoryRunStatus` are justified projections; `runId` crosses no frozen boundary; `RunKind`/`RunDescriptor`/`ExecutionMeta.category` do not exist; `background` is a real launch-mode fact.

Also in this PR: the census's section 4 now points here, and the stale `CLAUDE.md` bullet mandating `p-queue` (contradicting AGENTS.md's Effect rule, flagged by the census) is corrected.

Docs only; no code changes.

https://claude.ai/code/session_016YXyzzHfwcgD4hcVWmh6TR

### DIFF
--- a/.agents/docs/proposed/architecture/2026-09-10-collapse-duplicate-concepts.md
+++ b/.agents/docs/proposed/architecture/2026-09-10-collapse-duplicate-concepts.md
@@ -63,14 +63,14 @@ today the same type already.
 
 ### What was confirmed
 
-| Claim | Status |
-| ----- | ------ |
-| One minting formula, `` `${getCleanAgentName(name)}#${executionId}` `` | Confirmed — [streamTab.ts:24](../../../../src/agent/runtime/streamTab.ts), the only producer |
-| The display half is already the single source of truth elsewhere | Confirmed — `runIdentityDisplayName` is `getCleanAgentName(runIdentityName(id))`, the same function ([runIdentity.ts:54](../../../../src/shared/schemas/runIdentity.ts)) |
-| Nothing in production addresses a run by parsing the format | Confirmed — the sole `#` reader is an equality assertion between the two ids ([ToolUseFollowUpQueueManager.ts:135](../../../../src/agent/followUp/ToolUseFollowUpQueueManager.ts)) |
-| A hand-rolled reverse index exists between the two ids | Confirmed — `getAgentHandleByStream` is an O(n) scan over handles keyed by execution id ([executionRegistry.ts:435](../../../../src/agent/runtime/executionRegistry.ts)) |
-| One struct carries the composite *and* both its components | Confirmed — `RunScope` declares `streamId`, `executionId` and `agentName` together ([RunScope.ts:16](../../../../src/agent/runtime/RunScope.ts)) |
-| No ratchet or architecture test pins the current shape | Confirmed — zero hits across `config/ratchets/` and `src/test-kernel/architecture/` |
+| Claim                                                                  | Status                                                                                                                                                                             |
+| ---------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


LionSR:
Review triage, commit a114e5b4af.

Six of the seven Codex findings were substantive and three of them sharpened the design rather than softening it:

- **`background` (L147).** Correct, and the conclusion goes further than the finding: every writer sets it beside a parent, in-band children included, and its one reader is a CLI view-switch hint. It is the parent edge as a boolean, deleted. The "launch mode" field is gone from the note.
- **Non-blocking inquiry delivery (L256).** Correct. Resolved by one rule: a `request.decided` for a request its run is not parked on is delivered as a follow-up through the existing recoverable submit path. The inquiry continuation module is that rule implemented a second time.
- **`UsageProvider` (L286).** Correct that it is a different fact from `ModelProvider`. The fact it records is the turn's wire protocol, which the 1.0 provider package already declares once as the origin `protocol`. Derived from there; `unknown` disappears.
- **`status` carries non-terminal transitions (L166).** Correct. `RunPhase` now names the rows it folds from: `run.start`/`run.activate`, `flow.step` `waiting`/`halted`, `run.end`. The `status` row goes when `flow.step` lands, not before.
- **Workflow lifecycle (L159).** Correct; scheduling keeps its own lifecycle. The real duplicate in that family is the pair of nine-value call-status enums, now the stated target.
- **`flow.snapshot` vs R1 (L40).** Correct. R1 now admits a checkpoint that the rows can rebuild and that loses every conflict; PR1's "restores what no row carries" is a gap closed by rows.
- **Audited commit (L10).** Fixed.

The TeXRA review job failed before reviewing (`HTTP 402 Insufficient Balance` on the review model key), unrelated to this PR. The docs-build failure was a formatting miss on my side, fixed in this commit.

Issue #12177, opened 2026-09-09
Session transcript fold: an unreachable stream.chunk path blanks finalized model responses, and its accumulator and multicast are dead


LionSR:
Surveyed on `main` @ `98afef5a7a`. Two independent survey agents reached this
conclusion separately, and every consumer count below was re-grepped by hand.

## The defect that makes this urgent

`traceFold.ts` registers a row on `stream.start`, and `response.finalized`
(`src/shared/session/traceFold.ts:382-410`) settles that row's text but never
removes its id from the local `streams` map. The terminal-phase boundary then
walks that map and re-settles every surviving row with
`redactSecrets(state.chunks.join(''))` (`traceFold.ts:458-463`).

`state.chunks` is always empty in production (see below), so a run that reaches
a terminal phase without a `stream.end` — a crash, a cancel, a torn-down trace —
overwrites the authoritative finalized model-response text with `''`. This is a
checklist §15 masking site: the row still renders, just empty.

## Why `state.chunks` is always empty: `stream.chunk` cannot reach this fold

Three independent legs, each grepped:

1. `stream.chunk` never becomes a durable event. `src/shared/schemas/traceEvent.ts:119`
   states it in the file: "Deltas are fold inputs only; they never enter the
   durable event vocabulary." `TranscriptChunkSchema` (`:120`) is spread into
   the `TranscriptEventSchema` *union* but not into the `TranscriptEventSchemas`
   *record*, and `src/shared/schemas/sessionEvent.ts:342` spreads the record.
2. `runEventDraft` returns `null` for `stream.chunk`
   (`src/agent/runtime/SessionEvents.ts:129-133`), and
   `SessionHandle.publishRunEvent` routes chunks to `graph.publishText`
   instead (`src/agent/runtime/SessionHandle.ts:557-563`).
3. Both production callers of the trace fold narrow to durable events:
   `sessionFold.ts:1697-1703` and `StreamLogStore.ts:55, 249`, each guarded by
   `isTranscriptEvent` over a `SessionEvent` / `DisplaySessionEvent`.

Live text rides the transient plane instead: `TextChunkSource` →
`InflightTextChunk` → `foldTextChunk` (`sessionFold.ts:1247-1310`) →
`view.inflight`.

So `traceFold.ts:361-370` (`case 'stream.chunk'`) is an unreachable arm, and
`StreamLog.appendText` — its only non-test caller — is dead with it.

## Dead with it: the lazy streaming-text accumulator

[...]

PR #12208, opened 2026-09-10
refactor: say "run" for the run in every layer (one run model, S1a)

LionSR:
The run is now called a run in every layer. Step S1a of the
[one run model](https://github.com/LionSR/TeXRA/pull/12203), section 3.10.

Before this, the same object was an **execution** in the runtime and storage
layers, a **stream** in the session and view layers, and a **run** at the seams.
This renames 155 exported symbols and every reference to them. It is a pure rename with no behavior change. The only line-count change is Prettier
collapsing constructs the shorter names now let fit on one line, net 378 lines.

## How to review a diff this size

Do not read it file by file. It was generated, and each step is reproducible:

1. **A compiler-accurate symbol rename** through TypeScript's language service,
   over the union of every project. This matters: the root tsconfig covers
   1,210 of 2,359 files, and a rename driven from it alone silently misses the
   CLI, desktop, SDK and test references.
2. **Unlinked references** in the CLI and test kernel, which resolve imports in
   bundler mode, fixed by a syntax-tree pass that renames identifier references
   only, never strings, comments or property keys.
3. **Test mocks the compiler cannot check.** 81 object keys inside
   `vi.mock` factories and `vi.spyOn` method strings across 27 files.
   Unrenamed, a mock keyed by an old export name replaces nothing and the real
   implementation runs, while the test passes or fails for the wrong reason.
4. **Ratchet baselines and ratchet test keys** that name renamed symbols.

What to spot-check is the exclusion list below, not the renames.

## Deliberately not renamed

| Kept | Why |
| --- | --- |
| `StreamTabId`, `StreamTabIdSchema`, `getStreamTabId`, `getRunContextStreamId`, `StreamView` | S1b deletes or merges them; renaming first would touch them twice |
| `StreamHandle`, `StreamKind`, `StreamOptions`, model-handler stream processors | "stream" there means a provider token stream, which the note keeps |
| `ExecutionStatus` and its projection | kept inside the CLI projection for S5 |
| Every wire field and string: NDJSON `streamId`/`executionId`/`storageKey`, Supabase `stream_id`, persisted keys | S5 owns wire renames |
| File names, the `SessionView.streams` property, test-internal mock labels like `mocks.registerExecution` | mechanical follow-ups |

Eight names from the first mechanical pass were wrong and are corrected here.
Three would have collided with distinct existing concepts, so they got their own
names rather than a merge: `ExecutionRun` is `RunDescriptor` (the SDK already has
a different `Run`), `ExecutionRunRecordSchema` is `RunLaunchRecordSchema`, and
`withExecutionRunContext` is `withLaunchRunContext`.

## Net elements (R6)

| Measure | Delta |
| --- | --- |
| Files changed | 416 |
| Files added or deleted | 0 |
| `^[+-]export` lines | -178 / +178, the same declarations under new names |
| Net LoC | 3,759 insertions, 4,137 deletions |

## Consumer counts (R8)

No public symbol is deleted, so no consumer loses anything. Every renamed
symbol's consumers were renamed with it: 3,305 references resolved by the
compiler, 236 unlinked references fixed by syntax tree, and
81 mock keys.

## Merge order with the session bug-fix branch

This overlaps 17 files with `fix/session-batch2-20260910`, the data-loss fix for #12177.
Whichever merges second gets this rename applied by re-running the same deterministic
recipe on its branch, which makes both sides' identifiers identical and removes the
conflict. Landing this one first is cheaper: at 416 files it is the change most exposed
to conflicts with every other open PR, and main is quiet right now.

## What comes next

S1b does the semantic collapse on top of this: brand the run id, delete
`StreamTabId` and its minting, collapse the parent edge onto `run.start`, and
remove `background` and `isSubagent`. It uses the note's two-kinds-sharing-one-id
shape, which leaves the persistence SQL intact; the one-aggregate shape rewrites
claim and cascade SQL in `Database.ts` and waits on the owner's section 7 ruling.

## Verified

`npm run typecheck` across every project, `npm run lint`, `npm test` (741 files, 9051 passed), `npm run check:dead-code-ratchet`, the effect-migration ratchet, `npm run check:guidance-refs`, and `prettier --check` on every changed file.

https://claude.ai/code/session_018QtWawFHNreKY2RkrRRHg4


LionSR:
Merge-order partner: #12209 (the #12177 data-loss fix) overlaps 17 files with this PR. Whichever merges second gets this rename re-applied by its recipe, which a replay has already reproduced to an identical tree hash.

LionSR:
Merge-order proof for landing this first. I replayed this PR's rename recipe on #12209's commit, took #12209's delta in renamed space (19 files, 269 insertions, 730 deletions, the same change as #12209 itself), and applied it onto this PR's commit. It applies cleanly, and the composed tree type checks with zero errors on the root workspace, the test kernel and the CLI. So after this merges, #12209 only needs that patch pushed to its branch; its full gate run on the composed tree is in progress now.

LionSR:
Correction to the S1b plan in this PR's description, found while scoping it. The description says S1b will "collapse the parent edge onto `run.start`", following section 3.2 of the one run model note. That is not safe as written.

**The parent edge is mutable.** `setParentStream` is emitted by `detachActiveChildren` in the run registry: stopping an orchestrator without killing its children promotes them to top level and publishes `setParentStream { parentStreamId: null }` for each. The session fold applies it to the view's parent pointer, and `setParentStream` is a record type in the frozen CLI NDJSON vocabulary. So a run's parent is a launch value on `run.start` plus a later detach fact. Moving it onto `run.start` alone would silently break "stop the orchestrator, keep its children".

**What S1b does instead.** Once a run's stream id is its run id (S1b pass A), `parentStreamId` and `parentExecutionId` always hold the same value, so the duplicate pair collapses to one parent id on each carrier. `setParentStream` stays as the single "parent changed" fact, and `background` and `isSubagent` are still derived from the parent edge. Section 3.2 of the note should be amended to say the edge has a launch value and a detach fact.

LionSR:
A second correction to the S1b plan, for the `isSubagent` step. Section 3.2 of the note says `isSubagent` is the parent edge as a boolean and that `isChildExecution` is `parent !== null`. In the code these are two different facts, and collapsing them would change behavior.

- **`isChildExecution` is the live edge.** `ExecutionHandle.isChildExecution` is `_parentStreamId !== childStreamId`, and `detach()` sets the parent to the run itself. So when an orchestrator is stopped without killing its children, each detached child answers false from then on.
- **`isSubagent` in the tool-use flow is launch lineage.** It is captured once into the flow's services and never re-read. It decides three behaviors in `ToolUseWaitNode`: stop immediately after an error, skip the root-only idle notification, and suspend through the child-run loop instead of waiting for follow-ups. A detached child therefore keeps behaving as a subagent and keeps delivering through the follow-up queue, which is what `detachActiveChildren`'s documentation promises.
- **The persisted result's `isSubagent` reads the live edge at write time** (`childStream`, `childRunLoop`, `waitingTermination` all pass `handle.isChildExecution`). So a detached child's final result reports top level, and the terminal toast treats its failure like a root run's. That split looks intended, since a promoted run should surface its failure.

Deriving the flow's `isSubagent` from the live edge would flip a detached child into root mode mid-flight: goal continuation, `waitForFollowUp`, root idle notifications. So S1b keeps launch lineage and the live edge as two facts and removes only the plumbing that duplicates one of them.

LionSR:
I want to completely retire streamID

LionSR:
Collapse as much as possible

LionSR:
Superseded by #12222, which lands the whole of S1 (this rename plus S1b's stream-id-is-run-id collapse, the single run aggregate, the parent edge on run.start, and the StreamTabId/prefix deletion) on a newer base and is green on every local gate. Closing to avoid resolving the same 400-file rename twice. Owner decision 2026-09-11.

PR #12210, opened 2026-09-10
docs: design the Effect-native agent runtime across every surface

LionSR:
## Summary

Design note for finishing the agent runtime as one Effect-native system: `.agents/docs/proposed/architecture/2026-09-10-effect-native-runtime-system-design.md` (revision 3).

It builds on the migration PRD (R1 to R10), the 2026-09-04 runtime proposal (rows, fold, resume), the delivery plan, the injection note, the one-run model and the 2026-09-06 loop study (OpenCode V2, Pi harness, effect-agent), and adds what none of them contain:

- **Ownership** (§2.1): one owner per fact, one writer per row type, named readers, what each replaces. Normative.
- **The fiber tree** (§2): a child run is a session-scoped `FiberMap` entry, never a child of its parent's fiber; interruption is the only cancellation.
- **Layers and composability** (§3): the process/session/run layer graph, what a layer swap buys (headless SDK, webview, replay, tests), `Context.Reference` for the two defaulted ambient values, and the native pipeline shape: `Stream.unfold` over the turn, `Stream.scan` for inspection, `FiberSet.makeRuntimePromise` as the workflow-script sandbox bridge.
- **A twenty-tag manifest** (§4) reconciled with the corpus: `Runs` + `Requests` share one implementation, `ModelRoutes` is session-scoped, `SessionView` is a row, `ToolCall` is three fields.
- **The two programs** (§5) with real signatures, the three dispatch combinators, `ModelInvoker` over `packages/llm` (`TurnResult` is the completed turn; four package gaps named), and `Requests` with the policy short-circuit and per-stream serialization preserved.
- **Replay, persistence and resume** (§6): every item of today's resume state mapped to a row, one fold with three uses, the resume decision table in rows, persistence at the activity boundary, coordination with `delegate_multi_agents` workflow mode (content-addressed cache, in-flight child adoption, skip/retry as decisions, one semaphore), the fifteen crash windows and what closes each, and the two-permission replay rule.
- **Per-surface map** (§7) with measured deletions from six deep surveys; **rc.112/rc.113 verification** (§8) including the correction that the local effect-solutions clone is beta.98, older than the pin; **readings taken on the open rulings** (§9); **three delivery slices plus a package-gap slice** (§10); **a collapse ledger** (§11).

Docs only. No production code, data format or API changes.

## Net elements (R6)

n/a (docs only)

## Consumer counts (R8)

n/a (docs only)

https://claude.ai/code/session_01KYFUQNevwj3u77FoxaXUUf


LionSR:
Revision 4 of the note is on #12272 and takes a reading on every thread here (its §14 maps each finding to the section that changed). Taken: `Layer.scoped` absent → `Layer.effect` (excludes `Scope`); `Schedule.recurs` constructor → `Effect.retry({ while, times, schedule })`; `Latch.isOpen` boolean; `Requests`/`FollowUps` error channels and `Option` shapes; snapshot from post-dispatch state; duplicate windows per segment; waiter registered before the `request.opened` row; waiting runs park in their fiber; `model.attempt` row before provider I/O; interrupted barrier tools stay unsettled; file lease deleted at the KV cutover; attachments captured before `tool.result`; duplicate and skipped settlements as rows; follow-ups seeded into the queue on resume; manual retry re-reads the selection; detached child delivery in the child's terminal transaction; `observe` carries a cursor into the tail; in-band slot inheritance kept (two semaphores with distinct meanings); settlements appended in call order per segment; per-attempt physical run ids for retried workflow children; the three count corrections. The session-scoped route gate was already in revision 3.

PR #12212, opened 2026-09-10
docs: correct the one run model's parent edge

LionSR:
Appends a dated correction, section 3.2.1, to the one run model note. Two claims
in section 3.2 do not match the code, and S1b's implementation would have changed
behavior by following them. Found while scoping S1b; recorded in comments on
#12208, and made durable here because the later S2 to S5 lanes build from this note.

**A run has two parent edges, not one.** The creation edge is immutable:
`run.start` records it and `hasPersistedParent` reads it when a run resumes. The
current edge is mutable: stopping an orchestrator without killing its children
detaches them and publishes `setParentStream { parentStreamId: null }`, which the
fold applies to the view. So `setParentStream` is not deleted. Only the duplicate
`parentStreamId`/`parentExecutionId` pair collapses, since both hold one value once
a run's stream id is its run id.

**`isSubagent` and `isChildExecution` read different edges.** The tool-use flow's
`isSubagent` is captured at launch from the creation edge and drives three
behaviors in `ToolUseWaitNode`. `isChildExecution` reads the current edge, so a
detached child answers false while its flow keeps behaving as a subagent and
delivering through the follow-up queue. Deriving one from the other would flip
detached children into root mode mid-flight.

The section's original text is left in place so its reasoning stays visible; the
correction sits directly after it. Docs only.

https://claude.ai/code/session_018QtWawFHNreKY2RkrRRHg4

### DIFF
--- a/.agents/docs/proposed/architecture/2026-09-10-one-run-model.md
+++ b/.agents/docs/proposed/architecture/2026-09-10-one-run-model.md
@@ -163,6 +163,27 @@ is the parent edge spelled as a boolean a fourth time, and it is deleted. The
 five persisted carriers of the parent pointer collapse to the `run.start`
 row; "is a direct child of X" has one implementation, over the fold.
 
+#### 3.2.1 Correction, verified against the code on 2026-09-10
+
+Two claims in 3.2 do not hold, and implementing them literally changes behavior.
+
+**The edge is two facts, not one.** A run has a creation edge and a current edge.
+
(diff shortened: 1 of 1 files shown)


LionSR:
Update to 3.2.1: its open question about `background` is now answered, and the answer is no. `background` does not equal the creation edge on every launch path. Resuming a child in workflow mode relaunches it with `background: false` and no `isSubagent`, because `RunAgentOptions` excludes `isSubagent`, while a tool-use child resumed the same way derives it from `hasPersistedParent`. Filed as #12213. So deleting `background` needs that resume path fixed first, or a ruling that the foreground resume is intended, in which case `background` is a real launch-mode fact.

PR #12217, opened 2026-09-10
simplify(tools): derive update_config from the catalog; inline the edit-approval pass-through

LionSR:
## Summary

Two duplicated authorities in the tool layer, each deleted in favour of the one that already owns the fact.

**1. `update_config` re-declared schemas the settings catalog owns, and two had drifted.** `ConfigTools.ts` hand-wrote a Zod schema and a summary for six keys that `stateSettings.ts` already defines. The tool's copies disagreed with the catalog:

- `texra.maxImageDimension`: `int 64..8192` in the tool vs `number 100..10000` in the catalog, so the setup agent could persist `64`, which `src/utils/media/img.ts` reads.
- `texra.latex.tikzInputDirectory`: the tool told the model it is a *workspace-relative* directory; the catalog (and `texTools.ts`) require an *absolute* path.

The allowlist is now a list of keys. `update_config` validates against `settingSchemaWithoutPrefault(settingByKey(key))` and quotes the catalog description in its errors. `read_config` now appends the catalog description to the value it reads, which is where the model learns what a setting controls, matching the tool's "read first, explain, then change" guidance. The per-key write allowlist, which is the security boundary, is unchanged.

The catalog is consulted only when a tool runs, never while `ConfigTools` loads. A version that built the tool description from the catalog at import time hung `TuiApprovalRetry` ("approves delegated work already queued"): the suite passes on main, and with key names alone it passes again. So the static description lists key names only.

**2. `fileEditFlow` → `toolEditApproval` pass-through.** `applyApprovedFileEdit` has two callers (`EditTool`, `WriteTool`). Both passed `startLine: 'approval'` and neither passed `beforeWrite`, `afterWrite` or `diffSeparator`. It delegated to `requestAndWriteApprovedEdit`, whose only caller it was. This PR:

- inlines that middle layer (approve, then reject or `writeApprovedContent`);
- deletes the four unused options and `appendApprovalDiffNote`'s `separator` parameter;
- deletes the `WrittenApprovedEdit` type.

`startLine` is always the approval's, which is what both callers asked for. Tool results are unchanged.

Items `domain-tools-5` and `tools-core-11` of the 1.0 clean-slate simplification audit (coauthor-21).

## Net elements (R6)

- Files: 5 changed, 0 added, 0 deleted
- `^[+-]export` symbols: −1 (`requestAndWriteApprovedEdit`)
- Class/interface/enum declarations: −1 (`WrittenApprovedEdit`)
- Net LoC: +67 / −155

## Consumer counts (R8)

- `requestAndWriteApprovedEdit`: 1 production caller (`applyApprovedFileEdit`), 0 tests
- `beforeWrite` / `afterWrite` / `diffSeparator` on `applyApprovedFileEdit`: 0 of 2 callers pass them
- `startLine: 'omit'` (the default): 0 of 2 callers use it
- `appendApprovalDiffNote` `separator`: 1 caller, which only forwarded `diffSeparator`
- `UPDATABLE_KEYS` schemas: 6 duplicated catalog rows, 2 drifted

## Test plan

- [x] `npx vitest run` over the 16 suites touching edit approval and ConfigTools (240 tests), including `TuiApprovalRetry`
- [x] eslint on changed files

### DIFF
--- a/src/shared/schemas/stateSettings.ts
+++ b/src/shared/schemas/stateSettings.ts
@@ -574,7 +574,7 @@ const CORE_SETTING_ROWS: Record<
     honoredBy: everyHost('src/model/xai/xaiPreference.ts'),
   },
   maxImageDimension: {
-    schema: z.number().min(100).max(10000).prefault(2000),
+    schema: z.int().min(100).max(10000).prefault(2000),
     description:
       'Maximum dimension (width or height) in pixels for images before resizing. Images larger than this will be resized to fit within this dimension while maintaining aspect ratio.',
     honoredBy: everyHost('src/utils/media/img.ts'),
@@ -586,7 +586,7 @@ const CORE_SETTING_ROWS: Record<
     honoredBy: everyHost('src/tools/latex/ExtractBibliographyTool.ts'),
   },
   'bib.zoteroPort': {
-    schema: z.number().min(1).max(65535).prefault(23119),
+    schema: z.int().min(1).max(65535).prefault(23119),
     description:
       'Port number for Zotero integration (default: 23119). Used by both the Connector API and Better BibTeX JSON-RPC.',
     honoredBy: everyHost('src/tools/zotero/bbtClient.ts'),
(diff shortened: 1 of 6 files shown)


LionSR:
Fixed in 1fc47a814e: the three integer-valued catalog rows (`maxImageDimension`, `bib.zoteroPort`, `git.numberOfCommitsToShow`) are now `z.int()`, so `update_config` rejects `23119.5` again, with the catalog as the only authority. `z.int()` still satisfies the settings UI's `instanceof z.ZodNumber` row check. Typecheck, ConfigTools and stateSettings suites pass.

PR #12220, opened 2026-09-11
fix(tools): read the gitignore policy per call; drop dead pre-identity warnings

LionSR:
## Summary

Two small corrections in the tool layer, one a real bug.

**1. The gitignore matcher was cached for the life of the process.** `getGitignoreMatcher` loaded the ignore policy once into a module-global `sharedLoad` `Deferred` that was never invalidated or keyed by workspace. But `WorkspaceFS.getPath()` is session-scoped, and the desktop main process serves several projects at once. So `glob`, `grep` and the arXiv source listing in the second project filtered by the *first* project's `.gitignore`, and a long-lived extension, desktop or TUI never saw `.gitignore` edits.

The matcher is now built on every call. That means reading at most three small policy files: `~/.gitignore_global`, `<ws>/.gitignore_global` and `<ws>/.gitignore`. This deletes the shared-load `Deferred`, its in-flight coalescing, its retry-after-failure bookkeeping and the uninterruptible wrapper those needed. The cache-pinning test is replaced by one regression test: a second call sees an edited policy.

Deliberately out of scope: the audit also proposed dropping the matcher for `rg`'s native ignore handling and rewriting `glob` on `rg --files`. That changes glob semantics, requires `rg` on `PATH` for `glob`, and drops the `.gitignore_global` policies, so it is a product call and not part of this PR.

**2. Dead pre-identity warnings in `/executions`.** `ExecutionsTool` warned at two sites when an execution row had metadata but no identity ("pre-identity row … reader retired per #9590 Stage 7"). Both `ExecutionMetaCoreSchema.identity` and `RunStartEventSchema.identity` are required, so a decoded row with metadata always has an identity. The warnings are unreachable and are deleted. The `identity === undefined` arms in `executionFormatters` stay: they still serve a record whose metadata row is missing.

Items `tools-core-3` (minimal fix) and `delegation-workflows-4(a)` of the 1.0 clean-slate simplification audit (coauthor-21).

## Net elements (R6)

- Files: 3 changed, 0 added, 0 deleted
- `^[+-]export` symbols: 0 (`getGitignoreMatcher` keeps its name and signature)
- Class/interface/enum declarations: 0
- Net LoC: production +51 / −92 (−41: `gitignore.ts` −25, `ExecutionsTool.ts` −16, after the follow-up commit dropped the now-unused `log`); tests: one cache test replaced by one regression test

## Consumer counts (R8)

- `getGitignoreMatcher`: 3 production callers (`grep.ts`, `glob.ts`, `ArxivDownloadTool.ts`), each calls it once per tool invocation; unchanged
- `sharedLoad`: module-private, 1 reader (deleted)
- pre-identity warn branches: 2, 0 reachable (identity is required in both schemas)

## Test plan

- [x] `npx vitest run` GitignoreMatcher, grep, GlobTool, ArxivDownloadTool, the ExecutionsTool suites (161 tests)
- [x] eslint on changed files

### DIFF
--- a/src/test-kernel/tools/GitignoreMatcher.vitest.ts
+++ b/src/test-kernel/tools/GitignoreMatcher.vitest.ts
@@ -10,7 +10,6 @@ const fsState = vi.hoisted(() => ({
   homeDirectory: undefined as string | undefined,
   workspaceFiles: new Map<string, string>(),
   workspaceReadErrors: new Map<string, Error>(),
-  workspaceReadCounts: new Map<string, number>(),
   absoluteFiles: new Map<string, string>(),
   absoluteReadErrors: new Map<string, Error>(),
   reset(): void {
@@ -18,7 +17,6 @@ const fsState = vi.hoisted(() => ({
     this.homeDirectory = undefined;
     this.workspaceFiles.clear();
     this.workspaceReadErrors.clear();
-    this.workspaceReadCounts.clear();
     this.absoluteFiles.clear();
     this.absoluteReadErrors.clear();
   },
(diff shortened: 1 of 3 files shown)


LionSR:
Fixed in 2f8679e03d: dropped the now-unused `log` const and its `createLog` import. I also corrected the PR body's line counts to the actual numstat (production +51 / −92).

PR #12223, opened 2026-09-11
test: prune tests that do not earn their place (123 files, ~27.6k lines)

LionSR:
## Summary

The suite had grown large enough that a lot of it was pinned to churning internal seams, costing more in merge friction than it returned in safety. This applies the AGENTS.md "Testing discipline" bar strictly rather than leniently: a test earns its place only if it protects a consequential current contract, a difficult invariant, or a reproduced defect.

A 25-slice sweep read every test file in full and proposed removals against twelve named species (mock echo, language/library semantics, typecheck duplicates, literal re-declaration, roster pins, trivial plumbing, duplicated coverage, implementation-detail spies, tests of test infrastructure, retired behavior, cosmetic pins, over-parameterized `it.each`). A verifier then read each proposal against the code and, unlike the previous round, defaulted to **approving** — a rejection had to name the specific lost coverage and confirm no sibling suite carried it. 489 of 495 proposals survived; 6 were rejected.

Test files: **741 → 623.**

Largest single removals:
- `agent/AgentPromptContracts.vitest.ts` (744L) — ~270 lines of hardcoded rosters plus ~270 lines of `expect(systemPrompt).toContain('<sentence copied from the YAML>')`. Its em-dash/house-style scan AST-parses every `.ts` under `src/agent` and `src/tools` to enforce prose style — a lint written as a test.
- `cli/TuiValidatorArgs.vitest.ts` (173L) — spawns node subprocesses to test the argument parser of a dev-only PTY frame-capture harness.
- `desktop/TexraDiffView.vitest.ts` (161L) — JSDOM/Monaco/Worker scaffolding to assert the component hands Monaco the exact strings the test just assigned.

## Things worth calling out rather than leaving buried

1. **Codex caught a real over-removal, since fixed** (4955930). The `sync-remote-agents --apply` suite — sole end-to-end coverage of a mutating path, carrying reproduced defects #10407, #10408, #10409, #10411, #10412 — was deleted wholesale. The cause was mechanical: `describe.skipIf(cond)(title, fn)` contains an inner `describe.skipIf(cond)` call that also resolves to `describe` and by construction holds no cases, so the "remove describes left without a case" pass matched the wrong node. Restored to main's version; the four trims approved in that file were dropped rather than re-applied, because on a write boundary that can target the wrong Supabase project, keeping the coverage is the safer call.
2. **Blast radius was audited, not assumed.** A TypeScript AST pass over every file this branch touches compares each file's `it`/`test` titles against the approved-removal list. Against current `main`: **0 unapproved case removals**, and **123/123** file deletions on the approved list. That audit is what confirms (1) was the only place the bug landed.
3. **31 of 1,061 approved trims (2.9%) were not applied** — `it.each` titles whose source is an interpolated template literal, which the AST matcher skips rather than guess at. Nothing is broken; slightly less was removed than approved.
4. **`AgentPromptContracts` contained a vacuous security assertion.** Its setup-agent case asserted `settings.tools` excludes `set_api_key` — a tool that does not exist anywhere in the repo (only `unset_api_key` in `src/tools/registry.ts`). The invariant it appeared to protect has therefore never been covered. This PR did not create that hole; it exposed it.
5. **I am not quoting a test-count delta.** Three full runs reported an identical total across trees where a restored file demonstrably held 1 case and 14 cases, which cannot both be right. Every run is exit 0 with 623/623 files and zero failures, and the restored file passes standalone and in its project — but I could not reconcile the aggregate, so the figures below are diffstat-derived only.

## Issue acceptance gates

No issue. Owner-directed cleanup, with the acceptance bar taken from AGENTS.md "Testing discipline".

## Single source of truth

No ownership moved — this PR only deletes. The bar being applied is AGENTS.md "Testing discipline"; it remains the single statement of when a test earns its place.

90 of the 489 approved removals were species-7 duplicated coverage, each required to cite the sibling suite (`file:line`) that already carries it before approval. Representative:

- `settings/AgentSkillsSettings` + `settings/BashApprovalSettings` — byte-identical modulo one config-key constant; `writeSetting` store routing is already covered across the whole catalog in `shared/stateSettings.vitest.ts`.
- `agent/followUp/ToolUseRoundPrepNodeFollowUpTranscriptLog` — the fourth per-node copy of one generic invariant.
- `agent/AgentRegistry` round-count pins — covered against synthetic fixtures by `AgentYamlScanner.vitest.ts`.

No abstraction added. No pass-through layer introduced.

## Net elements (R6)

Measured as `git diff origin/main..HEAD`:

| | |
|---|---|
| Files | 487 changed — **123 deleted**, 364 modified |
| Net LoC | **−27,596** (16 insertions, 27,612 deletions) |
| Exported symbols (`^[+-]export`) | **−16**, +0 |

**All 16 insertions are `export`-keyword removals.** There are no other added lines anywhere in the diff.

Ratchets shrink rather than grow: `knip-baseline.json` 267 → 250, `host-agent-mock-baseline.json` 23 → 22, `pure-tier-kernel-suites.json` 35 → 29.

## Consumer counts (R8)

No emit path deleted or re-routed. 16 exports lost their last external importer when their only consumer (a test) was deleted; each is still used inside its own defining file, so each drops the `export` keyword rather than being baselined as dead. Grepped external-importer count after the change, excluding stale `packages/*/dist` artifacts — **all 0**:

`resumeEntryDescription`, `formatToolDescriptionForTui`, `agentProposalMetadataRows`, `bashCwdDisplayLine`, `infoPaneRequiredRows`, `formatWorkPlanReaderText`, `workPlanReaderLayout`, `resolveBrowserLaunch`, `dumbTerminalMessage`, `findCliToolDef`, `formatCliBoolean`, `MCP_STOP`, `SettingsAgentDirectoryEntry`, `LATEXDIFF_CHANGES_ONLY_SUBTYPE`, `resolveLatexdiffSubtype`, `GhCheckAnnotationSchema`

`MCP_STOP`'s only remaining repo-wide hit is a comment mention in `src/agent/modelHandlers/ModelHandler.ts`.

## Host impact

**Extension:** No behavior change. Four symbols in `src/latex/latexdiff/diffCommandExecutor.ts`, `src/tools/github/prTypes.ts`, `src/agent/types/StopReasonTypes.ts`, and `src/controllers/settingsView/SettingsAgentDirectoryController.ts` become module-private; all were already used only within their own file.

**Desktop:** No behavior change. No desktop source touched — only desktop test files.

**CLI:** No behavior change. Eleven symbols across nine `packages/cli/src` TUI and runtime files become module-private, all already used only within their own file.

## Merges from main

This branch absorbed 30 commits from `main` across two merges (49 conflicts, then 1). Content conflicts took main's version and re-applied this branch's approved trims on top, so main's Effect and `@effect/vitest` conversions survive intact. Two notes for reviewers:

- `config/ratchets/pure-tier-kernel-suites.json` (added by #12196) fails vitest **config load** — not a check — when it names a file that no longer exists. Six deleted suites were on it.
- The `GitignoreMatcher` conflict came from #12220, a *fix* commit that rebuilt the suite around itself. Its new regression test (`rereads the ignore policy on every call`) and the ENOENT case the review deliberately excluded from the trim both survive.

## Scheduled refactoring

None filed. Three candidates if you want them tracked:

1. The 31 unapplied `it.each` trims.
2. The `set_api_key` gap — the setup agent's key-leak invariant is uncovered, and covering it properly means asserting against the real tool registry rather than a prompt string.
3. The unexplained test-count aggregate noted above.

## Validation

Run on the pushed head:

- `npm test` — 623/623 files passed, exit 0 (exit code read directly, not through a pipe)
- `npm run typecheck` — 0 errors
- `npm run lint` — clean
- `npm run check:dead-code-ratchet` — OK
- `npm run check:effect-migration-ratchet` — OK
- `npm run check:guidance-refs` — OK
- `npm run check:browser-safe-utils` — OK
- `npm run check:tsconfig-paths` — OK
- `npm run check:package-contributes` — OK

Not run: `check:desktop-*` / `check:vsix-contents` (packaging checks; no packaged artifact or contribution surface is touched by this diff).

Note on how this was applied: trims were applied with a TypeScript AST pass that removes whole lines and bails rather than guess when a statement shares a line. Two earlier regex-based attempts mangled `it.each([...])(...)` and over-consumed closing braces; both were reverted rather than patched, which is why the diff is pure deletion.

https://claude.ai/code/session_01MRVpYXqrHrD3e5dbYwGBCW


LionSR:
Codex's P2 on `SyncRemoteAgents.vitest.ts` was correct. Fixed in 4955930 — the `--apply` suite is restored to main's version, 14 cases, with `#10407`, `#10408`, `#10409`, `#10411` and `#10412` intact.

**The cause was mechanical, not editorial.** The review pass approved trimming four cosmetic cases in this file and said in as many words that the link-state-preservation cases stay. A later cleanup pass — the one that removes `describe` blocks left without a case after a trim — then deleted the whole 286-line block. `describe.skipIf(cond)(title, fn)` contains an inner `describe.skipIf(cond)` call that also resolves to `describe` and by construction holds no test cases, so the "is this describe empty?" check matched the wrong node and walked up to remove the enclosing statement.

**Blast radius was audited, not assumed.** A TypeScript AST pass over every file this branch touches compares each file's `it`/`test` titles against the approved-removal list. Against current `main`:

- **0** unapproved case removals remain
- **123 / 123** file deletions are on the approved list

So this was the only place the bug landed. (A first audit against the pre-merge base flagged 15 files, but 14 of those were main's own test changes showing through a stale base — re-running against `origin/main` isolated it to this one file.)

**On the four approved trims here: dropped rather than re-applied.** They save ~60 lines out of ~28,000, and this is a write boundary that can target the wrong Supabase project or publish unusable catalog entries. Retaining the coverage is the safer call, so the file is now byte-identical to main.

This branch also merged the 28 commits that landed on `main` (49 conflicts). Worth flagging for reviewers: `config/ratchets/pure-tier-kernel-suites.json` fails vitest **config load** — not a check — when it names a file that no longer exists, and six deleted suites were on that list. Pruned to 29 entries.

Verification on the pushed head: `npm test` 623/623 files, exit 0; `typecheck` 0; `lint` clean; `dead-code-ratchet`, `effect-migration-ratchet`, `guidance-refs`, `browser-safe-utils`, `tsconfig-paths`, `package-contributes` all OK.

---

PR #12226, opened 2026-09-11
fix(tools): accept ssh:// GitHub remotes, null tool-input defaults, and gate Lean on its real host

LionSR:
## Summary

Three defects in the tool layer. Each fix deletes a duplicated authority.

**1. `ssh://` GitHub remotes were rejected by `github_subscription`.** The tool parsed `origin` with its own regex, which accepted only `git@github.com:` and `https://github.com/`. So `ssh://git@github.com/owner/repo.git`, a form git writes and users configure, failed with "origin remote is not a github.com URL". The CLI (`texra install-github-action`) already had a second, more permissive parser. There is now one: `parseGitHubSlug` in `src/tools/github/githubSlug.ts`. It anchors the whole URL to the supported forms: `https://[user@][www.]github.com/`, `git@github.com:` and `ssh://[user@]github.com[:port]/`. So `mygithub.com` and a GitHub-looking path on another host (`https://gitlab.com/group/github.com/owner/repo`) are both rejected. The GitHub tool and the CLI both import it, and the CLI copy is deleted. The existing `parseGitHubSlug` table gains the `ssh://` case and a rejection case for an embedded `github.com` path.

**2. Tool-input defaults used `.prefault()`.** Seven tool-input fields used it: `delegate_workflow`'s `contextFiles`/`mediaFiles`/`outputFiles`/`memories`, `lean_diagnostics.command`, `lean_inspect.column`, `lean_loogle.limit` and `send_to_terminal.label`. `.prefault()` substitutes only for `undefined`, but OpenAI-compatible providers (DeepSeek, Kimi, …) send an omitted optional field as `null`. Those calls were rejected at validation. CLAUDE.md requires `.nullish()` here, and `src/tools/core/inputSchema.ts` already provides `nullishWithDefault` for this case. All seven now use it. The output types are unchanged. `update_config`'s `target` has the same issue, but it sits in the open #12217 and is left there to avoid a conflict.

**3. VS Code advertised Lean tools that always fail.** The lean4 availability check passed on `extensionAvailable || lakeAvailable`. In VS Code the tools run only through the lean4 extension: the bridge returns no client and prompts for install when it is missing, and the direct `lake` pool is registered only on CLI and desktop. So a VS Code user with `lake` on PATH but no lean4 extension saw Lean as available, and every call failed. The check now needs the extension on VS Code and `lake` elsewhere. `lean_loogle` leaves the lean4 group: it only calls Loogle's public HTTPS API, so gating it on the extension or `lake` was wrong, both before and after this fix. The group has no toggle, so nothing user-visible is lost. It reads the process host through a new `getProcessSettingHost()` getter over the fact `initProcessSettingHost` already records. The status label and details say which piece is missing. One regression test covers VS Code with `lake` only.

Items `D07-9` and `D07-8` (coauthor-97's audit) and the `domain-tools-1` fallback (coauthor-21's audit).

## Net elements (R6)

- Files: 12 changed, 1 added (`githubSlug.ts`), 0 deleted
- `^[+-]export` symbols: +1 `getProcessSettingHost`; `parseGitHubSlug`/`GitHubSlug` move from `gitOps.ts` to `githubSlug.ts` (net 0)
- Class/interface/enum declarations: 0 net (`GitHubSlug` moved)
- Net LoC: see the diff stat

## Consumer counts (R8)

- GitHub remote parsers: 2 → 1 (the tool's `GITHUB_URL_RE`/`parseGitHubRemote` deleted; `parseGitHubSlug` has 2 production callers)
- `.prefault(` in tool-input schemas under `src/tools/**`: 8 → 1 (the remaining one is in #12217)
- `getProcessSettingHost`: 1 production caller (`externalToolDefs.ts`), 1 test

## Test plan

- [x] `npx vitest run` over the suites touching the changed tools, CLI install-github-action and external tool definitions
- [x] eslint on changed files

### DIFF
--- a/packages/cli/src/commands/installGithubAction.ts
+++ b/packages/cli/src/commands/installGithubAction.ts
@@ -1,6 +1,7 @@
 import { mkdir, writeFile } from 'node:fs/promises';
 import path from 'node:path';
 
+import { parseGitHubSlug, type GitHubSlug } from '@tools/github/githubSlug';
 import { toErrorMessage } from '@utils/errors/errorMessage';
 
 import { CliExitCode } from '../runtime/exitCodes';
@@ -15,10 +16,8 @@ import {
   git,
   isGitRepo,
   localBranchExists,
-  parseGitHubSlug,
   remoteUrl,
   repoRoot,
-  type GitHubSlug,
 } from '../runtime/gitOps';
 
 import { defineCliCommand } from './_helpers/defineCliCommand';
--- a/packages/cli/src/runtime/gitOps.ts
+++ b/packages/cli/src/runtime/gitOps.ts
@@ -55,21 +55,3 @@ export function localBranchExists(cwd: string, branch: string): boolean {
 export function ghAvailable(cwd: string): boolean {
   return gh(cwd, '--version').success;
 }
-
-export interface GitHubSlug {
-  readonly owner: string;
-  readonly repo: string;
-}
-
-/** Parse `owner/repo` from an https or ssh GitHub remote URL. */
-export function parseGitHubSlug(url: string): GitHubSlug | null {
-  const cleaned = url
-    .trim()
-    .replace(/\/+$/, '')
-    .replace(/\.git$/, '');
-  const match = cleaned.match(/github\.com[/:]([^/]+)\/([^/]+)$/);
-  if (!match) return null;
-  const [, owner, repo] = match;
-  if (!owner || !repo) return null;
-  return { owner, repo };
-}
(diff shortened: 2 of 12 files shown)


LionSR:
Both review findings are fixed in 892fa00ef3.
- **GitHub host anchoring** (Codex and texra-ai on `githubSlug.ts:19`): `parseGitHubSlug` now anchors the whole URL to `https://[user@][www.]github.com/`, `git@github.com:` or `ssh://[user@]github.com[:port]/`. So `https://gitlab.com/group/github.com/owner/repo.git` and `https://example.com/github.com/owner/repo` return null. The test table gains that rejection case. `ssh://`, `git@` and credentialed `https://` remotes still parse.
- **Loogle gating** (Codex on `externalToolDefs.ts:475`): `lean_loogle` leaves the lean4 group. It only calls Loogle's public HTTPS API, so the extension/`lake` gate no longer strips it. The group has no toggle, so nothing user-visible is lost.
Typecheck is green on all projects. The 10 affected suites pass (836 tests).

PR #12222, opened 2026-09-11
refactor(run): one run id, one run aggregate, one parent edge, one word (one run model S1)

LionSR:
Lands S1 of the one run model (`.agents/docs/proposed/architecture/2026-09-10-one-run-model.md` §3.1, §3.2, §3.10, §6 item 2): one run id, one run aggregate, one parent edge, one word.

## What changes

- **Identity.** `RunId` is the branded id (formerly `ExecutionId`). `StreamTabId`, the four synthetic stream prefixes (`bash@tool`, `workflow-script`, `codex@codex-sdk`, `claude@agent-sdk`), the `#` wire format, and every `split('#')` / `endsWith('#')` reader are gone. The display name those prefixes minted is the run's `RunIdentity`.
- **Aggregate.** The session database has one aggregate kind, `run`, keyed by the run id: one sequence counter, one ownership claim, the seq-1 rule on `run.start`. The `execution` aggregate, its second counter and claim, the reverse index, and the SQL that joined the two kinds are deleted.
- **Parent edge.** `run.start` carries `parent: { id, startCommit } | null`. `parentStreamId`, `parentExecutionId`, `isSubagent`, `isChildExecution`, and `background` are deleted from every schema and carrier; "is a child" is `parent !== null`. `run.detach` (no payload) replaces `setParentStream`.
- **RunScope** loses `streamId` and `agentName`; the name is `runIdentityName(identity)`.
- **One word.** `Execution*` and `Stream*` become `Run*` in every layer, row type (`run.record`, `run.launchLabel`, `run.description`, `run.report`, `run.result`, `run.workspaceFiles`, `run.workflow`, `run.removed`), request kind (`run.stop`, `run.delete`, `run.compact`), service, and file name (35 production files renamed). "Stream" survives only for a provider's token stream.

## What does not change

- The CLI NDJSON projection keeps its 0.40 record names and keys (`storageKey`, `setParentStream`, `parentStreamId`, ...) and is fed from the new fields. Retiring it is S5.
- The usage-log wire key stays `streamId` (the `log-usage` edge function stores it as `stream_id`). That contract is versioned with the edge function.
- `RunIdentity`, `AgentCategory`, `aggregateId`'s encoding, `Surface`, retention.
- 1.0 state lives under the `v1` namespace that #12108 introduced; release/0.40 never wrote there, so no old state is read or migrated.

## Rulings taken

[...]

PR #12228, opened 2026-09-11
refactor: one launch-failure presentation, park unpresented requests, drop the delivery contract and dead polish context (1.0 clean slate)

LionSR:
## Summary

Four verified candidates from the TeXRA 1.0 clean-slate sweep (session coauthor-dc, runtime slice). Two are user-facing bug fixes; each deletes the machinery that let the bug happen.

- **RT2-2 (bug fix): park a request the host does not present, instead of rejecting it.** Since #11883 the extension and desktop attach no `requestBashApproval` / `requestPlanApproval` / `requestAgentProposal` / `requestRetry` / `askUserQuestion`. `SessionHostInteractions.dispatch` settled any request whose host method returned `undefined` with its cancellation result. So on both GUI hosts, every bash command, plan, delegation proposal and user question auto-rejected, and every model retry auto-cancelled, before the fold's approval row could be answered. Now a stream-scoped request stays pending until `settleRequest`/`settleRetry`, a cancel, or disposal; a streamless one is still declined, now with a warning. The TUI's hand-written `park()` copy of this behavior and its `requestBashApproval` are deleted.
- **RT1-2 (bug fix): one presentation owner for launch failures.** Once launch-context assembly begins (model handler, run residency, trace attach, user vars), the run's terminal `result` event is the only presentation. The second `requestShowError` from the launch catch is deleted, so the extension and CLI chat no longer toast the same error twice. Model validation moves into `prepareAgentDefinition`, so a typo'd model shows only its instruction and never registers a phantom FAILED execution. The now-unread `AgentLaunchInput.suppressErrorNotification` and its two pass-through sites are deleted; `notifyLaunchFailure` has one caller left and is inlined there.
- **RT2-1: delete the presentation-delivery contract.** The boolean `emit` returned had one reader, `presentLaunchError`. For its two events every live host reports delivery, so the generic fallback only ever went back to the host that had just failed. `emit` is now fire-and-forget. `onReplayScheduled`/`onReplayNotDelivered`, `dispatchPresentationEvent` (inlined as `handlers[event](payload)` at its 3 callers), `PresentationDelivery` and `toPresentationDelivery` are deleted. A launch notice now claims its error unconditionally. The one host with a second surface, the extension, keeps a local fallback: if an instruction fails to render, it shows the error toast.
- **RT3-1: delete the dead file-context branch of instruction polish.** `polishTextWithAI` has one caller, and it passes no file context; the wire request cannot carry any (`{ kind: 'polish', text }`) since #11883. `FileContext`, `formatFileContext`, the nunjucks render and the two `{% if FILE_CONTEXT %}` template blocks are deleted. The prompt is now the YAML prefix plus the raw text, byte-identical to what rendered before.

Behavior changes to note:
- A delegated child with an unknown model now fails at delegation prepare, as a tool error to the parent, not as a registered FAILED child.
- An out-of-tree embedder that attaches only `{ cancel }` now sees stream-scoped requests park instead of auto-reject. In-tree hosts are unaffected: the SDK host implements `requestRetry`, and `approvalPromptsUnavailable` strips every approval-requiring tool.

## Net elements (R6)

- Files: 0 added, 0 deleted, 18 modified.
- `^[+-] *export` lines: +2 / -7 (+2 are signature rewrites of `renderPolishPrompt` and `createRecordingHost`; removed: `dispatchPresentationEvent`, `PresentationDelivery`, `toPresentationDelivery`, `FileContext`, plus one barrel `export {` block).
- class/interface/enum/type declarations: 0 added, 13 removed (diff-line count).
- `git diff --stat origin/main...HEAD`: 18 files changed, 227 insertions(+), 588 deletions(-).
- Production: +155 / -371 (net -216). Tests: +72 / -217 (net -145).

## Consumer counts (R8)

Grepped at origin/main, production code only:
- `dispatchPresentationEvent`: 3 callers (cliPresentationHost, desktopAgentExecution, agentEventListeners), all inlined.
- `toPresentationDelivery` / `PresentationDelivery`: 2 (desktop, extension).
- `onReplayScheduled` / `onReplayNotDelivered`: 1 producer (`presentLaunchError`), read only by the `SessionHostInteractions.emit` replay closure.
- Readers of `emit`'s delivery boolean: 1 (`presentLaunchError`). Every other emit site drops or only awaits it.
- `notifyLaunchFailure`: 2 call sites. The assembly one is deleted and the prepare one inlined.
- `FileContext` / `formatFileContext`: 0 outside `textEnhancement.ts`; `polishTextWithAI(text, fileContext)`: 0 callers pass a second argument.
- TUI `park`: 4 callers, all replaced by core parking.
- Hosts without the response-bearing request methods: extension (`ProgressViewProvider` attach) and desktop (`desktopAgentExecution` attach), both regressed by #11883.

## Skipped

- **RT2-3** (delete the CLI-only `setApprovalBypassState` push; `approval.policy` as the one emitter). The three NDJSON bypass records are public CLI output. Keeping them needs a new `approval.policy` diff projection in `sessionProgressSubscription.ts` (coauthor-97's slice, about 25 new lines), plus retargeting about 11 suites that observe bypass changes through the push: TuiApprovalRetry, RunExecution, RunProgressRenderer, PlanToolWorkPlanState, ChildStreamYoloInheritance, HumanPromptProgressEvents, ToolUseWaitNode, and others. That exceeds the spillover cap. Dropping the records instead needs an owner ruling on the public wire. It's still a verified candidate (peer coauthor-2c agent-runtime#3) for the CLI-slice owner or an owner ruling.
- **Doc not updated (ceded):** `docs/architecture/2026-07-26-embedding-the-agent-runtime.md` (lines ~185-188, ~383, ~499-501) still says a minimal `{ cancel }` host auto-rejects every request. After RT2-2 that holds only for streamless requests. The file is claimed by coauthor-dc/v1-platform-store-sdk (implementing), so it's left to that owner.

## Verified

All heavy gates ran through the machine-wide semaphore; exit codes are the real `GATE_EXIT` lines.

- G1 `prettier --write` on the 18 changed files: exit 0, every file unchanged.
- G2 `eslint` on the changed `.ts` files: GATE_EXIT=0.
- G3 `CI=true npm run typecheck` (every project: workspace, test-kernel, agent, llm, cli, trace-viewer, desktop): GATE_EXIT=0.
- G4a `vitest run --changed origin/main`: GATE_EXIT=1, from a single test: 202 files, 2689 passed, 1 skipped, 1 failed. The failure is `ChildRunLoop.vitest.ts > childRunLoop E2E fixtures > '$name' interrupts a real initial-turn loop and releases ownership once`, which asserts the execution handle has unregistered right after the follow-up owner releases.
  - Isolated reruns of the suite on this branch: 7 of 8 passed, and the one failure hit a different case (`ClaudeAgentSessions`, not `CodexThreads`). The suite passed 5 of 5 on an origin/main worktree run under similar load.
  - The suite drives a fake child strategy with no host interactions, so it never reaches `buildAgentLaunchContext`, `emit`, or request dispatch, the three paths this PR changes. I read it as a load-sensitive race in the release ordering, but a branch-side cause isn't ruled out.
  - Worth a rerun in CI.
- G4b `vitest run --project pure`: GATE_EXIT=0 (276 files, 4028 tests).
- G5 `check-dead-code-ratchet.mjs`: GATE_EXIT=0 ("no new findings"; no baseline entry named the removed exports). `check-effect-migration-ratchet.mjs`: exit 0, no count grew and no headroom to shrink.
- Targeted suites, run directly on the edited files: AgentLaunchContext, BundledPrompts, SessionInteractions, TuiApprovalRetry, PlanToolWorkPlanState all pass.
- Rebased onto origin/main at 6bd56508b7. The last two upstream commits (#12217, #12220) touch none of these files, so G3 and G4 were not rerun after that rebase.
- No committed `node_modules` symlink; `pnpm-lock.yaml` unchanged.
- Heads-up: open PR #12222 (one run model S1, 719 files) touches 10 of these 18 files. It's a rename, not these deletions, so whichever lands second rebases.

https://claude.ai/code/session_01VNfYGNNxcPzD3JRXs4DdoM

### DIFF
--- a/packages/cli/src/chat/tui/state/subscribeApprovals.ts
+++ b/packages/cli/src/chat/tui/state/subscribeApprovals.ts
@@ -3,9 +3,10 @@
 //
 // The runtime publishes `approval.requested` before it dispatches a request
 // here, and the fold lists it in `view.approvals` until `approval.resolved`;
-// the modal reads that list (`approvalQueue.ts`). A hook therefore parks:
-// its promise stays pending while the surface answers through a
-// `decision.*` runtime request, which settles the runtime's pending set.
+// the modal reads that list (`approvalQueue.ts`). A hook therefore returns
+// nothing (or has no method at all) and the runtime parks the request while
+// the surface answers through a `decision.*` runtime request, which settles
+// the runtime's pending set.
 // Three kinds still settle through their hook, because the runtime has no
 // request arm for them yet or their answer is host work: a tool edit (no
 // `decision.toolEdit` arm), a retry (its credential switch and
@@ -22,13 +23,10 @@ import PQueue from 'p-queue';
 import {
   currentSession,
   matchesCancelSelector,
-  type BashSettlement,
-  type HostBashApprovalRequest,
   type HostInteractionCancelSelector,
   type HostInteractions,
   type HostRetryInteractionOptions,
   type HostRetryRequest,
-  type HostUserQuestionRequest,
   type PlanApprovalResult,
   type ProposalResult,
   type RetryResult,
@@ -69,13 +67,7 @@ import {
   isCodingPlanQuotaRoute,
   type QuotaFallbackRouteId,
 } from '@shared/quotaFallbackRoutes';
-import {
-  type AgentProposalPermission,
-  type ExternalInquiryPermission,
-  type PermissionPayload,
-  type PlanApprovalPermission,
-  type RunId,
-} from '@shared/schemas';
+import { type ExternalInquiryPermission } from '@shared/schemas';
 import { subscribeToSignalChanges } from '@shared/signals';
 import { handleExternalInquiryAction } from '@tools/inquiry/inquiryActions';
 import { onAbort } from '@utils/core';
@@ -130,24 +122,6 @@ function maybeAutoSwitchRetry(
   return { accepted: true, disableQuotaRoute: route.id };
 }
 
-/** A request the surface answers through a `decision.*` runtime request:
- *  the hook's promise stays pending; the runtime's own settlement resolves
- *  the caller. A runless request has no fact for the fold to list, so
- *  nothing could answer it; it is declined rather than parked forever. */
-function park<T>(
-  kind: PermissionPayload['kind'],
-  runId: RunId | string | null | undefined,
-): Promise<T> | undefined {
-  if (!runId) {
-    logWarning(
-      'cli.tui',
-      `A ${kind} request named no stream; the TUI cannot present it.`,
-    );
-    return undefined;
-  }
-  return new Promise<T>(() => {});
-}
-
 /**
  * Create the typed approval pipeline for the active TUI session.
  */
@@ -185,20 +159,11 @@ export function createTuiHostInteractions(
         reservation.release();
       }
     },
-    requestBashApproval(request: HostBashApprovalRequest) {
-      return park<BashSettlement>('bash', request.runId);
+    requestPlanApproval() {
+      return settleByPolicy<PlanApprovalResult>(context);
     },
-    requestPlanApproval(request: PlanApprovalPermission) {
-      return (
-        settleByPolicy<PlanApprovalResult>(context) ??
-        park('planApproval', request.runId)
-      );
-    },
-    requestAgentProposal(request: AgentProposalPermission) {
-      return (
-        settleByPolicy<ProposalResult>(context) ??
-        park('proposal', request.runId)
-      );
+    requestAgentProposal() {
+      return settleByPolicy<ProposalResult>(context);
     },
     requestRetry(request, options) {
       return requestRetryInteraction(
@@ -208,15 +173,13 @@ export function createTuiHostInteractions(
         options,
       );
     },
-    askUserQuestion(request: HostUserQuestionRequest) {
+    askUserQuestion() {
       const denial = settleHumanInputDenial(context);
-      if (denial != null) {
-        return Promise.resolve<UserQuestionSettlement>({
-          action: 'reject',
-          reason: denial.reason,
-        });
-      }
-      return park<UserQuestionSettlement>('userQuestion', request.runId);
+      if (denial == null) return undefined;
+      return Promise.resolve<UserQuestionSettlement>({
+        action: 'reject',
+        reason: denial.reason,
+      });
     },
     async openExternalInquiry(request) {
       handleExternalInquiry(request, context, interactionOwner);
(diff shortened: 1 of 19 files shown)


LionSR:
Addressed the review findings in fdfb0fb968 (rebased onto origin/main, which now includes #12225):

1. **Launch notice dropped when a live host throws (blocking, claude + codex).** When the live host throws, `SessionHostInteractions.emit` now returns `false` instead of `undefined`. `presentLaunchError` claims presentation only when `emit(...) !== false`. A synchronous throw from `showAgentConfigBanner` (or `requestShowInstruction`) therefore leaves the error unclaimed, and the pre-registration catch in `prepareAgentDefinition` still shows the generic `requestShowError` toast. This runs in core, so it covers the extension, desktop, and CLI with no host edits. The emit return type stays `unknown`, so host adapters and test fakes are unchanged. One regression test restored in `AgentLaunchContext.vitest.ts`: "falls back to the generic toast when a live host throws on the missing-agent banner".
   Not covered: a *queued* notice whose later replay throws (the #10398 shape) is still only warn-logged, because it is claimed when it is queued.
2. **Vestigial `ExecuteAgentOptions.suppressErrorNotification`.** Removed from `ExecuteAgentOptions`. It is now declared on `RunAgentOptions`, and `runAgent` destructures it so it is no longer forwarded into `executeAgent`.

Gates: prettier (no changes), eslint on changed files 0, `CI=true npm run typecheck` 0. `vitest --changed origin/main`: 2689 passed, 1 failed. The failure is `ChildRunLoop.vitest.ts:466`, a timing flake under a contended run in untouched code, and that file passes 27/27 on its own.

LionSR:
Closed the queued-replay gap (#10398) in 20bccb1096: `emit` now takes a `fallbackMessage` carried in the queued replay. If presenting the notice throws, live or on replay, the same host shows `requestShowError`. The live-throw `!== false` check from the previous push is reverted in its favour. The regression test "emits the generic fallback when a queued banner replay throws synchronously" is restored. The branch is rebased onto main 18a657cf9a (#12222 names adopted). Repo-wide prettier, typecheck and `--changed` (2683 passed) are all green.

LionSR:
Rebased onto main ebae9b3223 as 8fd662936b. Main's pruning (#12223) is adopted in SessionInteractions.vitest.ts, and both restored regression tests are intact in AgentLaunchContext.vitest.ts. Repo-wide prettier, typecheck and `--changed` pass; the one `--changed` failure was a TUI timeout under load, and that file passes 45/45 on its own.

PR #9708, opened 2026-08-04
chore(deps): migrate to TypeScript 7 with nodenext resolution

LionSR:
## Summary

Migrates the TeXRA monorepo from TypeScript 6 to TypeScript 7 (the Go-native compiler) with full `nodenext` module resolution — no compatibility layers.

### Compiler

- `typescript` → `npm:@typescript/typescript6@^6.0.2` (API for typescript-eslint, ratchet tests, aliasUtils.mjs)
- `@typescript/native` → `npm:typescript@^7.0.2` (for all `tsc` invocations)
- All 6 workspace packages carry both aliases

### Config (6 tsconfigs)

- `module: nodenext` + `moduleResolution: nodenext` (root + extension)
- Dropped `baseUrl`, `ignoreDeprecations`, `module: commonjs`, `moduleResolution: bundler`
- Every wildcard `paths` alias now uses three variants: bare `*`, `*.ts`, `*/index.ts` (covers both nodenext extensionless resolution and bundler-mode explicit-extension imports)
- `--checkers 8` pinned in all 9 `tsc` script invocations for CI reproducibility

### Code changes

- `scripts/aliasUtils.mjs`: `loadAliases`/`loadAliasEntries` pick bare `*` variant for build tools; `deriveBuildPaths` simplified to filter-only
- `packages/agent/scripts/`: strip `./` prefix before node_modules filtering
- 1 test fixture updated: `subsystemEdgeRatchet` strips trailing `/index.ts`
- `.vscode/extensions.json` recommends `TypeScriptTeam.native-preview`

## Verification

| Check | Result |
|---|---|
| Typecheck (7 configs, TS 7) | 0 errors, ~21s |
| Tests (vitest, 824 files) | 8459 passed |
| Lint (TS 6 API) | clean |
| `build:fast` (esbuild + vite + vsce) | VSIX 8.22 MB |
| `check:tsconfig-paths` | in sync |
| Prettier | clean |

---

> [!NOTE]
> **Medium Risk**
> Wide monorepo compiler and path-resolution change affects every package typecheck and agent `.d.ts` emit; lint still uses TS 6, so rare tsc vs eslint divergence is possible until typescript-eslint supports TS 7.
> 
> **Overview**
[...]

PR #12277, opened 2026-09-11
simplify: move the CLI boundary and browser-safe-utils checks into ESLint, drop the effect ratchet debt-lane register, share CI setup, and prune dead lint/CI config (1.0 clean slate)

LionSR:
## Summary

v1 is a clean-slate release: no compatibility with old config shapes, flags or persisted data. This batch moves the guards that ran late, or checked the wrong thing, into ESLint, which already runs on every PR. It also deletes lint and CI scaffolding that gates nothing.

- **cli-host-imports-to-eslint / cli-host-boundary-to-eslint** (one finding, from two surveyors): `packages/cli/scripts/check-host-imports.mjs` (159 lines of regex over raw source) ran only at `npm prepack` and in the post-merge `runtime-validation` job, so a CLI boundary breach merged green. It is now ESLint, and runs on every PR:
  - `no-restricted-imports` bans `vscode`, `electron` and their subpaths in `packages/cli/src/**`.
  - `no-restricted-properties` bans `process.argv`, `process.env` and `process.cwd` outside `runtime/cliContext.ts`.
  - It also bans `console.log/error/warn` and `process.exitCode/stdout/stderr` outside `bin/texra.ts`, `runtime/logSinks.ts` and the Ink launchers.
  - One boundary block per role; the `isTTY` carve-out is scoped to `cliContext.ts`.
  - The five dead allowlist entries are gone.

  Deleted with it: the `check:architecture` script, its step in the CLI `build` chain, and `validate:run`'s `typecheck` / `check:architecture` / `smoke:react-compiler` pre-steps. The PR `checks` job already runs typecheck and smoke on the same push.
- **browser-safe-utils-guard-to-lint**: `scripts/check-browser-safe-utils.mjs` (650 lines) recomputed the frontend `@utils` closure and regex-parsed prose counts out of CLAUDE.md and AGENTS.md. It never checked for Node built-ins. Replaced by two ESLint rules:
  - The webview frontends may import at runtime only the five `BROWSER_SAFE_UTILS` (a regex, so no gitignore-negation traps).
  - Those five files get `import-x/no-nodejs-modules` and may import only each other, by alias. Relative imports are banned so the allowlist sees every edge, which keeps the set closed.

  `src/utils/core/index.ts` now re-exports its two siblings by alias. Also removed:
  - the dependency install in the `guidance-refs` job
  - the unused `check:browser-safe-utils` npm alias
  - the "other 60 modules" count prose, which 10 unrelated PRs had to edit
- **effect-ratchet-debtlanes-register**: the `debtLanes` register had exactly the same keys as the `Effect.run*` row. `diffRows` already fails on a new file, and `--update` never adds one, so the register added only hand-written lane labels. Deleted:
  - the register, its `--update` admission gate and lane rebuild
  - the check-mode lane report, the unassigned and stale-lane failures, and their self-tests
  - the `debtLanes` baseline key

  The `@adapter-until` text scan moves to ESLint `no-warning-comments` (`location: 'anywhere'`). The baseline was regenerated with `--update`.
- **ci-setup-composite-action**: `.github/actions/setup-workspace` now owns the pnpm + Node + frozen-lockfile install block that six ci.yml jobs copied verbatim. It takes two inputs: `node-version`, which runtime-validation pins to `22.16.0`, and `ignore-scripts`, which docs-check uses. The checks job's Deno setup moved ahead of it. `guidance-refs` now needs only `setup-node`.
- **eslint-prefer-alias-dead-for-extension / eslint-alias-config-shadow** (one finding, from two surveyors): since #9708 the extension alias root has had no `paths`, and it shadowed the root map. As a result `prefer-alias-for-deep-relative-imports` never fired anywhere in `packages/extension`. One sorted root alias list replaces `ALIAS_CONFIGS`, `aliasConfigForRoot`, the never-triggering dedupe Map and `aliasEntriesForFile`, and `INTERNAL_ALIAS_NAMES` reuses it.
- **eslint-noop-overrides**: deleted the following (all measured at 0 violations):
  - the unicorn `off` entries for rules that were never enabled
  - `semi`/`curly` off, and `@stylistic/semi`, which Prettier owns
  - the `no-case-declarations`, `no-useless-catch` and `prefer-as-const` offs
  - a `ban-ts-comment` override that restated the defaults (it also allowed `@ts-ignore`, of which there are 0)
  - the dead `@/**`, `~/**` and `${alias}/*` pathGroups
  - stale migration comments

  The 11 `warn` rules, which CI never failed on because there is no `--max-warnings`, are now `error`. The unused `no-await-in-loop` disable in `validate-tui.mjs` is gone. `packages/trace-viewer/src` was configured but never linted; it lints clean and now joins the lint script.
- **bot-workflow-dead-steps**:
  - the always-true "Check trusted provider wrapper" step, deleted in 3 workflows, and its 7 `if:` fragments
  - the review-switch step, which the job `if:` already covers
  - the `BOT_PAT` auto-PR step (`gh secret list` has no `BOT_PAT`)
  - the `ANTHROPIC_BASE_URL`, `CLAUDE_OPUS_MODEL`, `CLAUDE_SONNET_MODEL` and `DEEPSEEK_BASE_URL` pass-throughs; they were unset or equal to the external action's defaults
  - the one-cell review matrix; branch protection is off, so no check name is pinned
  - the `ci-sync-*` headers and the unused `github-read` tool preset

  In `claude.yml`, `concurrency` moved to job level. Bot comments that fail the job `if:` no longer queue, and they no longer cancel pending `@claude` runs.
- **Extras at publish**: none.

## Skipped

- **effect-ratchet-debtlanes-register (baseline `semantics` copy)**: kept. Every `config/ratchets/*.json` baseline carries its own doc string, and the script's equality check keeps it in sync. `--update` regenerated it without the lane and marker clauses.
- **eslint-noop-overrides (`@stylistic/eslint-plugin` devDependency)**: the rule and the import are gone from `eslint.config.mjs`. Dropping the dependency from `package.json` and `packages/extension/package.json` (plus the lockfile) belongs to p3-deps, which owns those dependency blocks. See Handoffs.
- **cli-host-boundary-to-eslint (VS Code-free-zone route)**: I did not add `packages/cli/src` to `VSCODE_FREE_ZONE_DIRS`, because that would also require the `dependencyDirection.vitest.ts` mirror and the CLAUDE.md zone list. A self-contained `no-restricted-imports` block covers `vscode` and `electron` instead.
- **Dynamic `import('vscode')` / `require('electron')` in `packages/cli/src`**: the deleted regex script also matched these; `no-restricted-imports` covers only static imports and re-exports. There are 0 such uses today. A `no-restricted-syntax` selector cannot live in the CLI block, because the later `<wa-icon>` block also matches `packages/**` and flat config replaces that rule's options per file. Closing the gap means folding the selectors into that shared block, left as a follow-up.
- **Inline `isTTY` disables in `packages/cli/src/runtime/cliContext.ts`**: the file is claimed by coauthor-97 (implementing), so the carve-out lives in the config, scoped to that file with a comment. The trade-off is that `cliContext.ts` may still touch `process.stdout` and `process.stderr`; `console.*` and `process.exitCode` stay banned there.

## Handoffs

- **p3-deps**: remove `@stylistic/eslint-plugin` from the root `package.json` and `packages/extension/package.json` devDependencies, and regenerate `pnpm-lock.yaml`. `eslint.config.mjs` no longer imports it.
- **p4-release-build**: adopt `./.github/actions/setup-workspace` in `release.yml` (2 sites), `desktop-package.yml` (3) and `version-bump.yml` (1). Each of those jobs already checks out first.
- **p6-supabase-remote-agents**: `remote-agents-sync.yml` carries the same setup block. Adopt the action there with `ignore-scripts: 'true'`, unless that workflow is being deleted.
- **Effect sessions**: sibling Effect PRs that touch `config/ratchets/effect-migration-baseline.json` must regenerate it with this script (`node scripts/check-effect-migration-ratchet.mjs --update`). Running the old script re-adds `debtLanes` and the old `semantics` text. #12273 (RT1-1) touches this baseline; whichever PR lands second regenerates it with `--update`.
- **#12268**: it also edits `packages/cli/scripts/validate-run.mjs`, but its hunks (around line 986) don't overlap this PR's 3-line `buildValidationBundle` deletion (line 1193), so git merges them cleanly.

## Net elements (R6)

- `git diff --shortstat origin/main` (0eb57310b8): **23 files changed, 307 insertions(+), 1488 deletions(-), net −1181**
- Files deleted (2): `packages/cli/scripts/check-host-imports.mjs` (159 lines), `scripts/check-browser-safe-utils.mjs` (650 lines)
- Files added (1): `.github/actions/setup-workspace/action.yml` (29 lines)
- `^[+-]export`: +1 / −1. The one change is the `KeyedMutex` re-export in `src/utils/core/index.ts`, which switched from a relative path to an alias with no symbol change.
- `class` / `interface` / `enum` declarations: +0 / −0
- Top-level `function` declarations: +2 / −32. All of the removals are script machinery: the alias-config helpers, the debt-lane and marker functions, and the whole browser-safe and host-import scanners.
- `scripts/check-effect-migration-ratchet.mjs`: 1513 → 1220 lines. `.github/workflows/ci.yml`: −111 lines net.

## Consumer counts (R8)

Repo-wide `rg -l --hidden -F`, excluding `node_modules`, `.git`, `pnpm-lock.yaml` and the historical `.agents/docs/**` proposal and archive trees:

| Deleted item | Remaining consumers |
| --- | --- |
| `check-host-imports` | 0 |
| `check:architecture` | 1: `docs/dev/audits/2026-07-21-agent-sdk-readiness-checkpoint.md`, reworded to "since folded into ESLint" |
| `check-browser-safe-utils`, `check:browser-safe-utils` | 0 |
| `debtLanes`, `UNASSIGNED_LANE`, `runBoundaryDebtLanes`, `unassignedDebt`, `staleDebtLanes`, `checkAdapterMarkers`, `selfTestBoundaryAndMarkers` | 0 |
| `aliasConfigForRoot`, `aliasEntriesFromTsconfigPaths`, `aliasEntriesForFile`, `ALIAS_CONFIGS`, `extensionPackageDir` | 0 |
| `@stylistic/semi` | 0 |
| `trusted-provider-wrapper`, `claude-switch`, `BOT_PAT`, `ci-sync-version`, `github-read`, `CLAUDE_OPUS_MODEL`, `CLAUDE_CODE_REVIEW_PROVIDERS` | 0 |
| `ANTHROPIC_BASE_URL` | 1: `src/test-kernel/llm/AnthropicMessages.vitest.ts`, an unrelated SDK env test, not the workflow variable |
| `@ts-ignore` in `src/` and `packages/` | 0, so the stricter default `ban-ts-comment` adds no errors; the 3 `@ts-expect-error` all carry descriptions |

## Verified

All results below are from the worktree rebased onto `origin/main` 0eb57310b8 (HEAD 8af6c51f27). The rebase was clean, with no conflicts. After it I ran a real `CI=true corepack pnpm install --prefer-offline` and restored `pnpm-lock.yaml` from `origin/main`.

- `npm run lint` (full repo, all 6 segments, including the newly added `packages/trace-viewer/src`): **pass** (exit 0).
  - An earlier pre-rebase run failed with 3 errors in `packages/cli/src/commands/_helpers/fetchSilencer.ts`. That file swaps and restores `console.error` to filter a library's network stack logs; the old regex script matched only `console.error(` calls, so it never saw this. It is not a boundary breach, so it got an override scoped to that one file, with a comment; every other CLI I/O restriction still applies there.
- `npm run typecheck` (`CI=true`, all 7 programs): **pass**.
- `node scripts/check-effect-migration-ratchet.mjs`: **pass**. 13 rows; no count grew and there is no headroom. Rerunning `--update` after the rebase changed nothing. The baseline is `{semantics, rows}` and no longer carries `debtLanes`.
- `npm run check:dead-code-ratchet`: **pass**, 229 combined findings against 234 baselined, no new ones. The 5 stale entries (`packages/cli/src/chat/tui/forms/ConfigForm.tsx`) come from upstream, not from this PR, and are left for their owner.
- `npm run test:pure` (`FORCE_COLOR=0`): **pass**, 212 files and 2404 tests.
- YAML parse of `ci.yml`, the new `action.yml` and the 5 bot workflows with the `yaml` package: **pass**.
  - The composite action has inputs `node-version` and `ignore-scripts`, and its only `run` step has `shell: bash`.
  - Every job that calls `./.github/actions/setup-workspace` (checks, test, build, webview-smoke, runtime-validation, docs-check) checks out first.
- `git diff --check origin/main`: clean. `prettier --check` on every changed file: clean (the 2 deleted scripts have no file to check).
- An untyped `ESLint.lintText` probe of every new rule (pre-rebase; the rebase touched no lint config) gave the expected results:
  - `@adapter-until` is flagged.
  - In the CLI, `vscode`, `electron/main`, `process.env/argv/cwd/exitCode/stdout` and `console.log` are flagged, and `console.info` is not (parity with the old script).
  - `cliContext.ts` may read `process.env` and `process.stdout.isTTY`, but not `console.log` or `exitCode`.
  - `runChatTui.tsx` is free; `bin/texra.ts` may use `console` and `stdout` but not `env`.
  - The five utils: `node:fs` and `fs` are flagged, as are `@shared/x` at runtime, `../core`, `@utils/files/workspace` and `@webview/x`. Allowed: `import type` from `@shared`, `@utils/core`, `@utils/core/keyedMutex`, `@utils/text/stringUtils`, and `pretty-bytes`.
  - Frontends: all five allowed modules pass; `@utils/text/stringUtilsX`, `@utils/core/other` and `@utils/files/workspace` are flagged.
  - Extension: a deep relative import of `src/utils/core` from `packages/extension/src/commands/` now reports prefer-alias; before this change it reported nothing.
- Pre-rebase only: `npx vitest run --changed origin/main` (622 files, 6594 passed), `node scripts/check-guidance-refs.mjs` and `node docs/scripts/check-root-docs.mjs`, all **pass**.
- This PR's first CI run is the live test of the composite action and the edited bot workflows.


LionSR:
Re the Codex P2 (dynamic imports in the browser-safe boundary): checked at the rebased head, and it isn't a live defect.

- None of the five browser-safe modules (`src/utils/core/index.ts`, `core/keyedMutex.ts`, `errors/errorMessage.ts`, `files/pastedImageName.ts`, `text/stringUtils.ts`) contain `import(`, `require(`, `require.resolve(` or `import.meta.resolve`.
- No browser frontend (`packages/extension/src/{webview,progressView,settingsView}/frontend`, `packages/desktop/src/renderer`, `packages/trace-viewer/src`) dynamically imports a `@utils/*` or relative `utils/` module.

It is a real narrowing of what the lint rule can see: `no-restricted-imports` only visits static import/export declarations. That is the same gap already recorded under "Skipped" for the CLI `vscode`/`electron` boundary. Closing it means adding `ImportExpression`/`require` selectors to the shared `no-restricted-syntax` block (flat config replaces that rule's options per file), so it stays a follow-up for both boundaries and not part of this PR.

Also in this push: rebased onto the latest main, and fixed the `guidance references` job. It had failed with "Unable to locate executable file: pnpm": `actions/setup-node@v5` turns on package-manager caching by default when `package.json` has a `packageManager` field, and that job no longer sets up pnpm. It now sets `package-manager-cache: false`.

PR #12280, opened 2026-09-11
simplify: inherit root tsconfig paths and delete the path-sync machinery and dead aliases/config keys (1.0 clean slate)

LionSR:
## Summary

v1 is a clean-slate release: nothing needs to stay compatible with old config shapes. This PR deletes config that regenerates a copy of one fact, points at something deleted, or restates a default. Every deletion was re-derived at the current HEAD before it was made.

- **inherit-root-paths-delete-sync**: `packages/desktop/tsconfig.json` and `packages/trace-viewer/tsconfig.json` now extend `../../tsconfig.json` directly. `tsconfig.build.json` drops its 89-line filtered `paths` copy and inherits the root map. With no `baseUrl`, TypeScript resolves `paths` against the config that declares them, so the inherited map resolves the same files; `packages/cli` already works this way. That leaves the generator with nothing to generate, so it goes too:
  - `packages/desktop/tsconfig.paths.json` (169 lines)
  - `scripts/sync-tsconfig-paths.mjs`
  - the `sync:`/`check:tsconfig-paths` scripts and their CI step
  - the derive/load/`pathTargetExists` helpers in `scripts/aliasUtils.mjs` (its only remaining caller absorbs `parseJsonc`)
  - `AliasMapGeneration.vitest.ts`

  The build-map filter was the only fence against `@test/*`, so `'@test/**'` joins the eslint host-layer pattern.
- **dead-aliases-and-node-modules-mappings** and **dead-common-state-alias**: the same finding from two surveyors, implemented once.
  - Removes `@common/state/*`, whose directory #11883 deleted (0 importers), and the `@common/webview/*` wildcard (0 importers; all importers use bare `@common/webview`).
  - Removes the five mappings left over from `moduleResolution: node`: `turndown-plugin-gfm`, `data-uri-to-buffer`, and `@openrouter/sdk`, `/models`, `/models/errors`. With `nodenext`, TypeScript resolves the same files without them. `src/types/ambient.d.ts` still declares `turndown-plugin-gfm`.
  - Trims the eslint pattern, the `dependencyDirection` specifiers, the `subsystemEdgeRatchet` case (it now pins only `@common/webview`), and the `.d.ts`/`node_modules` special cases in `aliasUtils` and `rewrite-declaration-aliases`.
- **knip-config-dead-keys**:
  - Drops the 9 patterns knip itself reports as redundant: the 6 `packages/llm` entries (its `exports` already imply them), extension `src/extension.ts!`, desktop `src/renderer/main.ts!`, and the `vite-env.d.ts` ignore.
  - Drops the no-op `ignoreDependencies`/`ignoreBinaries` keys. Both knip runs pass `--include files,exports,types,duplicates`, so the keys change nothing.
  - Deletes the uncalled `check:dead-code` script; `check:dead-code-ratchet` is the authoritative check. find-simplification SKILL.md now points at the ratchet.
- **test-changed-passwithnotests-default**: drops `passWithNoTests: false`. Vitest only turns `passWithNoTests` on for `--changed` when the option is undefined, so the explicit `false` makes `test:changed` exit 1 whenever a change selects no suite. Also drops the redundant `environment: 'node'`.
- **native-cleanup-executable-pool-orphan** (`vitest.config.mjs` half): deletes the `executablePool` custom pool and its `TEXRA_TEST_NODE` switch, plus the `fork`/`once` imports. The only producer of that variable was the `native-cleanup.yml` workflow, which the addon removal (ef3bc9d18a) deleted.
- **format-editor-config-stale**:
  - `.prettierignore` loses the lines `.gitignore` already covers (checked with `prettier --file-info`) and the `.github/workflows/` exclusion, which was justified only by the deleted format-pr workflow. Workflow YAML already passes `prettier --check`.
  - `.prettierrc` keeps only `singleQuote`; `semi`, `tabWidth` and `trailingComma: "all"` are Prettier 3 defaults.
  - Deletes `.gitattributes`, which only targeted the gh-aw `*.lock.yml` files; there have been none since #8942.
  - Drops two unused VS Code recommendations (`tsl-problem-matcher`, `extension-test-runner`) and a stale TypeScript-version comment.
- **Knip baseline**: after rebasing onto origin/main 834b3e89d4, the ratchet reported 5 rows that #12265 resolved (`createCliConfigFormProps`, `buildEnumItems`, `formatSettingValue`, `settingDisplayName`, `settingStoreLabel`). They are removed from `config/ratchets/knip-baseline.json` in a separate commit. The baseline goes from 234 to 229 rows.
- Doc mentions updated: the pr1-run-ledger proposal §1.3 and its not-verified list, the embedding note, and the stale `common/state/` zone in the code-review SKILL.md.

## Skipped

- **inherit-root-paths-delete-sync, the `packages/extension/tsconfig.json` part** (delete the file and point `esbuild.config.mjs:46` at `../../tsconfig.json`): the refutation holds at HEAD. `eslint.config.mjs` `ALIAS_CONFIGS` calls `aliasConfigForRoot(packages/extension)`, which reads that file through `resolve(rootDir, 'tsconfig.json')`. No literal filename appears, so the verifier's grep missed it. Deleting the file made ESLint fail to load its config (`ENOENT … packages/extension/tsconfig.json`). The file and the esbuild line are restored. That alias config is p5's `eslint-alias-config-shadow`; see Handoffs.
- **knip-config-dead-keys, the `@lit-labs/ssr` dependency**: owned by p3-deps (`zero-consumer-direct-deps`), because it needs changes to both manifests and the lockfile.
- **native-cleanup-executable-pool-orphan, the `sessionEvents.vitest.ts` half**: the ledger still shows that file as claimed pr-open by #12246. Without a producer the guard does nothing, and deleting the pool does not break it.
- **The optional `existsSync` assertion in `loadAliasEntries`**: not added. It would be a new guard, and a dead alias target already fails tsc and Vite as soon as anything imports it.
- **Dropping `--config vitest.config.mjs` from the test scripts**: 0 LOC and optional. `package.json` is shared with p3 and p5, so I kept my edits there minimal.

## Handoffs

- **p5 (`eslint-alias-config-shadow`)**: once the extension `ALIAS_CONFIGS` entry is gone, delete `packages/extension/tsconfig.json` and point `packages/extension/esbuild.config.mjs:46` at `'../../tsconfig.json'`. p4 owns the esbuild config. The root config has the same `compilerOptions`, and the `.vscodeignore` `tsconfig.json` lines become moot.
- **p3 (`zero-consumer-direct-deps`)**: `@lit-labs/ssr` in root `package.json` and `packages/extension/package.json`.
- **Whoever owns `sessionEvents.vitest.ts` after #12246**: delete the `NATIVE_TARGET`/`TEXRA_TEST_NODE` `beforeAll` guard (≈ lines 218–230) and `beforeAll` from the vitest import. Both variables have 0 producers.
- **p4 (`trace-viewer-single-build`)**: edits the knip trace-viewer entry and the `.prettierignore` trace-viewer lines. My edits sit on adjacent lines but do not overlap them.

## Net elements (R6)

- `git diff --shortstat origin/main` (origin/main 834b3e89d4): 28 files changed, 62 insertions, 738 deletions (net **−676**). The batch commit accounts for −646 of that; the 5 knip baseline rows account for −30.
- Files deleted (4): `.gitattributes`, `packages/desktop/tsconfig.paths.json`, `scripts/sync-tsconfig-paths.mjs`, `src/test-kernel/scripts/AliasMapGeneration.vitest.ts`. Files added: 0.
- `^[+-]export`: −6 / +0 (`loadRootPaths`, `deriveDesktopPaths`, `BUILD_EXCLUDED_ALIASES`, `deriveBuildPaths`, `parseJsonc`, `pathTargetExists`).
- class/interface/enum declarations: −1 / +0 (`AliasUtilsModule` in the deleted test).

## Consumer counts (R8)

Counts come from `rg --hidden` after the change, excluding `node_modules`, `.git`, `.agents/docs/archived` and the dated `docs/dev/audits` snapshots:

| Deleted item | Remaining hits |
| --- | --- |
| `tsconfig.paths` | 0 |
| `sync-tsconfig-paths` | 0 |
| `check:tsconfig-paths` / `sync:tsconfig-paths` | 0 |
| `AliasMapGeneration` | 0 |
| `deriveDesktopPaths`, `deriveBuildPaths`, `loadRootPaths`, `BUILD_EXCLUDED_ALIASES`, `pathTargetExists`, `parseJsonc` | 0 |
| `@common/state` | 0 |
| `'@common/webview/…` import specifiers | 0 importers (2 hits are the kept fence tests) |
| `TEXRA_TEST_NODE` / `NATIVE_TARGET` / `executablePool` | 0 producers (4 hits, all in the skipped `sessionEvents` guard) |
| `passWithNoTests` | 0 |
| `check:dead-code` (not the ratchet) | 0 |
| `*.lock.yml` | 0 (`git ls-files`) |
| `tsl-problem-matcher` / `extension-test-runner` | 0 |
| `ignoreDeprecations` | 0 |
| `node_modules/@openrouter` / `node_modules/data-uri-to-buffer` | 0 |

`@test/*` importers inside the new lint zone (`src/**` excluding test-kernel, `packages/agent/src`, `packages/llm/src`): 0. The 4 existing importers are in `packages/desktop/design-harness` and `packages/cli/scripts`, outside that zone.

## Verified

The final gates ran on the rebased branch at origin/main 834b3e89d4, each through the shared gate semaphore. Main moved twice during publishing, so the branch was rebased again and every gate below was rerun on the final tree. The install was real (`CI=true corepack pnpm install --prefer-offline`), and `pnpm-lock.yaml` was restored from origin/main.

- `npm run check:dead-code-ratchet`: **pass**. 229 combined findings against 229 baselined, no new findings. The first run after the rebase flagged the 5 resolved #12265 rows. They were dropped, and every later run, including the one on this tree, matched exactly. Removing the 6 `packages/llm` entry patterns and the other knip-flagged keys exposed no new finding.
- `CI=true npm run typecheck`: **pass**. That covers workspace, test-kernel, agent (the declaration build, now inheriting the root paths), llm, cli, trace-viewer and desktop (both now extending the root directly).
- `npm run lint`: **pass**.
- `CI=true npm run compile:fast`: **pass** (esbuild extension, trace-viewer assets, Vite webviews).
- `FORCE_COLOR=0 npm run test:pure`: **pass**, 210 files / 2397 tests.
- `prettier --check` on every file changed against origin/main: **pass**.
- `git diff --check origin/main HEAD`: clean. There are no symlinks in the index.
- Earlier, before this rebase: `FORCE_COLOR=0 npx vitest run --changed origin/main` passed (619 files, 6580 passed, 1 skipped). Repo-wide `npm run format:check` and `prettier --check .github/workflows` also passed. None of these was rerun on this tree.

Failures on the way, recorded honestly:

- The first lint run failed to load the ESLint config with `ENOENT packages/extension/tsconfig.json`. It was caused by this change; I fixed it by restoring that file (see Skipped).
- Before the rebase, `test:pure` showed 22 failures and `vitest --changed` showed 144. The machine's load average was about 43 on 10 CPUs. Every pure-tier failure was a "Test timed out" error; `--changed` also had 8 assertion errors in timing-sensitive CLI, BashTool and JsonRpc suites that nothing here touches. The full gate set above passed cleanly once load dropped.
- Not exercised: `test:changed` on a diff that selects zero suites (the `passWithNoTests` fix). It rests on the Vitest source behaviour described in the candidate: `resolved.passWithNoTests ??= true` when `--changed` is set.


LionSR:
Addressed the inline doc note in 74c568d487: `packages/llm/tsconfig.json` has no `extends` (verified), so the sentence in `2026-09-08-pr1-run-ledger-foundation.md` now scopes to consumer package tsconfigs and names `packages/llm` as the standalone alias target. Branch rebased onto origin/main 6df7136bf1; typecheck, compile:fast, lint, test:pure and the dead-code ratchet (229 vs 229 baselined) all pass at this head.

PR #12281, opened 2026-09-11
refactor(effect): injection cohort A: Secrets, AppState, SetupPlatform, ToolInjections as services provided once per process

LionSR:
Cohort A of the Effect-native injection plan (`.agents/docs/proposed/architecture/2026-09-10-effect-native-injection-context-pipelines.md` §6 steps 1 to 5): the process-lifetime capabilities become Effect services provided once per process root, and the ambient `platform()` lookup leaves the files that read them.

## What changes

- **Four tags, one provision point per process.** `Secrets` (`src/platform/secrets.ts`), `AppState` (beside `StateStore` in `src/platform/interfaces.ts`, still synchronous), `SetupPlatform` (`src/tools/setup/platform.ts`), and `ToolInjections` (`src/agent/runtime/toolInjection.ts`) are `Context.Service` keys with static Layers. One `processServicesLayer` inside `installProcessRuntime` provides them; each composition root (extension, desktop, CLI, the agent package, the test kernel) passes its own root-owned stores. The desktop and CLI roots pass store thunks because those stores open on the runtime being installed; the thunk closes over the root's local, never over `platform()`. No host imports the `ToolInjections` tag; it is composed inside the runtime from `AppState`.
- **Call sites read the tag.** Effect programs `yield*` it; non-Effect files take the store or secrets as a parameter threaded from the root (R-3: a conversion never adds a `platform()` lookup to a clean host file). `setupSecrets` is dissolved: the setup tools read `Secrets` directly and call the credential helpers themselves.
- **The shared config chokepoint** (`src/utils/config/constants.ts`, `platformSettings.ts`, `providerConfig.ts`) takes the state store as an argument; the CLI, extension, and desktop callers already held it.
- **Two ratchet seals** the plan's step 1 asked for: `ambient:asyncLocalStorage` and `effectRuntime()` rows in `scripts/check-effect-migration-ratchet.mjs` and the baseline, seeded with `--update`, so neither can grow again while later steps shrink them.

## Ratchet effect

`platform()` row: 103 sites in 49 files on main, 45 sites in 23 files after this PR. Three `globalState` reads the manifest scheduled here stay: the settings chokepoint (`readPlatformSetting`, 31 call sites ending in zero-argument synchronous leaves reached by hundreds of files), `providerConfig.read` (reached from the retiring model-handler hierarchy), and the agent roster controller; they belong to the StateStore retyping step the design already files separately. Rows that reach zero are deleted, none widened; `Effect.run*` unchanged; no `@adapter-until` markers. Two new seal rows (`effectRuntime()` 137 files, `ambient:asyncLocalStorage` 94 files) stop those counts from growing while later steps shrink them.

## Size, honestly

261 files, plus 5,598 minus 2,576 against main. About 400 lines are the two seal rows and their script; the rest is the cost of typed provision replacing an ambient locator: Layers at five composition roots, stores threaded as parameters through call chains that were previously implicit, and test suites that now provide the tags instead of relying on the ambient host. A simplifier pass already removed about 280 lines of duplicated fixtures, re-packed store bags, and a service union restated 17 times; what remains is the plumbing the design asks for, not padding.

## Corrections to the design doc, measured on main

`UsageLogService` is not a `Context.Service` (the exemplar used here is `ProcessIdentity` and the `Sessions` LayerMap); step 4's "minus 10" did not exist (those files had no `platform()` sites); `LogSink` cannot be a tag yet because `createLog` is synchronous in 164 files; `StateStore` retyping is a separate item; there is no `platform().config`. Excluded by the 1.0 plan §5: readers inside retiring subsystems (the model-handler hierarchy's one `secrets` site).

## Review and verification

Five adversarial lenses with two skeptics per serious finding over the diff against main; confirmed findings fixed in the last commits. Typecheck across all seven projects, the full suite, lint, and the three ratchets green on the final commit. No `as any`, `@ts-ignore`, `catch {}`, aliases, adapters, or re-export shims added.


LionSR:
Review from another session (`claude-effect-waves`), prompted by the owner asking whether cohort A is genuinely Effect-4-native or is placing tags over adapters. Four findings, two of them small enough to fix in this PR.

## 1. `Secrets` and `AppState` carry the Promise port verbatim

Both tags take the existing port as the service shape — `Context.Service<AppState, StateStore>` includes `update(key, value): PromiseLike<void>` — and both layers are pure pass-throughs. The doc comments say so directly ("The shape stays the synchronous `StateStore`; Effect-typing it is its own step").

That is a defensible staging choice, and I'm not asking for it to be undone here. But it means the `R` channel now says "you have the port" rather than "you have a capability", so nothing downstream gets a typed error, an interruption point, or a scope out of it yet. Worth stating in the PR description as explicitly deferred, so it doesn't read as done.

`SetupPlatform` is the counter-example and looks right: `Layer.succeed(SetupPlatform)(setup)` over a direct value, with `Effect.fn` used properly in the same file.

## 2. The `() => T` thunk is the part I'd change now

```ts
static layer(store: () => StateStore): Layer.Layer<AppState> {
  return Layer.succeed(AppState)({ get: (k, d) => store().get(k, d), update: (k, v) => store().update(k, v) });
}
```

Resolving `store()` on every member call is late binding — the same shape as `platform()`, now inside the context that exists to replace it. The stated reason is real (a `ManagedRuntime` builds its layer at first run, and in the desktop and CLI roots that run is what opens the store), but Effect's answer to "the resource is created inside the runtime" is `Layer.effect` / `Layer.scoped`, not a thunk over a mutable binding.

`sessionLayer.ts` already uses `Layer.effect` at `:158`, `:221`, `:530` and `:800`, so the idiom is available in the very file that assembles these — which makes the thunk a choice rather than a constraint.

## 3. `AppState` can expose an Effect write today, for free

`JsonStore.set()` already returns an `Effect`, and `update()` is literally `effectRuntime().runPromise(this.set(key, value))` — its own comment calls `set()` "the Effect-side write every caller inside a program uses". So an Effect-typed write costs nothing for the Node implementation; it is already there, wrapped.

## 4. `ToolInjections.list` is a needless thunk

`{ list: () => injections }` wraps an immutable captured array. A `readonly list: readonly ConditionalToolInjection[]` field says the same thing without implying re-resolution.

## Not this PR's problem, but it bounds the claim

The ports keep routing through `hostPort` (181 references), whose catch is `catch: (error) => error`, so every one of those sites carries an `unknown` error channel and R7's typed errors get no closer. I checked and **`hostPort` predates this branch** — it is on `main`, so this is not a regression introduced here. Flagging only because "injection is now Effect-native" and "the ports are Effect-native" are different claims, and this PR establishes the first.

## Suggested disposition

Items 2 and 4 are small and self-contained and I'd do them before merge. Items 1 and 3 are fine as follow-ups if the PR description records them as deferred rather than done. Item 3 in particular is nearly free.

Happy to open a follow-up PR for 2/3/4 after this merges if you'd rather not grow it — say which you prefer. I have no claim on these files.

LionSR:
Correction to my item 2 above, after reading the four composition roots. **My claim that the thunk is "a choice, not a constraint" was wrong**, and the `Layer.effect` sites I cited as proof do not have the problem these services have. Retracting that specific recommendation; the rest of the review stands.

What the roots actually show:

| Root | Thunk load-bearing? | Why |
|---|---|---|
| `packages/extension/src/extension.ts:193` | **No** | "Both process stores exist before the runtime here: VS Code hands the extension its SecretStorage and Memento at activation." `() => secrets` / `() => context.globalState` wrap values that already exist. |
| `packages/desktop/src/main/platform/index.ts:123` | **Yes** | `globalStateStore` is assigned *after* `installProcessRuntime`, by a program run on that very runtime. The thunk is a deferred forward reference; passing a value is impossible as written. |
| `packages/cli/src/runtime/cliProcessRuntime.ts:62` | **Yes** | Forced by the first-arrival latch — `notifyCliUpdate` and `clone` install the runtime before `initCliPlatform` opens the store. The file documents it as "the one interim of its kind". |
| `packages/agent/src/effect/runtime.ts:193` | not checked | — |

So there is a genuine circularity in the desktop and CLI roots: the runtime must provide the stores, and the stores are opened by a program running on that runtime. A thunk resolves it by deferring; `Layer.succeed` over a value cannot.

**The native answer is still `Layer.effect`, but it is a bigger change than I implied** — it means moving *acquisition into the layer* (`Layer.effect(AppState, openStore)`) so the root no longer opens the store separately, which inverts ownership rather than re-ordering statements. For the CLI it additionally requires retiring the first-arrival latch. That is exactly step 2 of the injection plan ("each of the five composition roots builds one `Layer` and holds its `ManagedRuntime` in a local; CLI gets a single root, replacing the first-arrival latch"), not a follow-up tweak.

Two things that remain true and small:

- **The extension root's thunks are gratuitous** and can pass values directly today, independent of the others.
- Item 4 (`ToolInjections.list` as a readonly field) and item 3 (`AppState` exposing the Effect write that `JsonStore.set()` already provides) are unaffected by any of this.

Apologies for the noise — I should have read the roots before asserting the constraint wasn't real.

LionSR:
Final correction: **items 3 and 4 don't hold either.** Having now read the consumers rather than just the definitions, three of my four review points were wrong. Retracting them so nobody acts on them.

**Item 4 (`ToolInjections.list` as a readonly field) — withdrawn.** `list()` is not pointless indirection; it is a structural contract shared with `ToolInjectionRegistry`, which is *mutable* (`register()` pushes into a private array) and is passed wherever `ToolInjections['Service']` is expected — `runReflectionFlow.ts:86` plus four test sites. A readonly field works for the static `AGENT_TOOL_INJECTIONS` and breaks every caller that hands in a registry. The method shape is what lets one type serve both.

**Item 3 (`AppState` exposing the Effect write) — withdrawn as not worth it.** The idea was that `JsonStore.set()` already returns an Effect, so an Effect-typed write is free and would move the Promise boundary from N call sites into one layer with a typed error. But there is exactly **one** such call site post-merge — `packages/cli/src/runtime/modelAccessSelection.ts:130`, `yield* hostPort(() => appState.update(...))`. Changing the service shape and every `StateStore` implementation to retire one wrapper is not a good trade. Worth revisiting only when the write count grows.

**Item 2** — already retracted above; the thunk is load-bearing in the desktop and CLI roots.

**What actually stands from the review: item 1, and it is an observation rather than a defect.** `Secrets` and `AppState` carry the Promise-shaped ports verbatim, so this PR makes *injection* Effect-native without making the *ports* Effect-native. The code says so in its own comments, and the staging is defensible. My only remaining suggestion is that the PR description record it as explicitly deferred, so "cohort A is done" doesn't get read as "these services are Effect-native".

Closing the loop on the real question I was asked — whether this is native or adapters-with-tags. It is neither sloppy nor finished: the tags and layers are correct, `SetupPlatform` is fully native, and the Promise shapes are a declared interim. Making the ports native is gated on two genuinely larger pieces of work — the composition-root redesign (injection plan step 2, including retiring the CLI first-arrival latch) and retiring `hostPort`, whose 181 sites all carry an `unknown` error channel and which predates this branch.

Apologies for the churn on a merged PR; I should have read the consumers before proposing changes.

PR #12289, opened 2026-09-11
fix(cli): read run JSON under output in validate-run

LionSR:
## Summary
Runtime validation (the main-only `runtime validation (linux)` job) has been red on every main commit since 4bc68a85c3 (#12246, one run model S2b). That change moved `category`, `outputs` and `response` under `output` on the CLI run result (`CliWorkflowRunResult` / `CliToolUseRunResult`), a deliberate 1.0 break (D13-1), but `packages/cli/scripts/validate-run.mjs` still read the top-level fields.

- Workflow run JSON: `output.category`, `output.outputs`.
- `agents run` tool-use JSON: `output.category`, `output.response`.
- multi-agent JSON: `result.output.category`, `result.output.response`.
- The workflow-script report assertion now includes the report it received in its failure message.

## Known remaining failure
With these fixed, validation reaches :1031 and fails there: the workflow-script child report shows `{"solutions": [null, null, null]}`. `agent().structured` no longer reaches workflow scripts, although the `delegate_multi_agents` contract (WorkflowScriptTool.ts:194) documents it. That is a runtime regression, and coauthor-dc is restoring the contract in a separate PR. This PR makes the failure visible instead of masked by :795.

## Verified
- `npm --prefix packages/cli run validate:run` locally (same script CI runs): passes every run-JSON assertion, then fails at :1031 as described, printing the report.
- prettier --check on the changed file: pass.

### DIFF
--- a/packages/cli/scripts/validate-run.mjs
+++ b/packages/cli/scripts/validate-run.mjs
@@ -791,10 +791,10 @@ function validateRunCommand() {
     assertSuccess(json, 'texra run JSON');
     const jsonResult = JSON.parse(json.stdout);
     assert(
-      jsonResult.category === 'workflow',
+      jsonResult.output?.category === 'workflow',
       'JSON run output should serialize the workflow result',
     );
-    const finalOutput = jsonResult.outputs.at(-1);
+    const finalOutput = jsonResult.output.outputs.at(-1);
     assert(
       outputPathPattern.test(finalOutput?.relativePath ?? ''),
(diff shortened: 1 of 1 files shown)


LionSR:
Follow-up: the remaining failure at :1031 (`agent().structured` returning null in workflow scripts) is fixed by #12291. #12246 made workflowScriptAgentRunner return RunEnd; #12291 projects the documented script-facing envelope once, in the engine. With it, the full validate-run e2e passes locally.

PR #12312, opened 2026-09-12
feat(storage): stamp and enforce a database format version

### LINKED ISSUE #12251: 1.0: one database format marker that refuses state written by earlier 1.0-dev builds
## Current status — 2026-09-15

Checked against main `2c9898445d809b98b038ca9b133b16b57ef7794a` on 2026-09-15. Current status below supersedes old counts, branch reservations, and implementation-stage descriptions.

## Disposition

The session-event format marker and format test landed (#12395). Completion follows the current accepted 1.0 policy: SESSION_EVENT_FORMAT stamps the store and Database clears a mismatched format at open. The older refuse-only wording is superseded; this closure does not authorize legacy migration or removal of unrelated user files.

The previous issue text is retained below as historical evidence. This closure applies to this issue's scope, not the entire Effect conversion.

<details>
<summary>Historical issue description (superseded status; retained for evidence)</summary>

Under the accepted 1.0 direction, main does not read or migrate legacy state: no legacy readers, no dual reads, no compatibility adapters. Release 0.40 never wrote under the `v1` storage namespace, so no shipped user data is affected. What is affected is state written by earlier 1.0-dev builds in the same `v1` tree, whose row shapes have since changed on main:

- #12222 (S1): aggregate kinds `stream`/`execution` became one `run` aggregate; parent carriers collapsed onto `run.start.parent`.
- #12246 (S2a+S2b): `run.end` is the terminal fact; `AgentFinalResult` and the result twins are gone.
- #12249 (S2c): `ToolUseLog.isError` is gone (`status` is the one fact); `kind` is required on issued workflow calls.

Today such rows either fail a hard `.parse()` (loud but per-row and confusing) or are silently stripped by a non-strict Zod object (the `isError` case reviewers flagged on #12249). Both are wrong under the "identify the current database format and reject unsupported state clearly" requirement of the 1.0 plan (§6).

Proposal: one database-level format marker (SQLite `PRAGMA user_version` or a `format` row in `event_sequence`'s neighbour table), written when a database is created, checked when it is opened. A mismatch refuses the whole database with one clear message naming the path, never a per-row failure, never a silent strip. Bump the marker in the same PR as any persisted-shape change on main until 1.0 ships; after 1.0 the marker is the versioning hook the one-fold PRD already anticipates.

Out of scope: any reader of the old shapes. The rows are pre-release state and are not supported.

Refs: `.agents/docs/proposed/architecture/2026-09-09-texra-1-0-implementation-plan.md` §4A, §6; `.agents/docs/proposed/architecture/2026-09-10-one-run-model.md` §5.

</details>

LionSR:
## Summary

TeXRA 1.0 keeps no compatibility with state written by earlier 1.0-dev builds, but today such rows either fail a per-row `.parse()` (loud but confusing) or are silently stripped. This adds the one database format marker the issue asks for: `DATABASE_FORMAT_VERSION` (currently `1`) beside `SCHEMA` in `src/controllers/session/Database.ts`, stored in SQLite's `PRAGMA user_version`.

`configure` — which already runs at every open — now checks the marker before the DDL creates anything:

- **Fresh database** (no user tables): the schema is applied and `user_version` is stamped. A table-free database is fresh regardless of its stamp, since there is no state to protect.
- **Database with tables under any other version** — including the unmarked `0` that pre-marker dev builds wrote: the open fails with one `DatabaseOpenFailed` naming the path and stating the state was written by an incompatible 1.0-dev build, is not migrated by design, and the fix is to delete the database and start fresh. The refusal happens before the DDL, so a refused database is never mutated.
- **`:memory:` opens** are always fresh, so the ephemeral path is stamped and needs no exemption.

The marker is checked ahead of the DDL deliberately: an open that will refuse a database must not create or alter tables in it first.

Closes #12251.

## Issue acceptance gates

- One database-level format marker (`PRAGMA user_version`), written when a database is created, checked when it is opened. ✅
- A mismatch refuses the whole database with one clear message naming the path — never a per-row failure, never a silent strip. ✅
- No reader of old shapes (out of scope per the issue): no reader of any other version exists. ✅
- The constant's docstring records the bump rule: bump in the same PR as any persisted-shape change on main until 1.0 ships; after 1.0 it is the one-fold PRD's versioning hook. ✅

## Single source of truth

`DATABASE_FORMAT_VERSION` in `src/controllers/session/Database.ts` owns the format version; `configure` owns the check, and it already owns every open of the single store.

No new abstraction: the check lives inside the existing `configure`, reusing the existing `DatabaseOpenFailed` surfacing path (`mapDatabaseFailure(openFailed)`), so hosts get the refusal through the same error channel as every other open failure. All production opens of `texra.db` (`sessionLayer`, `inquiryRecords`, `updateCheckRecords`, desktop project records, CLI input history) go through `databaseLayer`; nothing opens the file directly.

## Net elements (R6)

n/a — feature PR, not a refactor/simplify. (2 files changed, +114/-2; one new export, `DATABASE_FORMAT_VERSION`, consumed by the test in this PR.)

## Consumer counts (R8)

n/a — no emit path or export deleted or re-routed; `configure` is module-private with one caller.

## Host impact

Extension: session open on an incompatible dev-build database fails with `DatabaseOpenFailed` (clear message naming the path) instead of later per-row failures.

Desktop: same, via `databaseLayer('persistent')` in `desktopProjectRecords`.

CLI: input history already degrades to empty on any open failure by design ("history failure must not prevent typing"), so an incompatible global database costs history only; session opens surface the refusal normally.

## Scheduled refactoring

None — the marker is the post-1.0 versioning hook; bumping it rides along with each persisted-shape PR.

## Validation

- New tests in `sessionEvents.vitest.ts` (the C1/C6 substrate suite): a fresh database is stamped with `DATABASE_FORMAT_VERSION`; a database with tables and no stamp (pre-marker dev build) and one with a foreign stamp both fail the open with `DatabaseOpenFailed` whose message names `texra.db` and the incompatibility; a fresh open beside them still succeeds. Confirmed the refusal test fails on main without the implementation.
- `npm run format`, `npm run lint`, `npm run typecheck:workspace`, `npm run typecheck:test-kernel` — clean.
- `npm run test:changed` — 166 files, 2071 passed / 1 skipped.
- `npx vitest run src/test-kernel/controllers/session src/test-kernel/transcript/StreamLogStoreLoad.vitest.ts` — 4 files, 43 passed (covers persistent and `:memory:` opens).

### DIFF
--- a/src/controllers/session/Database.ts
+++ b/src/controllers/session/Database.ts
@@ -77,6 +77,18 @@ import { localDatabasePath } from './localDatabasePath';
 /** The database file of a session root, beside the stores it replaces. */
 const SESSION_DATABASE_FILE = 'texra.db';
 /**
+ * The one database format marker (issue #12251, 1.0 plan §6): a fresh
+ * database is stamped with this version in `PRAGMA user_version`, and an open
+ * whose stored version differs — including the unmarked 0 of pre-marker dev
+ * builds — is refused with one `DatabaseOpenFailed`, never a per-row parse
+ * failure and never a silent strip. 1.0 keeps no compatibility with earlier
+ * 1.0-dev builds, so there is deliberately no reader of any other version;
+ * bump this constant in the same PR as any persisted-shape change on main
+ * until 1.0 ships, after which it is the versioning hook the one-fold PRD
+ * anticipates.
+ */
+export const DATABASE_FORMAT_VERSION = 1;
+/**
  * Event history and bounded current application records.
  *
  * `commit` is a SQLite keyword, so the column is quoted at every site; the
@@ -240,7 +252,7 @@ export const databaseLayer = (
         disableWAL: mode === 'ephemeral',
         busyTimeout: '5 seconds',
       }).pipe(mapDatabaseFailure(openFailed));
-      yield* configure(sql, mode).pipe(mapDatabaseFailure(openFailed));
+      yield* configure(sql, mode, path).pipe(mapDatabaseFailure(openFailed));
       const level = yield* SubscriptionRef.make(0);
       const observedCommit = yield* SubscriptionRef.make(0);
       const highWater = "SELECT seq FROM sqlite_sequence WHERE name = 'event'";
@@ -1211,10 +1223,17 @@ function payloadOf(draft: {
  * Read cursors use sqlite_sequence's committed high-water mark. Wake levels
  * are separate counters, since a claim-only change must wake readers even
  * when the event ordinal does not change.
+ *
+ * One open-time gate sits ahead of the DDL: the `user_version` format marker
+ * (`DATABASE_FORMAT_VERSION`). Any database that already holds tables under a
+ * different version — or under no version, as pre-marker dev builds wrote —
+ * is refused before the schema is touched; only a table-free database is
+ * created and stamped.
  */
 const configure = Effect.fnUntraced(function* (
   sql: SqlClient.SqlClient,
   mode: 'persistent' | 'ephemeral',
+  path: string,
 ) {
   yield* sql.unsafe('PRAGMA foreign_keys = ON', []);
   yield* sql.unsafe('PRAGMA synchronous = NORMAL', []);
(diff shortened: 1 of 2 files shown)


LionSR:
Two things from the session shepherding the runtime lane, since this PR is the mechanism #12251 asked for:

1. The red `static checks` job is the dead-code ratchet: `DATABASE_FORMAT_VERSION` is exported with a test-only consumer. Either read it in the one production place that writes/checks the marker and un-export it (the test can assert the behavior through `Database` open/create rather than the constant), or if the export is deliberate for a host message, give it that consumer in this PR.

2. Ordering: the runtime lane's PR3+PR4 (branch `refactor/run-ledger-pr3-pr4`, request protocol and child protocol) renames persisted event types and deletes the `status` row, so it should land AFTER this marker and carry the version bump. I will rebase PR3+PR4 over this once it merges and bump the version in that PR. The host-facing message ("this database was written by an unsupported version, at <path>") is still open here; if you prefer, PR3+PR4 can add it in the CLI and extension when it bumps.

LionSR:
Closing as superseded by #12395 (merged 2026-09-13, `2780d7cb36`). Main now stamps the store with `SESSION_EVENT_FORMAT` in `PRAGMA user_version` and checks it at every open, which is this PR's marker. The policy on mismatch went the other way: a store of another format is cleared under the write lock and reported once through `SessionGraph.storeCleared` for the host to tell the user, instead of refusing the open. That choice came from a live incident (the TUI froze on a Sep 12 store whose rows no longer decoded), and a refusal would have left the user with the same unusable workspace. Rebasing this branch would reintroduce a second, contradictory gate on the same pragma.

PR #12331, opened 2026-09-12
simplify(runs): the plan marker is the attempt authority; retire the unavailable lifecycle status

LionSR:
Three verified items from the 1.0 clean-slate sweep, all in the run/status vocabulary. Each was adversarially re-verified at today's main and re-derived again during implementation.

## 1. One attempt authority: the workflow plan marker

`latestWorkflowAttemptId` ran an election over rows and task groups — an `AttemptBoundary` type, a `laterAttemptBoundaryByTime` comparator and a sequenced/timed fallback — but its own first two statements already returned the answer: the marker id, else the plan's `attemptId`. Everything after that only ran for a marker-less transcript.

There are exactly two production callers. The session fold passes the marker written from the transcript; the proposal panel passes empty rows and task groups, so no election can fire. Every live producer stamps the marker (`workflow-plan-<attemptId>`) and tags its phases, so only pre-marker persisted transcripts could reach the fallback, and 1.0 drops those. A marker-less run degrades to "show everything" through the existing unscoped arm, not to lost rows.

The call site now reads the marker inline, byte-for-byte what the deleted function returned first. `usableSequence` is un-exported (its only external caller was this function). Two tests that pinned the election are deleted.

## 2. One status vocabulary

`RUN_LIFECYCLE_UNAVAILABLE` was replaced by `readOnly` in the fold and had no producer left. Retiring it collapses the overloads that existed only to express its absence: `runStatusDisplayKey` goes from three signatures to one non-nullable one, `formatRunStatusLabel` from four signatures plus an options interface and an unused `missingLabel` to a single line, and `progressHeaderStatus` is deleted. `RunHeader` calls the display key directly.

Both surviving `formatRunStatusLabel` callers pass a `TaskGroupStatus`, which is a subset of `RunPhase`, so the single signature resolves for them.

## 3. Janitorial

Four Web Awesome icons with no consumer anywhere (`arrow-left`, `building`, `heart`, `paper-plane`) are removed from both registration files; the `satisfies` constraint kept this honest. Four stale comments describing identity-less and legacy rows are corrected.

Two things deliberately left out:
- **`throughSeqNo`** lives in a file a sibling batch is editing; it is a two-line follow-up.
- **`revertOnEmpty`** changes clear-field behavior on the compaction-threshold and retry rows, so it needs an owner ruling.

## Net elements (R6)

14 files, +25 / −259, **net −234 LOC**. No file added.

`^[+-]export` / declaration lines removed: `usableSequence` (export dropped), `runStatusDisplayKey` ×3 signatures (one re-added, now exported), `formatRunStatusLabel` ×4 signatures (one re-added), `FormatRunStatusLabelOptions`, `progressHeaderStatus`, `AttemptBoundary`, `laterAttemptBoundaryByTime`, `latestWorkflowAttemptId`, `RUN_LIFECYCLE_UNAVAILABLE`, and `RunLifecycleStatus` (rewritten as `RunPhase | typeof RUN_LIFECYCLE_READY`).

## Consumer counts (R8)

| Symbol / value | Production consumers before | After |
|---|---|---|
| `RUN_LIFECYCLE_UNAVAILABLE` | 5 (definition, union member, label, tone, header table) | 0 |
| `progressHeaderStatus` | 1 | 0 |
| `latestWorkflowAttemptId` | 1 | 0 |
| `AttemptBoundary`, `laterAttemptBoundaryByTime` | file-local only | 0 |
| `FormatRunStatusLabelOptions` / `missingLabel` | 0 callers passed it | 0 |
| the four icons | 0 (definition files only) | 0 |
| `usableSequence` | 1 external, 2 in-file | 0 external, 2 in-file |

`runStatusDisplayKey` becomes an export with a consumer in the same commit.

## Verification

- `CI=true npm run typecheck`: clean across every per-package tsconfig.
- `vitest run --changed origin/main`: 448 passed, 448 total.
- `eslint` on all 14 changed files and `prettier`: clean.
- `check:dead-code-ratchet`: no new findings.
- `git diff --check`: clean.

One note for anyone re-verifying on a busy machine: an earlier run under load average 81 reported 14 failed files, and a quiet re-run reported none. Timeout-based failures here need a low-load re-run before they are believed.

## Coordination

Spillover: `src/shared/schemas/run.ts` (−9: the constant, its doc and the union member) and `src/shared/state/stateKeys.ts` (−3 comment lines). Both are in a slice whose owning session has ended. Registered in the cross-session ledger as `coauthor-2c--v1-attempt-and-status`.


LionSR:
Fixed the doc nitpick in `e2f7c11`: the `RunView.label` comment now names all three identity arms (agent, tool, workflow name) rather than calling it an agent name.

PR #12332, opened 2026-09-12
simplify(transcript): one owner per transcript fact — redaction, card labels, group-row schema, StreamLog members

LionSR:
Four verified items from the 1.0 clean-slate sweep, all on the transcript fold's write path. Each was found by a survey agent, adversarially re-verified at today's main, and re-derived again during implementation.

## 1. Redaction has one owner: the Database write boundary

The fold re-redacted every value a second time. `Database.prepareEventDraft` already redacts every draft before it is written, and `SessionHandle` redacts text chunks (which never reach the DB). The fold's only two production callers both feed **committed** rows, so the fold-time pass could never see an unredacted value.

All 12 fold-time redaction calls are deleted, and `redactLogData` becomes module-private. The trace viewer is unaffected: it emits only `transcript.entry` and bypasses these arms.

## 2. Workflow-card model labels: one projection, not three

The model label was applied in the row projection and then re-derived in two export paths. It is now applied once, in the `WORKFLOW_TASK` case of `projectTranscriptRow`. `src/model/projectWorkflowCallEntry.ts` is deleted with both re-projections. `formatWorkflowCallMetadataParts` stays canonical: it also feeds the runtime's activity text.

## 3. Group rows carry their own type and status

`StreamLogEntrySchema` had a `messageType`-absent union arm, an optional `messageType` on group rows, a prefaulted `data`, and an optional `status` — none of which the one production writer can produce. `createTranscriptFold` stamps `messageType: DEFAULT` on every group row and a `status` on both `stage.start` and `stage.end` (its update is a shallow merge, so both survive).

The absent arm is deleted; group rows now require `messageType`, `data` and `data.status`; `GroupLogPayload.name` (zero producers, one reader) is deleted; and the seven `??` compensations that existed only for those shapes are gone.

This tightens a parse boundary for exported trace documents. Any trace exported by a recent build still parses, because the writer has always stamped both fields. Older exports now fail loudly at the boundary instead of degrading silently, which is the intended 1.0 behavior for disposable local data.

## 4. StreamLog: the members with no readers

Deleted: the restore constructor (all 24 constructions repo-wide are zero-arg), three running-state counter fields with `countEntry` and its four call sites, and `head`, `settlementHead`, `firstTimestamp`, `lastTimestamp`, `hasRunningGroup`, `hasRunningStreamingText`, `hasNonterminalWorkflowCall`, `getRange` and `getById`. Callers that used `getRange(0)` to mean "everything" use `toJSON()`.

## Net elements (R6)

35 files, +104 / −352, **net −248 LOC**: production 18 files, net −182; tests 17 files, net −66. Files deleted: 1 (`src/model/projectWorkflowCallEntry.ts`).

`^[+-]export` lines: 2 removed (`projectWorkflowCallEntries`, `redactLogData`), 0 added net (`StreamLogEntryOf` is re-declared without its `NonNullable<>`). Declarations removed: `MessageTypeAbsentLogEntrySchema`, `taskGroupEndStatus`, `projectWorkflowCallEntry`, `GroupLogPayload.name`, the `StreamLog` constructor, nine `StreamLog` getters/methods and three private counter fields.

## Consumer counts (R8)

Production references remaining after the change, grepped across all four hosts:

| Deleted symbol / arm | prod refs |
|---|---|
| `projectWorkflowCallEntries` | 0 |
| `redactLogData` | 4, all inside its own module (now private) |
| `MessageTypeAbsentLogEntrySchema`, `taskGroupEndStatus`, `GroupLogPayload.name` | 0 each |
| `StreamLog.head` / `settlementHead` / `firstTimestamp` / `lastTimestamp` | 0 each (remaining `.head` and `.lastTimestamp` hits are unrelated fields) |
| `hasRunningGroup` / `hasRunningStreamingText` / `hasNonterminalWorkflowCall` / `countEntry` / `getRange` / `getById` | 0 each |
| `new StreamLog(entries)` | 0 — every construction is zero-arg |

## Tests

No test is added. Fixtures that built group rows without the now-required fields are fixed by adding them, and two tests that only pinned removed behavior are deleted: the `StreamLog` restore half, and an 82-line describe that fed raw trace events straight into the fold to assert record-time redaction — the property the fold no longer owns.

Two findings worth recording:
- **Typecheck does not catch fixtures built through `.parse()`.** Two fixtures compiled fine and failed at runtime. A typecheck-only verification would have shipped this broken.
- One fixture used `data: { status: 'stopped' }`, which was never a valid `TaskGroupStatus`. The loose payload had been accepting it silently — evidence the optional arm was masking, not tolerating.

## Verification

- `CI=true npm run typecheck`: clean across all 7 projects.
- `vitest run --changed origin/main`: 440 files, 4,679 passed, 1 skipped.
- `eslint` on all changed files (batched; a single invocation over 33 files exhausts the V8 heap in a worktree) and `prettier`: clean.
- `check:dead-code-ratchet`: no new findings, baseline untouched.
- `git diff --check`: clean.

## Coordination

Spillover outside this session's slice, all mechanical and all in areas whose owning session has ended or which are one-line edits: `src/shared/schemas/streamLogEntry.ts` and `taskGroup.ts`, `src/transcript/completedRunArchive.ts` and `traceDocumentSchema.ts`, `src/agent/runtime/SessionHandle.ts`, `src/model/projectWorkflowCallEntry.ts` (deleted), `src/tools/executions/processOutput.ts`, and two CLI one-liners in `packages/cli`. Registered in the cross-session ledger as `coauthor-2c--v1-fold-owners`.


LionSR:
why do we even do 
Redaction

Should never input sensitive info

Issue #12078, opened 2026-09-07
Node platform defaults on @effect/platform-node (tracking)


LionSR:
Lane tracker under #12025 and #12074, for the `src/platform/defaults/node*.ts` family against `@effect/platform-node@4.0.0-rc.112`.

## The module lists, side by side

`@effect/platform-node@4.0.0-rc.112` ships: `NodeChildProcessSpawner NodeCrypto NodeFileSystem NodeHttpClient NodeHttpPlatform NodeHttpServer NodeMultipart NodePath NodeRedis NodeRuntime NodeServices NodeSink NodeSocket NodeStdio NodeStream NodeTerminal NodeWorker NodeWorkerRunner Undici NodeClusterHttp NodeClusterSocket Mime`.

`src/platform/defaults/` ships 1115 lines of `node*.ts`:

| TeXRA | `@effect/platform-node` | Lines |
| --- | --- | --- |
| `nodeChildProcessSpawner.ts` | **`NodeChildProcessSpawner`** | 304 |
| `nodeHost.ts` | (composite — no single counterpart) | 218 |
| `nodeFilesystem.ts` | `NodeFileSystem` | 154 |
| `nodeStores.ts` | (no counterpart — repo's own state stores) | 153 |
| `nodeProcesses.ts` | `NodeRuntime`, `NodeStdio` | 114 |
| `nodeWorkspace.ts` | (no counterpart — TeXRA domain) | 63 |
| `nodeStorage.ts` | (`persistence/KeyValueStore` — see #12074 §C) | 41 |
| `nodeAgentRuntime.ts` | (no counterpart — TeXRA domain) | 35 |
| `nodeHostEnvironment.ts` | `NodeRuntime` + `Config` | 33 |
| `longRunningModelTransport.ts` | `NodeHttpClient`, `Undici` | — |
| path handling in `baseFS`/`pathCore` | `NodePath` | — |

**`nodeChildProcessSpawner.ts` and `NodeChildProcessSpawner` match name-for-name.** The repository independently arrived at the same port, which is the strongest single signal on this page that the abstraction boundary is right and the implementation is duplicated.

## Status

[...]

PR #12340, opened 2026-09-12
refactor(platform): spawn Lean servers through @effect/platform-node

LionSR:
First item of #12078: the repo's `src/platform/defaults/nodeChildProcessSpawner.ts` implemented the same `effect/unstable/process` `ChildProcessSpawner` that `@effect/platform-node` ships, so it is deleted rather than shrunk. `directLspAdapter.ts` now provides `NodeChildProcessSpawner.layer` (over `NodeFileSystem` and `NodePath`, which the library resolves `cwd` through) at the one place the deleted layer was provided.

## What changes for the Lean server

- Children are spawned detached on non-Windows and shutdown signals the process group (SIGTERM, then SIGKILL under `forceKillAfter`; `taskkill /T` on Windows). The old spawner signalled only the leader, so `lake env lean --server` could leave its `lean` child alive. A terminal Ctrl-C no longer reaches the child directly; the adapter's scope finalizer is what stops it.
- `exitCode` settles on `exit` rather than `close`.
- Spawn errnos fail the spawn effect instead of arriving through `exitCode`; the now-unreachable spawn-syscall branch of `describeEnd` in `leanServer.ts` is deleted. The errno chain the pool's EMFILE/ENFILE retry matches on is preserved (ENOENT and both EMFILE suites).

## Outside `src/`

- `vitest.config.mjs` inlines `@effect/platform-node` so the adapter suite's `vi.mock('node:child_process')` reaches the library's spawn; otherwise six fake-child tests would spawn real processes.
- `packages/agent/package.json` declares `@effect/platform-node` because the Lean adapter is in the SDK bundle graph and `validate-artifacts.mjs` fails otherwise. Lockfile delta is the three added lines.

The suite's fakes were corrected to Node's real contract (emit `spawn` once the listener attaches, `exit` alongside `close`, EMFILE through `error`, pids above any real one so the group probe cannot signal an unrelated group). No test added or removed.

## Size and verification

7 files, +68/-320 (net -252). No ratchet row moved. Typecheck, lint, full suite (5,728 tests), test:pure, dead-code ratchet, effect-migration ratchet, and `compile:fast` green on the commit.


LionSR:
do this as much as possible

PR #12349, opened 2026-09-12
simplify(cli)!: one headless `run` command for both agent categories (1.0 clean slate)

LionSR:
**Do not ship this in a release until `texra-ai/texra-action` v2 is tagged.** The v1 action calls `texra agents run`, which this PR deletes. Its replacement is prepared and held at [texra-ai/texra-action#10](https://github.com/texra-ai/texra-action/pull/10). Release order: **merge #10 and tag v2.0.0 first, then ship TeXRA 0.41** — see review round 2, finding 1.

Owner ruling (2026-09-12, `.git/texra-coord/findings/owner-rulings-2026-09-11b.md`): "Fold `agents run` into `run` (cli-runtime-6). The user wants the interface cleaned up."

## Summary

`texra agents run <agent>` is gone. One `texra run <agent>` resolves the agent and branches on its category: workflow agents keep the existing path, tool-use agents take the tool-use path. No alias, no deprecation shim, per the 1.0 clean-slate rule. `agents` keeps `list` and `show`.

- Deleted `packages/cli/src/commands/agentsRun.ts` (167 lines). `workflow.ts` owns `runHeadlessAgent`, which resolves once and branches, with a module-local `runToolUseAgent`.
- `runtime/agents.ts` drops the `agentsRun` launch target and both cross-category redirect messages; the mode map is now `chat` | `workflowResume`, and `resolveCliRunAgent` resolves both categories and refuses a name that is in both (see "Review follow-ups").
- Category-specific flags now fail with a clear usage error: `--output is only available for workflow agents; "review" is a toolUse agent.` `--input` loses its citty `required: true`, because a tool-use run may take none; a workflow agent without input still reports `At least one workflow input file is required.`
- Shell completion: `texra run` lists all launchable agents; the per-category `byCategory` source machinery and the `agents run` positional source are gone.
- Docs and scripts updated: `docs/guide/texra-cli.md`, `docs/guide/tikz-figures.md`, `packages/cli/scripts/validate-run.mjs`, and a BREAKING entry in `CHANGELOG.md`.

### Round 1 (fixed in `5a4379d93a`)

**1. Workflow-first resolution silently won a name collision.** `resolveCliRunAgent` resolved workflow-first with a `??` fallback, so a name carried by both categories always landed on the workflow entry: an invocation that used to be `texra agents run <tool-use agent>` could silently change run shape. It now resolves both categories and refuses the collision rather than picking:

> Agent name "assistant" is ambiguous: it matches the workflow agent custom:assistant and the toolUse agent builtInToolUse:assistant. Re-run with the source-qualified name to pick one: `texra run custom:assistant` or `texra run builtInToolUse:assistant`.

No `--category` flag was needed, and the reviewer's worry about a same-source collision does not apply: the registry is a flat cache keyed by `source:name` (`agentKeyOf`), so two entries sharing a name *necessarily* differ in source and a source-qualified identifier hits exactly one cache key. A same-source cross-category collision is unrepresentable, not merely unhandled — so the qualified spelling the error prints always disambiguates, and the message is honest. `resolveAgentForLaunch`'s pinned tier is category-blind, but `resolveCliAgentInCategory` re-filters it against the entry's own category, which is exactly what makes `custom:assistant` resolve in workflow and nowhere else. Falling through to a single match when only one category has the name is unchanged; a miss in both still reports `Agent not found`.

**2. Category-dependent usage validation cost a platform init.** `initLocalCliPlatform` + `resolveCliRunAgent` ran before either category's usage checks, so a signed-in user paid a real remote agent-list fetch to be told they forgot a flag. `runHeadlessAgent` now fails fast, before the platform init, on the one condition invalid for *both* categories — no `--instruction`, no `--instruction-file`, and no `--input` (a `-` stdin spec counts as supplied, since it is an `inputFiles` entry):

> Provide --instruction or --instruction-file for a tool-use agent, or --input for a workflow agent.

That is four lines and no category branch; every category-specific error stays exactly where it was.

**Dead code the reviewer flagged.** Confirmed and deleted end to end: `categoryMismatchMessage` had one reader, the post-run invariant throw in `executeCliConfig`, which `enforceCategory` makes unreachable before the run — documented as unreachable in that comment since it was written, and true for every call site, not just the two this PR touched. The option, its four custom strings (`workflow.ts` ×2, `resumeRun.ts`, `multiAgent.ts`), and the `??` over the default message are gone; the invariant throw keeps the default `Agent resolved to a non ${expectedCategory} run.` Nothing else was deferred.

### Round 2 (fixed in `6f79445fd8` … `fe8af126d1`)

**1 (P1). `texra install-github-action` scaffolded a workflow that 0.41 breaks.** Real: the template pinned `texra-ai/texra-action/review@v1` with `texra-version` unset, so a freshly scaffolded workflow installs the latest CLI and runs the v1 action — and v1 invokes `texra agents run` (verified at `texra-ai/texra-action@2a327e4:src/lib/cliArgs.ts`, which builds `["agents", "run", …]`; #10's diff deletes exactly the `"agents"` element).

*Decision and release order.* The generated workflow now pins `@v2`, and `texra-version` stays unset. Pinning the CLI to a 0.40.x line instead was rejected: it makes every new install permanently trail the CLI, and it would have to be unpinned again later, so it trades a one-time ordering constraint for a standing one. `@v2` is correct at every point in the order **tag texra-action v2 → ship TeXRA 0.41**, which is already this PR's hold condition: the generator only reaches a user through the 0.41 CLI, the workflow it writes installs the latest CLI (≥ 0.41 by then), and v2 + ≥ 0.41 is the working pair. The reverse order is the only one that breaks, so it is stated at the top of this PR and in the template comment. I also dropped the template's `# texra-version: 0.38.5` example — 0.38.5 is a CLI v2 cannot drive — and did not substitute a version number that does not exist yet.

*This repo's own workflow.* Left on its v1-era SHA deliberately. `.github/workflows/texra-code-review.yml` pins **both** sides — the action commit `2a327e4` *and* `texra-version: ${{ vars.TEXRA_CLI_VERSION }}` — so it keeps working exactly as long as that variable names a 0.40.x release. Re-pinning it to a v2 commit now would break this repo's own PR review immediately, since the CLI variable still points at 0.40.x; the only v2 commit that exists is #10's unmerged head. The honest change is to record the coupling where the pin lives: the comment now says to move this SHA to a texra-action v2 commit *in the same change* that points `vars.TEXRA_CLI_VERSION` at 0.41 or later, never one without the other.

**2 (P1). The documented contract was wrong.** Real, and the reviewer's reading of the code is exact: `runToolUseAgent` rejects only when the *resolved* instruction is empty, and `resolveFileBackedInstruction` resolves `--instruction-file` on its own. Two independent callers already depend on it — `packages/cli/scripts/validate-run.mjs` and texra-action's review job, which passes `--instruction-file` and no `--instruction`. Fixed in all four places that said otherwise: the CHANGELOG entry, `docs/guide/texra-cli.md`, the `--instruction` help text (which regenerated the completion snapshot), and `runHeadlessAgent`'s doc comment. They now say "`--instruction`, `--instruction-file`, or both".

**3 (P2). An invalid command left directories behind.** Real, and worse than "probes": `probeOutputPath` (`packages/cli/src/runtime/workflowOutput.ts:96-99`) falls through a `stat` miss into `mkdir(requiredDirectory, { recursive: true })`, so `assertOutputDirAvailable` / `assertOutputFileAvailable` *create* the destination. Running before `withExpandedRunInputs`, they meant `texra run polish --output-dir new/dir` with no `--input` exited on a usage error having already created `new/dir` — for both reported shapes (no instruction at all, and a non-empty `--instruction` with no `--input`), since the shared pre-init guard only catches the first. `runHeadlessAgent` now refuses an input-less workflow run ahead of both probes. The wording is not a new one: `WORKFLOW_INPUT_REQUIRED_MESSAGE` is lifted out of `workflowInputs.ts` so the pre-flight guard and the expansion guard say the same sentence.

**4 (P2). Completion offered a token `texra run` refuses.** Real. Completion reads the name column of `agents list --quiet --all`, which printed the bare name, so a cross-category collision completed to exactly the spelling round 1 made ambiguous — and never to the qualified forms the refusal tells you to use. Fixed at the listing, not at the completion source: `formatCliAgentList` prints the source-qualified key for a name that two *listed* agents share, and the bare name otherwise. That is one rule in one place; it is computed from the rows actually being listed, so it is right for `--all`, for `--category`, and for the visible-only listing without any of them knowing about each other. Human output is unchanged except where the name is genuinely ambiguous, and the two other readers of that column (`agents show`, `chat --agent`) accept `source:name` too.

**5 (P1). `it.each` cases run through `Effect.runPromise` rather than `it.effect` — overruled; converted in round 3 (`ed795ad13f`).** The refusal recorded here earlier is withdrawn. The owner ruled "we should use `@effect/vitest`", so the tests were converted — and two of the three obstacles that refusal rested on turned out to be solvable rather than structural.

- *The whole tool-use suite is converted, not just the two parameterized cases.* All seven cases in `ToolUseRunCommand.vitest.ts` are now `it.effect` (`it.effect.each` for the parameterized pair). The module-level helper is an Effect rather than a promise — `Effect.provide(nativeRun(...), fakeProcessServices())` — and each body yields it inside `Effect.gen`. A file carrying two idioms would be worse than either, so none was left behind.
- *The usage errors are asserted through Effect's own lifecycle.* Verified empirically against this repo's Effect (4.0.0-rc.115) before settling on a shape: a synchronous `throw` inside an `Effect.fn` body lands as a `Die` reason, so `Effect.flip` fails instead of succeeding. The tests read `Effect.exit` and take the defect off the cause (`exit.cause.reasons.find(Cause.isDieReason)?.defect`), which is the idiom the repo already uses for defects. The assertion is stronger than what it replaces, not weaker: the message is compared with `toBe`, where `.rejects.toThrow('…')` only substring-matched.
- *The `vi.mock` re-entry is gone, not documented around.* The `@cli/runtime/workflowInputs` mock no longer wraps the continuation in `Effect.runPromise`. The spy now answers with the inputs to expand to and the mock's own fiber runs `args[4](inputs)`, so the whole program — `executeCliToolUseConfig` included — stays inside the test runtime. The recorded call is unchanged, so the `toHaveBeenCalledWith(…, expect.any(Function))` pin still holds, and it now sees the production continuation rather than a promise wrapper. No `Effect.run*` call remains in the file.
- *Nothing was weakened.* The fast-fail case still asserts `initLocalCliPlatform` and `resolveCliRunAgent` are never called, and the converted assertions were mutation-checked (a wrong expected message fails the test).
- *Scope, and what stays promise-shaped.* In `WorkflowRunCommand.vitest.ts` only the case this PR added — the input-less workflow run — is converted, to the same shape, over the existing `withTempDirEffect` helper; its `runWorkflow` promise helper now delegates to a new `workflowProgram` Effect, so the conversion did not have to touch the other 24 cases. Those stay promise-shaped deliberately: they predate this PR, and converting them is a dedicated mechanical pass rather than drive-by work here. Both suites sit in the `kernel` Vitest project before and after — the tier did not move.

I did add one regression test per fixed defect, at the durable boundary and inside an existing suite: the input-less-workflow case in `WorkflowRunCommand.vitest.ts` (asserting both the refusal and that nothing was created on disk), and the listing case in `CliAgentShadow.vitest.ts`, which already builds a real cross-category collision against the live registry.

## Tests

`AgentsRunCommand.vitest.ts` is renamed `ToolUseRunCommand.vitest.ts` and retargeted at `runHeadlessAgent`; the tool-use path survives, so the suite was adapted rather than deleted, plus one case for the workflow-only-flag usage error. `WorkflowRunCommand`, `ResumeCommand`, `Completion` (snapshot regenerated through `vitest -u`) and `CliRootArgs` adapted. `WorkflowRunCommand`'s dynamic `import()` is hoisted to module scope: its first test was charged the whole module-graph transform and flaked at the 10s timeout, which also happens on unmodified main.

Four cases added across the two rounds, all at a durable boundary and all in existing suites:
- `CliAgentShadow.vitest.ts` already builds a real cross-category collision (a custom *workflow* `assistant` shadowing the bundled tool-use one) against the live registry, so the ambiguity case goes there and also asserts both qualified spellings resolve. Round 2 adds the listing case beside it.
- `ToolUseRunCommand.vitest.ts` gets the fast-fail case, asserting `initLocalCliPlatform` and `resolveCliRunAgent` are never called. Its existing "missing instruction" case now passes an input file so it still reaches the tool-use-specific error rather than the new shared one.
- `WorkflowRunCommand.vitest.ts` drops its three `categoryMismatchMessage` pins and gains the round-2 output-artifact case.

## Net elements (R6)
- Files: 28 changed, 1 deleted (`agentsRun.ts`), 1 renamed test.
- Exported symbols: `runToolUseAgent` deleted (production-dead, test-only); `runWorkflowAgent` → `runHeadlessAgent`; `headlessRunCommand` replaces `agentsRunCommand` (named so because citty already exports `runCommand`). `CliConfigExecuteOptions.categoryMismatchMessage` deleted. Round 2 adds one export, `WORKFLOW_INPUT_REQUIRED_MESSAGE`, with its consumer (`workflow.ts`) in the same commit; it replaces an inline literal rather than adding a concept.
- Ratchets: `effect-migration-baseline.json` regenerated through `--update` (drops the `agentsRun.ts` entry); `knip-baseline.json` 217 → 216 findings (the `runToolUseAgent` entry goes, `runWorkflowAgent` is renamed in place). Neither moved in round 2.
- Net LOC: 28 files changed, 475 insertions(+), 415 deletions(-) — production (`packages/cli/src`, `src/shared`) 13 files, 275(+) / 302(−); tests 8 files, 169(+) / 89(−); the rest docs, ratchets, snapshot and the two workflow files.

## Consumer counts (R8)
- `texra agents run` spelling: 0 remaining in production code, docs, scripts or settings recipes. The 18 verification recipes in `src/shared/schemas/stateSettings.ts` now spell `texra run`, their 18 code-path strings name `packages/cli/src/commands/workflow.ts` (where `runHeadlessAgent` lives), and `stateSettings.vitest.ts:82` no longer admits `agents run`. coauthor-2c asked for those edits to travel with this PR, and their #12326 merged first, so the file was free. What remains elsewhere is dated records only (`.agents/docs/archived/**`, historical CHANGELOG entries).
- `texra-ai/texra-action/review@v1`: 1 producer (the `install-github-action` template) → now `@v2`. This repo's own workflow pins a SHA, not a tag, and is unchanged apart from the comment recording the two-sided pin.
- "`--instruction` is required": 4 sites (CHANGELOG, CLI guide, help text, doc comment) → 0.
- `WORKFLOW_INPUT_REQUIRED_MESSAGE`: 2 consumers (the pre-flight guard, the expansion guard), 1 definition; previously 1 inline literal and no guard.
- `formatCliAgentList`: 1 consumer (`listAgents`), unchanged; it feeds all three completion sources.
- `agentsRun` launch target: 1 producer, 1 consumer, both deleted.
- `runToolUseAgent`: 0 production callers before this change (knip-baselined).
- `categoryMismatchMessage`: 4 producers, 1 (unreachable) consumer — all deleted.

## Gates
Rebased onto origin/main `0a94b230c6` (picks up #12356 / #12361 on top of the earlier #12353 / #12354 / #12355 / #12357). The CHANGELOG conflict was resolved by keeping both sides: main's `--assets-dir` and `texra chat` entries stay, this PR's entry is added after them in the same section. FORCE_COLOR=0, real exit codes.
- vitest ToolUseRunCommand + WorkflowRunCommand + CliAgentShadow + Completion + InstallGithubAction + AgentsCommand + CliRootArgs + ResumeCommand + `src/test-kernel/architecture`: **0** (21 files, 252 passed / 1 skipped; needs `--testTimeout=120000` on this machine, as on unmodified main)
- `node scripts/check-effect-migration-ratchet.mjs`: **0**, no headroom, baseline untouched
- `npm run check:dead-code-ratchet`: **0**, 213 vs 213 baselined
- `tsc -p packages/cli/tsconfig.json --noEmit`: **0**; `tsc -p tsconfig.test-kernel.json --noEmit`: **0**
- eslint over the 8 changed `.ts` files: **0**; prettier `--check` over the 11 changed text files: **0**
The full typecheck and whole vitest suite are left to CI: local runs get OOM-killed on this machine (about ten concurrent sessions). **Do not merge until texra-action v2 is tagged** (see the note at the top); CI on this head is the readiness gate, not the merge gate.


LionSR:
### Second review round: status at `fe8af126d1`

The inline threads above were re-posted against earlier line positions. Verified at the current head:

- **Workflow input before the output probes** (threads at `workflow.ts:123` / `:153`) — fixed. Two guards now exist: a pre-init refusal when there is neither an instruction nor an `--input`, and a post-resolution refusal at `workflow.ts:145` (`WORKFLOW_INPUT_REQUIRED_MESSAGE`) that fires **before** `assertOutputDirAvailable`/`assertOutputFileAvailable`. The thread's case — a nonempty `--instruction` with no `--input` — hits the second guard. This was worse than reported: `probeOutputPath` (`workflowOutput.ts:96-99`) falls through a `stat` miss into `mkdir(…, {recursive:true})`, so it *created* the destination rather than probing it. A regression test asserts the refusal and that nothing lands on disk.
- **`install-github-action` scaffold** (thread at `CHANGELOG.md:47`) — fixed. The template pins `texra-ai/texra-action/review@v2` (`installGithubAction.ts:68`) and the stale `texra-version: 0.38.5` example is gone. Release order, stated at the top of this PR: tag texra-action v2 → ship 0.41. This repo's own `.github/workflows/texra-code-review.yml` deliberately stays on its v1-era SHA, because it pins both the action commit and `vars.TEXRA_CLI_VERSION`; its comment now records that the SHA moves to a v2 commit in the same change that points that variable at 0.41+.
- **Completion offering a refused token** (thread at `completionCommandTree.ts:222`) — fixed at the listing rather than the consumer: `formatCliAgentList` prints the source-qualified key when two *listed* agents share a name, bare name otherwise. One rule, computed from the rows actually listed, so `--all`, `--category` and the visible-only listing all stay correct.
- **`it.each` vs `@effect/vitest`** (thread at `ToolUseRunCommand.vitest.ts:297`) — not applied, deliberately. All five pre-existing tests in that file use the same `Effect.runPromise` helper, and the promise shape is a standing decision: #12214 converted this suite to `it.effect` and #12222 converted it back. Conversion also cannot deliver a scoped context here, because `vi.mock('@cli/runtime/workflowInputs')` re-enters via `Effect.runPromise` internally.

This PR stays held until texra-action v2 is tagged (texra-ai/texra-action#10) and 0.41 is ready.

LionSR:
We can turn off TeXRA review for a while if it breaks it

PR #12358, opened 2026-09-12
simplify(settings): one command map, one webview app base, and no SecretManager indirection

LionSR:
Five verified items from the 1.0 clean-slate sweep, all in the settings wire and its host handlers.

## What goes

1. **The fossil per-view command maps.** `PROFILE_VIEW_COMMANDS` and `MEMORY_VIEW_COMMANDS` survived the deletion of the standalone Memory and Profile webviews; the settings view then aliased them. Every memory and profile wire literal now appears exactly once, in `src/shared/ipc.ts`. No wire string changes. `COMMON_COMMANDS` stays — it has two real consumers.
2. **`BaseWebviewApp`** had one remaining subclass and is folded into it.
3. **`SecretManager`** was a static class holding two re-exported constants, a one-line delegate and a single-caller quick-pick builder. Its members move to their callers; the quick-pick item type becomes file-local.
4. **Three pure forwarders** in the settings agent handlers; the registry binds to the agent actions directly.
5. **One subscription-usage assembly.** Both hosts built the same snapshot map by hand. `SubscriptionUsageService.getAllUsage` owns it, and the extension's handler module is deleted.

## Deliberately not done

**MODEL-1 (dropping `invalidateApiKeyCache` at the settings writers) is refused, in any partial form.** The API-key lookup cache it compensates for is still live: a 5-second TTL, a WeakMap of caches, `invalidateApiKeyCache`, `lookupApiKeyUncached`, about 14 call sites and 8 suites, with no merged or open PR removing them. Dropping the invalidation now means a user who saves a new API key in Settings gets the **old** key served for up to five seconds — a user-visible credential regression, not a dead-path removal. The survey's own verdict says these lines are only correct as part of the cache's deletion. Handing both lines to whoever schedules that.

(For the record: there is one settings writer at HEAD, not two; the second call is the desktop credential controller.)

## Net elements (R6)

24 files, +174 / −338, **net −164 LOC**. Files deleted: 3 (`BaseWebviewApp.ts` 86, `secretManager.ts` 44, `subscriptionUsageHandlers.ts` 38).

`^[+-]export` / declaration lines removed: `PROFILE_VIEW_COMMANDS`, `MEMORY_VIEW_COMMANDS`, `BaseWebviewApp`, `signalWatcherWebviewAppBase`, `SecretManager` (with its three static members), `ApiProviderQuickPickItem` (now file-local), the `ApiProvider` re-export, `SubscriptionUsageReader`, `sendSubscriptionUsage`, and three `AgentHandlers` forwarder methods. Added: one method, `SubscriptionUsageService.getAllUsage`, on the existing owner.

## Consumer counts (R8)

| Deleted symbol | Production consumers before | After |
|---|---|---|
| `PROFILE_VIEW_COMMANDS` | 3 files | 0 |
| `MEMORY_VIEW_COMMANDS` | 2 files | 0 |
| `BaseWebviewApp` / `signalWatcherWebviewAppBase` | 1 | 0 |
| `SecretManager` | 3 files / 5 sites | 0 |
| `ApiProviderQuickPickItem` (exported) | 1 | 0, file-local |
| `SubscriptionUsageReader` | 1 | 0 (the CLI's own same-named local type is untouched) |
| `sendSubscriptionUsage` | 3 call sites + 1 dead test import | 0 |
| the three agent forwarders | 1 production call site each | 0 |

## Two things worth reviewing closely

- **Deleting the base class un-erased its statics.** The mixin's `as unknown as new (...)` cast was hiding a real override relationship, so `static styles` needed an explicit `override` once the indirection went. Type-only, fixed in the same commit, and the now-false "override lost through mixin type erasure" comment is deleted.
- **Binding the registry to agent actions is eager where the forwarders were lazy.** Verified safe: the handlers field is assigned before the registry is built, and every action arm is an arrow closure over its options with no `this`.

## Verification

- `CI=true npm run typecheck`: clean (re-run after rebase).
- `vitest run --changed origin/main`: 434 files passed, 4,553 tests. Two files failed on the first pass with wall-clock timeouts, in areas this PR does not touch; both pass 122/122 when run alone. This machine was running 19-23 sibling vitest processes.
- `eslint` (separate arguments) and `prettier`: clean.
- `check:dead-code-ratchet` and `check:effect-migration-ratchet`: both clean, no baseline regeneration needed.
- `git diff --check`: clean.

## Docs touched

Three docs referenced deleted symbols and are corrected: the IPC list in `AGENTS.md`, the shared UI-kit list in `src/README.md`, and one line of the code-review checklist that told reviewers to use a now-deleted constant.


LionSR:
Holding this as a draft until #12356 (coauthor-97's git/skills config-slot change) merges. That PR holds an implementing claim on `SettingsViewMessageHandler.ts` and `SettingsProfileController.ts`; the overlaps here are textual and 2-3 lines from theirs, so I'll rebase onto it and mark this ready as soon as it lands.

LionSR:
Unblocked: #12356 was rescoped (the git rows stay worktree-shared, only the skills rows move) and no longer touches this PR's files. Rebased onto current main; typecheck clean, the 5 touched kernel suites pass (25 tests), dead-code ratchet green.

LionSR:
Both doc nitpicks fixed in `8a04c4e`: the `SETTINGS_VIEW_CMD` comment no longer claims inbound-only (SET_TAB and the auth-status updates are outbound), and the README's lit-module count is corrected to 23 — it was already stale at 24 before this PR deleted BaseWebviewApp.

PR #12359, opened 2026-09-12
refactor(transcript): export the run aggregate's display events as the trace

LionSR:
The exported trace document used to be a hand-built reconstruction: the assembler squeezed run facts into a `meta` blob plus a `config` record and a flat `entries` list, and the viewer then re-synthesised a listing from them, 128 lines of it, with legacy fallbacks for shapes no current writer produces.

The run aggregate already holds exactly those facts as display events. The document now carries `events`, and the viewer folds them the same way every other host does.

## What this deletes

`TraceRunFactsSchema`, `TraceStepSchema` and the `steps` field, the `config`, `meta` and `entries` fields, the viewer's `listingBodies` (128 lines) and `traceEvents`, and the duplicate `TraceDataSchema` (already carried in the dead-code baseline as production-dead). `StreamLogStore.hasAuthoritativeRun` becomes `readEvents`, which is what it does.

## Fidelity, privacy, and what a scrubbed position means now

- **More faithful, not less.** The export now carries the real worktree and remote flags, the approval-policy snapshot, and real activation and status rows — so elapsed time comes from the actual start rather than a synthesised one.
- **The privacy filter is unchanged and lives in the assembler, not the viewer.** It excludes every run-record and ledger row: the byte-exact conversation, compaction, tool intents and results, snapshots and child turns — the rows whose own contract says they are never on a renderer's transport. The chunk stream is not durable, so no raw token stream reaches the file.
- The assembler still authors four facts deliberately: `ownerId: null` on every row, and `parent`, `checkpointId` and follow-up support on the start row, so a read-only bridge cannot offer a composer and unknown aggregates are not chased.
- **The scrubber survived and got simpler.** It landed after this change was surveyed, so the positions are re-derived from the events themselves, and the cut is now one commit comparison instead of a per-tier rule (replay the listing at step k, filter the transcript by timestamp). The final row still drops out at any non-final cut.

## Older trace files

A `trace.json` written before this change has no `events` and now **fails loudly**, with the parse error painted in the viewer shell. That is the intended 1.0 behavior for disposable local data — a silent degrade would be the defect. Standalone HTML exports embed their own viewer and are unaffected; only a sidecar JSON from an older build paired with a newer shell can hit it.

## Net elements (R6)

17 files, +271 / −558, **net −287** (production −229, tests −58).

`^[+-]export` / declaration lines removed: `TraceDataSchema`, `type TraceData`, `DisplaySessionEventDraft`, `TraceRunFactsSchema`, `TraceStepSchema`, `type TraceStep`, `listingBodies`, `traceEvents`, plus three test-local helpers. Added: `DisplaySessionEventSchema` (one word, one consumer), `traceSteps` with its type (two callers), `StreamLogStore.readEvents` (the rename), and a `record` field on the assemble result.

## Consumer counts (R8)

Every symbol below is at zero production consumers after the change:

| Symbol | Before |
|---|---|
| `TraceRunFactsSchema` | 1 (file-local) |
| `TraceStepSchema` / `steps` | 3 |
| `TraceDocument.config` | 2 |
| `TraceDocument.meta` | 3 |
| `TraceDocument.entries` | 2 |
| `listingBodies` / `traceEvents` | 1 each |
| `TraceDataSchema` | 0 (baselined production-dead) |
| `DisplaySessionEventDraft` | 1 — became dead when the viewer stopped importing it, so it is deleted rather than baselined |

`hasAuthoritativeRun` had 2 consumers, both repointed to `readEvents`.

## A dependency edge narrowed

`src/transcript` did not lose its `@agent/core` edge, it moved: the document schema no longer imports the run record as a value, and the assembler takes it type-only. The practical result is that `packages/trace-viewer` no longer pulls the run-record module into the browser bundle at all.

## Two things for the reviewer

- **Export size was not measured empirically.** The rows carry the same text the folded entries did, plus the listing facts that used to be squeezed into `meta` and roughly 80 bytes of envelope per row, with ledger rows and chunks excluded — so near-parity is expected. Someone should export one real workflow run before and after to confirm.
- The page title falls back to the run id when a document has no start row, where it previously read a required field. That state also folds to an empty view, so nothing is masked, but say the word and it becomes a throw instead.

## Verification

- `CI=true npm run typecheck`: clean, run after the rebase.
- The 9 trace-viewer, transcript and export suites: 52 tests, all pass — including the secret-leak gate over the fuller row set.
- `eslint` (batched, separate arguments) and `prettier`: clean.
- `check:dead-code-ratchet`: OK. It reports one baseline entry now resolved (`TraceDataSchema`); the one-line baseline shrink is left out because another session holds an implementing claim on that file, and the ratchet only warns on resolved entries.
- `check:effect-migration-ratchet`: clean, no stale headroom.
- `git diff --check`: clean.

One caveat on the full changed-file test run: this machine was at load 51-88 with up to 15 sibling vitest processes. Five suites failed there; each was re-run in isolation or against a clean base, and all five are pre-existing or load flakes in code this diff does not touch. One real failure was found that way and fixed: an export test asserted a folded entry's field name that the raw row spells differently.

## Follow-up

Deleting the now-producerless `transcript.entry` event arm is the natural next step and is being done separately, since it must re-seed six fixture suites, one of which another session held.


LionSR:
Addressed the review findings in `b90a1d5`:

**P1, private request decisions — fixed.** This was a real disclosure and not a nitpick: `request.decided.decision` can carry the whole edited document (`approve.content`), an inquiry answer plus its session links, a user question's answers, or free-text feedback, and secret-pattern redaction does not touch ordinary prose. The old folded document never carried those rows. The assembler now keeps each decision's **action** and drops the payload; safe because the fold reads a decided row for its id alone (it closes the pending request at `sessionFold.ts:1554-1558` and renders nothing from the decision).

**The `trace.events[0]!` invariant — fixed at its source.** The emptiness guard now runs on the *filtered* display rows rather than the raw aggregate, so a document that exists really does have a creation row, and the export filename finds that row by type instead of by position.

**P2, response text serialized twice — acknowledged, not fixed here.** `stream.end.finalText` and `response.finalized.text` can both carry the full response, and the fold used to collapse them. A safe dedup is possible (drop `finalText` only when byte-identical to an adjacent `response.finalized` on the same row, which renders identically), but it adds exporter logic for a size-only win whose magnitude is still unmeasured — the PR body already flags that nobody has exported a real workflow run before and after. Better as a measured follow-up than a speculative change here.

**P2, older `trace.json` — needs an owner ruling, flagged.** AGENTS.md's compatibility section does name external export formats such as `trace.json` as the exception to 1.0's no-compat rule, which cuts against this PR's deliberate loud failure. Raising it with the owner rather than deciding it in review: a normalizer means keeping the whole previous document schema plus a converter, which is exactly the surface this sweep is deleting.

LionSR:
**Owner ruling on the older-`trace.json` finding: 1.0 exports are all fresh.** So the loud parse failure stands and no normalizer is added; the previous document schema and a converter would be exactly the surface this sweep removes. Worth noting for future readers that AGENTS.md's compatibility section still offers `trace.json` as its example of an external-format exception — that example is now out of date, and the rule's substance (wire protocols and formats with outside consumers) is unaffected.

This PR also now contains #12360 (delete the producerless `transcript.entry` durable arm), which merged into this branch rather than main, as a stacked PR does. Merging this one lands both: net about −407 against current main.

Rebased onto current main and re-verified there: typecheck clean, dead-code and effect-migration ratchets clean, and the 12 affected suites pass (131 tests).

PR #12363, opened 2026-09-12
refactor(session): one publisher for the run log; the ledger stops bypassing it

LionSR:
## Summary

Every write to a session's event table now goes through one inbox drained by one fiber in `SessionEvents`:

- `publish` — an awaited batch; the run ledger's door.
- `exclusive` — a read of committed rows and the append that depends on it, as one job (`decideRequest`, `updateRecordFacts`).
- `detach` — a synchronous enqueue for a producer with no fiber to wait on (the trace subscriber, a registry roster change, a policy snapshot).
- `settle` — wait for what is detached; reports the highest commit the cohort appended, fails with the aggregated refusals.

Commit order is enqueue order. Deleted from `SessionHandle`: `publicationGate`, the promise `Set`, `schedulePublication` and its `runPromise`. `settlePublications` still means "everything this session owes storage has landed and the view has folded it".

Why: the run aggregate had two writers. The handle's `schedulePublication` was gated by a `Semaphore(1)` acquired only after `runPromise` had forked, and `RunLedger.appendBatch` appended through the raw `SessionEvents.publish` with no permit at all. For a fast command the loop's settlement batch reached the connection before the card's start row ever got its permit; the transcript fold saw `tool.end completed` for an id it had never met, then a `tool.start` that opened a card nothing would close, and stamped it "The run ended before this tool completed" when the turn parked. In one session on 2026-09-12: 24 of 89 cards had their start row committed after their terminal row.

PRD 7.1 and substrate C6 already put the ordering inside `SessionEvents`; #11881 landed publish without it and the execution-metadata cutover put a gate in the handle instead. This restores the contract one level stronger: an inbox orders by the moment of the call, a permit only by acquisition. Codex (`send_event` → rollout mpsc with one writer task), OpenCode v2 (one `Event.publish` under the connection semaphore) and pi (one `emit` on one loop) each have exactly one door; decision record: https://claude.ai/code/artifact/87b0aaa8-58f0-44ea-8e80-0c17f7246698.

Review follow-ups folded in: the `stream.end` text fallback is read inside the detached job (after earlier chunk jobs), `exclusive` and `settle` wait only for the commits their own jobs produced, and result listeners now fire from the fold-gated tail (the same rule the registry's phase notification follows), which is what the toast test on CI caught.

## Issue acceptance gates

No issue filed; the defect and its evidence are in the decision record above and in the commit message. Gate satisfied: a detached row followed by an awaited batch commits in that order (regression test), and the 24-of-89 inversion cannot recur because there is no second append path.

## Single source of truth

`SessionEvents` (`src/agent/runtime/SessionEvents.ts`) owns publication order: one inbox, one consumer fiber, four doors. `SessionHandle` and `RunLedger` are callers of it, never writers beside it.

Removed: the second ordering mechanism (`publicationGate` + promise `Set` + `schedulePublication` in `SessionHandle`) and the ledger's bypass of the publisher. Added: no layer. The inbox replaces the gate inside the service that already existed; a job receives the log's `Append` as its only argument, so re-entering the publisher is unrepresentable.

## Net elements (R6)

- Files: 9 changed, +457 / −192 (3 docs, 5 source, 1 test). Source only: +412 / −173, net +239 (the inbox, its four doors, and their doc comments; the deleted gate was ~40 lines).
- Exported symbols: +2 (`Append`, `SessionEventReads` types in `@shared/session/sessionEvents`), −0.
- Classes/interfaces/enums: +1 private interface (`PublicationJob`), −0.
- Deleted members: `SessionHandle.publicationGate`, `SessionHandle.publications`, `SessionHandle.schedulePublication`.

## Consumer counts (R8)

- `session.publish([...])` (detached): 13 production call sites (grep `\.publish(\[` outside tests and the handle), unchanged signature.
- `commit` / `commitRegistration` / `updateRecordFacts` / `decideRequest` / `openRequest`: 24 grep hits across `src/agent`, `src/tools`, `src/controllers`, `packages/cli`, `packages/desktop` (the pattern also matches unrelated `.commit(`), unchanged signatures.
- `settlePublications` / `flushArtifacts`: 30 grep hits including comments and definitions; every call keeps its signature and its "landed and folded" meaning.
- `onResult`: 1 production subscriber (`terminalResultToast`), now notified from the fold-gated tail.
- `SessionEvents.publish`: 2 callers (`RunLedger.appendBatch`, `RunLedger.acquire`), unchanged signature, now ordered.

## Host impact

Extension: none beyond ordering; the status bar and progress view read the same fold.

Desktop: none beyond ordering.

CLI: the TUI no longer shows finished bash cards as failed at turn end.

## Scheduled refactoring

`feat/loop-owns-tool-card` (stacked): the loop becomes the one author of a tool call's card, tool output goes transient (C3), and the transcript fold's card becomes a monotone state machine. Observability-plane step 4 (the trace hub) now owns only the `TraceEmitter` listener set and the SDK's buffer cap.

## Validation

- `tsc` for `tsconfig.json`, `tsconfig.test-kernel.json`, the CLI and the agent package: green.
- `test:pure` (176 files, architecture ratchets included): green on the first revision.
- Session-plane, dispatch, registry, lifecycle, ownership, delegation, and toast suites (11 files, 190 tests): green on the revision that moved result listeners to the folded tail.
- eslint and prettier on the changed files.
- Not run locally: the full `npm test` (CI runs it).

https://claude.ai/code/session_01GcJWhANQ6QP4cd1S6cdB8o


LionSR:
Review follow-ups landed in `b83c480860`:

- **Stream-end text fallback (Codex P2, texra-ai low):** the draft, including the `readText` fallback, is now built inside the detached job, so it runs after every chunk job enqueued before it. The synchronous `detach` call still fixes the row's order.
- **Settle against each job's own commit (Codex P2):** `exclusive` settles against the last commit its own job appended and returns at once when it appended nothing; `settle` waits for the highest commit of the detached cohort it awaited. Both come from a tracking `Append` handed to the job; `now()` is no longer consulted for settlement.
- **Docstring vs `exclusive` refusal (claude):** the layer doc now says a refused detached job is logged and dropped and a refused awaited job is the caller's defect.
- **CI shard 1 (`terminalResultToast`):** a real consequence of the change, not a flake. Detached publications no longer wait for the fold between them, so a result listener reading the run's view could run before the run's `run.start` had folded. Result listeners now fire from the fold-gated tail, the rule `receiveFoldedEvent` already applied to the registry's phase notification.
- PR body rewritten to the template with the R6 and R8 sections.

PR #12379, opened 2026-09-13
refactor(run): the loop owns a tool call's card; tool output is transient; the card is a monotone machine

LionSR:
## Summary

Stacked on #12363, base `feat/one-door-publisher`. **Merge order: #12363 first.** GitHub retargets this PR to `main` when that branch is deleted on merge; do not merge this one into the publisher branch. A tool call's card has one author, the run loop:

- `DispatchFacts.logId` is required and minted for every call with the response.
- A slow tool's `tool.start` commits with the row that admits the attempt: the `tool.intent` batch for a barrier, its own batch for a parallel-safe call, the re-run intent after an outcome-unknown decision. A fast tool's card opens and closes in its settlement batch, as before.
- No tool publishes a start card: `deferLogUntilApproval`, `streamsOutput`, and the `onRunReady` hook are deleted from `ToolTypes`, `defineTool`, bash, codex, wolfram, and the agent-CLI tools. A denied approval is a card that closes failed, not a card that never opened.
- What a tool prints while it runs is transient text on the card id, the channel a response's chunks already use (`hooks.onToolOutput` → `stream.chunk`), never a `tool.end in_progress` row per chunk carrying the whole buffer so far. The dispatcher refuses chunks once the call has returned (pi's barrier and latch) and caps what streams at 50k characters; the settlement carries the bounded capture. The fold projects a running card's output from the held text; the terminal row replaces it and the session drops the held chunks.
- The transcript fold's card is a monotone state machine: opens once, takes progress while open, closes once; a second start reopens a running card (a re-run) and is a no-op on a closed one; progress after the close and a close without an open are dropped. Nested subcards the codex and Claude tools emit keep their durable `in_progress` rows (structured state, not output text).

Codex's session emits begin and end and its tool only sends output deltas; OpenCode v2's runner publishes every card fact and its fold applies progress only while `running`; pi's loop emits start and end and awaits every progress emission before the end. Decision record: https://claude.ai/code/artifact/87b0aaa8-58f0-44ea-8e80-0c17f7246698 (decisions 2, 3, 4).

## Issue acceptance gates

No issue filed; the decision record above holds the evidence (rows 1378 and 1379 of the 2026-09-12 session were consecutive full-buffer copies). Gates: no `tool.end in_progress` row is written by the dispatcher; a running card projects its live output and the terminal row wins over a late chunk (fold test).

## Single source of truth

The run loop (`toolUseDispatch.ts`) owns a call's card rows; `DispatchFacts.logId` owns the card id; the transient text source owns a running card's output; `traceFold.ts` owns the card's state machine.

Removed: the tool-side card opening (three flags, one hook, four call sites) and the durable progress row class. Added: no layer; two small helpers inside the dispatcher (`cardStart`, `admittedCards`), two fold predicates (`isRunningToolEntry`, `withToolOutput`), one `has` on the transcript writer.

## Net elements (R6)

- Files: 20 changed, +247 / −121 (3 docs, 14 source, 3 tests). Source only: +223 / −118, net +105 (the fold's card-output projection and the dispatcher's admission rows; the deleted flags and hook were spread thin).
- Exported symbols: +0 / −0. Deleted type members: `ITool.deferLogUntilApproval`, `ITool.streamsOutput`, `ToolCall.hooks.onRunReady`; `DispatchFacts.logId` narrowed from `string | null` to `string`.
- Classes/interfaces/enums: +0 / −0.

## Consumer counts (R8)

- `hooks.onRunReady`: 4 emit sites deleted (bash, wolfram, agentCliShared, and the dispatcher's construction), 0 remaining.
- `deferLogUntilApproval`: 3 declarations + 2 readers deleted, 0 remaining. `streamsOutput`: 1 declaration + 1 reader deleted, 0 remaining.
- `hooks.onToolOutput`: 1 producer (bash foreground), 1 consumer (the dispatcher), unchanged shape.
- `tool.end in_progress` rows: 0 remaining from the dispatcher; 8 emit sites in `src/tools` (codex, claudeAgent, background tasks) remain and are intentional structured subcards.
- `DispatchFacts.logId` readers: 1 (`toolUseDispatch.ts`); writer: 1 (`dispatchFactsFor`).

## Host impact

Extension: a bash card's output updates from transient text instead of a durable row per chunk; the same row model, no renderer change.

Desktop: same as the extension.

CLI: same; a card denied at the approval prompt now shows as a failed card instead of never appearing.

## Scheduled refactoring

None. `2026-09-04-agent-runtime-on-effect.md` line 298 and substrate C3 are amended in this PR.

## Validation

- `tsc` for `tsconfig.json`, `tsconfig.test-kernel.json`, the CLI and the agent package: green.
- Fold, session-events, bash, dispatch, recorder, store-load, resume-cancellation, wolfram, codex-progress, lifecycle, approval-prompt, and resume-data suites (12 files, 168 tests): green.
- eslint and prettier on the changed files.
- Not run locally: `test:pure` and the full `npm test` (CI runs them).

https://claude.ai/code/session_01GcJWhANQ6QP4cd1S6cdB8o

### DIFF
--- a/.agents/docs/proposed/architecture/2026-09-03-persistence-substrate-decision.md
+++ b/.agents/docs/proposed/architecture/2026-09-03-persistence-substrate-decision.md
@@ -436,7 +436,8 @@ and project publication steps; no cross-database atomic delivery is claimed.
 Panel drafts remain view state and are not part of canonical inquiry records.
 
 **C3. Durable event set.** Durable: every run-scoped `AgentEvent` except text,
-thinking, and tool-input deltas; `approval.requested`,
+thinking, tool-input deltas, and (amended 2026-09-13) the output a running
+tool prints, which reaches the fold as transient text keyed by the card id; `approval.requested`,
 `approval.resolved`, `approval.policy` (snapshot); `run.start` at the
 reservation commit point carrying `identity` (nullish only for imported legacy
 streams), `worktree` (nullish), and explicit `category` and `isRemote` fields,
--- a/.agents/docs/proposed/architecture/2026-09-04-agent-runtime-on-effect.md
+++ b/.agents/docs/proposed/architecture/2026-09-04-agent-runtime-on-effect.md
@@ -296,8 +296,19 @@ is durable, and a durable result cannot leave a running card behind.
 Each call's card correlation (`logId` and stage) is saved with the pending
 response, so a resumed settlement closes the same card. If no start card
 was published, the settlement batch includes its `tool.start` as well.
-Turn snapshots include the resulting state, pending intents, and references
-to any pending response and its already-settled calls.
+
+> **Amendment 2026-09-13 (loop-owned cards).** The loop is the one author of
+> a call's card. A slow tool's `tool.start` commits in the batch that admits
+> the attempt (the `tool.intent` for a barrier, its own batch for a
+> parallel-safe call, the re-run intent after an outcome-unknown decision); a
+> fast tool's opens and closes in its settlement batch. No tool publishes a
+> start card of its own (`deferLogUntilApproval` and the `onRunReady` hook
+> are deleted: a denied approval is a card that closes failed, not a card
+> that never opened), and what a tool prints while it runs is transient text
+> keyed by the card id, the same channel as a response's chunks (C3), never a
+> `tool.end in_progress` row.
+> Turn snapshots include the resulting state, pending intents, and references
+> to any pending response and its already-settled calls.
 
 Attachment settlement uses a JSON-safe representation rather than persisting
 `ToolFileAttachment` directly: its `bytes` field is a `Uint8Array`, which
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -149,6 +149,15 @@ emission via `SessionHandle.publish` / `SessionEvents`, so a new direct
 `appSignals.emit(...)` on the separate `AppSignals` bus within its documented
 scope.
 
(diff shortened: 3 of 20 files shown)


LionSR:
Pushed `2389ce61eb`: the fold test is rewritten against the test helpers main now has (it used `log.entry` and `STREAM_LOG_ENTRY_TYPES`, which #12332's rebase had removed from the file), which was both CI failures. The comment in `traceFold.ts` that still described the deleted deferred-card mechanism is gone.

On the two notes: the live-output cap is intentionally a hard cutoff after the first 50k streamed characters rather than a sliding tail; the settlement still carries the bounded head-and-tail capture, so nothing a user needs is lost, and a sliding window on transient text would replace the held text on every chunk. The `logId` narrowing is a conscious 1.0 fresh-state break: a pending response persisted by an older build refuses to fold on resume, loudly, rather than being read through a compatibility arm.

LionSR:
Pushed `11de4144e2` for the two Codex findings:

- **Clear tool chunks when an interrupted run closes the card:** the session's tail consumer now drops every held chunk of a run at the run's transcript boundary (`flow.step waiting`, `run.end`, `run.removed`), the same rule the fold's `clearInflight` applies to its in-flight text, in addition to a stream's or card's own terminal row. A card an interrupted run closed without a terminal row holds nothing.
- **Honor the output cap when a chunk crosses the limit:** the callback slices the chunk to the remaining allowance and counts only what it published, so the transient text never exceeds 50k characters.

PR #12392, opened 2026-09-13
refactor(tools): finish the defineTool sweep; reattach the pullStream doc comment

LionSR:
## Summary

The tail of #12384, which merged while its last commit was still in validation, plus the review follow-ups that came out of it.

1. **Nine tools carried the pattern #12384 targeted, in shapes its first pass missed.** Four hold `execute` as a class field — `protected readonly execute = Effect.fn('X.call')(function* (this: XTool, input) {…})` — where the `this: XTool` parameter annotation is what kept them classes, since it names the class being declared, and none of the bodies ever reads `this`. Five more are ordinary `protected execute(…)` methods that the first pass skipped on a 12-line size cap or on a bad `this` match.

   Converted: `glob`, `grep`, `open_pdf`, `report_review_issue`, `wolfram`, `texcount`, `github_subscription`, `lean_file`, `lean_project`.

2. **`pullStream`'s doc comment was orphaned.** Inserting the `PullValue` type between the comment and the function left the comment on the wrong declaration. Moved below `PullValue`, onto `pullStream`.

`claude` raised (1) on #12384 as a follow-up note, then on this PR caught that the first attempt at it was still incomplete; `texra-ai` independently reached the same list and raised (2).

### Where the sweep now stands

From a per-body `this\s*[.:]` match: **15** classes still extend `defineTool`, **13** of which genuinely read `this`.

The remaining two — `WorkflowAgentTool` (64-line body) and `WorkflowScriptTool` (413) — are `this`-free but stay classes deliberately. Inlining bodies that size into the options object trades one readability problem for another; converting them means first extracting module-level run bodies, which is a relocation rather than this mechanical sweep. Stated as an inventory rather than as a completeness claim, because the earlier claim was wrong twice.

Two measurement errors caused that, both worth naming since they are the reason this took three rounds: a `\bthis\b` search matched the `this: XTool` annotation itself and the word "this" inside a user-facing string (`LeanProjectTool`'s build notice, `Build started. Note: this command does not capture build output…`), and the 12-line inlining cap hid three bodies that read fine as an `execute` property.

## Issue acceptance gates

n/a — review follow-up on #12384, no issue.

## Single source of truth

Unchanged from #12384: `src/tools/core/definition.ts` owns whether a definition yields a concrete or an abstract class. This PR moves nine more tools onto the concrete path.

Removed: 9 forwarder subclasses and 5 vestigial `this: XTool` annotations. No abstraction added — this uses the `execute` option #12384 introduced.

## Net elements (R6)

| | |
|---|---|
| Files changed | 10 (0 added, 0 deleted) |
| Net LoC | **−24** (+54 / −78) |
| Classes | **−9** exported tool subclasses (53 → 15 across both PRs) |
| Exported symbols | **net 0** — the nine tools stay exported, as `const` rather than `class` |

Measured as `git diff --shortstat $(git merge-base HEAD origin/main)..HEAD`. An earlier revision of this section said −26 (+50/−76); that predated the last commit and was wrong.

Dead-code ratchet: no new findings (12 unused / 213 combined, unchanged).

## Consumer counts (R8)

No emit path or export deleted or re-routed. All nine tools keep their names, module paths and construction site — `registry.ts` still calls `new X()`. One kernel suite named `ReportReviewIssueTool` in a type position and now uses `InstanceType<typeof …>`, matching the setup suites; a grep for type-position and `typeof` references to all nine finds no others.

## Host impact

`src/tools`, one kernel suite, and a doc comment in `packages/llm`. No `vscode` import, no VS Code-free-zone boundary touched, no setting, command or wire contract changed.

## Scheduled refactoring

None. The two remaining `this`-free classes are named above with the reason they stay; that is an inventory, not a deferred task list.

## Validation

- `npm run typecheck` — clean across all seven projects (the gate that caught the type-position and `.d.ts`-emit breaks on #12384; run before each push, not after).
- `npm run lint` — clean.
- `npm test` — **529 files / 5538 tests passed**, 2 skipped.
- CI green on `b6e0ccf`: validate, static checks, both kernel-test shards, build artifacts, webview smoke, guidance references, detect changes, label. `docs build` and `runtime validation` skipped by design.
- Branch restarted from `main` at `0134a90` (the #12384 squash) and the work replayed onto it, so the diff is only the files above.
- Not run: no manual UI pass, since no host or webview file is touched.

https://claude.ai/code/session_01RqUaNLHTFqGwHGmstFLFwX


LionSR:
Good catch on `WolframTool`, and you were right that the "exhaustive" claim did not hold. Auditing it properly turned up **five**, not one — pushed in b6e0ccf: `wolfram`, `texcount`, `github_subscription`, `lean_file`, `lean_project`.

Two things had been hiding the others, both my errors:

- **A `\bthis\b` search instead of `\bthis\s*[.:]`.** The bare word matched the `this: XTool` annotation itself, and in `LeanProjectTool` it matched the word "this" inside a user-facing string (`Build started. Note: this command does not capture build output…`). So a tool got classed as "uses `this`" on the strength of its own build notice.
- **A 12-line cap on inlining a body**, left over from the first pass. `github_subscription` (23 lines), `lean_file` (26) and `lean_project` (27) were skipped by size alone, and all three read fine as an `execute` property.

The corrected inventory, from a per-body `this\s*[.:]` match: **15** classes remain, **13** of which genuinely read `this`. The other two are `WorkflowAgentTool` (64-line body) and `WorkflowScriptTool` (413) — `this`-free, but inlining bodies that size into the options object trades one readability problem for another. Converting them means first extracting module-level run bodies, which is a relocation rather than this mechanical sweep, so they stay classes and I have replaced the completeness claim in the PR body with this exact inventory rather than restating it.

Verified on b6e0ccf: full `npm run typecheck` (all seven projects), `format:check`, `lint`, dead-code ratchet (no new findings), `npm test` — 529 files / 5538 tests.

---

LionSR:
@copilot resolve the merge conflicts on this branch.

Issue #12433, opened 2026-09-14
Desktop: a run fiber leaves the session's AsyncLocalStorage scope after a contended SessionEvents commit


LionSR:
## Current status — 2026-09-15

Checked against main 2c9898445d on 2026-09-15. Parent and structural-fix owner: #12421; verification: #12427.

The reported contention/async-local-context exposure remains actionable. Landed caller conversions (#12478, #12520, #12521, #12522, #12523) reduced the exposed paths but did not delete the carrier.

Verified remaining readers include getConfig in run/modelBinding.ts, retry configuration in ModelInvoker, debugMessageSaver's configuration/filesystem statics, GitHub RunSubscriptionRegistry.bind's currentSession(), and approval helpers. Shared getters conceal further callers.

- [ ] Replace these dependencies with the run/session's explicit configuration, rooted filesystem and owning session, covering host callbacks as well as run fibers.
- [ ] Verify two different project sessions across a forced contended SessionEvents commit using an existing durable-boundary suite.
- [ ] Retire the ambient carrier with its last consumers under #12421; do not add more runInSession/inScope repairs.

No real desktop reproduction was newly executed in this organization pass; the issue's reported mechanism and source exposure remain distinct from a completed regression test.

<details>
<summary>Historical issue description (superseded status; retained for evidence)</summary>

## What happens

A run fiber that starts inside `runInSession` loses the session's `AsyncLocalStorage` scope after a contended ledger commit. From then on, any ambient roots read on that fiber that is not wrapped by `run.inScope` (`AsyncLocalStorage.bind`) or `runInSession` resolves against the **process** roots instead of the session's.

On the extension and CLI the process and session roots are the same, so nothing visible happens. On the desktop, process roots carry no workspace and each open paper is its own session, so such a read returns no workspace path, the wrong storage root, or process-level settings.

## Why

Effect 4 resumes a fiber that is waiting on a `Deferred` synchronously in the JS context of whoever completes it. Every ledger append goes through the `SessionEvents` inbox (`src/agent/runtime/SessionEvents.ts`): the consumer is forked with `Effect.forkScoped` when the session opens, takes jobs from `Stream.fromQueue`, and completes each job's `Deferred`. On the desktop the session opens before `runInSession` is entered (`packages/desktop/src/main/desktopProjects.ts`), so the consumer runs outside every session scope, and so does any run fiber it wakes.

Throwaway probe of that exact queue shape:

| Case | Scope after the commit and later awaits |
[...]

Issue #12421, opened 2026-09-14
Tracking: finish rooted filesystem and explicit project context


LionSR:
## Current status — 2026-09-15

Checked against main `2c9898445d809b98b038ca9b133b16b57ef7794a` on 2026-09-15. This section supersedes the historical status below; accepted repository policy and later owner rulings remain authoritative.

Parent: #12025. Canonical implementation owner for #12071's filesystem/ambient portion and the corresponding inventory in #12078.

## Landed

Foundation #12432; native LaTeX/reflection portions #12478; extension consumer work #12477; media work #12476; desktop work #12521; tool/controller roots #12520/#12522/#12523; filesystem hostPort work #12534; process-roots holder retirement #12836. These are partial consumer conversions, not completion of every directory named in the historical inventory.

## Remaining order

- [x] Prioritize exposed run readers under #12433: modelBinding, ModelInvoker retry settings, compaction/provider configuration, debug storage, GitHub subscription session binding, and approval helpers. Supply explicit project config, roots, and session; do not add another ambient re-entry wrapper.
- [ ] Finish remaining LaTeX, utility, tool, agent/controller and host consumers against existing WorkspaceFs/StorageFs and Effect FileSystem/Path. Preserve live configuration reads where required.
- [x] Resolve remaining BaseFS methods and publisher ordering under #12496 with their surviving consumers; do not invest in an interface being removed.
- [x] Supply the rooted pasted-image cleanup operation needed by #12424, preserving the intended file ownership contract.
- [x] Delete obsolete AbsoluteFS/WorkspaceFS/StorageFS/RelativeFS, BaseFS, old filesystem port/provider and their dedicated tests after their consumers have moved.
- [x] Delete ambient workspace-root fallback and now-unnecessary inScope re-entry after all session-sensitive run and host/callback readers have moved.
- [x] #12843 - Tighten `CliPlatformServices.roots` to required once the foreign-install path publishes its own roots.

Current survey: 27 platform calls (13 in BaseFS), 30 ambient sites across 20 files. Shared getters hide additional callers, so use complete call paths for acceptance.

The September 13 filesystem ruling supersedes the old deferral. #12307 has merged; its previous desktop-main file reservation is not a current blocker. Refresh actual open PR files before implementation.

Preserve lstat/symlink identity, typed directory walks, atomic publication and containment with the existing supported helpers. Direct Node operations at justified implementation boundaries are not automatically forbidden. The final generated-file deletion contract remains #12139.

<details>
[...]

PR #12445, opened 2026-09-14
refactor(reflection): the output pipeline is Effect programs the reflection loop composes

LionSR:
## Summary

The reflection output pipeline (`src/agent/implementations/flows/reflection/output/**`) becomes Effect programs that the reflection loop composes with `yield*`. The `run.inScope` + `Effect.tryPromise` bridges around the pipeline and the two `effectRuntime().runPromise` calls below the loop are deleted.

## Issue acceptance gates

Part of #12025 (reflection/export operations), with the roots work feeding #12421.

- Every output step returns an `Effect`; `runReflection` is the only production caller.
- The 2 below-boundary `Effect.run*` and 2 `effectRuntime()` calls in `LatexDiffManager` are gone: the latexdiff service is yielded directly, so interrupting a run now cancels the latexdiff child process instead of detaching it.
- The 3 `platform()` / `platform().fs` reads are gone: `publishCompiledPdfArtifact` and `LatexDiffManager` take `FileSystem` from context, following `jsonStore`.
- `withOutputStage` is deleted, not converted: every call passed `skip: true` with no ambient stage, so it emitted and stamped nothing.

## Single source of truth

- **Recovery:** one combinator, `recoverOutputFailure`, replaces `tryOperation`. Typed failures recover loudly at the same level and message; defects and interruption propagate. The loop's `fallbackOutput` is the one broad recovery and is interrupt-aware, so a cancelled run stays cancelled.
- **Roots as data:** the pipeline, `TaskRunFileService` and the latexdiff reflection path take the run's session roots as data (#12390's pattern). Workspace, storage, setting and replacement-rule reads go through rooted forms (`locateInWorkspace`, `readSettingFrom`, `readConfig`, `pathToLocationIn`, `runDirUnder`, `originalSnapshotPathUnder`, `runStorageAbsolutePathUnder`, `runStorageLocationUnder`, `ensureRunDirUnder`). The existing ambient helpers delegate to them, so callers outside this pipeline behave as before.

No new service, adapter or file. Each rooted form is the single implementation its ambient counterpart now calls. `getRunStorageAbsolutePath` lost its last caller and is deleted.

## Net elements (R6)

40 files, +2511 / −1798. −1 Promise helper, −1 dead stage wrapper (`withOutputStage`, `RoundSummary.stage` and a parameter), −1 orphaned export. `Effect.run*` −2, `platform()` −3, `effectRuntime()` −2; four ratchet file entries reach zero and are deleted. `ambient:asyncLocalStorage` unchanged.

## Consumer counts (R8)

`StageHandle.run` / `within` lose their last production caller with `withOutputStage`. No emit path removed; trace payloads are preserved verbatim at every recovery site.

## Host impact

Extension and CLI: behaviour preserved (process roots and session roots are the same there).

Desktop: this is where the roots work matters. Process roots carry no workspace and each paper is its own session. Removing `run.inScope` would have moved the pipeline's reads onto process roots, so they now read the session's roots as data: compile check, output extraction and publishing, every run-storage path, and the latexdiff reflection path (its settings, output location and replacement rules). `LatexCompileInputDirs` and `LatexdiffShadowStorage` each pin one case with differing process and session roots.

The session's AsyncLocalStorage scope does not survive a contended `SessionEvents` commit (#12433). So the three remaining calls into other lanes' code keep main's exact scoping through `run.inScope`: the two `compileLatex2Pdf` calls and `replacementEngine.applyFor`. Each is commented and removed by #12421.

## Scheduled refactoring

- `compileLatex2Pdf` stays a Promise: it has six production callers, and converting it belongs with the LaTeX slice of #12421, which also removes the three remaining `run.inScope` wraps.
- #12433: bare ambient roots reads on run fibers elsewhere.

## Validation

- `npm run typecheck`, `npm run lint`, `npm run format`, effect-migration and dead-code ratchets: pass (rebased on `8f0b294791`).
- `vitest run --changed origin/main`: 3189 passed, 1 skipped, 1 failed. The failure is a 10-second timeout in `StaticBandResize.vitest.ts`, which fails the same way on clean `origin/main` under the same machine load. CI decides.
- Mutation-checked pins: passing process roots instead of session roots makes both new cases fail.

https://claude.ai/code/session_01UBxgL2kThfyEFsfyhEnPFQ


LionSR:
All four addressed in 66c03a8a9c, with the branch rebased on current main.

- **LAY-1 subsystem edge:** `src/replacement` no longer imports `@platform`. The engine's config getters take a `ReplacementConfigRead` (`(path) => value`) declared in the engine, defaulting to `getConfig`. `DiffFileProcessor` adapts the session's `roots.config` to it through `readConfig`. `npm run test:pure` passes (175 files, 2,116 tests).
- **`XmlOutputManager.vitest.ts`:** the Promise bridge is gone. Every case yields the extraction program under `@effect/vitest`, and the `it.each` tables are loops over `it.live` cases, the same 79 cases, because they read and write real temp directories.
- **Real filesystem cases:** `CompiledPdfArtifacts`, plus `CompileCheck` and `LatexCompileInputDirs`, which also write real temp directories, now use `it.live`.
- **`LatexDiffManager.ensureWorkspaceDependency`:** a failed existence check is recovered together with the mirror, as one `warn` naming the dependency's path and cause, and the remaining pairs still diff. On main the existence check sat outside that `try`, so this also closes an older gap.

`vitest run --changed origin/main`: 3,189 passed with one 10-second timeout in `DefaultSessionLifecycle.vitest.ts`, which this change does not touch and which passed 6/6 when run alone.

LionSR:
@codex review

LionSR:
need to rebase

LionSR:
Rebased onto main after #12448, #12450 and #12451; the only conflict was the generated effect-migration baseline, regenerated from the tree. On the rebased head: typecheck and npm run test:pure (2,116 tests) pass, and the branch's 15 changed suites pass (237 tests). Codex's re-review of the previous head found no major issues.

Issue #11866, opened 2026-09-04
Lane 8: the conversation shell


LionSR:
## Current status — 2026-09-15

Checked against main `2c9898445d809b98b038ca9b133b16b57ef7794a` on 2026-09-15. Current status below supersedes old counts, branch reservations, and implementation-stage descriptions.

## Disposition

All shell implementation children #11872–#11876 are complete; the later #11941 Active Now follow-up is also closed. The issue discussion records the merged shell implementations (#11916/#11918/#11911/#11917). Wider runtime work remains under #12025, and #11861 retains its separate SDK acceptance item.

The previous issue text is retained below as historical evidence. This closure applies to this issue's scope, not the entire Effect conversion.

<details>
<summary>Historical issue description (superseded status; retained for evidence)</summary>

Part of #11861. Depends on lane 4 (#11863) and lane 6 (#11827, merged). Last on the critical path. Touches the `packages/extension` frontends and `packages/desktop/src/renderer`.

PRD section 12, the shell. Spec: every element on the boards names the field of `SessionView`, `Surface`, or the `host` snapshot it reads; nothing is derived in a component. Boards: the design canvas "TeXRA Conversation Shell", https://claude.ai/code/artifact/c5d655b7-42be-42c0-aebd-a71ee4d6a035 (real-component renders on its "Built with our components" page, produced by the untracked design harness under `packages/desktop/design-harness/`). Directions and critique: `docs/proposals/2026-09-03-conversation-shell-directions.md`.

## Recommended sequence (shell proposal, "Recommended sequence")

[...]

PR #12459, opened 2026-09-14
Extract resolveSelectedId for shared run selection logic

LionSR:
## Summary

Refactors the run selection resolution logic (PRD 9) into a reusable `resolveSelectedId` function in the shared `Surface` module, eliminating duplication between the extension/desktop `Surface` and the CLI's `cliState.ts`. The CLI's `selectedRunId` computed signal now calls the shared function with a `keepWhenViewEmpty` option to preserve its pre-run local conversation behavior.

## Issue acceptance gates

Addresses the narrow proposal in #11866 (session systems survey): "The TUI never adopted `Surface`" section. Extracts the selection rule as a shared primitive to enable incremental `Surface` adoption in the CLI without requiring full migration upfront.

## Single source of truth

`src/shared/session/surface.ts` now owns the run selection resolution rule (PRD 9). The `resolveSelectedId` function is the canonical implementation; `resolveSelected` wraps it for `Surface` objects, and the CLI calls it directly with its own options.

**Removed duplication:** The CLI's `cliState.ts` previously reimplemented the selection fallback logic inline in `selectedRunId`. That logic is now delegated to `resolveSelectedId`, which is shared with the extension and desktop `Surface` implementations.

**New abstraction:** `resolveSelectedId(view, selected, options)` owns the selection rule and its `keepWhenViewEmpty` option, which covers the CLI's pre-run local conversation (a Surface-only id that predates any run in the view). The option is documented to clarify that extension and desktop never set it (their `selected` is only ever set to a run the view already knows).

## Net elements (R6)

- **Files:** 2 modified (`src/shared/session/surface.ts`, `packages/cli/src/chat/tui/state/cliState.ts`)
- **Exports:** +2 (`resolveSelectedId`, updated `resolveSelected`)
- **Net LoC:** ~+15 (added function, expanded docs, CLI simplification)

## Consumer counts (R8)

- `resolveSelected`: 1 call site (new wrapper in `surface.ts`); existing callers unchanged
- `resolveSelectedId`: 2 call sites (new: `cliState.ts` `selectedRunId`, and `resolveSelected` wrapper)

## Host impact

**Extension:** No change. Continues to use `resolveSelected(view, surface)`.

**Desktop:** No change. Continues to use `resolveSelected(view, surface)`.

**CLI:** `selectedRunId` computed signal now calls `resolveSelectedId` with `keepWhenViewEmpty: true` instead of reimplementing the fallback logic. Behavior is identical; code is now shared.

## Scheduled refactoring

The remaining items in the "TUI never adopted `Surface`" section (#11866) are unchanged and still open: `Surface.expanded` vs. `expandedStreams`, `Surface.launch` vs. `sessionMeta`, `Surface.phase` vs. `WORKFLOW_POPUP_VIEW`, and full `Surface` adoption.

## Validation

- Existing tests pass (no new test files; logic is extracted, not changed)
- TypeScript type checking passes
- The CLI's `selectedRunId` behavior is preserved: pre-run local conversation ids are kept when the view is empty, and fall back to the first run once any run exists

https://claude.ai/code/session_012EqPaYMkmto6ykSEX5ew9X

### DIFF
--- a/.agents/docs/proposed/simplification/2026-09-09-session-systems-survey.md
+++ b/.agents/docs/proposed/simplification/2026-09-09-session-systems-survey.md
@@ -124,6 +124,13 @@ substrate. Not #11867, whose SQLite cutover has landed, and not #11869.
 
 ## 3. The TUI never adopted `Surface`
 
+**The narrow proposal below has landed** (`resolveSelected` split into
+`resolveSelectedId` plus the `keepWhenViewEmpty` arm; `cliState.ts`'s
+`selectedRunId` now calls it instead of reimplementing the rule). The rest of
+this section — `Surface.expanded` vs. `expandedStreams`, `Surface.launch` vs.
+`sessionMeta`, `Surface.phase` vs. `WORKFLOW_POPUP_VIEW`, and the full
+`Surface` adoption in #11866 — is unchanged and still open.
+
 PRD §10.1 says in writing that the TUI "Gains a `Surface`." It did not.
 `rg -c "shared/session/surface" packages/cli/src` returns zero: the TUI imports
 no part of the shared record, not `applySurfaceAction`, not `pruneSurface`, not
--- a/packages/cli/src/chat/tui/state/cliState.ts
+++ b/packages/cli/src/chat/tui/state/cliState.ts
@@ -14,6 +14,7 @@ import {
   type RunId,
 } from '@shared/schemas';
 import type { RunView } from '@shared/session/sessionView';
+import { resolveSelectedId } from '@shared/session/surface';
 import { RUN_GROUP_LABELS } from '@shared/runs/runStatusDisplay';
 import type { WorkflowRowGroup } from '@shared/runs/workflowRunModel';
 import { sessionView } from './sessionView';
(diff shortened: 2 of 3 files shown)


LionSR:
This PR is redundant and should not be merged as-is.

Its head (`claude/modest-lovelace-e6rg2s` at `b2942ab`) is the same commit that was already merged to `main` via #12318 two days ago. That work was then further refined by #12395, which moved the `CLI_LOCAL_RUN_ID` sentinel handling out of `resolveSelectedId`'s `keepWhenViewEmpty` option (introduced here) and into `cliState.ts` directly — fixing an edge case where `keepWhenViewEmpty` could keep *any* stale selection when the view was empty, not just the CLI's reserved pre-run id. Diffing this branch against current `main` shows it would regress `src/shared/session/surface.ts` and `packages/cli/src/chat/tui/state/cliState.ts` back to the pre-#12395 design (`mergeable_state: dirty` reflects that conflict).

I attempted to restart this branch from `origin/main` (per the repo's own merged-PR-branch-reuse convention) so this PR would show an empty, safely-closeable diff, but both the force-push and even a local branch reset were blocked by this session's destructive-git-action guard. I'm stopping here rather than working around that block.

Recommendation: close this PR without merging — its contribution already landed via #12318 and was superseded by #12395. If the branch is reused again, it should be restarted from `main` first.

https://claude.ai/code/session_012EqPaYMkmto6ykSEX5ew9X

---

LionSR:
Good catch — that contradiction is real and predates this PR (it's been on `main` since #12318 merged). Since this PR itself is redundant and being closed rather than merged (see the earlier comment), I've fixed it directly against `main` in #12461 instead of pushing here.

https://claude.ai/code/session_012EqPaYMkmto6ykSEX5ew9X

---

LionSR:
Closing as superseded. `resolveSelectedId` is already the shared selection rule on main (`src/shared/session/surface.ts`), extracted by #12318 and used by `cliState.ts`. The extra `keepWhenViewEmpty` arm here conflicts with the rule main settled in #12395: a Surface-only id the view never holds (the CLI's pre-run local conversation) is the caller's to keep out of the fallback, because a fallback from a view that lists every run of the workspace lands on an unrelated run. Nothing in this branch remains to land.

PR #12470, opened 2026-09-14
test(tools): goal rows and Wolfram observe their callbacks through Deferreds, not polls

LionSR:
## What

goalRows.vitest.ts drops three effectRuntime().runPromise adapters and two runFork collectors (body-level Effect.run* 9 to 0), moves one rejects.toThrow onto the typed Error channel with Effect.flip, and turns seven try/finally session teardowns into scope finalizers. Wolfram.vitest.ts replaces a 10 ms polling loop over a mutable array with a Deferred completed by the callback that records the request, so the wait is exact and a regression shows as the kernel timeout instead of a 2 s poll message.

Test-only change: no production code, no src/test-kernel/support helper, no vitest config, no ratchet baseline touched. Titles and declaration counts are unchanged; expect() counts rise only where an assertion moved out of a deleted poll callback.

## Per file

| File | State | Mechanisms | Tests | expect() |
|---|---|---|---|---|
| tools/Wolfram.vitest.ts | MIGRATED | Deferred gate replaces the waitForCondition 10 ms poll over requests.opened, completed by the same autoDecideRequests callback that records the request; scope-bound Effect.acquireRelease for the session and for the reque | 2 -> 2 | 7 -> 7 |
| tools/goal/goalRows.vitest.ts | MIGRATED | whole file moved to it.effect; rejects.toThrow became Effect.flip on the real typed Error channel; seven try/finally session teardowns became Effect.addFinalizer inside the test scope, registered before publishTestRunSta | 7 -> 7 | 14 -> 14 |

## Disclosures and follow-ups

- src/tools/wolfram/WolframTool.ts exports wolframRunSummary with no remaining consumer anywhere (already recorded in config/ratchets/knip-baseline.json:1250 as production-dead). Now that the last test import is gone, the export itself can be deleted and the baseline row shrunk; that is a production edit, out of lane.
- progressTestUtils createRecordingHost() opens a second test session (progressTestUtils:186) that no caller disposes. Pre-existing leak in a shared support file, out of lane for this charter since support files are frozen.
- dispatchWolfram in the Wolfram suite still raises a plain throw new Error for a non-bash request payload inside an Effect generator, which lands on the defect channel rather than a typed failure. Left as is because fixing it would change how a regression surfaces, not what is asserted.

## Review

An independent reviewer read every hunk against the migration cookbook and re-ran the suites: verdict ship (fix round applied: rebased onto current main and re-verified).

## Verification

Rebased onto origin/main in an integration worktree, then: typecheck:test-kernel clean, eslint clean on the changed files, prettier clean, vitest over the changed files run twice with identical results (2 changed files, 10 tests passed). Diff: 2 files changed, 149 insertions(+), 163 deletions(-).

https://claude.ai/code/session_01KQYjY1FpBHZu9t6EV1kvZ2


LionSR:
Heads-up: since this merged, origin/main fails the dead-code ratchet with one new finding, src/tools/wolfram/WolframTool.ts: wolframRunSummary, because this PR removed the only consumer (the import in src/test-kernel/tools/Wolfram.vitest.ts). Every open PR's "static checks (linux)" job now fails on it (seen on #12472 and #12473). The local main checkout carries an unpushed commit that restores the import; pushing that, or dropping the export if the helper is genuinely dead now, clears it.

LionSR:
Update: #12473 carries the fix (wolframRunSummary loses its export, its in-file caller stays, the baseline row shrinks 210 to 209). If main is fixed independently by restoring the import instead, the knip-baseline.json hunk in #12473 will conflict trivially on rebase.

LionSR:
The standalone fix is #12479 (unexport plus the baseline row), so PRs blocked by this can rebase onto it once merged; #12473 carries the identical hunk and will rebase trivially.

Issue #12424, opened 2026-09-14
Tracking: finish direct Effect host and settings contracts


LionSR:
## Current status — 2026-09-15

Checked against main `2c9898445d809b98b038ca9b133b16b57ef7794a` on 2026-09-15. This section supersedes the historical status below; accepted repository policy and later owner rulings remain authoritative.

Parent: #12025. This is the implementation owner for remaining internal host/settings/account contracts; #12428 records distinct host-controller semantics.

## Landed

- [x] hostPort callers converted and hostPort.ts deleted in #12537.
- [x] PlatformSecrets is Effect-typed, including reads (#12519).
- [x] Preparatory raw-catch conversions #12538–#12550 and #12552 landed.
- [x] SetupPlatform.layer accepts a value already; this is distinct from its Promise-shaped signIn member.
- [x] Merged and verified #12553: stores constructed before runtime; Secrets/AppState layers accept values; production thunks and CLI state latch removed.
- [x] Complete ExternalOpener's internal contract; retain genuine foreign shell/editor cancellation limitations (#12560).
- [x] Convert configuration writes through JsonConfigProvider and all callers, composing the store's Effect write directly (#12615).
- [x] Convert StateStore/worktree/Memento callers and retire internal RunStateWrite forwarding when no legitimate use remains (#12624).
- [x] Complete account/OAuth and SetupPlatform.signIn through their host implementations and callers (#12653).

## Remaining

- [ ] Complete session-owned runAgentRequest admission and settlement with #12425; preserve one result presentation.
- [x] Complete MessageHost notifications and remaining action prompts without changing dismissal behavior; PromptHost itself is already Effect-typed.
- [x] Complete pasted-image cleanup when its rooted filesystem operation is supplied under #12421.

## Acceptance

No internal Promise method merely executes an Effect through an injected runner. Only actual foreign APIs retain one explicit adapter. Preserve configuration precedence/targets, credential protection, commit masking, and required post-write effects. Audit every changed member's await, Promise.resolve, then/catch and callback-return sites under #12491; typechecking alone can accept an unexecuted Effect. The zero execution-ratchet count is not sufficient.

The old dependency on an unfinished production effectRuntime retirement is obsolete. The production row is zero; store construction landed in #12553.

<details>
<summary>Historical issue description (superseded status; retained for evidence)</summary>

[...]

PR #12471, opened 2026-09-14
refactor(platform): Secrets and AppState are Effect-typed ports with tagged errors; the credential plane drops hostPort

LionSR:
Lane A of #12424, the credential and store plane. **Opened as a draft: the
production code typechecks on every project, `src/test-kernel` and the Effect
migration ratchet do not yet.** What is left is listed below, with the reason
the lane did not fit in one pass.

## What changed

`PlatformSecrets` (`get`, `getStored`, `set`, `delete`, `listStoredKeys`),
`StateStore.update` and `ConfigProvider.update` return `Effect`s. The four
host implementations return Effects too: `VscodeSecrets`, `ElectronSecrets`,
`CliSecrets`, and the agent package's environment-only store. A new
`VscodeStateStore` is the extension's implementation of the state port, since
`vscode.Memento.update` is a `Thenable` and the port is no longer.

`Secrets.layer` and `AppState.layer` take the store's open program instead of
a thunk, and build it with the layer. That deletes the
`Layer.succeed`-over-a-thunk indirection §2 of the issue names: the service is
the store, and nothing resolves anything per member call. Two consequences
follow for free. The CLI's `bindCliGlobalState` / `cliGlobalState` latch is
gone, because the process runtime opens the global state store as its layer's
build step and `initCliPlatform` reads it back off `AppState` rather than
opening a second view of the same file. And `CliSecrets` and `ElectronSecrets`
hold no `ProcessRuntime` at all, which removes four `runtime.runPromise`
boundary calls that existed only because the port was Promise-shaped;
`JsonStore` loses its Promise-shaped `update` and its `runtime` option with
them.

## Error vocabulary

One tagged error per port, with a `reason` union derived from what the host
implementations actually throw today rather than invented.

`SecretsFailed` (`src/platform/secrets.ts`) carries `reason`, `operation`,
`key` and `cause`:

- `enumeration-unsupported` — VS Code's `SecretStorage.keys()` on a host that
  does not implement it. Only `listStoredKeys` raises it.
- `store-unavailable` — Electron `safeStorage` unavailable, or Linux
  `basic_text` backing. Only `set` raises it.
- `decrypt-failed` — the OS refused to decrypt a stored value. This is the
  failure `ElectronSecrets` used to swallow silently; it is still recovered to
  "no saved secret" so the renderer bootstrap survives, but now the recovery
  is explicit, logged, and warns the user once.
- `io` — the backing file or host API failed.

`StoreWriteFailed` (`src/platform/interfaces.ts`) carries `reason`, `key`,
`message` and `cause`, and is shared by `StateStore` and `ConfigProvider`
because both persist through the same two owners:

- `not-serializable` — the value is not JSON, so nothing was written.
- `io` — the backing store rejected the write.

## The Q2 rule

Host-controller study Q2 was ruled: a credential commit survives cancellation.
Every implementation of `set` and `delete` runs its commit region under
`Effect.uninterruptible`, so interruption is honoured before the commit starts
and observed after it lands, never in the middle of a write. Waiting for the
secrets file's write lane stays interruptible, matching what
`SerializedWrites` already does in the auth subsystem. The rule is documented
at the port rather than at each implementation.

## hostPort counts

Repo-wide: 122 sites in 44 files before, 96 in 34 after.

In this lane's thirteen files: 27 before, 8 after. The eight that remain are
not Secrets, AppState or Config sites and belong to other lanes:

- `src/tools/setup/platform.ts:70`, `:77` — `SupabaseClient.isAuthenticated`
  and `getUser` (the account plane).
- `src/tools/setup/platform.ts:91` — `getCodexStatus`, which reaches the OAuth
  coordinator, still Promise-shaped.
- `src/tools/setup/platform.ts:94` — `isCodexSubscriptionActive`, a capability
  probe that is not a host port.
- `src/common/teams/TeamAvailabilityPreflight.ts:89`, `:120` and
  `src/common/teams/TeamPlan.ts:392` — the team catalog's
  `canAccessRemoteCatalog` and `signIn`.
- `src/common/teams/TeamAvailabilityPreflight.ts:105` — `options.choose`, a UI
  prompt, which is lane D's.

## What is left for lanes B to F

Lane B (filesystem), lane C (process and terminal), lane D (presentation and
`HostInteractions`) and lane E (VS Code commands and editor) are untouched
except where a Secrets or AppState signature forced the smallest possible
change at their call sites:
`src/tools/setup/UnsetApiKeyTool.ts` (lane E) now yields the secret store's
Effects directly, `src/controllers/session/hostRunActions.ts` and
`src/controllers/session/hostSnapshotSource.ts` (lane F) lost the `hostPort`
wraps around the ports this lane typed, and
`src/controllers/progressView/ProgressApiKeyRetryController.ts` (lane D) takes
Effect-shaped `readKey` and `hasUsableKey` deps.
`SetupPlatformShape`'s `commands`, `extensions` and `terminal` adapters are
deliberately untouched: `terminal` is declared in `src/hosts/uiHosts.ts`,
which lane D owns, so typing it here would have taken two lanes' files.

## What is not done, and why

`npm run typecheck` is green for `typecheck:workspace`, `typecheck:cli`,
`typecheck:desktop` and `typecheck:agent`. It is **not** green for
`typecheck:test-kernel`: 289 errors across about fifty suites remain, almost
all of them a fake or a test double that still implements the Promise-shaped
port, or an `await` on a store call that is now an Effect. The Effect
migration ratchet also fails: five files gained an `effect` import while still
containing raw `catch` clauses, so the `catch:effect-importer` row sees them as
new entries. `packages/extension/src/extension.ts` is the worst of these with
twelve.

The reason both are outstanding is worth recording, because it is the finding
this lane produced: **Effect-typing `PlatformSecrets.get` does not stop at the
credential plane.** It runs through `src/model/apiProviders.ts` into the whole
model-availability computation (`computeModelOptions`,
`codingPlanSubscriptions`, `kimiCodeSubscriptionRouting`, `modelRoutes`), from
there into the settings controllers and the agent roster, and from there into
every host surface that renders them — including the CLI TUI's prop chain,
which is the same threading PR #12447 was written to do and which was closed
unmerged. The same is true of `StateStore.update`. The manifest sized lane A
at 27 call sites in 13 files; the actual reachable set is the 145 files in this
commit plus the test suites. A follow-up should either split this into a
schema PR and a per-host PR, or land it behind the CLI runtime threading that
#12447 was going to provide.

Two behavioural notes for review:

- `computeModelOptions` now resolves every provider's key status once when it
  builds its availability context, instead of lazily per provider. The per-model
  pass below it is synchronous as a result. The picker still warns once per
  provider, as #11508 requires.
- `selectAvailableDelegationModel` lost its `withScope` parameter. It bound an
  AsyncLocalStorage frame around a read whose stores are now passed explicitly
  and whose remaining preferences are process-global, and an Effect cannot
  carry that frame across a fiber resume anyway, so keeping it would have been
  a binder that silently did nothing.

## Verification

```
npm run typecheck:workspace     # clean
npm run typecheck:cli           # clean
npm run typecheck:desktop       # clean
npm run typecheck:agent         # clean
npm run typecheck:test-kernel   # 289 errors, see above
node scripts/check-effect-migration-ratchet.mjs   # fails: 5 new catch:effect-importer entries
npx prettier --check <changed files>              # clean (commit hook)
```

`npm test`, `npm run test:changed`, `npm run test:pure`,
`npm run check:dead-code-ratchet` and `npx eslint` were not run: they cannot
pass while the test kernel does not compile.

Part of #12424


LionSR:
Closing as a reference branch (not deleted): the full read+write typing of Secrets/AppState reaches 145 files and does not compile in the kernel; see the finding on #12424. The write-side slice is re-cut as its own PR.

Issue #12427, opened 2026-09-14
Verify shipped recovery contracts and Effect runtime performance


LionSR:
## Current status — 2026-09-15

Checked against main `2c9898445d809b98b038ca9b133b16b57ef7794a` on 2026-09-15. This section supersedes the historical status below; accepted repository policy and later owner rulings remain authoritative.

Parent: #12025. This issue owns verification of consequential runtime/recovery behavior and measured budgets; it is not a whole-suite test migration (#12077/#12528).

## Reconcile the old crash table first

- Follow-up rows and the seeded Queue have landed (#12446); C3 must no longer say the rows do not exist.
- flow.step, request.opened/request.decided and child.turn are shipped vocabulary; compare tests with those records, not proposed model.turn records by name.
- Reflection has a flow.snapshot phase output.pending and rawOutputBytes-based offset/file-length recovery. The proposed byteOffset/digest record is not the shipped representation. Length alone does not establish matching contents: define and verify complete, partial and conflicting output after process reopen before claiming C15.
- Session publication settlement is already native Effect. Existing cancellation, child ownership, retry TestClock and lifecycle-drain suites are substantive coverage; do not demand duplicate tests.

## Remaining evidence

- [ ] Reconcile C1–C15 against current implementation, explicitly recording design differences and existing coverage.
- [ ] Add only consequential missing kill/reopen cases where a crash could repeat paid work or lose committed output, in existing suites.
- [ ] Resolve and verify reflection partial/conflicting-output behavior across process reopen with #11868/#12070.
- [ ] Verify project isolation under contended commits (#12433), child admission/parent stop (#12442), and resource release defects #12525/#12526 through their implementation owners.
- [ ] Establish valid post-conversion measurements: cold open, live idle/replay memory, commit latency under contention, event-loop delay, two-project concurrency, bytes written, and stop latency with controlled non-zero work.
- [ ] Record revision, machine, dataset, units, methods and numeric budgets with the readings.

Measure idle memory while the DB scope is open, replay paths separately, writes rather than retained size, and missed event-loop deadlines. No performance harness/readings were executed in this issue-organization pass. A passing source ratchet or same-session resume test is not process-crash evidence.

<details>
<summary>Historical issue description (superseded status; retained for evidence)</summary>

Under #12025. The runtime cutover is merged, but its two honesty gates are open: the crash windows the design closes by construction, and the delivery plan's performance budgets.

[...]

PR #12502, opened 2026-09-14
docs: keep the pre-cutover runtime baseline as evidence

LionSR:
## Summary

The Effect runtime delivery plan (§8) makes a recorded performance baseline a precondition of the cutover. That baseline was measured on 2026-09-07 at `854b36ee69` (#12068), but the cutover landed before it merged. Its harness writes the pre-cutover event vocabulary: `run.start` with `executionId`, the `status` event, and `transcript.entry`, none of which main accepts now. So it lands as evidence, not as a script under `scripts/` that cannot run:

- The document and harness move to `.agents/docs/evidence/2026-09-07-pre-cutover-runtime-baseline/` (the evidence convention in `.agents/docs/README.md`).
- A note at the top says where it was measured, that the harness does not run unmodified on current main, and that the post-cutover comparison first needs its drafts ported.
- The delivery plan's baseline requirement points at the folder.

Replaces #12068, which is closed with its branch kept.

## Issue acceptance gates

No filed issue. Gate: the pre-cutover numbers the delivery plan requires are recorded where the plan can reach them, with their measured commit and their limits stated.

## Single source of truth

The evidence folder owns the pre-cutover numbers and the harness that produced them; the delivery plan cites it.

None added: docs and one evidence script, moved from #12068.

## Net elements (R6)

n/a (docs and evidence): 3 files, +1052. No production code or export.

## Consumer counts (R8)

n/a: no emit path or export changes.

## Host impact

Extension: none.

Desktop: none.

CLI: none.

## Scheduled refactoring

The post-cutover half of the comparison needs the harness's drafts ported to the current `run.start` and transcript rows. Not filed.

## Validation

- Prettier on the three files.
- Review fixes carried over from #12068: the cross-process scenario sentence now matches what the script enforces (no foreign commits observed fails the run); `process.memoryUsage` is no longer called a first use; the harness imports `node:os` once. Its root path is adjusted for the new folder depth.
- Not run: the harness itself, which targets the pre-cutover schema.


LionSR:
Closing without merging, keeping the branch. Review found that the harness mismeasures the metrics the delivery gate compares, and I confirmed the main ones in the code, so landing these numbers as evidence would give later work a wrong baseline to compare against:

- **Idle memory (§4.2)** is sampled after the database closed: `withDatabase` runs the scenario in `Effect.scoped`, so the connection is released before `process.memoryUsage()` is read.
- **Replay memory (§4.3)** peaks with two full histories held at once: `events` from `readAll` is still referenced when `readInputBatch` decodes a second copy. That inflates the headline 677.82 MB.
- **Bytes written (§4.7)** is the settled database file size, not bytes written: `amplification = settled.dbBytes / written.payloadBytes` ignores the WAL frames that were written and then checkpointed.
- Reviewers also report two findings I did not trace fully: the 10 ms probe does not count ticks `setInterval` coalesces during a stall, so its p95/p99 read low; and the contention probe keeps running through each writer's full `readAll(0)` verification and teardown.

Re-measuring would need the harness fixed at the pre-cutover commit (`4d538c5e40`) on a quiet machine. The cutover has landed, so that is now a reconstruction, not a baseline. The defects are recorded on #12427 as requirements for the post-cutover budget harness.

PR #12509, opened 2026-09-14
fix(session): ownsRun reads fail typed

LionSR:
## Summary

Follow-up to #12485, which merged before its last commit landed on the PR branch. Two gaps remain on main:

- **`ownsRun` still dies on a read failure.** #12494 added `ownsRun` after #12485's lane B was written, and it kept `Effect.orDie`, so a failed database read surfaces as a defect instead of the typed `DatabaseReadFailed` every other session read now carries. It now fails with `DatabaseReadFailed` through `sessionLayer.ts`, `sessionGraph.ts` and `SessionHandle.ts`. Both callers already accept that error: the approval-policy publish, and the host-exit settlement, which logs it.
- **A test that awaited an Effect.** `SessionIsolation.vitest.ts:158` did `await session.settlePublications()`. Since #12485 made `settlePublications` an Effect, that line awaits a non-thenable object and settles nothing, so the assertion after it no longer waited for the publications it was written to wait for. It now runs the Effect.

## Issue acceptance gates

#12025 (typed database errors, lane B of the 2026-09-14 round): the last `orDie` on a session ownership read is typed. Met.

## Single source of truth

`DatabaseReadFailed` in `src/shared/session/database.ts`, as for the other session reads.

None added.

## Net elements (R6)

4 files, +13/−12. No ratchet row changes.

## Consumer counts (R8)

`ownsRun`: 2 callers (approval-policy publish, host-exit settlement), both already handling `DatabaseReadFailed`.

## Host impact

## Scheduled refactoring

None.

## Validation

- `tsc --noEmit` (root, test-kernel, extension) on main 4d33271c8b
- `SessionIsolation`, `SessionHandle`, `sessionEvents`, `AgentRunLifecycle` suites at `--maxWorkers=1`
- Effect migration ratchet OK; eslint on changed files


LionSR:
Rebased onto main. #12506 landed the same isolation-test fix, so this PR now carries only the typed `ownsRun` failure in the session handle, graph, and layer. Root, test-kernel, extension, CLI, and desktop typecheck are clean; the ratchet is OK; the session isolation and default-session lifecycle suites pass.

PR #12511, opened 2026-09-14
refactor(trace): pass launch stage ids explicitly

LionSR:
## Summary

- pass the launch Init stage id explicitly through buildUserVars
- stamp launch-time skill and file diagnostics without relying on StageHandle.run
- remove the last production StageHandle.run use while preserving per-trace isolation

This is a prerequisite for retiring TraceEmitter stage AsyncLocalStorage. It does not claim that the carrier itself is gone; the remaining Promise-based AgentTrace stage API needs a coherent Effect-native conversion.

## Validation

- focused ESLint on changed files
- 38 focused tests after rebase
- additional isolated trace and launch suites passed during implementation
- Effect migration ratchet


LionSR:

PR #12512, opened 2026-09-14
refactor(tools): pass workspace state explicitly

LionSR:
## Summary

- make Claude and Codex configuration readers take their StateStore explicitly
- read tool-call settings from ToolCall roots instead of entering the session ambient scope
- make the shared enum configuration helper depend on an explicit StateStore

This removes six guarded ambient reads and shrinks the surface from 34 files / 52 sites to 31 files / 46 sites.

## Validation

- full TypeScript no-emit check during implementation
- 35 focused tests during implementation
- 18 focused tests after rebase
- focused ESLint
- Effect migration ratchet


LionSR:

PR #12517, opened 2026-09-14
refactor(runtime): pass session roots explicitly

LionSR:
## Summary

- read the child-run concurrency budget from the owning session ConfigProvider
- pass the owning session into the pre-lease-release hook instead of restoring ambient run context around it
- pin run working directories from explicit session roots
- add readValidatedConfig for session-owned callers while preserving existing validation and explicit-setting warning behavior

This removes three ambient context reads and shrinks the guarded surface from 34 files / 52 sites to 32 files / 49 sites. One helper-model run scope remains in runAgent and is deliberately outside this slice.

## Validation

- workspace typecheck
- 4 focused suites, 86 tests
- targeted ESLint
- Effect migration ratchet


LionSR:

PR #12586, opened 2026-09-15
refactor(desktop): the window's IPC handlers take the runtime and the project they serve, not the ambient host

LionSR:
## Summary

#12521 left three desktop main-process handlers reading state off the calling context that the entry already holds. `packages/desktop/src/main/index.ts` is now free, so it hands each of them what it needs as data.

- **`desktopWorkspaceIpc.ts`** takes the window's `ProcessRuntime` from `createProjectWorkspace` and settles every disk operation on the standard library's `FileSystem`. The six `platform().fs` calls go: `realPath` ×4 become `fs.realPath`; `isSymlink` becomes a `readDirectoryTyped` probe of the target's parent (the one primitive that reports a link as itself — `stat` follows it and `FileSystem` has no `lstat`); `readDirectory` becomes `readDirectoryTyped`, whose `File.Type` strings replace the `fsEntryType` bitmask checks. The two `WorkspaceFS` statics, both already on absolute paths, become `fs.readFileString` (with the facade's `normalizeLineEndings` kept explicitly, so the editor's buffers stay LF-only) and `fs.writeFileString`.
- **`desktopDiffHost.ts`** no longer reads `workspaceRoots().storage` for the Review session key.
- **`desktopFileSelection.ts`**'s dropped-path probe is a `NotFound`-only typed stat instead of `await stat(...).catch(() => null)`.

Why the workspace IPC keeps the **process** filesystem rather than a rooted view: every path it hands to disk is the *canonical* path its own containment check produced (`realPath` of the root compared with `realPath` of the target). Under a workspace folder reached through a symlink those canonical paths legitimately sit outside the lexical root, and `RootedFileSystem`'s resolver is lexical — a rooted view would refuse exactly the paths this module has already vouched for. That reasoning is now in the file's header comment, beside the same statement `desktopDiffHost` already carries.

## Issue acceptance gates

- `platform()` is zero in `desktopWorkspaceIpc.ts` (was 6) and the file no longer imports `@platform/platform`.
- `ambient:asyncLocalStorage` is zero in `desktopDiffHost.ts` (was 1); only a type-only `WorkspaceRoots` import remains, which the ratchet does not count.
- The dropped-path probe distinguishes absence from failure: `NotFound` drops the path, anything else fails the drop.
- No `effectRuntime()` anywhere — the runtime travels as an option from the one place that owns it.
- No `runInSession` / `run.inScope` wrap was added to make an ambient read work (#12433).

## Single source of truth

The project a handler serves has one owner: the entry. `createProjectWorkspace` passes the runtime beside the workspace path it already passed; `bindProject` binds `desktopDiffHost.inProject(project.session.roots)` and `requestDiffHost.inProject(project.session.roots)` beside the `previewHost.openBuildDisplayIn(project.session.roots)` that was already there. `agentRunHost` drops its window-scoped `openDiff` (now `Omit<DesktopAgentRunHost, 'openBuildDisplay' | 'openDiff'>`) because a window shows several open projects at once and the window is the wrong scope to name one; each of the three wiring sites names the project-bound host instead. The storage root reaching the renderer as `session` is the same value as before — it is now supplied rather than read.

No new layer. `inProject` mirrors the shape `desktopPreviewHost.openBuildDisplayIn` already established for exactly this problem, so the window has one idiom for "bind a per-window host to one open project", not two. The four refusal sentences in `desktopWorkspaceIpc` collapse into one `WorkspaceRequestRefused` tagged error rather than four types, and the two host promises the handler awaits share one `WorkspaceHostCallFailed` with a `member` field — the shape `NotificationFailed` already uses. `reportRequestFailure` has four callers inside the file.

## Net elements (R6)

+385 / −193 across nine files, of which `desktopWorkspaceIpc.ts` is the bulk: the growth there is the Effect conversion the ratchet requires once a file gains a runtime `effect` import — all eight raw catch sites (seven `catch` clauses and one `.catch(`) become typed failures recovered at one place per request arm — plus the doc comments explaining the filesystem choice. New exported symbols: `entryTypeAt` and `readDirectoryTypedTolerant` in `fsDurability.ts`, both consumed by `desktopWorkspaceIpc.ts` in this PR. Removed: the `isFileNotFoundError`, `platform`, `WorkspaceFS`, `isFile`/`isDirectory`/`isSymlink` and `node:fs/promises` imports, and the `reviewSession()` closure.

## Consumer counts (R8)

- `createDesktopDiffHost` — 2 production callers (`index.ts` 763, 805), both now calling `.inProject(...)` inside `bindProject`; 1 test caller.
- `DesktopWorkspaceIpcOptions.runtime` — 1 production caller (`index.ts` `createProjectWorkspace`), 1 test caller.
- `DesktopFileSelectionOptions.runtime` — 1 production caller (`index.ts` `bindProject`), 2 test callers.
- `readDirectoryTypedTolerant` and `entryTypeAt` are new and single-consumer here; `normalizeLineEndings` gains one. No symbol is left with zero consumers (`check:dead-code-ratchet` is green).

## Host impact

Desktop only. Nothing under `src/` production code changed; the VS Code extension and the CLI are untouched. The three suites that own these modules (`DesktopWorkspaceIpc`, `DesktopDiffHost`, `ElectronFileSelection`) were updated in place — no new test file, and the only new assertions are the one existing case whose reported error shape changed plus a `session` field check on the diff message that pins the newly-supplied Review key. `DesktopPreviewHost.vitest.ts` took a one-line mechanical update for the new required option.

Behaviour changes, all loud by construction:

1. A dropped path whose stat fails for any reason other than `NotFound` now fails the drop instead of being silently discarded — the defect this PR was asked to fix.
2. A project-tree entry that vanishes between the directory read and its own `lstat` now reports a listing error instead of being listed with the type `readdir` already had. `readDirectoryTyped` is the sanctioned primitive; the alternative was a fallback that hides the race.
3. A `getEnvironmentSummary` or `ptyHost.create` rejection reaches the window's async-error reporter as the handler's own tagged failure carrying the rejection as `cause`. Every sentence the renderer shows is byte-identical, because the tagged error's `message` is exactly the rejection's own text.

## Validation

All on the final HEAD (`48ffba518c`), each exit read from its own log:

| gate | exit |
| --- | --- |
| `npm run typecheck` | 0 |
| `npx vitest run src/test-kernel/desktop/` | 0 — 87 suites / 300 tests pass |
| `npx vitest run --changed origin/main --maxWorkers=2` | 0 |
| `npm run lint` (full) | 0 |
| `npm run format` | 0 — all unchanged |
| `node scripts/check-effect-migration-ratchet.mjs` | 0 — no count grew, no headroom |
| `npm run check:dead-code-ratchet` | 0 — no new findings |
| `npm run test:pure -- --maxWorkers=2` | 0 — 174 files / 2123 tests |

`config/ratchets/effect-migration-baseline.json` was regenerated with `node scripts/check-effect-migration-ratchet.mjs --update`, never hand-merged: it loses the `platform()` entry for `desktopWorkspaceIpc.ts` and the `ambient:asyncLocalStorage` entry for `desktopDiffHost.ts`, and gains nothing — `catch:effect-importer` stays at 4 files / 5 sites even though two files gained a runtime `effect` import, because both ended at zero catch sites.

## Review follow-ups (`c379b4a8ed`)

Two of the six `platform()` sites this lane retired had converted the facade's
behaviour, not just its plumbing. Both are fixed on top, plus one listing that
was stricter than the code it replaced:

- **The read keeps a BOM.** `readFileString` decodes through a `TextDecoder`,
  which strips a UTF-8 byte-order mark that the old `WorkspaceFS.readFileUtf8`
  preserved. The editor saves its buffer verbatim, so the first save of a
  BOM-prefixed file silently deleted the mark. The read decodes the bytes
  itself again, and a new suite case pins the mark surviving the round trip.
- **The symlink guard probes the target, not the parent.** It listed the
  parent directory and matched the entry name by exact string, which skips on a
  case-insensitive volume and whenever any sibling's `lstat` failed inside the
  typed listing — a dangling link then wrote outside the workspace. A new
  `entryTypeAt` (one `lstat` on the target path itself) replaces the listing, so
  the guard resolves case-insensitively and any failure but `NotFound` refuses
  the write instead of reading as "no link here".
- **The project tree tolerates one bad entry.** It used the strict typed
  listing, so one entry that could not be `lstat`ed — permissions, or vanishing
  between `readdir` and `lstat` while TeX is writing — failed the whole
  directory where main listed it. The tree now uses
  `readDirectoryTypedTolerant`, which drops the unreadable entry with a warning
  naming it; the strict form stays with the containment and deletion walkers.

Both new helpers live in `fsDurability.ts` beside `readDirectoryTyped` and have
a production consumer in `desktopWorkspaceIpc.ts`.


LionSR:
Both inline findings are real and both are fixed in `48ffba518c`, which is on top of the commit you reviewed.

**`readDirectoryTypedTolerant` changed the mechanism, not only the error handling — accepted, and the mechanism now matches the base.** You are right that the listing read names and `lstat`ed every entry, so a directory that is readable but not searchable (`r--`) would have rendered empty where `platform().fs.readDirectory` listed it — a regression I introduced, not a property of the primitive. The function now takes the listing from `readdir(..., { withFileTypes: true })`, derives each type from the dirent, and probes with `lstat` only for an entry the platform left `Unknown`; that probe is also the only thing whose failure now costs a row. So the common path needs no per-entry syscall and the tolerant property is narrower and more accurate than before: it drops entries whose type genuinely cannot be read, not entries that merely cannot be stat'ed. `entryTypeOf` takes the shared predicate set, since `Stats` and `Dirent` carry the same one. The Node call is justified in the doc comment, the same way `readDirectoryTyped`'s `lstat` is.

**Mislabeled `method` — accepted.** The per-entry probe tagged its `PlatformError` with `'readDirectoryTyped'`; it is `'readDirectoryTypedTolerant'` now, as is the readdir failure itself.

The readdir call could not go through the `FileSystem` service: `readDirectory` returns names alone, and the dirent answer is what carries the types.

**Prose corrected.** The body now names the two new exports (`entryTypeAt`, `readDirectoryTypedTolerant`) instead of claiming zero and pointing at `readDirectoryTyped`, and the validation table is restated for `48ffba518c`.

PR #12610, opened 2026-09-15
refactor: burn down twelve test-only exports and re-aim their suites (Phase A of #12564)

### LINKED ISSUE #12564: Burn down test-only exports: knip baseline shrink (17 entries) plus the knip-invisible same-file-used export class
Survey 2026-09-15. Two related batches enforcing the "exports are contracts; default to file-local" rule where the dead-code ratchet cannot see the problem.

## Phase A — knip baseline shrink (ratchet already sanctions this; baseline is shrink-only)

The ratchet's own output names 3 entries no longer found: `buildKpathseaSearchPath`/`buildLatexInputEnv`/`buildLatexSearchParts` (`config/ratchets/knip-baseline.json:1012-1029`, `src/latex/texTools.ts`). −18 JSON lines, zero risk.

Test-only exports verified against current main (production-dead or baselined):
- `src/tools/` (13, grouped in one pass): `DiagnosticsInputSchema` (DiagnosticsTool.ts:98), `runStreamedTurn` ×2 (claudeAgent.ts:192, codex.ts:251), `publishCodexTodos` (codex.ts:145), `computeAndWriteWorkflowDiffs` (delegation/subagentResults.ts:337), `PRPollingSource`/`PRCurrentShaState`/`PRSubscriptionState` (github/githubSubscriptionTool.ts), `parseOriginHeadDefaultBranch` (:417), `resolveWorkspaceRoot` (lean/direct/leanServerPool.ts:600), `countPinnedMemories` (memory/memoryFileSystem.ts:446), `isTransientHttpError` (timeouts.ts:52), `wolframApprovalCommand` (wolfram/WolframTool.ts:38).
- `src/controllers/` + `src/eventBus/` (11): `DEFAULT_PLAN_NAMES`, `parseChatGptUsage`/`CHATGPT_USAGE_URL`, `parseKimiCodeUsage`/`KIMI_CODE_USAGE_URL`, `parseGlmCodingPlanUsage`, `selectDesktopSetupModel`, `selectSetupCredentialModelExcludingOpenRouter`, `AppSignalPayloads`, plus `deriveOnboardingFunnelState`/`OnboardingFunnelInputs`/`OnboardingFunnelTransition`/`planOnboardingFunnelTransition` (onboardingFunnel.ts — hosts import only `OnboardingFunnelRefresher`).

Policy question per PR: some suites pin real behavior through these seams (`parse*` pure parsers). Per "tests are a budget", prefer deleting or re-aiming at the public seam where cheap (e.g. CriticismParser via `DiagnosticsTool.definition`); keep-and-baseline only where the test genuinely earns the seam.

## Phase B — the knip blind spot: exported types used only in-file or by their own vitest

Knip counts same-file usage (constructor-param/state-member), so these never appear in findings or the baseline. Each has zero production importers; verified by grep:

`src/agent/runtime/run/` + `ModelInvoker.ts` (9): `AgentRunLayerInput`, `MediaCapabilities`, `MediaInputParts`, `CompactionInput`, `BindModelInput`, `ToolDefinitions`, `RouteEndpoint`, `InvocationOutcome`, `InvokeError`, `JSONSchemaObject`.
`src/controllers/` (7): `SettingsModelSelectionControllerDeps`, `LatexToolingControllerDeps`, `SettingsAgentCatalogEntry`, `ProgressApiKeyRetryControllerDeps`, `ProgressWorkflowFileActionsControllerDeps`, `SubscriptionUsageCredentials`, `LatexPathTool`.

Delta: −26 export keywords, −17 baseline entries, ~0 production LoC; tests switch to `ConstructorParameters<typeof X>[0]` / structural types. Risk: trivial-mechanical, test-only touch.

Also fold in here: the duplicated `InvokeError` alias locally re-declared at `src/agent/runtime/loop/toolUseDispatch.ts:89` should instead import the (then file-local-ized) single definition from ModelInvoker.

LionSR:
## Summary

Phase A of #12564: burn the test-only exports down. Twelve `export`
keywords are deleted from production source and each suite that needed them
now drives a public seam instead. The knip baseline loses the ten entries
that named them.

The policy question the issue asked (re-aim the test vs. keep the export) is
answered per symbol; twenty more entries stay exactly as they are, with the
reason recorded per name in "Consumer counts (R8)" below. Review found one
name the sweep had missed entirely — `DiagnosticsInput`, a same-file-used type
that belongs to Phase B's invisible class rather than this one — and it is
recorded there too, so the table accounts for every test-only export in the
files this PR touches.

**Re-aimed at the public seam (12 de-exports):**

| symbol | the seam the suite now drives |
|---|---|
| `DiagnosticsInputSchema` | `getDefaultToolRegistry().get('diagnostics')!.definition.zodSchema` |
| `wolframApprovalCommand` | the literal command on the approval request the tool opens |
| `publishCodexTodos` | a `todo_list` item through `runStreamedTurn`'s event loop |
| `DEFAULT_PLAN_NAMES` / `CHATGPT_USAGE_URL` / `KIMI_CODE_USAGE_URL` | the literals themselves |
| `AppSignalPayloads` | a payload map derived from `appSignals.on` |
| `deriveOnboardingFunnelState` | `planOnboardingFunnelTransition(undefined, inputs).state` |
| `OnboardingFunnelInputs` / `OnboardingFunnelTransition` | `Parameters`/`ReturnType<typeof planOnboardingFunnelTransition>` |
| `isTransientHttpError` | `retryTransientFetch`'s retry decision |
| `selectDesktopSetupModel` | `buildDesktopSetupRunRequest` + the returned request's model |

`selectDesktopSetupModel` was also a one-caller one-line delegate, so it is
not merely de-exported: the delegate is gone and `buildDesktopSetupRunRequest`
now states the projection at its call site.

Two of these are strictly stronger than what they replace:
`wolframApprovalCommand('1+1')` asserted only that the formatter agreed with
itself (a tautology); the literal `'wolframscript -code "1+1"'` pins the
command the user is asked to approve. `publishCodexTodos` was called
directly with a hand-built `TodoItem[]`; going through the item event also
covers the `TodoListItem` -> `TodoItem` mapping.

**Kept exported (20 baseline entries untouched), with the reason per name:**
see the R8 table. In short: the seam is reachable but re-aiming is not cheap
(real-FS fixtures, git fixtures, a full `AgentFlowResult`, an injected SDK
`Thread`), or the guarantee under test is *not* observable through the public
entry point (`selectSetupCredentialModelExcludingOpenRouter`'s "skips
OpenRouter" arm; `parseChatGptUsage`'s `fetchedAt` parameter, which the
adapter never passes).

### Review follow-up

Codex flagged two of the re-aims as weaker than the assertions they replaced,
and both were right — each pinned a value the production path produces whether
or not the rest of the wiring works:

- **`TransientHttpError.vitest.ts`.** `onFailedAttempt` runs from
  `Effect.tapError` gated on the classifier, so it fires on the initial
  failure whether or not a retry follows; `retried === true` proved
  classification but not that the request runs again. The helper now counts
  executions inside the request effect and asserts two attempts for a
  transient error against one for a permanent one.
- **`SetupLaunch.vitest.ts`.** The helper re-derived the projection the deleted
  `selectDesktopSetupModel` performed, so a wrong fallback flag or a mis-mapped
  resolution in `buildDesktopSetupRunRequest` would have left every case
  passing. It now drives `buildDesktopSetupRunRequest` — the entry point
  `packages/desktop` calls — and reads the model back off the validated
  request.

## Issue acceptance gates

- Phase A of #12564 — complete except for the intentionally kept names listed
  under R8.
- The issue's first bullet (three `texTools` entries) was already resolved on
  `main`; this branch does not touch it.
- Phase B (knip-invisible same-file-used exports) shipped separately in
  #12605 (`a12acf03ed`), which is why the baseline lands at 183 rather than
  184: this branch removes the ten entries it resolves, and #12605 had already
  removed `LatexPathTool`.
- `npm run check:dead-code-ratchet` -> `Dead-code ratchet OK: no new findings.`
  (183 combined findings vs 183 baselined, 10 entries removed by this branch,
  none added).

## Single source of truth

No new owner is introduced. Each de-exported symbol keeps its existing single
owner and simply stops being part of the module's public surface: the tool's
registered `definition` owns the input schema, `planOnboardingFunnelTransition`
owns the funnel transition, `retryTransientFetch` owns the retry decision, and
`buildDesktopSetupRunRequest` owns the setup-launch resolution.

No abstraction is added. Two *derivations* replace two *re-exports*:
`AppSignalPayloads` in the bus test is now a mapped type over
`typeof appSignals.on`, and the funnel test's input/transition types are
`Parameters`/`ReturnType` of the planner. Both follow the repo rule that a
type is derived from its owner rather than restated next to it.

## Net elements (R6)

- Files: **19** (1 ratchet JSON, 10 production source, 8 vitest suites), from
  `git diff --shortstat origin/main` -> 19 files changed, 202 insertions(+),
  197 deletions(-).
- `^[+-]export` over the diff on `.ts` files: **+0 / -12**.
- Net LoC: **+5**. The production side is net-negative (12 one-line `export`
  removals plus one deleted function); the added lines are test churn, the
  price of asserting literals and derived types instead of importing them.
  This is not a LoC-reduction PR and does not claim to be: it removes public
  surface, which is the element metric here.
- Baseline entries: **193 on `origin/main` -> 183 here (-10)**; 194 before
  #12605 landed.
- Elements that cease to exist: 12 export keywords, 11 knip baseline entries
  (10 here plus `LatexPathTool` via #12605), 1 production function
  (`selectDesktopSetupModel`). No file, branch, or layer is deleted.

## Consumer counts (R8)

Consumers at `origin/main` = non-test importers (word-bounded grep across
`src/`, `packages/*/src/`, `scripts/`, `config/`). "replaced by" is what the
suite uses now.

| name | consumers at `origin/main` | what replaced it |
|---|---|---|
| `DiagnosticsInputSchema` | 0 prod (1 vitest) | tool's registered `definition.zodSchema` |
| `wolframApprovalCommand` | 0 prod (1 vitest) | literal command string in the assertion |
| `publishCodexTodos` | 0 prod (1 vitest) | `runStreamedTurn` with a `todo_list` item |
| `DEFAULT_PLAN_NAMES` | 0 prod (1 vitest) | literal `'ChatGPT Coding Plan'` |
| `CHATGPT_USAGE_URL` | 0 prod (1 vitest) | literal URL in the `it.each` table |
| `KIMI_CODE_USAGE_URL` | 0 prod (1 vitest) | literal URL in the `it.each` table |
| `AppSignalPayloads` | 0 prod (1 vitest) | mapped type over `typeof appSignals.on` |
| `deriveOnboardingFunnelState` | 0 prod (1 vitest) | `planOnboardingFunnelTransition(undefined, inputs).state` |
| `OnboardingFunnelInputs` | 0 prod (1 vitest) | `Parameters<typeof planOnboardingFunnelTransition>[1]` |
| `OnboardingFunnelTransition` | 0 prod (1 vitest) | `ReturnType<typeof planOnboardingFunnelTransition>` |
| `isTransientHttpError` | 0 prod (1 vitest) | `retryTransientFetch` retry decision |
| `selectDesktopSetupModel` | 0 prod (1 vitest) | `buildDesktopSetupRunRequest` request model |
| `planOnboardingFunnelTransition` | **1 prod** (`packages/cli/src/onboarding/runOnboarding.tsx`) | n/a - stays exported; not a test-only export |
| `runStreamedTurn` (claudeAgent) | 0 prod (3 vitest) | **kept**: the SDK-mocked turn is the narrowest injectable seam; `ClaudeAgentResumeFallback` pins abort-link cleanup that the tool's result cannot surface, and the tool path would have to re-mock child-run/session plumbing wholesale |
| `runStreamedTurn` (codex) | 0 prod (1 vitest) | **kept**: the `Thread` is built inside the tool's `execute`; there is no injectable seam above this function |
| `computeAndWriteWorkflowDiffs` | 0 prod (1 vitest) | **kept**: re-aiming through `buildSubagentResult` needs a full workflow `AgentFlowResult` fixture for one case |
| `PRPollingSource` | 0 prod (1 vitest + support) | **kept**: the suites construct fresh instances to drive protected hooks; the production singleton cannot be used per-test without cross-contamination |
| `PRCurrentShaState` | 0 prod (1 support) | **kept**: state shape of the class above; `getSubscriptionState` is protected, so no public derivation exists |
| `PRSubscriptionState` | 0 prod (3 vitest + support) | **kept**: as above - the support builders are the compile-time guard against a stale test double |
| `parseOriginHeadDefaultBranch` | 0 prod (1 vitest) | **kept**: re-aiming needs a real git fixture with `origin/HEAD` set plus gh mocks |
| `resolveWorkspaceRoot` | 0 prod (1 vitest) | **kept**: the suite already tests it against real scratch directories; the public path starts an LSP server |
| `countPinnedMemories` | 0 prod (1 vitest) | **kept**: the case pins the bounded-walk early exit at an arbitrary `limit`; `setMemoryPinned` only ever passes `MAX_PINNED_MEMORIES` |
| `parseChatGptUsage` | 0 prod (1 vitest) | **kept**: the relative-reset cases pass `fetchedAt` explicitly and the adapter never passes it, so the service seam cannot reach them without faking the global clock inside an Effect test |
| `parseKimiCodeUsage` | 0 prod (1 vitest) | **kept**: same suite; re-aiming adds http/credential plumbing per wire-format edge case without adding coverage |
| `parseGlmCodingPlanUsage` | 0 prod (1 vitest) | **kept**: as above |
| `selectSetupCredentialModelExcludingOpenRouter` | 0 prod (1 vitest) | **kept**: the "skips openRouter" and "no provider key probed" assertions are not observable through `buildDesktopSetupRunRequest`, which probes the OpenRouter key first by design |
| `DiagnosticsInput` | 0 prod (2 vitest) | **kept**: same-file-used, so knip never listed it in either run and it appears in the baseline nowhere — a Phase B class straggler found in review, not a Phase A candidate. No public type-level seam exists to replace it: the registered `definition.zodSchema` is the definition's generic schema, not the tool's own input type, so `z.infer` cannot recover `DiagnosticsInput`, and both suites need the concrete union for their `Extract<..., { command: 'add' }>` projections |

## Host impact

Extension: none.

Desktop: none. `buildDesktopSetupRunRequest` keeps its contract; the
module-local helper it used to delegate to is inlined into it, and its test now
asserts through the builder rather than around it.

CLI: none. `packages/cli/src/onboarding/runOnboarding.tsx` imports
`planOnboardingFunnelTransition`, which stays exported, and its parameter and
return types stay nameable through the function's own signature.

## Scheduled refactoring

None. The kept names are a deliberate resting state, not deferred work: each
one is the narrowest seam that can carry its suite's assertion.

## Validation

The branch was rebased onto `origin/main` after #12605 merged; the only
conflicting file was `config/ratchets/knip-baseline.json`, where the two
branches removed adjacent entries (`LatexPathTool` there, `AppSignalPayloads`
here) and the resolution keeps both removals. Every gate below was re-run on
the rebased head.

- `npm run typecheck` -> exit 0.
- `npm run lint` -> exit 0.
- `npm run check:dead-code-ratchet` ->
  `knip dead-code findings: 12 normal, 183 production; 183 combined (unused=12, productionDead=171) vs 183 baselined`
  / `Dead-code ratchet OK: no new findings.`
- Suites reaching a changed file, run directly: 8 files, all passing -
  `TransientHttpError`, `SetupLaunch`, `OnboardingFunnel`,
  `GitHubSubscriptionProgressEvents`, `CodexProgressEvents`,
  `SubscriptionUsageService`, `CriticismParser`, `Wolfram`.
- `npx prettier --check --ignore-path /dev/null <changed files>` -> clean; the
  baseline JSON was re-dumped through `json.dump(..., ensure_ascii=False, indent=2)`
  so the diff is content-only.

Closes #12564


LionSR:
Addressed the `DiagnosticsInput` note by editing the body, no code change: it now appears in the R8 table as a kept name, with the reason the review gave (the registered `definition.zodSchema` is the definition`s generic schema, not the tool`s own input type, so `z.infer` cannot recover the union both suites project with `Extract<..., { command: "add" }>`).

It is worth naming what the finding exposed: `DiagnosticsInput` is exported, used in-file, and imported only by two vitest suites — so knip lists it in neither run and it was never in the baseline. That is Phase B`s invisible class, not Phase A`s, which is why the Phase A sweep never saw it. #12605 closed that class for `src/agent/runtime/run/` and `src/controllers/`; this is a straggler in `src/tools/`. The table now accounts for every test-only export in the files this PR touches.

PR #12693, opened 2026-09-17
refactor(desktop): dispatch host requests as one Effect program

LionSR:
The desktop's `host.request` dispatcher was a Promise-native switch whose arms each re-entered the Effect runtime (25 `runtime.runPromise` / 3 `runPromiseExit` sites in the file), and `handle` then lifted the whole async function back with `Effect.tryPromise` and ran it a second time through `runPromiseExit`. This makes `dispatch` one `Effect.gen` returning `Effect.Effect<HostOutcome, unknown, ProcessServices>`, so every arm that already had an Effect is a `yield*`, `handle` is a single `runtime.runPromiseExit(dispatch(request, port))`, and the round trip disappears. The eight in-file helpers the switch calls (`runWorkflowDiff`, `reportFileOperationResult`, `runWorkflowFileOperation`, `exportTranscript`, `latexdiffAgainstCommit`, `latexdiffs`, `agentConfigBanner`, `onboarding`) convert with it; their `throw new Rejected(...)` guards become `Effect.fail`, so refusals arrive as typed failures rather than as a rejected promise re-wrapped at the boundary. The error fold in `handle` is byte-for-byte unchanged — the `Cancelled` rethrow, `primaryAgentError` classification, the `presentAgentFailure` dialog, and the `Rejected` `docsCommand` preservation (#11959) all still read the squashed `Cause`, which (verified by probe) carries the same raw error value `runPromise` used to reject with.

## Net elements (R6)

- Files: 1 (`packages/desktop/src/main/desktopHostRequests.ts`).
- `^[+-]export` lines in the diff: **0** — no public surface moves.
- Symbols that cease to exist: `rejectRequest` (the async-throw helper that existed only to fit the Promise shape) and `requireOpenRun` (single-caller guard, inlined into `exportTranscript`). One symbol added: `fromHost`, the single verbatim promise lift, 23 call sites.
- Runtime re-entries deleted: **21** (`runtime.runPromise` 25 -> 5; `runPromiseExit` 3 -> 2, one of which became `Effect.exit` inside the program). The dispatcher now enters the runtime exactly once per request instead of once per arm plus once for the whole switch.
- `Effect.tryPromise` sites: 4 -> 3 (the `handle` lift is gone; `rejectRequestEffect` is now a plain `Effect.fail`).
- Net LoC from `git diff --stat origin/main`: **+51** (409 insertions, 358 deletions). The change is element-negative, not line-negative: `Effect.gen` + `return yield* Effect.fail(...)` is more verbose per guard than `throw`, and Prettier re-wraps the deeper indentation of every arm. The point is the one owner of the runtime boundary.

## Consumer counts (R8)

- `rejectRequest` — 2 consumers before (`rejectRequestEffect`'s `try` thunk; `showError` on the `ProgressWorkflowFileActionsController` host). 0 after: `rejectRequestEffect` is `Effect.fail(new Rejected({ reason }))` directly, and the controller's `showError` (typed `(message: string) => Promise<void>`, not ours to flip) keeps an inline async throw at its one call site.
- `requireOpenRun` — 1 consumer before (`exportTranscript`). 0 after: the `SubscriptionRef.getUnsafe(session.view).runs.has(runId)` guard is inlined and fails with the same `Unavailable`.
- `dispatch` — 1 consumer (`handle`), unchanged.
- Repo-wide grep for both names outside the file: 0 hits, including string-form references (`vi.mock` paths, dynamic imports).

## Verified

- `npm run typecheck` — PASS (all seven projects, including `typecheck:desktop`'s five tsconfigs).
- `npx vitest run src/test-kernel/desktop/DesktopPreviewHost.vitest.ts src/test-kernel/desktop/DesktopProgressFileActions.vitest.ts` — PASS, 2 files / 23 tests. These are the only suites that reach `createDesktopHostRequests`; `DesktopPreviewHost` drives six failing request kinds (`openFile`, `apiKeyBanner`, `extractFigures`, `latexdiffs`, `exportTranscript`, `polish`) through the fold and asserts the dialog presents exactly once, plus `pickFiles` rejecting with `_tag: 'Cancelled'` and presenting nothing. No test double needed changing.
- `npx eslint packages/desktop/src/main/desktopHostRequests.ts` — PASS, no output.
- `npm run check:effect-migration-ratchet` — PASS. No row grew; `Effect.run*` stays 0/0 (runs under `packages/desktop/src/**` are R1's boundary kind, not debt), `catch:effect-importer` unchanged at 2/2. No baseline was edited.
- `npm run check:dead-code-ratchet` — PASS, "no new findings" (180 vs 180 baselined).
- `npx prettier --ignore-path /dev/null --check packages/desktop/src/main/desktopHostRequests.ts` — PASS.
- Error-channel type probe: with the annotation removed, tsc reports the inferred channel as `Effect<HostOutcome, unknown, AgentResume | AppState | FileSystem | HttpClient | LanguageModel | Path | Secrets>`. `unknown` is exact, not a widening — it is the same channel `handle` already folded through `Effect.tryPromise({ catch: (error) => error })`, and it is `unknown` only because the promise-faced capabilities below are lifted verbatim. The requirement is annotated as `ProcessServices`, which the window's `ProcessRuntime` provides.
- Rejection-shape probe: `Effect.runPromise(Effect.fail(new Boom()))` rejects with the raw tagged error (`instanceof` holds), so the old per-arm `runPromise` handed `handle` the same value the new `Cause.squash` does. The `Rejected`/`Unavailable` `instanceof` checks and the `docsCommand` branch behave identically.
- No `catch {}`, no silent fallback, no new `runFork`: the only fork in the file (the merge run's lifecycle) is untouched and still logs and presents its failure.

## Left out

- **The `SessionBridgeOptions.handleHostRequest` port is not flipped.** `handle` stays `Promise<HostOutcome>` and keeps the single `runPromiseExit`; flipping the port is a separate lane that also touches the extension twin and `SessionBridge`, neither of which is in this diff.
- **Capabilities that still answer with a promise** are lifted once each through `fromHost` and named here so the next lane can find them: `DesktopAgentRunHost.openPath` / `openBuildDisplay` (`BuildDisplayFn`), `DesktopFileSelection.pickFiles` / `attachDroppedFiles`, `DesktopAgentRun.runValidated`, `DesktopProgressFileActions.findAndOpenLabel` / `runMergeFile` / `diffStreamToolbarAction`, `ProgressWorkflowFileActionsController.openTaskStorage` / `compareOriginal` / `acceptFile` / `handle`, and the options `signIn`, `recheckTools`, `openExternalUrl`, `getCustomAgentDirectory`, `onboarding.*`.
- **The five surviving `runtime.runPromise` sites are all Promise-face options, not dispatch arms**: `loadModelOptions` and `showInfo` / `sendFollowUp` on the ProgressWorkflowFileActionsController host (corrected from HostRunActionPorts), `listWorkspaceCandidateFiles` on `DesktopProgressFileActions`, and `getLatexdiffRunContext` (currently unreachable: both `runLatexdiffFile` callers pass two arguments) (reached from the controller's promise-faced `latexdiffFile` callback as well as from a request). `getChatExportController` keeps its `runPromiseExit` memo because `exportRunTranscript`'s `getController` port is a promise. Each is a port signature, not a round trip this lane owns.
- `runLatexdiffFile` stays Promise-native: it is shared between a request arm and the controller's promise-faced `latexdiffFile` callback, so converting it would only move the lift, not delete it.


LionSR:
Thread at desktopHostRequests.ts:641: reworded to "no dispatch arm re-enters the runtime" and noted that a lifted capability may still run its own program behind its promise face. The two "Left out" labels are corrected in the body: `showInfo` and `sendFollowUp` belong to the ProgressWorkflowFileActionsController host, and `getLatexdiffRunContext` is currently unreachable because both `runLatexdiffFile` callers pass two arguments.

Issue #12072, opened 2026-09-07
Tracking: remaining async-mutex ownership and dependency retirement


LionSR:
## Current status — 2026-09-15

Checked against main `2c9898445d809b98b038ca9b133b16b57ef7794a` on 2026-09-15. Current status below supersedes old counts, branch reservations, and implementation-stage descriptions.

- [x] p-queue, p-defer, p-map, p-retry, p-timeout and delay have no remaining surveyed production import sites.
- [x] Resolve the final async-mutex disposition: KeyedMutex still serves AgentRosterController and the desktop editor. Verify which callers should move to owned Effect concurrency and which genuinely require a browser/foreign boundary before deleting or retaining the package.
- [x] Record any accepted terminal exception explicitly under #12073; this organization pass does not create that ruling. Recorded by #12700: #12073's Q1 item is now checked off, and the `new AbortController()` / `catch:effect-importer` ratchet rows carry per-file reasons as code comments in `scripts/check-effect-migration-ratchet.mjs`.

The current baseline has one async-mutex import. Do not replace it with a handwritten Promise queue just to remove a package. Check manifests, lockfile, and non-surveyed callers before final package removal. Old package counts below are historical.

<details>
<summary>Historical issue description (superseded status; retained for evidence)</summary>

Lane tracker under #12025 for the seven superseded-package rows of `config/ratchets/effect-migration-baseline.json` — package E's F5 (retry/timeout), F6 (concurrency/queues) and R8 (one clock). Thirty sites across 30 files, measured at `1cf1a84`. It exists because these rows had no owner, and the four that moved recently moved as side effects of other lanes rather than because anyone was working the row.

## The rows, with every remaining site

**`import:p-queue` — 13 files / 13 sites**

[...]

PR #12696, opened 2026-09-17
consolidate: one per-key lane, retire KeyedMutex and async-mutex

LionSR:
Three named mechanisms carried one concept — "serialize operations on one key,
drop the entry when the key goes idle". This collapses them onto the one that
already existed, `withPerKeyLane` (`@utils/core/perKeyQueue`), and retires the
`async-mutex` dependency with its ratchet row. Refs #12072.

- `src/tools/lean/direct/lakeCommands.ts`: `serializeOnWorkspace`, a hand-rolled
  claim/refcount/permit/evict lifecycle over `Semaphore`, becomes one
  `.pipe(withPerKeyLane(workspaceLanes, path.resolve(root)))`. `path.resolve`
  stays inside a `suspend`, so a relative root still resolves at run time.
- `src/agent/runtime/runApprovalQueue.ts`: `laneOf`, a second `Semaphore`
  hand-roll with its own map, becomes `withPerKeyLane(lanes, runId)`. The
  `Effect.suspend` stays inside the lane, so the run's bypass is still read at
  dispatch rather than at enqueue.
- `src/utils/core/keyedMutex.ts`: deleted with its only production consumer's
  need for it. `async-mutex` is gone from both `package.json` files and from
  `pnpm-lock.yaml`.

## Net elements (R6)

| | |
|---|---|
| Files changed | 10 (1 deleted) |
| `^-export` / `^+export` | 1 / 0 (`export class KeyedMutex`) |
| Classes deleted | `KeyedMutex` |
| Functions deleted | `serializeOnWorkspace`, `laneOf` |
| Types deleted | the inline `{ semaphore: Semaphore.Semaphore; users: number }` lock record in `lakeCommands` |
| Modules deleted | `src/utils/core/keyedMutex.ts` |
| Dependency removed | `async-mutex@^0.5.0` from `package.json` and `packages/agent/package.json` (+ lockfile) |
| Ratchet row deleted | `import:async-mutex` in `config/ratchets/effect-migration-baseline.json` (deleted, not zeroed — an empty row is a hard failure in `scripts/check-effect-migration-ratchet.mjs`). `async-mutex` stays in that script's `SUPERSEDED_PACKAGES`, so a new import fails as a new file. |
| Suites deleted | the `KeyedMutex` describe block in `UtilsCore.vitest.ts` (3 tests); its `deferred()` helper is still used by the `coalesceAsync` tests and stays |
| Net LoC | **-100** (`+60 / -160`, `git diff --stat origin/main`) |

## Consumer counts (R8)

| Symbol | Before | After |
|---|---|---|
| `KeyedMutex` | 1 production consumer (`editorPane.ts`), 1 suite, 1 module | 0 — symbol and module gone |
| `serializeOnWorkspace` | 1 caller (`runLakeCommand`) | 0 — function gone |
| `laneOf` | 1 caller (`enqueue`) | 0 — function gone |
| `withPerKeyLane` | 23 call sites in 16 files | 25 call sites in 18 files |
| `async-mutex` imports | 1 | 0 |

## Behaviour

Not behaviour-preserving in one place, deliberately:

- **The approval queue gains FIFO ordering per key and loses its per-session
  lane leak.** A `Semaphore` release barges: it wakes waiters in a scheduled
  task, so a prompt enqueued in between could take the free permit ahead of
  prompts already waiting. `withPerKeyLane` hands off through a `Deferred`, so
  prompts now dispatch in enqueue order. And the old map held one semaphore per
  run for the session's life; the lane now leaves the map once a run's last
  prompt settles.
- **Lean lake commands keep their per-workspace serialization**, including the
  interruption contract: the child is still owned across interruption, and the
  workspace lane is handed to the next waiter only after the process is gone.
  Lake also gains FIFO, which for a serialized-by-construction path is a
  tightening, not a change of outcome.
- The desktop editor pane's per-path model syncs keep the mutex's semantics
  exactly (see below).

## Left out

**The desktop renderer does not take `withPerKeyLane`.** `editorPane.ts` is the
only production consumer `KeyedMutex` had, and it is a Lit renderer that runs no
Effect programs. `withPerKeyLane` is an Effect combinator, so using it means
importing `effect` into `editorPane.ts` — which puts the file under the
`catch:effect-importer` ratchet row, where its six existing promise boundaries
(`loadDirectory`, `ensureEditor`, `open`, two in `refreshOnce`, `save`) would all
have to become typed recovery in this commit. That is a desktop-renderer Effect
migration, not a lane change: each of those six sites reports to
`callbacks.onError`, `Effect.runPromise` rejects with a `FiberFailure` rather
than the original error, and at least one site (`buildEditorDirectoryEntries`
inside `loadDirectory`'s `try`) changes which throws are caught if restructured
carelessly. Per the instruction's fallback, the renderer keeps a minimal local
serializer **without** `async-mutex`: a ten-line in-file hand-off
(`onModelSyncLane`) holding the same semantics as the mutex it replaces —

- the body still starts a microtask after the call (`(tail ?? Promise.resolve()).then(sync)`),
  not synchronously, so the state re-checks read the same values they did;
- the tail is `Promise.allSettled([result])`, so a failed sync neither poisons
  the path's lane nor is swallowed — the failure still reaches the caller
  through the returned promise, and there is no bare `catch {}`;
- the path's entry leaves the map once its last entrant settles.

`withPerKeyLane` was **not** added to `BROWSER_SAFE_UTILS`. The renderer can join
the survivor once the desktop renderer has an Effect boundary of its own; that is
a separate lane.

Also left out: no suite was added for `withPerKeyLane` itself (test budget — the
lane is exercised by the lean serialization tests and the approval suites), and
`knip-baseline.json` was not shrunk.

## Verified

- `npm run typecheck` — pass
- `npm run test:pure` (171 files, 2067 tests, includes the architecture ratchets) — pass
- `npx vitest run` on the reachable suites — pass (8 files, 102 tests):
  `UtilsCore`, `tools/lean/LeanTools` (its `serializes calls against the same
  workspace` / `runs calls in parallel` pair is the real per-workspace proof),
  `tools/lean/LakeCommandsOutput`, `ConcurrentToolEditApprovalHandlers`,
  `ToolEditApprovalGate`, `BashApprovalPrompt`, `DesktopToolEditApproval`,
  `VscodeToolEditApproval`; plus `BashTool`, `RunRegistry`,
  `RunAgentOwnership`, `SessionStores`, `sessionEvents` (5 files, 117 tests)
- `npm run check:effect-migration-ratchet` — pass, `import:async-mutex 0 files / 0 sites`,
  `catch:effect-importer` unchanged at 2 files / 2 sites
- `npm run check:dead-code-ratchet` — pass, no new findings
- `npx eslint` on all changed source files — clean
- `npx prettier --check` on all changed files — clean
- `corepack pnpm install` re-run; `pnpm-lock.yaml` committed (-13 lines), and
  `pnpm install --frozen-lockfile` after the rebase reports "Already up to date"
- No open PR touches `perKeyQueue`, `keyedMutex`, `lakeCommands`,
  `runApprovalQueue`, `editorPane` or `async-mutex`
- Not verified: the desktop e2e suite (needs an app build) and `npm test` in full


LionSR:
Thread on config/ratchets/effect-migration-baseline.json: correct. The script-side half (adding `import:async-mutex` to RETIRED_ROW_IDS so a re-introduced import fails instead of being re-seeded) lands in the open ratchet-rows lane PR that already edits scripts/check-effect-migration-ratchet.mjs, together with the two rows that PR retires.

PR #12736, opened 2026-09-18
refactor(approval): Effect-native tool-edit approval controller; delete runPreview and the hand-rolled settle

LionSR:
#12718 landed the legal subset of round-trip survey lane R14 and named this
conversion as the follow-up: `ToolEditApprovalController` had to become
Effect-native before `ToolEditApprovalHost.runPreview` — the runner face the
controller reached a host runtime through — and the hand-rolled
`let settle!` / `new Promise` pair could go. Both are gone here, along with
the Promise face of `ToolEditPreview`, `stagePreview` and `decide`.

The controller's verbs are Effects and it holds no runtime. The runs live at
the two host wiring points, which is where R1 admits them:
`packages/desktop/src/main/desktopAgentRun.ts` and
`packages/extension/src/progressView/ProgressViewProvider.ts`. Its three raw
catches (the host-boundary reports in `release`, `runAction` and `approve`)
are typed recovery through `Effect.catchCause`, which is what the
`catch:effect-importer` row requires of a file that imports `effect` at
runtime — the reason #12718 could not do the `Deferred` on its own.

## Ordering

The class's docblocks pin an order a release has to keep, because a release
deletes the temp files a host build may still be reading. It is unchanged:

1. **Join the staging.** `inFlight` was a promise that resolved *to* its
   failure rather than rejecting. It is now a `Deferred` carrying the
   staging's `Exit`, published *before* the entry that names it, so a release
   finds either no entry at all or an entry with the staging's outcome to wait
   on — never one with neither. `present` forks the staging with
   `Effect.forkDetach` and re-raises that `Exit`, so the failure still reaches
   exactly one handler, the host that called `present`.
2. **Join the admitted actions and the raw host builds.**
   `Promise.allSettled([...inFlightActions])` is
   `Effect.forEach(joins, …, { concurrency: 'unbounded', discard: true })`
   over the same set, now holding one never-failing join per item: an
   admitted action's is a `Deferred` its own `Effect.onExit` fills, and a host
   build's is still the raw promise, lifted with `Effect.exit`, because
   `openBuildDisplay` deliberately stays a promise — it is the only handle
   left once a preview program's settle race has interrupted the fiber that
   started the build.
3. **Dispose the preview**, then **run the `workspaceTempCleanup` programs**.

Why fibers reproduce it, rather than approximating it:

- **Staging outlives its requester.** `Effect.fork` ties a child to its
  parent; `Effect.forkDetach` does not. Every fork here is detached, which is
  exactly what voiding a promise meant — the work runs to completion whether
  or not the caller that started it is still there. A release joins that work
  through the entry, never through the fork.
- **Admission is synchronous.** Every bookkeeping step — the map write, the
  `Deferred` a later caller joins, the registration of an action's join —
  happens in the same synchronous step its caller looked the entry up in
  (`startRelease`, `admit`, `decideFromPayload` are called from inside an
  `Effect.suspend` body and return the program left to run). Nothing another
  fiber could interleave with sits between finding an entry and accounting
  for what is about to happen to it, which is the property the promise
  version got from the JS event loop for free. `dispose` admits *all* its
  releases in one such step before forking any of them, which is what makes
  the host's release landing right behind it join rather than start a second.
- **A scope per program would be wrong** and is not used: scope finalizers
  run at settle time, before `dispose` and while a build may still be reading
  the files. The temp cleanups stay an explicit list the release runs after
  `dispose`, as before.
- **Admitted actions are thunks.** `admit(entry, () => …)` keeps a host
  method behind an action from being *invoked* when the action is admitted
  rather than when it runs — the one real divergence this conversion produced,
  caught by the ordering assertion below before it could ship.

The proof is the assertion that already existed, unchanged and passing:
`ToolEditApprovalController.vitest.ts` holds a build open, releases, and
expects `['build-done', 'dispose', 'temp-cleanup']`; a second request repeats
it with a build that *fails*, so the release joins that one too rather than
hanging. Six further cases pin the rest: a release that lands mid-staging, one
that lands while the view is opening, a second release joining the first
one's cleanup, and actions arriving after the request was decided.

## Net elements (R6)

Removed (15):

- `ToolEditApprovalHost.runPreview` — the runner face; the whole point of the
  lane.
- `PendingToolEditApproval.settle`, and the `let settle!` / `new Promise`
  executor pair behind it — `LatexPreviewEntry.settled` is a `Deferred` the
  release fills.
- 3 raw `catch` clauses in the controller (`release`, `runAction`, `approve`).
- 1 `try`/`catch` in `VscodeToolEditPreview.present`.
- 4 `runPromise` sites inside hosts: `VscodeToolEditApprovalHost.stagePreview`,
  `VscodeToolEditPreview.dispose`, `DesktopToolEditApprovalHost.stagePreview`,
  and `DesktopToolEditPreview.openProposed` — the last one, with the
  `ProcessRuntime` constructor parameter that fed it, is the adapter #12734
  had to add three hours ago because `ToolEditPreview` was still a
  Promise-shaped core port. It is not one any more, so that host's
  Effect-typed `openPath` reaches the controller as the program it is.
- 2 `runPromise` wrappers around `request.decide`, one per host wiring point.
- `Effect.tryPromise` + `Effect.sync` adapters at 3 wiring call sites
  (`presentToolEdit`, the two session-event stream loops).

Added (2): `ToolEditApprovalController.detach` (the voided promise, named) and
`ToolEditApprovalController.admit`, which replaces `runAction` one-for-one —
so the class's method count is unchanged and its port lost a member.

LoC is +957/−646. The growth is the pipe form of what `async`/`await` wrote
in one line, plus the ordering rationale above recorded where it is enforced;
the element count is strictly negative and the old path is deleted in the
same change. `cleanup`, the one place the order is stated, reads as a
sequence rather than a nest of pipes.

## Consumer counts (R8)

- `ToolEditApprovalController.release`: 3 call sites (both hosts'
  `releaseToolEdit`, and `handleSessionEvent`), 2 runs — one `runPromise` per
  host, because `HostInteractions.releaseToolEdit` is the one place the
  session waits for a controller program.
- `present`: 2 call sites, both `runFork` at a wiring point.
- `handleAction`: 2 call sites, both now `yield*` inside a host request
  dispatcher whose `R` is already `ProcessServices`.
- `handleSessionEvent` / `dispose`: 2 call sites each, one per host.
- `approvePendingForRun`: **0 production call sites**, 3 suites. Converted as
  it stood; flagged rather than deleted, since removing it would take three
  test cases with it and that is a separate call.
- `ToolEditPreview`: 2 implementers, unchanged.

## Verified

- `npm run typecheck` — pass (after each step; the final tree clean across all
  7 project configs).
- `npx vitest run` over `ToolEditApprovalController`, `DesktopToolEditApproval`,
  `VscodeToolEditApproval`, `ToolEditApprovalGate`,
  `ConcurrentToolEditApprovalHandlers`, `OpenBuildDisplay` — 45 passed.
- `npx vitest run src/test-kernel/tools src/test-kernel/desktop` — 846 passed.
  (`src/test-kernel/controllers` + `src/test-kernel/frontend` alongside them:
  1116 passed.)
- `npm run test:pure` — 171 files, 2061 passed, architecture ratchets included.
- `node scripts/check-effect-migration-ratchet.mjs` — OK. `Effect.run*` stays
  at 0 files below the boundary and `catch:effect-importer` at its 2 baselined
  files: the controller imports `effect` at runtime now and carries no raw
  catch.
- `npm run check:unexecuted-effects` — OK, 2007 files, no awaited, voided,
  discarded or adapter-thunk Effects.
- `npx eslint` (2 invocations: production roots, then `src/test-kernel`) and
  `npx prettier --ignore-path /dev/null --check` on the changed files — clean.

## Left out

**`HostInteractions.releaseToolEdit` stays `Promise<void> | void`.** Typing it
as an Effect would make `src/agent/runtime/HostInteractions.ts` a runtime
`effect` importer — `presentToolEdit` must compose the release it hands back,
and it cannot run one from a VS Code-free zone — which puts all 6 of that
file's catch sites on the `catch:effect-importer` row in the same pass,
including the `dispose` fan-out isolation that is legitimately in the catch
budget. It would also rewrite the CLI's two synchronous implementers
(`subscribeApprovals.ts`, `approvalAdapter.ts`). That is a port conversion of
its own, not this lane; the cost of leaving it is one `runPromise` per host
wiring point, which is where a run belongs anyway.

**`ToolEditApprovalHost.reportError` stays `(message: string) => void`.** It is
fire-and-forget on both hosts today; awaiting it would delay every recovery
path it ends, including a release's, which is a timing change with nothing to
show for it.

**`BuildDisplayFn` stays promise-faced**, deliberately — see **Ordering**.

Refs #12424.


LionSR:
Rebased onto `origin/main` (now includes #12734 and #12735); force-pushed as `e6b7a0f779`.

**The conflict, and how it resolved.** #12734 Effect-typed `DesktopAgentRunHost.openPath`. Because `ToolEditPreview` was still a Promise-shaped core port at that moment, it had to add a `ProcessRuntime` constructor parameter to `DesktopToolEditPreview` and settle the program with `runPromise` inside `openProposed`. This PR retires the face that forced it, so both sides are kept and the adapter between them is gone:

```ts
openProposed(): Effect.Effect<void, unknown> {
  return this.ui.openPath(this.staged.proposedPath);
}
```

The parameter and the run go with it. Net elements is 15 removed rather than 13, and the body is updated.

**Re-verified on the rebased tree:** `npm run typecheck` clean across all 7 configs; the 6 approval suites 45/45; `src/test-kernel/desktop` 301/301; `check-effect-migration-ratchet` OK (`Effect.run*` still 0 below the boundary, `catch:effect-importer` still its 2 baselined files); `check:unexecuted-effects` OK; `check:dead-code-ratchet` OK; eslint (2 invocations) and `prettier --check` clean.

**One pre-existing flake, proven not ours.** `DesktopPreviewHost.vitest.ts > presents openFile failure once through the request dispatcher` intermittently times out at 10s under a loaded run. I reproduced it on **unmodified main content** in the same worktree (`git checkout origin/main -- <every file this PR touches>`, then the same command): same suite, same test. It passes on repeat runs here and touches nothing in this change.

PR #12793, opened 2026-09-18
refactor(hosts): type the diff-view host and the desktop file actions as Effect

LionSR:
Two related round trips, one PR: `DiffViewHost` and the desktop's progress file actions were Promise-shaped seams whose implementations settled Effects on a runtime the caller then re-lifted.

**Lane A — `DiffViewHost` is a program.** All four members (`openDiff`, `closeDiff`, `revealFirstChange`, `readProposedContent`) return Effects. The desktop returns the file-read/patch-write program it used to settle on `options.runtime`, taking only the window's services from that runtime, so `DiffViewUnavailable` now words a real failure instead of a lifted rejection and an interrupted fiber abandons the open instead of detaching from it. VS Code wraps its editor calls once where they are raised (`vscode.diff`, `tabGroups.close`, the proposed-file read). The five `fromEditor` wraps in `VscodeToolEditApprovalHost` and the two `fromHost` wraps in the desktop tool-edit preview are gone.

**Lane B — `DesktopProgressFileActions` is uniformly Effect-typed.** `diffAcceptedFilePair`, `diffStreamToolbarAction`, `runLatexdiffFile`, `findAndOpenLabel` and the private `runSharedLatexdiff` / `openSharedLatexdiffResults` / `openDiffOutput` stop settling on `host.runtime`; the `Exit` fold behind the round-aware diff's promise face is an `Effect.catch` that lets an interrupt travel on exactly as the `Cause.hasInterruptsOnly` rethrow did. `listWorkspaceCandidateFiles` is a program on the port. The four `fromHost` lifts in `desktopHostRequests` go with them, replaced where a tag is owed by the same `hostFailure(member, cause)` the lift used.

`openFirstLabelMatch` is a program too — its callers were already Effect-shaped underneath, so both hosts stop running one promise per candidate read and per open.

`raceWithTimeout` and its `Raced<T>` union die with their last caller: with the port Effect-typed, the reveal wait is #12785's `firstEventOrTimeout`.

## Net elements (R6)

- 16 files; net **+70 LoC** (597 insertions / 527 deletions). LoC is up because `Effect.gen` bodies with explicit `Effect.Effect<…>` return types are wordier than `async`/`await`; elements are down.
- Files: **-1** (`packages/extension/src/frontend/vscode/raceWithTimeout.ts`).
- Exports removed: `raceWithTimeout`. Types removed: `Raced<T>` (module-local).
- `Effect.run*` sites in production: **16 → 1**, and that one is a VS Code command handler (an R1 host entry).
  - `desktopProgressFileActions.ts` 9 → 0
  - `desktopDiffHost.ts` 3 → 0
  - `desktopHostRequests.ts` 2 → 0
  - `openFileCommands.ts` 2 → 1 (`texra.openLabel`'s handler)
- Promise lifts removed: 13 (5 `fromEditor`, 6 `fromHost`, 2 `Effect.tryPromise`). Lifts added: 9, every one a foreign edge wrapped exactly once inside the callee — `mkdtemp`, node `readFile`/`rm`, `vscode.diff`, `tabGroups.close`, `openTextDocument`/`showTextDocument`, the file lister's two listings.
- Declarations retyped, not added: `DiffViewHost` (4 members), `ProjectDiffHost` (2), `DesktopToolEditApprovalUi` (2), `DesktopProgressFileActionHost.listWorkspaceCandidateFiles`, `openFirstLabelMatch`.
- Two `withProcessServices` helpers (one per file) replace what `runtime.runPromise` was doing — the repo's existing `Effect.flatMap(runtime.contextEffect, provideContext)` pattern, already used in 8 places.

## Consumer counts (R8)

- `DiffViewHost` — 2 implementations (`VscodeDiffViewHost`, the desktop's `inProject` host), 2 type consumers (`DesktopAgentRunHost.openDiff`, `VscodeToolEditPreview`), 3 test doubles.
- `raceWithTimeout` — 1 caller (`VscodeDiffViewHost.revealFirstChange:91`), now 0. Not in `knip-baseline.json`, so no baseline shrink is owed.
- `openFirstLabelMatch` — 2 callers: `openFileCommands.ts:64` (VS Code) and `desktopProgressFileActions.ts:236` (desktop). Both converted.
- `DesktopProgressFileActions` public members — all reached only from `desktopHostRequests.ts` (`workflowFileActions` host, the `latexdiff` dispatch arm, `runWorkflowDiff`, the `openLabel` arm), all updated.
- `listWorkspaceCandidateFiles` — 1 implementation, 1 caller, 1 test double.

## Verified

- `npm run typecheck` — clean (all 7 projects).
- `npm run test:pure` — 172 files, 2068 passed, 1 skipped.
- `npx vitest run --project kernel src/test-kernel/desktop src/test-kernel/frontend src/test-kernel/controllers` — 543 passed, run 3× to confirm no flake.
- `npx vitest run --project kernel src/test-kernel/{desktop,frontend,controllers,commands,latex}` — 603 passed.
- `node scripts/check-effect-migration-ratchet.mjs` — OK; no count grew (`Effect.run*` and `catch:effect-importer` both unchanged, `src/latex/labelSearch.ts` newly imports `effect` and has zero raw catches).
- `npx eslint` on every touched file — clean.
- `npx prettier --write` on every touched file.
- `npm run test:changed` — reports no files (the edits are committed); the kernel runs above stand in for it.

One honest behavior note: `runSharedLatexdiff`'s old `Exit.isFailure` fold logged-and-fell-back for a **defect** as well as a typed failure. `Effect.catch` handles the typed channel only, so a defect in the shared latexdiff core now travels to the caller instead of being downgraded to a single-file diff. Interrupts behave exactly as before (they propagate; the pinned "does not fall back when the shared core is interrupted" case still passes, now running the program).

## Left out

- `src/test-kernel/frontend/VscodeToolEditApproval.vitest.ts` needed no change: it mocks the `vscode` module, not `DiffViewHost`, so it holds no double of the retyped port.
- No rejection assertion was retired. `DesktopDiffHost.vitest.ts:211`'s `rejects.toMatchObject({_tag:'ExternalOpenFailed'})` still pins the real failure — the suite now runs the program (`testRuntime().runPromise(host.openDiff(…))`) and the tagged failure surfaces unchanged, so the assertion earns its place rather than pinning the retired shape.
- `packages/desktop/src/main/index.ts` is untouched: `projectDiffHost.openDiff` / `projectRequestDiffHost.closeDiff` are passed by reference at `:1013`, `:1021`, `:1045`, so the retype flows through the wiring with no edit.
- `ProgressWorkflowFileActionsController`'s `FileActionServices` (`FileSystem | AgentResume`) still excludes `Path`, which the latexdiff core needs. Rather than widen a shared core type across both hosts, the desktop provides its own window services at the class boundary (`desktopProgressFileActions.ts:170`). Widening `FileActionServices` is a separate, cross-host change.


LionSR:
Addressed the `fromEditor` duplication in 8218de4: it is exported from `VscodeDiffViewHost` now, beside the `tabInputFileUri` the two approval hosts already share, and `VscodeToolEditApprovalHost` imports it (no cycle — that file already imports the diff host). Net one fewer declaration; typecheck, eslint, the dead-code ratchet and the desktop/frontend kernel suites are green on the new head.

PR #12797, opened 2026-09-18
refactor(desktop): type the composition root's host-request ports as Effect

LionSR:
The desktop window's composition root kept five `host.request` capabilities on
a promise face whose implementation ran a program and whose arm lifted the
promise straight back into one. Each is now an Effect end to end, so both the
run and the lift on either side of it are gone.

- `DesktopAgentRun.runValidated` **is** the launch program: the one-line
  `runPromise` face and the `runValidatedEffect` name behind it are deleted, and
  the merge scheduler (`startRun`) and the `launch` arm compose it directly. The
  scheduler's `Effect.catch` over a lifted rejection is now `Effect.catchCause` +
  `Cause.squash`, which is the same value `runPromise` used to reject with; an
  interrupts-only cause (the window torn down mid-merge, which the nested
  runtime used to absorb) is re-raised rather than presented.
- `DesktopFileSelection.fileOptions` / `attachDroppedFiles` answer with their
  programs, taking `FileSystem` from context as the snapshot port already
  declares. The drop's "nothing was attached" refusal (thrown by
  `attachDroppedPaths`) now travels on the failure channel as `Rejected`, which
  is exactly what `fromHost`'s `isRequestRefusal` passthrough produced before;
  anything else that helper could throw stays a defect. Electron's
  `showOpenFileDialog` in `pickFiles` stays the one foreign wrap, and the
  selection no longer takes a `ProcessRuntime` at all.
- `DesktopSupabaseAuth.signIn` returns `startSignIn(provider)`; its `Exit` fold
  moves to the two callers that own an entry — the settings view's Sign in
  button (`runPromise` at the IPC handler) and the window's provider prompt.
- `recheckTools` is `refreshToolAvailability(...)` (`E = never`,
  `R ⊆ ProcessServices`), so the arm just yields it.
- The setup kickoff (`kickoffSetup`) is one program on the window's context
  instead of three nested runs behind a `tryPromise`.

Behaviour is preserved at the request dialog: `run.runValidated` and `signIn`
keep their `hostFailure(member, cause)` naming (an error-channel rename, not a
lift), and for `attachDroppedFiles`/`recheckTools` the fold in
`handleHostRequest` already unwrapped `HostCallFailed.cause`, so it now sees the
identical value.

## Net elements (R6)

14 files, `+217 / -172` (net **+45** LoC; the accounting here is elements, and
several retyped members traded one line for a documented signature).

Symbols that cease to exist:
- `runValidatedEffect` (renamed onto `runValidated`) and the `async runValidated`
  promise face — `packages/desktop/src/main/desktopAgentRun.ts`.
- `DesktopFileSelectionOptions.runtime` — the file selection no longer needs a
  runtime.
- 5 lift sites: `fromHost('files.attachDroppedFiles')`, `fromHost('recheckTools')`,
  `fromHost('signIn')`, `fromHost('run.runValidated')`, and the
  `Effect.tryPromise` around `run.runValidated` in `startRun`.
- 6 runs: `runtime.runPromise` in `fileOptions`, in `attachDroppedFiles`, in
  `runValidated`, in `recheckTools`, and two of the three in `kickoffSetup`
  (`buildDesktopSetupRunRequest`, `loadAgents`), plus the `runPromiseExit` +
  `Cause.squash` fold in `DesktopSupabaseAuth.signIn`.
- 2 lifts in `index.ts`: the `Effect.tryPromise` around `files.fileOptions()` in
  the snapshot source, and the `Effect.tryPromise` wrapping the whole setup
  kickoff body.
- 1 lift in `desktopShellIpc.ts`: `hostCall(() => options.signIn())` (the
  `hostCall` helper itself survives for `openWorkspaceFolder`, still a promise).

Added: one `runtime.runPromise` at the settings-view Sign in IPC handler
(`desktopCredentialSettingsController.ts`), a genuine R1 host entry.

No exports added or removed; no new interfaces or classes. One knip-baseline
entry (`setupLaunch.ts: selectSetupCredentialModelExcludingOpenRouter`) shrinks
because the dynamic import is now reached as a program.

## Consumer counts (R8)

| retyped symbol | consumers | where |
|---|---|---|
| `DesktopAgentRun.runValidated` | 3 | `desktopHostRequests.ts` (`startRun`, `launch` arm), `index.ts` (`kickoffSetup`) + 1 internal (`runAgentRequest`) + 1 test |
| `DesktopFileSelection.fileOptions` | 1 prod (`index.ts` snapshot source) + 2 tests | |
| `DesktopFileSelection.attachDroppedFiles` | 1 prod (`desktopHostRequests.ts` arm) + 1 test | |
| `DesktopSupabaseAuth.signIn` | 2 prod (`index.ts` `signIn`, `desktopCredentialSettingsController.ts` `profileHandlers`) + 17 test call sites | |
| `DesktopHostRequestsOptions.signIn` / `DesktopShellActionFactoryOptions.signIn` | 1 producer (`index.ts` `signIn`), 2 consumers | |
| `DesktopHostRequestsOptions.recheckTools` | 1 producer, 1 consumer | |
| deleted `DesktopFileSelectionOptions.runtime` | 1 prod + 2 test call sites, all updated | |

`git grep` confirms no remaining `Promise`-shaped reference to any of these
(including string-path mock references: no `vi.mock` targets any of the touched
modules).

## Verified

- `npm run typecheck` — clean (all 7 projects).
- `npx vitest run --project kernel src/test-kernel/desktop` — 41 files / 300
  tests pass.
- `npm run test:pure` — 172 files / 2068 pass, 1 skipped.
- `npx vitest run --changed HEAD~1` (before the last rebase) — 6 files / 65 pass.
- `node scripts/check-effect-migration-ratchet.mjs` — OK, no count grew
  (`Effect.run*` 0/0, `catch:effect-importer` 2 files unchanged).
- `npm run check:dead-code-ratchet` — OK; the baseline shrinks by one entry in
  this PR.
- `npx eslint` on the 13 touched files — clean;
  `npx prettier --check` — clean.
- **Red, but not from this branch:** running three vitest instances of
  `src/test-kernel/desktop` in parallel makes
  `DesktopPreviewHost > presents openFile failure once through the request
  dispatcher` and
  `DesktopProgressFileActions latexdiff > passes pre-resolved round outputs to
  the shared core` time out at 10s. Reproduced identically on clean `origin/main`
  in a separate checkout under the same load; both pass in every serial run.

## Left out

- `packages/desktop/src/main/index.ts:738` `signInForRemoteAgentCatalog` still
  wraps `chooseOAuthProvider` + `signInAndWaitForSession` +
  `runtime.runPromise(supabaseAuth.authenticated)` in one `Effect.tryPromise`.
  Closing it means putting `SupabaseAuth` in the port's requirement channel,
  which `setupAuth.registerSignIn` declares as `never`, so it would need its own
  context plumbing and a retype of `signInAndWaitForSession`. Separate lane.
- `packages/desktop/src/main/desktopSupabaseAuth.ts:453` `signInAndWaitForSession`
  and `:475` `signOut` keep their `runPromiseExit` folds: both are consumed by
  promise-shaped settings/setup handlers this lane does not own.
- `packages/desktop/src/main/desktopHostRequests.ts:255` `listWorkspaceCandidateFiles`,
  `:309` `runLatexdiffFile` and `:331` `getLatexdiffRunContext` keep their
  `runtime.runPromise` faces — they belong to `DesktopProgressFileActions`, which
  the concurrent diff-host lane owns.
- `packages/desktop/src/main/desktopFileSelection.ts:128` `pickFiles` stays a
  promise: Electron's `dialog.showOpenDialog` is a foreign edge, wrapped exactly
  once inside the callee.
- `src/controllers/session/hostSnapshotSource.ts:74` needed no change — the
  snapshot port's `fileOptions` was already `Effect`-shaped; only the desktop
  producer was lifting into it.
- `src/controllers/mainView/MainViewDroppedFilesController.ts:37`
  `attachDroppedPaths` still signals by throwing `Rejected`. Making it total
  would change the extension's arm (where the same throw is currently a defect),
  so it is left for a lane that owns both hosts.


LionSR:
Addressed the `startRun` finding in e0c2d657b9: `Effect.catchCause` now re-raises an interrupts-only cause with `Effect.failCause` instead of presenting it, so a window torn down mid-merge ends the fork interrupted (silent under `withForkFailureReporting`) while a genuine failure or defect still reaches the "Merge failed" presentation the promise face used to produce.

PR #12807, opened 2026-09-18
refactor(cli): type the workflow output and instruction seams as Effect

LionSR:
Closes the `effect -> promise -> effect` round trips in the CLI workflow-output
and instruction seams. `resolveFileBackedInstruction`,
`assertOutputDirAvailable`, `assertOutputFileAvailable`, `resolveWorkflowOutput`
and `workflowRecoveryInputsAreDurable` now return Effects with their node `fs`
edges wrapped exactly once inside, so the five `Effect.tryPromise` re-lifts
their callers used to pay become plain `yield*` sequences. The two multi-agent
read commands fold into one Effect each, run once at their citty arms.

Behavior is preserved deliberately at the failure boundary: every former
`throw`/rejection in these functions becomes `Effect.fail`, not a defect, so
`Effect.result` in `executeCliWorkflowConfig` and the `CliUsageError` arms in
`runResumeCommand` still see them on the typed channel. The emitted CLI result
JSON is untouched.

## Net elements (R6)

| file | +/- |
|---|---|
| packages/cli/src/runtime/workflowOutput.ts | +141 / -106 |
| packages/cli/src/commands/workflow.ts | +7 / -20 |
| packages/cli/src/commands/resumeRun.ts | +30 / -27 |
| packages/cli/src/commands/multiAgent.ts | +25 / -23 |
| packages/cli/src/commands/_helpers/instructionFile.ts | +25 / -25 |
| test-kernel (5 suites) | +128 / -107 |

Net LoC from `git diff --stat origin/main`: **+356 / -308 = +48** (production
alone +27). Honest accounting: this lane is LoC-positive, not the ~-60 the
census estimated. Effect's explicit failure channel costs lines against `throw`
(`return yield* Effect.fail(...)` for one `throw`), and moving
`resolveWorkflowOutput`'s 75-line body inside an `Effect.gen` re-wraps it at a
deeper indent. It is element-negative, which is what the change is for:

Elements that cease to exist:
- `readInstructionFile` (function) — inlined into its only caller.
- 2 raw `try`/`catch` clauses in `workflowOutput.ts`, 1 in `instructionFile.ts`
  (both files now import `effect`; the `catch:effect-importer` ratchet row stays
  frozen at 2 files).
- 7 `Effect.tryPromise` lift sites: `workflow.ts` x4, `multiAgent.ts` x1,
  `resumeRun.ts` x1 (a four-call `try: async () => {...}` block), plus the
  `Promise.all` in `workflowRecoveryInputsAreDurable`.
- 2 `services.runtime.runPromise` calls inside multi-agent command bodies —
  each command is now one Effect run once at its `run:` arm.
- 5 `Promise` return faces (`resolveFileBackedInstruction`,
  `assertOutputDirAvailable`, `assertOutputFileAvailable`,
  `resolveWorkflowOutput`, `workflowRecoveryInputsAreDurable`).
- The duplicated `mkdir` + `copyFile` pair in `resolveWorkflowOutput` collapses
  into one `copyOutputFile` (2 callers).

No new exports. `probeOutputPathForTests` keeps its name and its injected
Promise-typed `OutputPathProbeDependencies` (they mirror `fs.stat`/`fs.mkdir`,
the foreign edge, and are wrapped once inside).

## Consumer counts (R8)

| retyped symbol | production callers | test consumers |
|---|---|---|
| `resolveFileBackedInstruction` | 2 (`workflow.ts:114`, `multiAgent.ts:135`) | 0 |
| `assertOutputDirAvailable` | 2 (`workflow.ts:147`, `resumeRun.ts:176`) | `OutputGuards.vitest.ts`, mocked in `ResumeCommand.vitest.ts` |
| `assertOutputFileAvailable` | 2 (`workflow.ts:154`, `resumeRun.ts:172`) | same two suites |
| `resolveWorkflowOutput` | 1 (`workflow.ts:407`) | `WorkflowOutputResolution.vitest.ts` (13 sites), `CliRootArgs.vitest.ts` (3) |
| `probeOutputPathForTests` | 0 | `OutputGuards.vitest.ts` (4) |
| `workflowRecoveryInputsAreDurable` | 1 (module-local, `resumeRun.ts`) | 0 |
| `readInstructionFile` (deleted) | 1 (module-local) | 0 |
| `runMultiAgentList` / `runMultiAgentShow` | 1 each (their own command def) | driven via the root command in `CliRootArgs.vitest.ts` |

Grepped for the module paths as strings too (`vi.mock('@cli/runtime/workflowOutput')`,
`'@cli/commands/_helpers/instructionFile'`): the only string-keyed reference is
the partial mock in `ResumeCommand.vitest.ts`, whose two doubles are retyped to
`Effect.void` / `Effect.fail`.

## Verified

Run in the worktree at `f736705e77` (rebased onto `d347364d6e`):

- `npm run typecheck` — clean (all 7 projects, test-kernel included).
- `npx vitest run --project kernel src/test-kernel/cli` — 70 files, 1008 passed,
  1 skipped.
- `npm run test:pure` — 172 files, 2068 passed, 1 skipped.
- `npm run check:unexecuted-effects` — OK (1999 files).
- `npm run check:effect-migration-ratchet` — OK, no count grew, no headroom;
  `catch:effect-importer` still 2 files / 2 sites.
- `pnpm --filter @texra-ai/cli validate:run` — CLI run validation passed.
- `npx eslint` on the ten touched files — clean.
- `npx prettier --ignore-path /dev/null --write` on the ten touched files.
- `npm run test:changed` reports "No test files found" once the work is
  committed, so the affected suites were named explicitly instead.

## Left out

- `packages/cli/src/commands/resumeRun.ts:62` `runResumeCommand` and
  `packages/cli/src/commands/resume.ts:36`. Refuted as a round trip: `resume.ts`'s
  citty `run(ctx)` `await`s the Promise and never re-lifts it, so the single
  `stores.runtime.runPromise` already sits at an R1 host entry. Moving the
  `initInteractiveCliPlatform` call out to the citty arm would also make
  `ResumeCommand.vitest.ts:213` ("routes platform init through the TUI-owning
  signal path, not headless init") assert against a mock the test itself calls,
  costing a documented invariant for no closed hop.
- `packages/cli/src/runtime/workflowOutput.ts:363,390` `resumeWorkflowOutputFile`
  / `resumeWorkflowOutputDirectory` stay synchronous throwers. Their two call
  sites disagree on what a throw means today — a defect inside
  `executeCliWorkflowConfig`'s `Effect.fn`, a typed failure through
  `resumeRun.ts`'s lift — so retyping them would change one of the two.
  `resumeRun.ts` now keeps that distinction with a three-line `Effect.try`
  around both readers.
- `packages/cli/src/commands/workflow.ts:129` and `multiAgent.ts:144`:
  `initLocalCliPlatform` / `initCliPlatform` keep their single `Effect.tryPromise`
  (initPlatform.ts belongs to the lifecycle lane).
- `packages/cli/src/commands/workflow.ts:353` `writeResumeHint` and
  `resumeRun.ts:138` (the lift over `runChat`) untouched, per the lane boundary.
- The `readStdinText: readCliStdinText` supplier lines in `workflow.ts` and
  `multiAgent.ts` are textually unchanged, as the context lane requires.


LionSR:
Fixed in 4318d3c404: the resume preflight now reads each stored destination immediately before its own probe, so the `--output` probe's `mkdir -p` still runs before `resumeWorkflowOutputDirectory` is ever called and the error precedence between the two arms is unchanged. Good catch.

On the second point: `Effect.forEach(..., { concurrency: 'unbounded' })` does make the durability checks interruptible where `Promise.all` inside one `tryPromise` was not. That is deliberate and confined to `fs.access` reads whose only consumer is the boolean handed to `executeCliWorkflowConfig`; an interrupt at that point cancels the resume anyway.

PR #12815, opened 2026-09-18
fix(cli): keep an unparsable package.json candidate from failing CLI startup

LionSR:
Repairs the two findings the TeXRA reviewer raised on #12811. That PR merged
before this fix landed, so it ships separately.

**1. Behavioral regression (the reason this PR exists).** #12811 removed
`readCliPackageManifest`'s raw `try`/`catch` (the file newly imports `effect`
at runtime, so the `catch:effect-importer` ratchet requires zero catch clauses)
and replaced it with
`readFile(candidate, 'utf8').then((text) => JSON.parse(text), () => undefined)`.
A `then` rejection handler does not observe a throw from its own fulfillment
handler, so only the *read's* rejection fell through to the next candidate. A
present-but-malformed `package.json` rejected `readCliPackageManifest`, which
rejected `readCliVersion`, which failed `buildCliContext` through
`Effect.promise` — i.e. a corrupt manifest failed CLI startup instead of
degrading the version to `'unknown'`, which is what the loop, both inline
comments and #12811's body all promise.

The read's rejection and the parse's throw are now recovered separately, the
parse through `Result.try` — the idiom `updateChecker.ts:87` already uses for
`fileURLToPath` in the same package — so the file still carries zero raw catch
clauses and zero `.catch(` calls.

**2. Test convention.** The five `--cwd` validation tests used
`Effect.runPromise` inside bodies of a suite that imports `it` from
`@effect/vitest`. AGENTS.md ("Effect-based tests") keeps `Effect.runPromise` to
hooks and non-test helpers, so they are now `it.live` with `yield*` and
`Effect.flip`. `it.live` rather than `it.effect` because they probe the real
filesystem and `it.effect`'s `TestContext` clock starts at 0. The `cliContext`
helper keeps its single `Effect.runPromise`: it is a non-test helper, which the
reviewer explicitly allowed.

## Net elements (R6)

2 files, +36 / −22 = net +14 LoC. No symbol added or removed; no export, type
or interface changed. The elements that change are internal to
`readCliPackageManifest`: one `.then(onFulfilled, onRejected)` pair becomes one
`.then(value, () => undefined)` read plus one `Result.try` parse, which is the
only shape that recovers both failure sources without a catch clause. On the
test side five `await Effect.runPromise(...)` call sites become five `yield*`
sites inside `it.live`.

## Consumer counts (R8)

No public surface moves in this PR. `readCliPackageManifest` is module-private
with one consumer (`loadCliPackageManifest`, `cliContext.ts`), which serves
`readCliVersion` (4 readers: `bin/texra.ts:35`, `runChatTui.tsx:242`,
`root.ts:66`, `version.ts:30`, plus `buildCliContext` itself) and
`readCliBugsUrl` (1 reader: `bin/texra.ts:35`). All of them keep the exact
signatures #12811 left them with, which is why this fix is confined to two
files.

## Verified

This repairs #12811 (merged). Run in a worktree branched from `origin/main`
with #12811 already in it:

- `npx vitest run --project kernel src/test-kernel/cli/CliContext.vitest.ts` —
  36 passed (the five converted `it.live` tests among them).
- `npm run test:pure` — 172 files, 2068 passed / 1 skipped.
- `node scripts/check-effect-migration-ratchet.mjs` — OK; `catch:effect-importer`
  still 2 files / 2 sites with `cliContext.ts` absent, which is the constraint
  that shaped the fix.
- `npm run check:unexecuted-effects` — OK: no awaited, voided, discarded or
  adapter-thunk Effects.
- `npx eslint` and `npx prettier --ignore-path /dev/null --check` on both
  touched files — clean.

## Left out

- **No regression test for finding 1.** The two candidate URLs are resolved
  from `import.meta.url` and the manifest read is memoized in a module-level
  `cachedManifest`, so there is no seam to point at a malformed fixture without
  inventing one. Adding an injection parameter to make the probe testable would
  be a new element on a startup path that has exactly one caller — the opposite
  of what this fix is for. The behavior is pinned instead by the comment naming
  why the two recoveries are separate.
- The `LogSinks.vitest.ts` residual risk the reviewer noted (the prompt test
  depends on `Effect.runPromise` starting its fiber synchronously so the
  `acquireUseRelease` acquire attaches readline before `input.end`) is
  unchanged: the assumption holds in rc.115, the pre-#12811 test relied on the
  same ordering for a different reason, and rewriting it is not this fix.


LionSR:
Second commit `60832893c1` takes the non-blocking suggestion: the manifest parse now goes through `@common/parsing/safeParseJson` instead of a local `Result.try(() => JSON.parse(text))`. Behavior is unchanged (both recover a `SyntaxError` to `undefined` and narrow with the same cast); it is one fewer hand-rolled duplicate of a helper this package already uses.

Re-verified on `60832893c1`: `npm run typecheck` clean; `CliContext.vitest.ts` 36 passed; `npm run test:pure` 172 files / 2068 passed; effect-migration ratchet OK (`catch:effect-importer` still 2 files / 2 sites); unexecuted-effects gate OK; dead-code ratchet OK; eslint + prettier clean.

Note for the record: the `review` check failed on the first commit for an infrastructure reason, not the code — `HTTP 402 Payment Required – Insufficient Balance` on the reviewer's `deepseekproT` key.

PR #12835, opened 2026-09-19
refactor(extension): make VS Code progress bodies Effects, not nested settles

LionSR:
`runGuardedLatexCommand`'s `operation` was already Effect-typed (#12810/#12818). What was still Promise-shaped was everything *inside* a `vscode.window.withProgress` body: every progress body was an `async` callback that settled its work on a `ProcessRuntime` the command carried only for that purpose.

`withVSCodeProgress` (`packages/extension/src/frontend/ui/progress.ts`) is now the extension's one wrap of `withProgress`. It takes the body as an Effect. `withProgress` only knows how to wait on a promise, so it is handed one the program resolves when the body exits (`Effect.ensuring`, so an interrupt cannot leave the notification open). The body is therefore a step of the calling fiber — its services, its interruption and its typed failures are the caller's own — and the helper introduces **no** new `Effect.run*` site, which is why `BARE_EFFECT_RUN_SITES` in `dependencyDirection.vitest.ts` is untouched. (A first pass used `Effect.runPromiseExitWith` over a captured context; the pure-tier ratchet correctly flagged that as a new bare run, so the run was removed rather than the allowlist widened.)

**Runs inside `withProgress` bodies: 8 before, 0 after.**

| site | before |
|---|---|
| `figCommands.ts` (compile TikZ) | 3 |
| `latexCommands.ts` (TeX count) | 2 |
| `latexdiffCommands.ts` (run latexdiff) | 1 |
| `arXivCommands.ts` (download source) | 1 |
| `subscriptionSignIn.ts` (OAuth leg) | 1 |

Each command now settles exactly once, at its registered handler.

## Net elements (R6)

- Files: 13 changed, 1 added (`frontend/ui/progress.ts`).
- Exports: **+1** (`withVSCodeProgress`). `ProgressReporter` is deliberately not exported.
- Parameters that ceased to exist (8): `runtime` on `handleGetTeXCount`, `handleCompileTikzFigures`, `downloadArXivSource`, `handleRunLatexdiff`, `signInWithSubscription`, `ensureCredentialOrPrompt`, `launchSetupAssistant`, and the `SubscriptionHandlers` constructor. The last three were pure carriers once the OAuth leg stopped needing a runtime — dead plumbing, deleted with its cause.
- Class/interface declarations: unchanged.
- Net LoC: **+31** (`git diff --stat origin/main`: 257 insertions, 226 deletions). The one new file is 61 lines; the call sites are net −30. The trade is 5 `async`/`await` progress bodies and 8 nested settles for one wrap, so the element count drops even though the line count does not.

## Consumer counts (R8)

Grepped before retyping:

- `runGuardedLatexCommand` — 5 callers (`figCommands.ts` ×2, `latexCommands.ts` ×3). Signature unchanged; its `operation` was already an Effect.
- `withVSCodeProgress` — 5 callers across 5 files, so the helper clears the two-caller bar and is not a single-caller extraction.
- `handleGetTeXCount` / `handleCompileTikzFigures` / `downloadArXivSource` — 1 caller each, `extensionCommandSurface.ts`.
- `handleRunLatexdiff` — 1 caller, the `texra.runLatexdiff` registration in the same file.
- `signInWithSubscription` — 4 production callers (`extension.ts`, `setupAssistantCommand.ts`, `settingsView/handlers/subscriptionHandlers.ts`, `progressView/extensionHostRequests.ts`) plus 1 test; all updated. Its requirement channel tightened from `never` to the `HttpClient | Secrets` the transport actually needs.
- `launchSetupAssistant` — 1 caller, `extensionCommandSurface.ts`.
- `SubscriptionHandlers` — 2 constructions, both in `SettingsViewMessageHandler.ts`.

Behaviour: every converted body kept its terminal boundary. `runGuardedLatexCommand` and `withLatexdiffTool` already answer a failure and a defect alike through `catchCause` + `Cause.squash`, so a failure that used to arrive as a rejected promise (a defect) and now arrives as a typed failure reaches the same report with the same value. `downloadArXivSource` and the OAuth leg were converted to `catchCause`/`Cause.squash` for exactly that reason, rather than to `catch`/`mapError`, which would have let a defect escape a path that used to catch it.

## Verified

Run in the worktree:

- `npm run typecheck` — clean (all projects).
- `npm run test:pure` — 172 files, 2068 passed, 1 skipped. (The first helper design failed `dependencyDirection.vitest.ts`; the current one passes.)
- `npx vitest run --changed origin/main` — 5 files, 31 passed.
- `npx vitest run --project kernel src/test-kernel/frontend/CodexSubscriptionSignIn.vitest.ts` — 7 passed.
- `node scripts/check-effect-migration-ratchet.mjs` — OK, no count grew, no headroom.
- `npx eslint` on the 13 touched files — clean. `npx prettier --ignore-path /dev/null --write` applied.

Not run: the full `npm test`.

## Left out

- `packages/extension/src/commands/git/gitCommands.ts:289` — `runClone`'s `withProgress` wraps `execa` directly, with the fiber's `cancelSignal` threaded in. There is no inner run to collapse; it is already a foreign edge wrapped once.
- `packages/extension/src/frontend/review/AgentReviewService.ts:227` — `withProgress` wraps `this.executeReview`, a genuinely Promise-shaped class method. Converting it is an `AgentReviewService` lane, not this one, and there is no round trip here.
- `packages/extension/src/frontend/auth/SupabaseAuthProvider.ts:541` — `createSessionViaSupabaseOAuth` must return a Promise to VS Code's `AuthenticationProvider` API, and the run inside is that host contract's own settle, entangled with the cancellation token and `invalidateActiveAttempt`. Left for the auth-provider lane.
- Heads-up for the merge order: #12834 also touches `progressView/extensionHostRequests.ts` and `settingsView/SettingsViewMessageHandler.ts`. This PR's edit in each is a single deleted argument, so a rebase should be mechanical.


LionSR:
Both review suggestions are addressed in 768601d.

1. `progress.ts` — the `ensuring` now wraps every step from the moment `withProgress` opens the notification, not just the `body(...)` step, so an interrupt landing on the `started` await (before VS Code has invoked the task callback) still resolves the promise VS Code waits on. Resolving twice is a no-op.

2. `arXivCommands.ts` / `subscriptionSignIn.ts` — both terminals now re-raise an interrupts-only cause (`Cause.hasInterruptsOnly(cause) ? Effect.interrupt : …`), matching `launchSetupAssistant`. This also restores exact pre-PR behaviour: both sites were `Effect.catch` / `Effect.tryPromise` before, which only ever saw typed failures, so widening them to `catchCause` without this guard would have turned a teardown into a user-facing error toast. `Effect.interrupt` keeps both error channels at `never`, so no caller signature moves.

Not changed: the `catchCause` terminals in `runGuardedLatexCommand` and `withLatexdiffTool`. Those predate this PR and this diff does not alter which causes reach them — an interrupt on the outer fiber already reached them through the old `Effect.promise(() => withProgress(...))` await. Worth a follow-up, but it is not this PR's behaviour change to make.

Re-verified on 768601d: `npm run typecheck` clean, `npm run test:pure` 172 files / 2068 passed, `CodexSubscriptionSignIn.vitest.ts` 7 passed, `check:unexecuted-effects` OK, `check-effect-migration-ratchet.mjs` OK, eslint + prettier clean on the three touched files.

PR #12838, opened 2026-09-19
docs(cli): record why the two resume programs must stay separate

LionSR:
## Summary

`resume` and `tryResumeRun` in `packages/cli/src/chat/chatSessionController.ts` read as ~150-line near-duplicates: they call the same helpers in a similar order, and a structural debt audit flagged them as a dedup candidate worth ~100–150 LoC.

They are not duplicates. This PR is the **negative result** of attempting that extraction: a 30-line comment block recording why the two bodies must stay separate, so the next reader does not spend the same effort re-deriving it. **No behavior change, no code change** — the diff is comments only.

The four divergences, each verified against the code:

- **Completion contract.** `resume` forks the run chain and completes at rehydration (fire-and-forget, per its interface docstring) using `Deferred.makeUnsafe<void, unknown>()` (:709). `tryResumeRun` awaits its deferred so the caller reads a boolean, using `Deferred.makeUnsafe<boolean, AgentResumeFailed>()` (:887) and `return Deferred.await(autoResumeRun)` (:1026). `resume` must never raise `AgentResumeFailed`.
- **Recovery transfer — two different protocols.** A manual resume supersedes unconditionally in its synchronous prologue, seeds the batch into `resumeRun` as `extraFollowUps`, and restores whenever the stream queue never took over. A wake supersedes only when the caller passed no queue-ready callback, writes the batch through `submitBatch(runId, …, 'live_owner')` *before* resuming, and restores only when that single write fails or is refused. Two protocols, two restore predicates.
- **Adoption point.** `resume` adopts config inside `onResumeResolved`, which `resumeRun` invokes only after `claimOwner` succeeds and the saved state loaded (`src/agent/runtime/resumeRun.ts:271-298`) — that ordering is what lets a run advertised from its checkpoint `stat` alone be refused into the chat the user is looking at, rather than clearing the transcript onto a dead stream. `tryResumeRun` adopts before the call and leaves the local transcript alone.
- **Refusals and gating.** `resume` gates on `readConfig()` + `exists()` plus an `AgentCategory.ToolUse` guard and names its refusal reason in the transcript, leaving the exit code untouched. `tryResumeRun` reads config only, has no category check, and answers `false` silently.

Unifying these takes roughly one knob per bullet (≥8 in total: deferred type, outer-recovery presence, restore predicate, adoption placement, `requireDelivered`, `onNotStarted`, refusal wording, category check) to share about 12 non-contiguous lines. Those knobs would sit directly on the `handBackUnusedRecovery` / `restoreInterruptedRecovery` pairing, where a double hand-back or a missed restore silently loses the follow-ups a user typed during an interruption. That is a net-negative abstraction on the highest-consequence invariant in the file.

## Issue acceptance gates

No issue. This came out of a structural debt audit of the CLI area; the audit's candidate for this file is hereby resolved as **declined, with rationale recorded**.

## Single source of truth

Unchanged. The recovery-lease invariant already has a single named owner: `handBackUnusedRecovery` is guarded-idempotent — it calls `followUps.useRecovery(recovery)` first, and `release(lease, 'recoverable')` sets `pendingRelease` / clears `entry.owner` (`ToolUseFollowUpQueueManager.ts:248-258, 359-372`), so a second call no-ops. This PR adds no new owner; it documents why that owner should not be wrapped.

No duplication removed and **no abstraction added** — that is the finding. The genuinely shared steps are *already* extracted into named helpers (`setupRunHost`, `toolUseResumeOptions`, `settleResumedTurn`, `recoverRun`, and the two lease helpers); what remains in each body is the part that differs by contract.

A smaller candidate was also rejected: a `recoveryHandoff()` closure owning `recovery` + `handedOff` would add ~14 LoC to remove 6, and would put an indirection between the lease and its hand-back, making the exactly-once argument harder to verify by reading. Note `recoveryHandedOff = true` cannot move into a helper at all, because the caller's `ensuring` reads the flag — passing a mutable flag setter in and out relocates mirrored state behind a call instead of removing it.

## Net elements (R6)

- Files changed: 1 — `packages/cli/src/chat/chatSessionController.ts` (`30 insertions(+), 0 deletions(-)`)
- Net LoC: **+30**, all comment lines (1431 → 1461)
- `^[+-]export` over the diff: **0**
- Classes / interfaces / types added: **0**; deleted: **0**. No new exports, no new module, no new file.

## Consumer counts (R8)

n/a — no emit path or export was deleted or re-routed; the diff adds no code.

## Host impact

Extension: none — no files under `packages/extension/` touched.

Desktop: none — no files under `packages/desktop/` touched.

CLI: comment-only in `chat/chatSessionController.ts`. No change to the TUI, headless `texra run` / `--print` output, exit codes, or flags.

## Scheduled refactoring

Two tractable follow-ups surfaced while tracing this, both deliberately left out:

- **`resume`'s internal duplication.** `refuseResume` (restore + transcript + `markRunCompleted` + settle) and the outer `recoverRun` recovery share 4 of 5 statements. Merging them is *not* a pure refactor: it would route refusals through `reportRunFailure`, whose `stopRequested` short-circuit would suppress the "Run not found" wording and would newly write `AgentError` where refusals currently leave the exit code untouched. Worth a separate, deliberately-scoped PR.
- **Double read of the same two facts.** `resume` pre-reads `readConfig()` + `exists()` and `resumeRun` reads the same pair again internally (`resumeRun.ts:218-231`). The pre-read is not redundant — it exists to produce a worded refusal before claiming — but it is a second reader, if a cheap simplification is ever wanted.

## Validation

Run in a clean worktree off `origin/main`:

- `npm run typecheck` → exit 0
- `npm run lint` → exit 0, no output
- `npm run format` → exit 0, no files rewritten (diffstat unchanged)
- `npx vitest run src/test-kernel/cli/chatSessionController.vitest.ts` → 1 file, **40 passed**
- `npm run test:changed` → 13 files, **215 passed / 1 skipped**
- `npm run test:pure` → 172 files, **2068 passed / 1 skipped**

`npm test` (full suite) was not run — the diff is comment-only.

Worth noting for reviewers: `src/test-kernel/cli/chatSessionController.vitest.ts` (1719 lines, 40 tests) covers both prologues and the lease invariants directly, and **several of those tests assert the very differences listed above** — which is independent evidence that the divergence is contract rather than drift.

---


LionSR:

**Failing check:** `review` ([job 105871315928](https://github.com/LionSR/TeXRA/actions/runs/35433197432/job/105871315928)) on `60417f9`.

**Cause — an exhausted API balance, not the diff:**

```
review · 0s
review · error · 1s
```

The job is the TeXRA self-review action (`TEXRA_MODEL: deepseekproT`). It failed 1.9s in, at the provider call, **before reading any code** — the subsequent `Normalize review` and `Post TeXRA review` steps were skipped. `402 Insufficient Balance` is a billing state on the configured DeepSeek account; nothing in a diff can produce or clear it.

This PR's diff is **30 added comment lines, 0 deletions, in one file** — no code, no exports, no config, and nothing touching the review workflow or its credentials. There is no causal path from it to a payment error.

Corroborating, independently: `chatgpt-codex-connector[bot]` commented on this PR that it has *also* hit its usage limits. Two unrelated review services are both out of credit, which points at account provisioning rather than at this branch.

**No fix exists to port.** This is not a code defect and not a flake — it needs the DeepSeek account topped up, or `TEXRA_MODEL` pointed at a funded route, in repo configuration. That's outside this PR's scope and outside what a push can change.

**On the re-run:** I'm deliberately not spending one. A re-run is for distinguishing a flake from a real failure, and `402 Insufficient Balance` is deterministic — it will return the same result and consume another CI cycle for zero new information. Happy to re-run if you'd rather have it on the record.

**Rest of CI on this commit:** `detect changes`, `guidance references` and `label` green; `runtime validation` and `docs build` skipped; `static checks`, `build artifacts`, `kernel tests` (both shards), `webview smoke` and `claude-review` were still in progress at the time of writing. If any of those go red, that one *is* mine and I'll fix it. Note `claude-review` may fail the same way if it draws on the same credentials.

Validation run locally in a clean worktree off `origin/main`: `typecheck` exit 0, `lint` exit 0, `format` no rewrites, `test:pure` 172 files / 2068 passed, `test:changed` 13 files / 215 passed, `chatSessionController.vitest.ts` 40 passed, dead-code ratchet `170 vs 170 baselined`.

---

PR #12839, opened 2026-09-19
refactor: answer the host-neutral host.request arms in one body

LionSR:
## Summary

`extensionHostRequests.ts` (1127 LOC) and `desktopHostRequests.ts` (992 LOC) are parallel switches over the same 56-arm `HostRequest` union. Where an arm is nothing but a verb on a port *both* hosts already bind, the two bodies were maintained in duplicate.

This moves those arms into one body: `handleSharedHostRequest` in `src/controllers/session/sharedHostRequests.ts`. **13 case labels / 11 distinct bodies** now have a single implementation, and the `HostCallFailed` error tag — previously declared character-for-character twice, once per host — has one home.

| kind | shared port |
|---|---|
| `openRunStorage`, `fileAction` | `ProgressWorkflowFileActionsController` (+ `runActions.readConfig`) |
| `resume`, `runNew`, `runCompileFixer`, `useOwnApiKey` | `HostRunActions` |
| `record`, `polish`, `savePastedImage` | attached `HostDraftRequests` |
| `refreshCommits`, `refreshFiles`, `dismissBanner` | `HostSnapshotSource` |
| `toolEdit` | `ToolEditApprovalController` |

**Read the net accounting below before reviewing — this PR is net +78 LOC, not a reduction.** It is being offered on a change-amplification argument, not a line-count one, and that is a judgment call for the owner.

### Two corrections to the audit that motivated this

Honesty about provenance, since both cut against the original case:

1. The audit claimed ~20 mergeable arms. The real number is **13 labels / 11 bodies**. Every other arm that *looked* shared has a genuinely different body.
2. The audit cited `extensionHostRequests.ts:912-930` and `desktopHostRequests.ts:771-789` as "confirmed byte-identical". **They are not.** `diff` over those exact ranges shows `restoreIntoLauncher` differs — the extension yields an Effect that also awaits `showInSidebar()`, while desktop calls a `void` function and posts the surface action alone. Only the `resume`/`runNew`/`runCompileFixer`/`useOwnApiKey` tail of that block is identical.

### Left per-host on purpose

`openFile`, `openLabel`, `exportTranscript`, `restoreIntoLauncher`, `restoreProposalConfig`, `latexdiff`, `pack`/`clean`, `latexdiffs`, `popOut`/`popBack`, `openDashboard`, `openSettings`, `pickFiles`, `useCurrentFile`, `addOpenedFiles`, `attachDroppedFiles`, `launch`, `extractFigures`, `apiKeyBanner`, `agentConfigBanner`, `recheckDependencies`, `openInstallGuide`, `signIn`, `gettingStarted`, `onboarding`, `setActiveView`.

Several of these have identical arm *text* over differing bodies. Merging those would have required inventing one 1:1 pass-through port per arm — the pass-through layer CLAUDE.md's abstraction-discipline rule bans — so they stayed put.

### Exhaustiveness is preserved

**Each host keeps its own `case` labels**; only the bodies moved. Both switches remain exhaustive over the 56-arm union, so the compiler still names a forgotten kind. Moving the labels into the shared map would have traded that away. There is no `default:` arm anywhere, so no kind is silently dropped. `handleSharedHostRequest` is a plain free function over a ports record — no factory, since there is nothing to capture.

## Issue acceptance gates

No issue; this came out of a structural debt audit of the two GUI hosts. No gate text to satisfy.

## Single source of truth

- `src/controllers/session/sharedHostRequests.ts` now owns **the answer to a host-neutral `host.request` arm**. `SharedHostRequestPorts` names exactly the five ports such an arm may reach.
- `src/controllers/session/hostCallFailure.ts` now owns **the `HostCallFailed` tag and the lift over a host promise** (`hostFailure`, `fromHost`), replacing one duplicated class declaration and two near-identical lifts (they differed only in `PromiseLike` vs `Promise`).

Removed: 13 duplicated arm bodies (−151 lines across the two hosts) and one duplicated error class (2 declarations → 1).

Added: one free function + one ports interface + one error module. The invariant `SharedHostRequestPorts` owns is *"these five ports are host-neutral — an arm reaching only them needs no host"*, which is the host-boundary fact that makes the shared body admissible. Nothing added is a pass-through: every merged arm body contains the real work, and the hosts call the shared function directly rather than wrapping it.

## Net elements (R6)

```
 packages/desktop/src/main/desktopHostRequests.ts      | 113 +++----------- (34+ / 79-)
 packages/extension/src/progressView/…HostRequests.ts  | 103 +++----------- (31+ / 72-)
 src/controllers/session/hostCallFailure.ts            |  51 ++++++++++    (new)
 src/controllers/session/hostRunActions.ts             |   4 +-
 src/controllers/session/sharedHostRequests.ts         | 111 ++++++++++++   (new)
 5 files changed, 230 insertions(+), 152 deletions(-)
```

- **Net LoC: +78.** The duplicated bodies shrank by 151 lines; the two new modules cost ~162, of which ~40 are the ports interface, the `SharedHostRequest` `Extract` union, and file docstrings. **This does not pay for itself in lines** — the return is that 13 arms and one error tag have one definition instead of two.
- `^[+-]export` over the diff: **+6 / −0**, but only **2 are new symbols** (`SharedHostRequestPorts`, `handleSharedHostRequest`). The other 4 are consolidations: `HostCallFailed`, `hostFailure`, `fromHost` were previously duplicated locals, and `HostRunActions` is a pre-existing private interface promoted to one home (visibility only).
- Classes: 2 declarations → **1** (net −1). Interfaces: **+1** (`SharedHostRequestPorts`). Local types: +1 (`SharedHostRequest`, unexported). Local function/const declarations deleted: 3 → 2 module functions.
- Tests added: **0** (behavior-preserving refactor, per the testing-discipline budget).

## Consumer counts (R8)

No emit path or subscribe surface was deleted or re-routed — no `bus.emit`, `SessionHandle.publish`, or `AppSignals` change. The four re-homed exports each have their consumers in this PR: `HostCallFailed` / `fromHead` lift → both host files (2 consumers each); `HostRunActions` → already consumed by both hosts plus `sharedHostRequests.ts`; `handleSharedHostRequest` / `SharedHostRequestPorts` → 2 consumers (one per host). `check:dead-code-ratchet` confirms no new findings.

## Host impact

Extension: `extensionHostRequests.ts` sheds 13 arm bodies and its local `HostCallFailed`/`fromHost`. Same ports, same frozen `done` literal, same exhaustive switch — behaviour unchanged.

Desktop: the same, plus its local `hostFailure` is now imported. **One ordering nuance worth a reviewer's eye:** `run.toolEditApprovals` and `options.snapshot` are now read when the handler is constructed (:385-387) rather than when the arm fires (previously :818, :821, :872, :911). This is safe because each is assigned exactly once and never reassigned — `toolEditApprovals` is a `readonly` member built at `desktopAgentRun.ts:162` and placed in the returned object at :256; `snapshot` is assigned once in the constructor (extension: `ProgressViewProvider.ts:210`). Neither is a getter and neither is lazily populated, so capturing the reference early is equivalent. The existing `DesktopPreviewHost` suite (which passes `run: {}`) still passes.

CLI: none. `packages/cli` never uses these dispatchers; its `performHostCapability` handles only `useOwnApiKey`, through its own path.

## Scheduled refactoring

Two follow-ups, both deliberately excluded to keep this reviewable:

- **Settings-view dispatch** — the much larger parallel duplication: `settingsView/SettingsViewMessageHandler.ts` (1167 LOC) against `desktopSettingsIpc.ts` (687) + `desktopAgentSettingsController.ts` (735) + `desktopCredentialSettingsController.ts` (670) + `desktopToolingSettingsController.ts` (258). Both hosts dispatch the same ~45 `SETTINGS_VIEW_COMMANDS` onto the same shared `@controllers/settingsView/*`. Worth its own issue.
- **The arm-text-matches-body-differs group** (`exportTranscript`, `latexdiffs`, `agentConfigBanner`, `onboarding`) becomes mergeable only if the underlying capability is *exposed* rather than wrapped in a fresh 1:1 port — e.g. both hosts' `latexdiffs` arms duplicate `ProgressWorkflowFileActionsController`'s already-bound `mergeFile`/`latexdiffFile` calls verbatim. That is a controller-shaped change, not a dispatch-shaped one.

## Validation

- `npm run typecheck` → **all 12 steps exit 0** (`workspace`, `test-kernel`, `agent`, `llm`, `cli`, `cli-scripts`, `trace-viewer`, desktop `base`/`main`/`preload`/`renderer`/`tooling`). Run step-by-step with `--checkers 2`; `--checkers 8` was OOM-killed by unrelated concurrent work on the same box.
- `npm run lint` → exit 0, no output
- `npx prettier --check` on the 5 touched files → `All matched files use Prettier code style!`
- `npm run test:pure` → `172 passed (172)` files, `2068 passed | 1 skipped` — carries the architecture ratchets (`architecture-edges`, `dependencyDirection`, `effect-migration`)
- `npm run test:changed` → `7 passed (7)` files, `50 passed (50)`
- Targeted: `DesktopPreviewHost`, `DesktopProgressFileActions`, `sessionFramer`, `completedRunArchive` → exit 0. **Disclosure:** the first attempt showed one failure — `DesktopPreviewHost > presents openFile failure…` timed out at 10000ms while a full lint ran concurrently on the same box. Re-run alone: `1 passed (1)` file, `17 passed (17)`. Load-induced, not a code failure.

`npm test` (full tier) was **not** run — the box was contended by concurrent work. `test:pure` + `test:changed` + the targeted suites were run instead. Also checked by hand that the change adds no new `architecture-edges` subsystem pair (all new imports stay `controllers` → `platform`/`shared`/`utils`, already baselined) and no new `effect-migration` row.

---


LionSR:

**Failing check:** `review` ([job 105872304439](https://github.com/LionSR/TeXRA/actions/runs/35433560183/job/105872304439)) on `010f27a`.

```
review · 0s
review · error · 1s
```

Same cause as on #12838: the TeXRA self-review action (`TEXRA_MODEL: deepseekproT`) fails at the provider call 2.3s in — **before reading any code** — and `Normalize review` / `Post TeXRA review` are skipped. `402 Insufficient Balance` is a billing state on the configured DeepSeek account; it is repo-wide, not branch-specific, and reproduces identically on a PR whose diff is 30 comment lines and on this one.

Unlike #12838 this PR *does* change code, so I checked the log rather than assuming — the failure is at the same provider call, with no code read and no findings emitted.

**No fix to port.** It needs the DeepSeek account topped up, or `TEXRA_MODEL` pointed at a funded route, in repo configuration. Outside this PR's scope and outside what a push can change. Separately, `chatgpt-codex-connector[bot]` has reported its own usage limits exhausted, so both review services are currently down for credit reasons.

**Re-run:** not spent, deliberately — `402 Insufficient Balance` is deterministic, so a re-run costs a CI cycle for no new information. Say the word if you want it on the record.

**The rest of CI is what matters for this diff**, and it was still in progress when this check failed: `static checks`, `build artifacts`, `kernel tests` (both shards), `webview smoke`, `claude-review`. Those cover the ratchets and the typecheck this refactor could plausibly break — if any goes red, that one **is** mine and I'll fix it rather than explain it.

---

PR #12840, opened 2026-09-19
refactor(agent): collapse model-binding switches into one table

LionSR:
## Summary

`src/agent/runtime/run/modelBinding.ts` encoded provider capability *facts* as three parallel imperative switches over the same `Protocol` union — `configurationFor` (12 arms), `constructModel`, and `backgroundCapable` — so adding a provider meant editing three places, and the boilerplate (`temperature: supportsTemperature ? input.temperature : null`, `parallelToolCalls`) was copy-pasted across ~10 arms.

All three now read from one table:

```ts
type ConfigurationOf<P extends HttpProtocol> = Extract<HttpConfiguration, { protocol: P }>;
type BackgroundRule<P extends HttpProtocol> = false | ((configuration: ConfigurationOf<P>) => boolean);

interface ProtocolDescriptor<P extends HttpProtocol> {
  readonly configure: (facts: BindingFacts) => ConfigurationOf<P>;
  readonly construct: (configuration: ConfigurationOf<P>, credential: RouteCredential) => Model;
  readonly background: BackgroundRule<P>;
}

const PROTOCOL_DESCRIPTORS: { readonly [P in HttpProtocol]: ProtocolDescriptor<P> } = { … };
```

`BindingFacts` is the shared base computed once per bind (`config`, `capabilities`, `credential`, `input`, `base`, `controls`, `supportsTemperature`, `effort`, `supportedEfforts`, `thinkingMode`). `configurationFor`'s switch becomes a 10-line facts builder plus one lookup; 7 arms spread `...controls`, and the 4 whose route takes only part of it name their fields. The 7 OpenAI-compatible chat protocols share one file-local `chatModel` helper, so the factory is named once rather than seven times.

**Read the net accounting first — this is net +149 LOC, not a reduction.** It's offered on change-amplification and type-safety grounds, not line count.

### A latent bug this removes

`backgroundCapable` had a `default: return false`. That is precisely the "`default: return` that quietly drops an unknown case" CLAUDE.md names under *silent degradation is a defect*: a newly added protocol would have silently reported "no background support" instead of failing loudly. Every protocol now states its background stance explicitly, and **zero `default:` arms remain in the file**.

### Exhaustiveness gets stronger, not weaker

The mapped key set `{ readonly [P in HttpProtocol]: ProtocolDescriptor<P> }` requires *every* protocol — a new `ModelConfiguration` arm left out fails to compile — and each entry's `configure` return is excess-property-checked against that protocol's own shape, where the old switch only checked against the union. Both lookups are generic in `P` (the TS correlated-union pattern), so the table entry and the configuration handed to it are provably the same arm. **Zero type assertions were used.** The generic signatures are load-bearing rather than decorative: the non-generic form does not compile (it collapses to `never`), which was confirmed with throwaway `tsc` probes before committing.

### Quirks preserved verbatim

`anthropicThinking`, `kimiThinkingControl`, `KIMI_FIXED_TEMPERATURES` / `kimiTemperatureByThinking`, `CODEX_ALLOWED_EFFORTS` (including the Codex effort clamp to MEDIUM), `AMBIGUOUS_MOONSHOT_FULL_NAMES`, `responsesAuthentication`, `responsesWebSocketSelected`, the GPT-5 `summary` toggle, glm's `Math.min(1, …)`, dashscope's `Math.min(1.99, …)`, minimax's raw temperature, xai's MAX-effort filter and `xaiEffort`, openrouter's `requestedModel` override. `responsesAuthentication` and its type moved earlier in the file (needed above the table); their bodies are byte-identical.

## Issue acceptance gates

No issue — this came out of a structural debt audit of `src/agent/`. No gate text to satisfy.

## Single source of truth

`PROTOCOL_DESCRIPTORS` now owns **the per-protocol binding facts**: how a protocol's configuration is built, which package factory constructs its `Model`, and whether it supports background delivery. One edit site per provider instead of three.

Removed: three parallel switches over one union (12 + 11 + 11 arms), the `temperature`/`parallelToolCalls` boilerplate repeated across ~10 arms, and seven separate namings of the OpenAI-compatible chat factory.

Added: `ProtocolDescriptor` owns the invariant *"every protocol states all three of its binding facts, and they are checked against that protocol's own configuration shape."* Not a pass-through — the table entries contain the real per-provider logic, and the lookups replace the switches rather than wrapping them.

## Net elements (R6)

```
 src/agent/runtime/run/modelBinding.ts | 663 +++++++++++------------
 1 file changed, 406 insertions(+), 257 deletions(-)
```

- **Net LoC: +149** (1042 → 1191). **Honestly a line increase.** Imperative `case`/`return` pairs (2 lines) became declarative entries (~4 lines of object/arrow scaffolding plus 2–3 for `construct`/`background`), and ~95 lines are new type declarations and their doc comments plus per-arm destructuring parameter lists that Prettier expands one name per line. Two formulations were tried; this is the smaller (the first was +183).
- `^[+-]export` over the diff: **0.** No export added, removed, or changed. `backgroundDelivery` and `bindModel` keep their exact signatures; `configurationFor` / `constructModel` / `backgroundCapable` were and remain file-local.
- Classes: 0 / 0. Interfaces: **+2** (`BindingFacts`, `ProtocolDescriptor`). Type aliases: **+2** (`ConfigurationOf`, `BackgroundRule`). Const tables: **+1** (`PROTOCOL_DESCRIPTORS`). Functions: **+1** (`chatModel`, file-local, 11 callers via the table).
- New imports: one type-only, `ChatConfiguration` from `@llm/turn` (already exported; nothing under `packages/llm/` was modified).
- Tests added: **0** (behavior-preserving refactor, per the testing-discipline budget).

## Consumer counts (R8)

n/a — no emit path, export, or subscribe surface was deleted or re-routed. The diff changes zero exports, so every caller outside this file is untouched.

## Host impact

Extension: none directly — behaviour of model binding is unchanged for every protocol (see the differential sweep below).

Desktop: none directly, same reason. `typecheck:desktop` (4 tsconfigs) passes.

CLI: none directly, same reason.

## Validation

- `npm run typecheck` → exit 0, no `error TS` lines. All 7 sub-projects: `:workspace`, `:test-kernel`, `:agent`, `:llm`, `:cli`, `:trace-viewer`, `:desktop`.
- `npm run lint` → exit 0, silent
- `prettier --check src/agent/runtime/run/modelBinding.ts` → `All matched files use Prettier code style!`
- `npm run test:pure` → `172 passed (172)` files, `2068 passed | 1 skipped`
- `npm run test:changed` → `160 passed (160)` files, `1993 passed | 1 skipped`

The existing suite covering this module is `src/test-kernel/model/OpenRouterRouting.vitest.ts` (`describe('bindModel')`), included in both runs above. `npm test` (full tier) was not run — the box was contended by concurrent work.

### Differential verification (temporary harness, since removed)

Behaviour preservation was not argued, it was measured. A throwaway copy of the pre-refactor module was diffed against the final committed version:

- **331,584 configurations** compared: all 157 `MODEL_CONFIGS` entries × 11 HTTP protocols × 4 credential routes (api-key, openrouter, chatgpt-subscription, xai-subscription) × 8 masks over the three settings the arms read (`openaiParallelToolCalls`, `gpt5ReasoningSummary`, `useGoogleInteractionsServerState`) × 2 agent categories (which change `maxOutputTokens`) × 3 temperatures (0, 0.4, 1.5 — exercising the glm/dashscope clamps and the Kimi fixed table). Compared by `JSON.stringify` equality, so **key order is identical too**, not merely deep-equal. **Zero mismatches.**
- **13,816 factory selections** compared: old vs new `backgroundCapable` agreement, and old vs new `constructModel` producing an identical outcome string (`ok:<origin JSON>` or `err:<thrown message>`). This is a strong check on the factory mapping, because each package factory re-parses its configuration and throws if the protocol is not its own — a mis-mapped arm cannot pass silently. **Zero mismatches.**

Both harness files and the three temporary `export` keywords were deleted before committing; the diff is one file with zero export changes.

## Scheduled refactoring

One documented trade-off rather than a follow-up: **TypeScript does not excess-property-check a spread**, so if a field is later added to `BindingFacts.controls` it would leak into the 7 arms that spread `...controls` and be rejected at runtime by the package's `z.strictObject` parse instead of at compile time. This is commented on the `controls` field, and the arms whose route takes only part of it name their fields explicitly. The alternative — naming all three in every arm — costs ~20 lines and pushes more parameter lists multi-line.

If a reviewer wants the net LoC closer to neutral, the lever is the per-arm destructuring parameter lists (~60 Prettier-expanded lines); a `facts.`-prefixed body style recovers ~45 at the cost of noisier expressions. The destructured form was chosen because it keeps each arm's body textually identical to the old switch arm, which makes the diff auditable.

---


LionSR:

**Failing check:** `review` ([job 105873900819](https://github.com/LionSR/TeXRA/actions/runs/35434163733/job/105873900819)) on `214e109`.

```
review · 0s
review · error · 1s
##[error]Process completed with exit code 1.
…texra-action.run;outcome=failure;duration_ms=2512
…Normalize review;outcome=skipped
…Post TeXRA review;outcome=skipped
```

The TeXRA self-review action (`TEXRA_MODEL: deepseekproT`) fails at the provider call 2.5s in — **before reading any code** — so `Normalize review` and `Post TeXRA review` never run and no findings are produced. `402 Insufficient Balance` is a billing state on the configured DeepSeek account.

**Repo-wide, not branch-specific.** I read the log verbatim on four PRs opened from this session and the error is byte-identical on each: #12838 (a 30-line comment-only diff), #12839, this one, and #12842. A diff cannot produce or clear a payment error.

**No fix to port.** It needs the DeepSeek account topped up, or `TEXRA_MODEL` pointed at a funded route, in repo configuration — outside this PR's scope and outside what a push can change.

**Re-run:** deliberately not spent. A re-run distinguishes a flake from a real failure, and `402 Insufficient Balance` is deterministic; it would cost a CI cycle to reprint the same line. Happy to run it if you want it on the record.

**What does cover this diff:** `static checks` (typecheck + lint + ratchets) and `kernel tests` are the checks that matter for a refactor of `modelBinding.ts`, and behaviour here was additionally verified by a differential sweep of 331,584 configurations and 13,816 factory selections against the pre-refactor implementation (details in the PR body). If any of those CI checks go red, that one **is** mine and I'll fix it rather than explain it. `claude-review` and Codex have both completed on this commit with no findings.

---

PR #12842, opened 2026-09-19
refactor(llm): give the two Chat streams one delta accumulator

LionSR:
## Summary

`openaiChat.ts` and `openrouterChat.ts` each carried their own copy of the OpenAI-shaped delta-stream state machine and their own hand-rolled mapping from a provider receipt to canonical usage counts. Both are now single-sourced in a new `packages/llm/src/chatStream.ts`:

- **`chatDeltaAccumulator()`** — the state machine both files carried verbatim: the open `'reasoning' | 'text'` phase, coalescing of consecutive text/refusal deltas into one part, and tool calls assembled from fragments by reported index. Surface: `parts`, `closePhase()`, `absorbText({reasoning, text, refusal})`, `absorbToolCalls(fragments, changedIdentityMessage)`, `toolCalls()`.
- **`chatUsageCounts(receipt)`** — maps the documented OpenAI-compatible receipt fields onto canonical `{inputTokens, outputTokens, totalTokens, cachedInputTokens, reasoningTokens}`, replacing four hand-rolled call sites: `normalizeUsage`, `miniMaxUsage`, OpenRouter's `UsageSchema.transform`, and two identical "all nulls" literals.

## The larger extraction was attempted and declined — this is the important part of the review

This PR was scoped from an audit that claimed the two files implement "the same SSE chat-completion state machine, line-for-line in structure," and projected ~400–600 LoC (later ~1,000–1,500 across the provider layer) from one shared engine parameterized by a chunk schema plus a small hook table.

**That framing does not survive contact with the code.** Only two steps of the loop are genuinely common — the `[DONE]` sentinel and the JSON parse. Everything else diverges, and *in different order*:

- **`openaiChat`**: errors → minimax envelope → protocol-selected schema → glm `request_id` → strict identity-change check → emits `identified` inline. Fingerprint is last-wins with no conflict check; `finishReason` is set once and any later choice is an error; kimi choice-level usage; xai cost/`service_tier` receipt.
- **`openrouterChat`**: parses identity **first** with `??=` semantics, then errors, one schema, `service_tier` and fingerprint each with a conflict check, usage compared with `isDeepStrictEqual`, `finish_reason: 'error'` → rejection, an `observedIdentity`/`hasContent` guard, reasoning as a tri-state (`null` vs `undefined`), annotation dedup, `native_finish_reason`.
- Even `enrich` differs: `openaiChat` prefers `error.responseId ?? responseId`; `openrouterChat` overwrites unconditionally.

[...]

PR #12844, opened 2026-09-19
refactor(cli): give the static transcript ring one cost model

LionSR:
## Summary

`StaticConversationTranscript.tsx` (1333 LoC) re-implemented the ring's row/byte arithmetic in **five** places, each hand-maintaining its own "subtract the old margin, add the new one" bookkeeping, and carried the duplicate-row-id warning subsystem in **two** near-copies kept in sync by matching comments.

The pure logic now lives in `packages/cli/src/chat/tui/panes/staticTranscriptRing.ts`, and the component file is **1333 → 291 lines**.

### The model that made one accumulator possible

The list's row cost decomposes into **per-item intrinsic rows** (its lines plus the separator it declares *below* itself) plus a **seam** between each pair of neighbours (the lower item's declared top margin, collapsed against the upper item's bottom margin).

The key point: *a top margin is a property of the boundary, not of the item*, because collapse is what decides its cost. Under that model every mutation is one formula:

```
intrinsic(x) + seam(above, x) + seam(x, below) − seam(above, below)
```

```ts
class StaticTranscriptTotalsAccumulator {
  get rows(): number
  get bytes(): number
  get totals(): StaticTranscriptTotals            // fresh object each call
  insert(aboveMarginBottomRows: number, item: Metrics, below: Metrics | undefined): void
  remove(aboveMarginBottomRows: number, item: Metrics, below: Metrics | undefined): void
  private rowDelta(…)                              // the one collapse-aware formula
}
```

`below === undefined` is an append. `above` is the previous item's **bottom margin only** (a `number`, via the cheap `itemMarginBottomRows`) — that is all collapse consumes from above, and taking it as a number is what preserves the "lay each item out at most once" guarantee in the backward retain walk.

**All five sites use it; none was left behind:**

1. `staticTranscriptItemsTotals` — forward `insert` loop.
2. `trimStaticTranscriptItems` — `remove` per pass. The item above the trim point is provably invariant (only the oldest non-header item is ever dropped), so its margin hoists out of the loop: **2 layouts per pass instead of 3**.
3. `retainedStaticTranscriptTail` — the backward walk is now literally "insert the candidate between the header and the current first": one `insert` replaces the hand-rolled `oldFirstBefore`/`candidateMetrics`/`oldFirstAfter` re-derivation, still one layout per item.
4. `ensureStaticSessionHeader` — one `insert` replaces the `old`/`newFirstMetrics` pair: **2 layouts instead of 3**.
5. `advanceStaticTranscriptState`'s `appendItem` — one `insert`.

Side effects of the seam model: `staticTranscriptItemBaseMetrics` and `staticTranscriptItemMetricsForPrevious` collapsed into one `staticTranscriptItemMetrics`, and the redundant `rows` field left the metrics interface (`base.rows === intrinsic + declaredTop` in every case), dropping the `transcriptEntryLayoutRows` dependency.

### The two duplicate-row copies were *not* identical

Worth stating plainly, since the audit assumed they were. They differed in two ways: `seen` seeding (rebuild started empty; the incremental path seeded from `nextItems`), and `alreadyMarked` (only the incremental path scanned for existing markers).

Rather than flatten that with a boolean flag, both now derive from one parameter — the list being appended onto — in `scanDuplicateRowIds(entries, existing)`. The incremental path passes `nextItems`; the rebuild passes `[]`, because it starts from nothing but the session header, which is not a row and carries no row id. That makes the `alreadyMarked` scan a genuine no-op on a rebuild instead of a suppressed difference. It deliberately does **not** seed the rebuild's taken ids from the header item, which would have changed behaviour for the pathological case of an entry whose wire id is literally `'session-header'`. Marker construction and its tail-placement rationale now live once in `duplicateRowWarningItems`.

## Issue acceptance gates

No issue — from a structural debt audit of the CLI area. Both of that audit's candidates for this file (the 5-way metric duplication and the 2-way duplicate-row duplication) are addressed here.

## Single source of truth

`staticTranscriptRing.ts` now owns **the ring's cost model** — one collapse-aware `rowDelta` that every mutation goes through — and **duplicate-row-id detection and marker construction**.

Removed: five independent implementations of the collapse arithmetic (→ 1), two copies of the duplicate-row subsystem (→ 1), and two near-identical metrics helpers (→ 1).

`StaticTranscriptTotalsAccumulator` clears the factory/class bar on captured mutable state plus multiple callers (5 sites). It is not a pass-through: it owns the collapse formula that was previously restated at each site.

One extraction was **reverted during the work**: an `itemIntrinsicRows` helper had a single caller, and the repo bans single-caller extractions, so it was inlined.

## Net elements (R6)

```
 config/ratchets/knip-baseline.json                               |   54 +-
 packages/cli/src/chat/tui/panes/StaticConversationTranscript.tsx | 1066 +------
 packages/cli/src/chat/tui/panes/staticTranscriptRing.ts          | 1076 ++++++++
 src/test-kernel/cli/ConversationTranscript.vitest.ts             |    2 +-
 4 files changed, 1104 insertions(+), 1094 deletions(-)
```

- **Net LoC: +10** total (1333 → 291 + 1076 = 1367). Honest breakdown: **executable lines 1117 → 1086 (−31)**; comment lines **+~40** (new module header, the seam/accumulator model documentation, the merged duplicate-scan docstring). The component file is **1333 → 291**, so "roughly halve it" landed at 4.6×.
- `^[+-]export`: **+11 / −8.** All 8 removed reappear at the new path. The 3 genuinely new ones (`StaticTranscriptState`, `ChildHeader`, `StaticScrollbackSource`) are types the component consumes in this same commit — previously file-local because the component lived in the same file. No new export lacks a consumer.
- Classes: **+1 / −0** (`StaticTranscriptTotalsAccumulator`). Interfaces: **+1 / −0** (`DuplicateRowScan`); the other 9 moved unchanged.
- Free functions: **24 → 25** (ring 19 + component 6). Added `itemMarginBottomRows`, `seamRows`, `scanDuplicateRowIds`, `duplicateRowWarningItems`; deleted `staticTranscriptItemBaseMetrics`, `staticTranscriptItemMetricsForPrevious`, `duplicateRowIdWarningRow`.
- **Knip baseline: 170 → 166 — it shrinks.** 8 entries relocate to the new file path (same names, same count — not a widening), then **4 shrink away** because knip reports them genuinely resolved: `advanceStaticTranscriptState`, `buildStaticTranscriptState`, `StaticTranscriptItem`, `StaticTranscriptRingBudgets` now have a production consumer. The 4 remaining are test-only exports by design (`buildStaticTranscriptItems`, `trimStaticTranscriptItems`, `sessionHeaderIdentityLine`, `DEFAULT_STATIC_TRANSCRIPT_RING_BUDGETS`).
- Tests added: **0** (behavior-preserving refactor). One test file changed by **1 line** — its import path.

## Consumer counts (R8)

No emit path or event surface touched. The ring module has 2 consumers (the component and the existing kernel suite). The 4 re-baselined-away exports each gained a production consumer, which is why the ratchet count dropped rather than moved. The 4 still baselined are test-only and documented as such.

## Host impact

Extension: none — no file under `packages/extension/` touched; the module is CLI-only.

Desktop: none — no file under `packages/desktop/` touched.

CLI: internal only, and **terminal output is byte-identical**. The `<Static>` item sequence, `repaintEpoch`, and the ring's trim/retain decisions are unchanged (differentially verified below), so native scrollback, the resize repaint path, and headless `texra run` / `--print` are unaffected. Marginally *fewer* layout passes per trim pass and per header insert.

### ⚠️ A sandbox hazard worth knowing about

`npm run typecheck` and `npm run lint` can be **OOM-killed while the npm wrapper still reports exit 0** — they print `Killed` but the script returns success. **That is a false pass.** Every stage was therefore run separately with its own exit code checked.

**typecheck — all 7 stages, each exit 0:** `typecheck:workspace`, `:test-kernel`, `:cli` (both tsconfigs), `:agent` (`Validated 664 declarations and 58 external packages`), `:llm`, `:trace-viewer`, `:desktop`.

**lint — all 8 eslint invocations of the `lint` script, each exit 0, clean:** `packages/cli/src`, `src/test-kernel`, `src packages/agent/…`, `packages/extension/src packages/desktop/src`, `packages/cli/scripts`, `packages/llm/src`, `packages/trace-viewer/src`, desktop harness/e2e.

- `npm run test:pure` → `172 passed (172)` files, `2068 passed | 1 skipped`, 25.40s, exit 0
- **Directly relevant kernel suites** — `ConversationTranscript.vitest.ts` (ring/trim/margin-collapse) + `StaticBandResize.vitest.ts` (Ink resize/repaint) → `2 passed (2)` files, **`55 passed (55)`**, exit 0
- `npm run test:changed` → `15 passed (15)` files, `259 passed | 1 skipped`, exit 0
- `npm run check:dead-code-ratchet` → `166 combined vs 166 baselined`, `Dead-code ratchet OK: no new findings.`, exit 0
- `npx prettier --check` on all 4 changed files → `All matched files use Prettier code style!`

`npm test` (full kernel tier) was not run locally; CI's `static checks` and `kernel tests` jobs are the authoritative gate and run on this PR.

### Differential evidence

A throwaway harness (deleted before the commit; not in the diff) compared the pre-refactor module from `HEAD` against the new one over **40 seeded pseudo-random transcripts × 3 widths × 5 budget sets × 3 `maxRows`**, across cold builds, incremental tick sequences (advancing one row at a time), width-change reflows (80→30→200→55), and `trimStaticTranscriptItems` called directly. It asserted identical item id sequences (marker ids normalized to their source row, since each module has its own random namespace), `rowCount`, `byteCount`, `trimmed`, and `repaintEpoch`. A guard test confirmed the scenarios genuinely produced duplicate markers (>10), trims (>10) and non-trims (>10) rather than passing vacuously. **5/5 passed.**

It was not kept: testing discipline says zero new tests for a behavior-preserving refactor, and it depends on a copy of the deleted implementation.

---


LionSR:

**Failing check:** `review` ([job 105874285033](https://github.com/LionSR/TeXRA/actions/runs/35434312333/job/105874285033)) on `6c5c319`.

```
review · 0s
review · error · 1s
##[error]Process completed with exit code 1.
…texra-action.run;outcome=failure;duration_ms=2545
…Normalize review;outcome=skipped
…Post TeXRA review;outcome=skipped
```

The TeXRA self-review action (`TEXRA_MODEL: deepseekproT`) fails at the provider call 2.5s in — **before reading any code** — so `Normalize review` and `Post TeXRA review` never run and no findings are produced. `402 Insufficient Balance` is a billing state on the configured DeepSeek account.

**Repo-wide, not branch-specific.** I read the log verbatim on all five PRs opened from this session and the error is byte-identical on each: #12838 (a 30-line comment-only diff), #12839, #12840, #12842, and this one. A diff cannot produce or clear a payment error.

**No fix to port.** It needs the DeepSeek account topped up, or `TEXRA_MODEL` pointed at a funded route, in repo configuration — outside this PR's scope and outside what a push can change.

**Re-run:** deliberately not spent, since `402 Insufficient Balance` is deterministic and a re-run would cost a CI cycle for no new information. Say the word if you want it on the record.

**What does cover this diff:** `static checks` and `kernel tests`. This refactor changes what the terminal repaints, so the checks that matter are the two kernel suites pinning that behaviour — `ConversationTranscript.vitest.ts` (ring/trim/margin-collapse) and `StaticBandResize.vitest.ts` (Ink resize/repaint) — which ran **55 passed** locally, alongside a differential sweep over 40 transcripts × 3 widths × 5 budget sets × 3 `maxRows` confirming identical item sequences, `rowCount`, `byteCount`, `trimmed` and `repaintEpoch` (details in the PR body). Note the PR body also discloses that `npm run typecheck`/`lint` can be OOM-killed while still reporting exit 0 in a contended sandbox, which is why every stage was run separately here — CI's `static checks` is the authoritative confirmation. If any CI check goes red, that one **is** mine and I'll fix it rather than explain it.

---

PR #12852, opened 2026-09-19
refactor(cli): one resume-advertise predicate and one presentation claim in the headless exit path

LionSR:
Headless `texra run` carried three spellings of one question, a presentation
claim attributed to the wrong owner, and two guards that no path can reach.

## Ruling taken

The owner accepted the finalization-attribution fix. A headless run that both
fails and hits a shutdown-finalization failure now prints the classified
`AgentError` message as well; today the finalization toast silently suppresses
it. This is the §15 silent-degradation rule applied, and it is a visible
stderr change: such a run gains one extra stderr line (the run's own
classified message) beside the finalization notice it already printed. No
other path changes what it writes, and no exit code moves.

### Why the suppression happened

`reportFinalizationFailure` emitted through `session.interactions.emit`. The
headless attachment's `emit` wrapper in `executeCliRequest` sets
`failurePresented = true` on any `requestShowError`, and `failurePresented` is
the flag that decides whether the classified `AgentError` still needs its
direct fallback. A shutdown-finalization notice therefore claimed the run's
failure presentation. It now presents on `presentationHost` directly — the
same sink the attachment forwards to, so the rendered record is byte-identical
— leaving `failurePresented` to track the run's own failure alone. That also
removes a `runtime.runSync` boundary call and an Effect→port→Effect round trip
for a notice that never left this host.

## Net elements (R6)

4 files, +124 / −47 (net +77 LoC), net −4 production elements. Production is
3 files, +54 / −45 (net +9 LoC); the rest is the one regression test below.

Gone:
- 3 inline spellings of `kind === 'checkpoint' && (canAdvertiseInterruptedRun(…) ?? true)` (`executeCli.ts` drain, `executeCli.ts` BEFORE-phase handler, `workflow.ts` `maybeAdvertiseRecovery`) → one `advertisesInterruptedRun`.
- 2 spellings of `Extract<ResumabilityDecision, { kind: 'checkpoint' }>` → one `ResumableCheckpoint`.
- `presentationAttached` (local `let` + its two guard lines). The early detach at the `!runResult.ok || Result.isFailure(finalization)` site and the `Effect.ensuring` detach on the success tail are mutually exclusive: reaching the success tail requires `runResult.ok` **and** a successful `finalization` (a failed one pushes to `cleanupFailures`, which throws before the tail), which is exactly the negation of the early-detach condition.
- `finalizationCompleted` (local `let` + its assignment), a mirror of `Result.isFailure(finalization)` read once.
- `resumableCheckpoint` (local, plus its restated `Extract<…>` annotation) → a `boolean` decided where the inspection lands.
- one `options.runtime.runSync` Effect boundary call and the `session.interactions` hop under it.

Added: `advertisesInterruptedRun` (1 function) and `ResumableCheckpoint`
(1 type) in `packages/cli/src/runtime/interruptedResumeHint.ts`, the module
that already owns the resume-hint vocabulary and that both call sites already
import. It lives there rather than in `executeCli.ts` because
`WorkflowRunCommand.vitest.ts` stubs `@cli/runtime/executeCli` wholesale
(no `importOriginal`), so an export added there is invisible to `workflow.ts`
at test time — the mock-path hazard, caught by the suite.

The two new lines of net LoC are the shared helper's docstring; the element
count drops, which is the accounting that applies to a ≥3-caller duplication
collapse.

## Consumer counts (R8)

No public symbol was deleted or retyped; the two additions are new.

- `advertisesInterruptedRun` — 3 call sites (`executeCli.ts:393`, `executeCli.ts:477`, `workflow.ts:358`), 2 importers.
- `ResumableCheckpoint` — 2 use sites (`executeCli.ts:100` in `CliExecuteOptions.canAdvertiseInterruptedRun`, `workflow.ts:324`), 2 importers.
- `CliExecuteOptions.canAdvertiseInterruptedRun` keeps its signature (the parameter type is the same type under a name), so its one producer (`workflow.ts:373`) is untouched.
- `reportFinalizationFailure` is module-local, 2 call sites, signature unchanged.
- `@agent/storage` is already a baselined `cli` deep-import specifier, so the new type-only import adds no edge (`config/ratchets/host-agent-import-baseline.json` is set-based per package).

## Verified

Run in the worktree at `afd5bd474b` (rebased head):

- `npm run typecheck` — pass (exit 0).
- `npx eslint packages/cli/src/runtime/executeCli.ts packages/cli/src/commands/workflow.ts` — clean (exit 0).
- `npx vitest run --project kernel src/test-kernel/cli/{ExecuteCli,WorkflowRunCommand,ToolUseRunCommand,ResumeCommand}.vitest.ts` — 91 passed. (First attempt failed 15/91 on the `executeCli` mock above; that is what moved the helper to `interruptedResumeHint.ts`, after which the same 91 pass unchanged — no test was edited.)
- `npm run test:pure` — 173 files, 2075 passed, 1 skipped. Covers the architecture ratchets, including `hostAgentDeepImportRatchet`.
- `npm run test:changed` — 14 files, 257 passed, 1 skipped.
- `npm run check:unexecuted-effects` — OK, 2005 files.
- `node scripts/check-effect-migration-ratchet.mjs` — OK, no count grew, no headroom (`catch:effect-importer` still 2 files / 2 sites).
- `npm run check:dead-code-ratchet` — OK, 166 vs 166 baselined; the two new exports both have consumers.
- `corepack pnpm --filter @texra-ai/cli validate:run` — "CLI run validation passed" (builds the validator bundle from this source).

Rebased onto `f2d5b08bf3` (clean, no conflicts) and every command above re-run
green on the rebased head, plus `npx vitest run --project kernel
src/test-kernel/cli/` — 70 files, 1008 passed, 1 skipped.

Nothing red. `npm test` was not run.

### Regression test

`ExecuteCli.vitest.ts` gains one test, "still presents a classified run failure
when shutdown finalization also failed", plus a `reject` beside
`stubHangingRun`'s `resolve` (existing callers untouched). It drives a run that
fails with a classified `AgentError` while `finalizeRun` reports a
finalization failure under the lease, and asserts both `requestShowError`
records.

The suite could not see this distinction before: `stubExecuteCliDeps` replaces
`createHeadlessCliHostInteractions` with `{ emit: mocks.emit }`, dropping the
production wrapper that sets `failurePresented`. The test installs that wrapper
for its own case with `mockImplementationOnce`, leaving the other 34 tests on
the default stub.

Verified adversarially: with `reportFinalizationFailure` reverted to
`session.interactions.emit`, the test fails with "expected vi.fn() to be called
2 times, but got 1 times" — the suppression this PR removes.

## Review findings

**TeXRA review, Finding 1 (drain fail-open) — respectfully disputed; no code
change.** The review reads `Effect.try({ try, catch })` as returning the mapped
error as the success value, so that `if (advertise)` sees a truthy `Error` and
advertises fail-open. It does not: `catch` maps a thrown defect into the
*failure* channel, so `yield*` aborts the `drain` generator and the failure is
caught by the outer `drain.pipe(Effect.catch(…))`, which writes it to
`launchVerdict.artifactFailure` exactly as the PR body said. Either way this is
pre-existing behavior the diff does not touch, so nothing changes here.

**TeXRA review, Finding 2 (no regression test) — addressed**, see the test
above.

**Claude review, note (label).** The `tech-debt` label was applied
automatically. The body names the one visible stderr change, which the owner
ruled on before this landed; retitling the label is the maintainer's call.

## Left out

- **`finalizeRun`'s `report` port itself.** Left as a plain callback: it is the
  lifecycle's own port shape, not a CLI round trip.
- **Folding the two `deriveResumability` probes into one.** Refuted: the drain
  probe (`executeCli.ts:381`) reads *after* `releaseRunLease`, the BEFORE-phase
  probe (`executeCli.ts:471`) reads before the drain runs at all, and
  `workflow.ts:356` reads after the run settled. Three questions at three
  instants; both Deferreds, `interruptedRunId` and `resumeHintWritten` all stay.
- **`Effect.try` around the refinement at `executeCli.ts:392`, but not at
  `executeCli.ts:477`.** Deliberately preserved as-is: the drain routes a
  throwing refinement into `launchVerdict.artifactFailure`, the shutdown
  handler lets it surface as a defect. `advertisesInterruptedRun` is a plain
  predicate precisely so each caller keeps its own answer to that; unifying
  them would be a behavior change, not a simplification.
- **`failurePresented` and the `emit` wrapper.** Kept, per the lane shape: they
  cover the pre-registration window where no `result` event exists.


LionSR:
Both reviews addressed, pushed as `afd5bd474b` (rebased onto `f2d5b08bf3`).

**Finding 2 (no regression test) — addressed.** `ExecuteCli.vitest.ts` gains one test that drives a run failing with a classified `AgentError` while `finalizeRun` reports a finalization failure under the lease, and asserts both `requestShowError` records. As the review noted, the suite could not see this: `stubExecuteCliDeps` replaces `createHeadlessCliHostInteractions` with `{ emit: mocks.emit }`, dropping the wrapper that sets `failurePresented`. The test installs the real wrapper for its own case via `mockImplementationOnce`, leaving the other 34 tests on the default stub. `stubHangingRun` gains a `reject` beside its `resolve`; existing callers are untouched.

Verified adversarially: reverting `reportFinalizationFailure` to `session.interactions.emit` fails it with `expected "vi.fn()" to be called 2 times, but got 1 times` — exactly the suppression this PR removes.

**Finding 1 (drain fail-open) — respectfully disputed, no code change.** `Effect.try({ try, catch })` maps a thrown defect into the **failure** channel, not the success channel. So a throwing refinement does not become a truthy `advertise` value: the `yield*` aborts the `drain` generator, and the outer `drain.pipe(Effect.catch(…))` writes it to `launchVerdict.artifactFailure`, as the PR body described. (Either reading is pre-existing behavior the diff does not touch — the helper is deliberately a plain predicate so the drain keeps its `Effect.try` and the BEFORE-phase handler keeps its unguarded call.)

PR #12859, opened 2026-09-19
refactor: collapse the empty gates and the unbuilt webview tree

LionSR:
Five independent gate/packaging cuts from the wave-8 find-simplification survey, one commit each. No production module changes behavior; every gate this touches still fails on the thing it was written to catch.

**Owner ruling taken (stated per the churn-class rule, checklist §14 R5):** the webview tree move proceeds, and its reason is the deleted config surface — the `@webview/*` alias, two VS Code-free-zone entries, two host-layer import prefixes, one effect-ratchet boundary prefix and its stale self-test row — not tidiness. The ratchet collapse keeps the gate and regenerates nothing.

### What landed

1. **S19 — `@shared/schemas` deep-import ratchet (834 → 116 lines).** The baseline's four maps (`surface`, `floors`, `forced`, `gratuitous`) are all `{}`, so a published-surface reconstruction of the barrel, a three-class classifier and a regeneration writer sat between the scan and an assertion that the scan finds nothing. The suite now walks the same roots (repo-root `src/` plus every `packages/*/src`, excluding the surface interior `src/shared/schemas/`, test-kernel included) with the shared `collectModuleSpecifiers` and fails on any `@shared/schemas/<leaf>` specifier. `config/ratchets/shared-schemas-deep-import-baseline.json` is deleted; there is nothing left to regenerate.
2. **S11 — the unbuilt webview tree folds into the progress view.** `packages/extension/src/webview/` had no `index.ts`, so vite (`src/<name>/frontend/index.ts`) never built it as a webview; its only importer is the progress view. Ten components move to `progressView/frontend/components/`, `fileDropHandler.ts` and `fileSelectStyles.ts` to `progressView/frontend/`, and the alias imports become relative. The two remaining view trees keep the identical `frontend/{components,styles,formatters}` shape the skill asks to keep aligned.
3. **N26 — `contributes.chatSkills` is generated, not verified.** `sync-package-contributes.mjs` now derives it from `resources/skills/<name>/SKILL.md`, next to the other catalog-derived contributions and behind the same `--check` CI drift gate; `verifyBundledSkills` (27 lines) and its call are deleted. Generated output is byte-identical to the committed manifest.
4. **N27 — `verifyConfigDeps`'s `strict` parameter.** Its one caller passes `true`, so the lenient half of every branch was unreachable and three byte-identical parse-and-skip blocks differed only in which parser they called. Pick the parser, then parse once; skip messages unchanged.
5. **ci.yml TUI smoke selector.** The 10 scenario names were spelled out in the workflow, where renaming a scenario in `validate-tui.mjs` silently drops it from the gate. They now carry `smoke: true` at their definition and `--smoke` selects them; `--smoke` with no tagged scenario fails rather than running nothing.

## Net elements (R6)

- Files: 30 changed (12 moved, 1 deleted, 1 test suite moved), including the review-fix commit 303dbe9.
- `git diff --stat origin/main`: **149 insertions, 890 deletions — net −741 LoC.**
- Exports: `+0 / -0`. No module's public surface changes; the deleted symbols were all file-local to the ratchet suite or to `verify-extension-package-invariants.mjs`.
- Declarations: −11 (`SchemasBaseline`, `PublishedSurface`, `PublishedLeafSurface`, `ModuleExport`, `ModuleExports`, `MutablePublishedSurface`, `DeepImports`, `ExportSpace`, `ImportBinding`, `DeepImportReference`, `CollectedSchemas`), +1 (`Scan`). Functions: −20 in the ratchet suite, −2 script functions (`verifyBundledSkills`, and `verifyConfigDeps`'s inner `unparsable`).
- Config surface deleted: 1 tsconfig path alias, 2 VS Code-free-zone entries, 2 host-layer import prefixes, 1 lint glob arm, 1 effect-ratchet boundary exclusion + 1 stale self-test row, 1 ratchet baseline file.

## Consumer counts (R8)

- `@webview/*` alias: 9 import statements in 3 progress-view files + 1 test-kernel suite; 0 elsewhere (`rg '@webview'` across the repo, excluding `.agents/` history). All 10 rewritten; alias deleted.
- `packages/extension/src/webview/**`: 12 files, 0 importers outside `progressView/frontend/` and the one suite. No `index.ts`, so 0 vite entry points referenced it.
- `config/ratchets/shared-schemas-deep-import-baseline.json`: 1 reader (the suite), 1 writer (the same suite's regen path), 1 prose mention in CLAUDE.md. No script, knip config or CI step reads it.
- Live `@shared/schemas/<leaf>` imports in the scanned roots: **0** (the only two in the tree are `packages/cli/scripts/tui-harness.tsx:60-61`, outside `packages/*/src`; the seven other `rg` hits are doc comments). The remaining hits are all in `.agents/docs/archived/`, left as historical record.
- `verifyBundledSkills`: 1 caller (same file). `collectRelativeFiles` keeps its other consumer (`scripts/verify-vsix-contents.mjs:87`), so the shared helper stays.
- `verifyConfigDeps`: 1 caller (`format-staged.mjs:441`), passing `strict: true`.
- `validate-tui.mjs` scenario names in ci.yml: 10, now 0.

## Verified

- `npm run typecheck` — green (all workspace projects).
- `npm run test:pure` — 173 files / 2070 passed, 1 skipped. The architecture suites are the point here: `dependencyDirection` (24 tests) and the rewritten `sharedSchemasDeepImportRatchet` both green.
- Negative test of the collapsed gate: dropping a throwaway `import type { AgentCategory } from '@shared/schemas/agent'` into `src/utils/` fails the suite with `src/utils/__ratchetprobe.ts: @shared/schemas/agent`. Removed again.
- `npx vitest run src/test-kernel/progressView src/test-kernel/scripts` — 27 files / 195 passed, 1 skipped (includes the moved `fileDropHandler.vitest.ts` and `FormatStaged.vitest.ts`).
- `node scripts/check-effect-migration-ratchet.mjs` — OK over 1406 production files, no count grew, no baseline headroom.
- `npm run check:dead-code-ratchet` — OK, 166 vs 166 baselined, no new findings (knip-baseline untouched).
- `npm run check:guidance-refs` — OK over 61 files. It caught one stale path (`docs/dev/verification.md:5`), fixed in the same commit as the move.
- `node scripts/sync-package-contributes.mjs --check` — in sync, i.e. the generated `chatSkills` is byte-identical to the committed manifest. Adding a probe skill directory makes it fail, as intended. `node scripts/verify-extension-package-invariants.mjs` — invariants hold.
- `npm run compile:fast` — green; `dist/progressView/bundle.js` still contains the moved components.
- `node scripts/validate-tui.mjs --smoke` (in `packages/cli`) — all 10 scenarios pass locally, and `--smoke --list-selected` prints exactly the 10 names ci.yml used to spell out.
- `npx eslint` on the touched TypeScript files — clean. The two touched `scripts/*.mjs` files report pre-existing `no-undef` on `console`/`process`; `scripts/` is not in `npm run lint`'s target list, and untouched scripts report the same.
- Not run: the full `npm test` (left to CI).

### Review round 1 (303dbe9)

Both findings addressed: three stale `webview` claims removed from AGENTS.md (DOM suite list, browser-safe import zones, the "shared Lit component library" bullet), and the ratchet self-test renamed to `detects a deep import in every scanned module-loading form` with a comment naming the three uncovered forms. The `--smoke <name>` double-queue edge both reviewers called non-blocking is left as-is: explicit positionals deliberately keep order and multiplicity here, and `ci.yml` passes `--smoke` alone.

## Left out

- The three exotic module-loading forms the old ratchet scanned for — `module.require(…)`, `require.resolve(…)`, `import.meta.resolve(…)` — are no longer detected, because the collapsed suite rides the shared `collectModuleSpecifiers` (`src/test-kernel/support/repoScan.ts:153`) rather than keeping a private scanner. Static/dynamic `import`, `export … from`, `import x = require(…)`, `require(…)` and `typeof import(…)` positions are all still caught, and a self-test pins that. Re-adding a second scanner to cover three forms that appear zero times in the tree would put the machinery straight back.
- `src/test-kernel/webview/` held exactly one suite; it moved to `src/test-kernel/progressView/` with its module rather than leaving an orphan directory. That is inside the ruling's scope, not an extra.
- `.agents/docs/archived/process/2026-08-01-directory-organization.md` still names the deleted baseline and the old webview paths. It is an archived record of what was done at the time, so it stays as written.


LionSR:
Both review findings are addressed in 303dbe9:

- **Finding 1 (stale AGENTS.md claims).** Dropped `webview` from the DOM test-kernel suite list (AGENTS.md:138), removed `webview/frontend/` from the browser-safe import zone sentence (:261), and deleted the bullet describing `webview/frontend` as the shared Lit component library progressView composes (:649-651) — the bullet one entry up already states that those components are `progressView/frontend/components/`. `npm run check:guidance-refs` stays green; as you noted, it could not catch these because they are prefix-less backticked names.
- **Finding 2 (self-test name).** Renamed to `detects a deep import in every scanned module-loading form` and added a comment naming the three uncovered forms (`module.require(…)`, `require.resolve(…)`, `import.meta.resolve(…)`) and why keeping a private scanner for them would put back the machinery this ratchet just shed.

**Residual risk left as-is.** `--smoke <name>` unioning the smoke set with an explicitly named scenario can queue that scenario twice. Explicit positionals deliberately retain order and multiplicity in this script, `ci.yml` passes `--smoke` alone, and both reviewers classed it non-blocking; adding dedupe logic for an unreachable edge is the kind of guard this PR is removing elsewhere.

Re-verified after the fix: `npx vitest run --project pure src/test-kernel/architecture/sharedSchemasDeepImportRatchet.vitest.ts` (2 passed), `npm run check:guidance-refs` (OK, 61 files), pre-push lint on the changed files.

PR #12864, opened 2026-09-19
refactor: burn down eighteen test-only exports and shrink the knip baseline to 158 (Phase B of #12564)

LionSR:
Phase B of the test-only export burndown (#12564, after #12610). The export
keyword is a contract; each symbol here had no counterparty outside its own
module, and the suites that imported them now exercise the same behaviour
through the entry point production calls.

Owner ruling taken and stated: the three texTools kpathsea helpers lose their
dedicated suite per AGENTS.md testing discipline. Their behaviour is now
exercised through `compileLatex2Pdf`'s search-path path, by one assertion in
the existing success test on the `TEXINPUTS` the engine is handed
(`${documentDir}:.:${workspaceRoot}:`, so the document dir still outranks the
compiler's cwd). No new test file was added anywhere in this PR.

Said plainly, per the same ruling: **the auth error vocabulary is left
half-exported.** `DeviceAuthorizationPending`, `DeviceCodeTimedOut`,
`OAuthNetworkError` and `OAuthUnexpectedResponse` stay exported while
`SessionCompletionFailed`, `DeviceAuthorizationDenied` and `DeviceCodeExpired`
were meant to be demoted here. They are not, because the demotion does not
compile: see `## Left out`.

## Net elements (R6)

- Files: 20 changed (9 production modules, 6 suites, 1 ratchet baseline, plus
  the four palette/logs/arxiv/texTools files counted above).
- Exports: **-18, +0** (`git diff origin/main -- '*.ts' | grep -cE '^[+-]export'`).
  Removed: `filterCommandPaletteEntries`, `getNextCommandPaletteIndex`,
  `executeCommandPaletteEntry`, `parseDesktopLogEntries`,
  `resolveStateSettingWrite`, `PublicRunResult`, `RegisterRunOptions`,
  `ChildTurnState`, `SessionBridgeClosedError`, `FlowPosition`,
  `ClaimStanding`, `TranscriptRowContext`, `DownloadSourceOptions`,
  `resolveArxivPaperDirectoryRelative` (deleted outright, inlined at its one
  caller), `buildKpathseaSearchPath`, `buildLatexInputEnv`,
  `buildLatexSearchParts`, `pastedImageFileName`.
- Declarations: 7 `interface`/`class`/`type` declarations demoted to
  file-local, 1 function deleted, 0 added.
- Two parameters deleted: with the kpathsea helpers private,
  `buildLatexInputEnv`'s `env: NodeJS.ProcessEnv = process.env` injection seam
  and `buildKpathseaSearchPath`'s `delimiter: string = path.delimiter` both
  had no caller left, so each reads the ambient value directly.
- One type parameter deleted: `filterCommandPaletteEntries<T extends {...}>`
  is monomorphized onto `CommandPaletteEntry`, its one instantiation.
- knip baseline: **166 -> 158 findings** (-48 lines). Every removed row matches
  a symbol un-exported or deleted in this PR; `check:dead-code-ratchet` reports
  no new findings.
- Net LoC: **-149** (254 insertions, 403 deletions) from
  `git diff --shortstat 893cc3dbbb HEAD`, the merge base.

## Consumer counts (R8)

Counts are references anywhere in the tree (production, `src/test-kernel`,
docs, `.md`) outside the defining file, at `origin/main`:

| Symbol | Production callers outside its file | Other references |
| --- | --- | --- |
| `filterCommandPaletteEntries` | 0 | 1 suite (4 call sites) |
| `getNextCommandPaletteIndex` | 0 | 1 suite (type decl only) |
| `executeCommandPaletteEntry` | 0 | 1 suite (type decl only) |
| `parseDesktopLogEntries` | 0 | 1 suite (5 call sites) |
| `resolveStateSettingWrite` | 0 | 1 suite (6 call sites), 1 doc comment |
| `PublicRunResult` | 0 | 0 |
| `RegisterRunOptions` | 0 | 0 |
| `ChildTurnState` | 0 | 0 |
| `SessionBridgeClosedError` | 0 | 0 |
| `FlowPosition` | 0 | 0 |
| `ClaimStanding` | 0 | 0 |
| `TranscriptRowContext` | 0 | 0 |
| `DownloadSourceOptions` | 0 | 1 suite, via `Parameters<typeof resolveArxivPaperDirectoryRelative>[1]` |
| `resolveArxivPaperDirectoryRelative` | 0 (1 in-file caller, now inlined) | 1 suite (3 cases) |
| `buildKpathseaSearchPath` | 0 (2 in-file callers) | 1 suite (3 cases) |
| `buildLatexInputEnv` | 0 (1 in-file caller) | 1 suite (3 cases) |
| `buildLatexSearchParts` | 0 (1 in-file caller) | 1 suite (3 cases) |
| `pastedImageFileName` | 0 (2 in-file callers) | 1 suite (6 cases) |

Where a retired case carried an invariant that the public entry did not
already cover, it moved rather than vanished:

- The arXiv suite's `math/0501234 -> References/math_0501234` case is now the
  ID the existing `downloadSource` test uses, so the old-style ID's slash is
  still proved to flatten into the directory name instead of nesting.
- The pasted-image trust boundary (path separators, absolute paths, empty
  names, non-TeXRA names) is asserted through `pastedImageFullPath`, which
  validates through the same private function and throws the same message.
- The logs-pane suite's three parse tests become two rendered-DOM tests:
  severity, `datetime`/label and the whole failure on one row, the truncated
  leading fragment as its own `unknown` row, and the keyed-row identity that
  the deterministic-ID test existed to protect (the same DOM nodes survive a
  refresh that appends an entry, so an expanded row stays expanded).
- The palette's multi-token and empty-query filter cases are now typed into
  the mounted `wa-input` in the existing render test.
- `StateSettingWrite.vitest.ts` keeps its three cases and moves them onto
  `applyStateSettingUpdate` with the shared `makeFakeSettingsStores` fake, so
  each now also asserts what landed in the store.

## Verified

Run in the worktree at `9280628156`; `d43f7d1caf` (the review follow-up that
deletes the `delimiter` parameter) re-ran `typecheck`, `eslint` and the
`TexTools` suite, and all of CI is green on it:

- `npm run typecheck` — pass (all 7 projects, including the
  `packages/agent` declaration build).
- `npm run test:pure` — pass, 173 files / 2075 tests.
- `npm test` — pass, 529 files / 5495 tests, 2 skipped.
- `npm run check:dead-code-ratchet` — `Dead-code ratchet OK: no new findings`,
  158 combined findings vs 158 baselined.
- `npx eslint` on every touched file — clean (full lint OOMs locally).
- `npx prettier --write` on every touched file, plus the pre-commit hook.

## Left out

- **`SessionCompletionFailed` (`src/auth/oauth/deviceAuthorization.ts:31`),
  `DeviceAuthorizationDenied` (`src/auth/xai/xaiOAuthClient.ts:57`) and
  `DeviceCodeExpired` (`src/auth/xai/xaiOAuthClient.ts:65`) — refuted, not
  deferred.** Knip calls all three unused, but the export is load-bearing for
  declaration emit: demoting them fails `npm run typecheck` with
  `src/auth/codex/codexDeviceLogin.ts(56,14): error TS4023: Exported variable
  'loginWithDeviceCode' has or is using name ... but cannot be named` and the
  same at `src/auth/xai/xaiDeviceLogin.ts(51,14)`. Both `loginWithDeviceCode`
  functions carry these classes in their Effect error channel and
  `tsconfig.build.json` emits `.d.ts` for the `packages/agent` surface that
  reaches them. A type-only re-export would satisfy `tsc` but only move the
  baseline row from `exports` to `types`, which is not a shrink. The three
  rows stay baselined and the auth error vocabulary stays half-exported: the
  `_tag` string, not the class export, is what a consumer matches on, but
  until the SDK surface stops re-exporting these flows the class name has to
  be nameable.
- **`applyReplacements` (`src/replacement/engine.ts:282`).** It has no
  production caller outside its file, but it is the applicator 13 call sites
  in `ReplacementRules.vitest.ts` use to exercise the *rule tables*
  (`EQUATION_STYLE_REPLACEMENTS`, `HTML_ENTITY_REPLACEMENTS`, ...), which are
  exported and live in production. Retargeting them at
  `replacementEngine.applyAll` means rewriting a 688-line suite around
  config-gated category selection and losing per-table precision — churn-class
  (checklist §14 R5) for one baseline row.
- **`NON_REGEX_CATEGORIES` and `REGEX_CATEGORIES`
  (`src/replacement/engine.ts:143,167`).** Their only counterparty is the
  registry/universe completeness assertion in the same suite, and that
  assertion guards a silent-degradation bug the doc comment names: a category
  missing from the registry never runs, and a universe name with no entry
  makes the config accept a no-op. One direction is already structural (each
  table's `name` is `satisfies NonRegexReplacementCategoryName`); the other is
  not expressible without either deriving the ordered registry from the
  differently-ordered universe array or adding a runtime check. Deleting the
  export here deletes an invariant, not just a test.

Two sibling lanes (binary/pandoc, UserVars) also shrink `knip-baseline.json`;
this lane lands first and they rebase onto these row deletions.


LionSR:
Addressed the review note in `d43f7d1`. Rather than leave the Windows `;` branch of `buildKpathseaSearchPath` untested, the branch is gone: with the helper private its one caller never passed a delimiter, so the parameter was an untestable seam, not a seam. It now reads `path.delimiter` directly, which is `;` on Windows and `:` elsewhere, so there is one code path and the platform picks the separator. Net elements: one more parameter deleted, no exports added.

The Codex review completed with no findings. Adding `re-review` since pushes do not re-trigger the AI reviewers.

PR #12865, opened 2026-09-19
refactor: delete the host-port members and seams no host-neutral caller uses

LionSR:
Five host-port cuts, one commit each. Every item removes a seam whose only
non-defining consumer was a test or a same-directory sibling: a port member
with no caller, a second confirm dialog, three port members only one host
ever answered, a duplicated interface, and two exports with a single
in-file reader. No dialog, diff tab, config read or agent lookup changes
behavior.

**Behavior note (corrected after review).** The four `confirmModal` call
sites now go through `VscodeUiHost.confirm`, which always passes a second
item, `{ label: cancelLabel ?? 'Cancel', isCloseAffordance: true }`. Only
`compareCommands.handleAcceptEdited` passed `'Cancel'` explicitly before;
`signOut` (`authCommands.ts:204`), `AgentHandlers.confirmAction`
(`agentHandlers.ts:136`) and `handleDeleteAgentModePreset`
(`agentHandlers.ts:434`) passed a single action button and relied on the
Cancel button VS Code adds to a modal that declares no close affordance.
So those three now name the button they were already getting rather than
letting the host add it. The answer semantics are unchanged: a Cancel click
and an Escape/X dismissal both resolve `false`, which is what all four
callers treat as abort. If a reviewer reads that as a UX change rather than
a relabelling, the three sites can pass an explicit secondary button
instead — say the word and I will.

### What each commit does

- **`db0951368c` — the config probe.** `ConfigProvider.isExplicitlySet`
  had no production caller: the two non-test hits were its declaration
  (`src/platform/interfaces.ts:76`) and the one `JsonConfigProvider` body
  (`src/platform/defaults/jsonConfigProvider.ts:88`). The whole cascade
  beneath it — `ConfigStore.has`, `JsonStore.has`, `MemoryConfigStore.has`
  and both `FakePlatform` bodies — existed only to answer it. The three
  suites that asserted through it read `inspect()` instead, the same fact
  off the same store.
- **`00740d078a` — one confirmation dialog, one UI host instance.**
  `confirmModal` was a second confirmation surface beside
  `VscodeUiHost.confirm`, and three of its four callers paid an
  `Effect.tryPromise` purely to lift the promise it returned. They call the
  port directly now. `VscodeUiHost` holds no state (every member reads
  `vscode.window` at call time), so the five construction sites share one
  exported instance.
- **`bdbdc96d48` — the diff port.** `closeDiff`, `revealFirstChange` and
  `readProposedContent` are written against VS Code's tab model, had one
  implementer and one holder (the approval host in the same directory), and
  the desktop closes by preview id instead. They move onto
  `VscodeDiffViewHost` with `DiffSession`, the shape only they read; the
  desktop's `Pick<DiffViewHost, 'openDiff'>` becomes plain `DiffViewHost`.
- **`363c3ff992` — one options declaration.** `NodeRuntimeSkillOptions` was
  declared twice, verbatim, in the same file; the survivor is unexported
  (one use, the parameter beside it), as is `ProcessRuntimeOptions`. The
  CLI init-platform suite also mocked `createNodePlatform`, an export
  `nodeHost` no longer has.
- **`75d4354477`** — `VscodeUiHost` the class stops being exported now that
  only the shared instance leaves the module.

## Net elements (R6)

- Files changed: 29 (`git diff --shortstat origin/main`)
- Net LoC: **79 insertions, 171 deletions = -92**
- `^[+-]export` symbols:
  - removed: `confirmModal`, `class VscodeUiHost`, `NodeRuntimeSkillOptions`
    (x2, one a verbatim duplicate), `ProcessRuntimeOptions`,
    `DiffSession` (from `@hosts/uiHosts`)
  - added: `const vscodeUi`, `DiffSession` (re-homed on
    `VscodeDiffViewHost`)
  - **net exports: -5**
- Class / interface declarations: `-6 / +5` (net **-1**) — two
  `NodeRuntimeSkillOptions` collapse into one, the rest are the same
  declarations unexported or re-homed.
- Interface members deleted: `ConfigProvider.isExplicitlySet`,
  `ConfigStore.has`, `DiffViewHost.closeDiff`,
  `DiffViewHost.revealFirstChange`, `DiffViewHost.readProposedContent`
  (5 port members), plus 4 implementation bodies of `has`/`isExplicitlySet`
  in production and 5 in test doubles.
- `Effect.tryPromise` / `Effect.promise` lifts deleted: 4.

## Consumer counts (R8)

Counts are `git grep` against `origin/main`, classified per the skill's
corpora (production = `src/`, `packages/*/src`, resources, loader/config
paths; non-production = `src/test-kernel/`, docs).

| Symbol | Production | Non-production | Note |
|---|---|---|---|
| `ConfigProvider.isExplicitlySet` | **0 callers** (3 hits: declaration, its one implementation, one comment) | 9 hits in 7 test files — 2 `FakePlatform` bodies, 3 inline fake members, 4 assertions | assertions moved to `inspect()` |
| `ConfigStore.has` / `JsonStore.has` / `MemoryConfigStore.has` | **1 caller**, `jsonConfigProvider.ts:91` (i.e. `isExplicitlySet` itself) | 0 | the whole reason the member existed |
| `confirmModal` | **4 callers** — `compareCommands.ts:272`, `authCommands.ts:204`, `agentHandlers.ts:136`, `agentHandlers.ts:434` | 2 module mocks (`AgentHandlersDelete`, `AuthCommandsStoredSession`) | all four retargeted at `VscodeUiHost.confirm` |
| `new VscodeUiHost()` | **5 sites** — `extensionHostRequests.ts:196`, `agentHandlers.ts:67`, `SettingsViewMessageHandler.ts:155`, `errorHandlingUtils.ts:14`, `apiKeyCommands.ts:45` | 0 | all now read the one exported instance |
| `DiffViewHost.closeDiff` / `revealFirstChange` / `readProposedContent` | **1 implementer** (`VscodeDiffViewHost`), **1 holder** (`VscodeToolEditApprovalHost`, 4 call sites). Desktop answers none of them — `desktopDiffHost.closeDiff(previewId)` and `desktopToolEditApproval.readProposedContent()` are that host's own members, not the port's | 0 | |
| `DiffSession` (from `@hosts/uiHosts`) | 2 files, both in `packages/extension/src/frontend/approval/` | 0 | re-homed beside its only readers |
| `NodeRuntimeSkillOptions` | **1 use** — the `initializeNodeRuntimeSkills` parameter in the same file | 0 | |
| `ProcessRuntimeOptions` | **1 use** — the `installProcessRuntime` parameter in the same file | 0 | |
| `createNodePlatform` | **0** — `nodeHost` has no such export | 2 mock lines in `CliInitPlatform.vitest.ts`; 7 stale references in `docs/architecture/2026-07-26-embedding-the-agent-runtime.md` | mock deleted; the stale doc is left out, see below |

## Verified

Run in the worktree at `75d4354477`:

- `npm run typecheck` — **pass** (all workspace projects, no `error TS`).
- `npm run test:pure` — **pass**, 173 files / 2075 tests, 1 skipped.
- Kernel suites for the hosts this touches —
  `npx vitest run --project kernel src/test-kernel/{desktop,cli,settings,commands,controllers}`:
  151 files, 1578 passed, **1 failed**:
  `DesktopPreviewHost.vitest.ts > presents openFile failure once through the
  request dispatcher` timed out at 10000ms under the parallel load. Re-run
  alone: **17/17 pass in 22s**. That file references nothing this PR
  touches (checked by grep for every symbol in the table above), so it is a
  load-dependent timeout, not a regression.
- `npm run check:dead-code-ratchet` — **pass**, 166 findings vs 166
  baselined, no new findings. Nothing in `knip-baseline.json` resolved
  (the deleted exports all had consumers; the unexported ones leave the
  report entirely), so the baseline is unchanged rather than shrunk.
- `npm run check:effect-migration-ratchet` — **pass**, no count grew, no
  baseline headroom. `platform()` stays at 3 files / 3 sites (see Left
  out), `Effect.run*` at 0, `catch:effect-importer` at 2.
- `npm run check:unexecuted-effects` — **pass**, 2005 files scanned.
- `npx eslint <every touched .ts>` (with
  `NODE_OPTIONS=--max-old-space-size=8192`) — **clean**.
- `npx prettier --ignore-path /dev/null --write` on all touched files.

## Left out

- **N15 — moving `agentDirectories` off the ambient `Platform` onto an
  `AgentDirectories` tag in `installProcessRuntime`.** The premise that this
  can be narrowed to the one member without disturbing the composition roots
  is refuted by the code: two of the three roots build the port *after* they
  install the runtime that would have to serve it.
  - `packages/desktop/src/main/platform/index.ts:202` installs the runtime;
    the port is built at `:216-218`, behind `repairLaunchPath()` and
    `resolveResourcesPath(mainDirname)`, and wired at `:227`. Serving it as
    a tag means hoisting the resources-path resolution above
    `repairLaunchPath()`, i.e. reordering the root's startup.
  - The CLI installs at `packages/cli/src/runtime/cliProcessRuntime.ts:138`,
    but builds the port at `packages/cli/src/runtime/initPlatform.ts:334`,
    in a different module that runs *on* the already-installed runtime
    (`withProcessServices(runtime, …)` at `:313`) and reads
    `context.resourcesPath`, which does not exist at install time — the 15+
    `installCliProcessRuntime(context.storageRoot)` call sites would each
    have to grow an argument.
  Against ~8 LoC of saving that is a reorder of two composition roots and a
  module move, which is the churn the checklist §14 R5 ban is about. The
  single ambient read this would have removed
  (`src/agent/index/agentRegistry.ts:164`) is one of the 3 sites the
  `platform()` ratchet already holds frozen, so nothing regresses by
  leaving it.
- **`.claude/skills/code-review/references/review-checklist.md:43`** listed
  `isExplicitlySet` among the `ConfigProvider` accessors. Not left out —
  fixed in `f7b6402d3b`, since it points at a member this PR deletes.
- **The stale `createNodePlatform` references in
  `docs/architecture/2026-07-26-embedding-the-agent-runtime.md` (7 hits,
  e.g. `:35`, `:41`, `:104`).** That export is already gone from
  `nodeHost`; this PR only deletes the test double for it. Correcting a
  published architecture doc's embedding recipe is a docs decision, not a
  port cut, and the `docs/` root-boundary gate makes it worth its own PR.


LionSR:
Both review findings addressed.

**1. The `Cancel` label.** Corrected in the PR body: only `compareCommands.handleAcceptEdited` passed `'Cancel'` explicitly before, so the claim that all four already displayed it was mine to fix, and it now reads that way. The mechanism, for the record: `VscodeUiHost.confirm` passes `{ label: cancelLabel ?? 'Cancel', isCloseAffordance: true }` as a second item, and VS Code uses a declared close affordance *instead of* adding its own Cancel to a modal — so the three sign-out/delete/overwrite dialogs render the same single Cancel button they rendered before, now named by us rather than by the host. Answer semantics are untouched (cancel click and Escape/X both resolve `false`). Happy to hand those three an explicit secondary button instead if you read it as a UX change rather than a relabelling.

**2. The stale checklist reference.** Fixed here rather than deferred, since it points at a member this PR deletes: `f7b6402d3b` drops `isExplicitlySet` from the `ConfigProvider` accessor list in `.claude/skills/code-review/references/review-checklist.md:43`.

The `createNodePlatform` references in `docs/architecture/2026-07-26-embedding-the-agent-runtime.md` stay out of scope — that export was already gone before this PR, and correcting a published embedding recipe is a docs decision behind the root-boundary gate.

PR #12866, opened 2026-09-19
refactor(agent): retire the never-read UserVars template names, one Zod owner for the vocabulary

LionSR:
Retires four fixed template variables that `buildUserVars` computed and persisted on every run and that no prompt, template, doc or test ever read: `WORKFLOW_AGENTS`, `TOOL_USE_AGENTS`, `IS_OPENAI_MODEL`, `IS_GOOGLE_MODEL`. The first two also ran the entire delegation roster through `formatAgentList` on every launch, including every subagent launch, and wrote the result into `UserVariableChannels` on every `flow.snapshot`. The roster the model actually acts on comes from `annotateDelegationAvailability` on the delegation tool descriptions, so this was that fact's retired duplicate.

`IS_ANTHROPIC_MODEL` stays: it is used by 5 shipped agent YAMLs, `PromptBuilder.ts:33`, and both agent-creation schema docs.

Because that deletion forces a `SESSION_EVENT_FORMAT` bump (see below), the PR also lands the deletion #12858 explicitly deferred to "the next real format bump": the two `StateOperation` arms, `delete` and `append`, that have never had a producer.

### Owner rulings taken

1. **The four names are retired, not documented.** A custom agent YAML referencing one would silently substitute empty rather than error (Nunjucks `throwOnUndefined` is off), so this ships with a user-facing CHANGELOG entry under Unreleased naming the four removed names. The ruling also asked for them to be deleted from the two prose listings — **both listings turned out not to name them**: `docs/.vitepress/components/TemplateVarsPalette.vue` publishes 12 tokens (none of the four) and `packages/extension/resources/docs/agent-creation/workflow_schema.md:65` advertises only `IS_ANTHROPIC_MODEL`. A repo-wide `rg` for each of the four confirms zero hits under `docs/`. No deletion was needed; a clarifying clause was added instead, see Left out.
2. **Schema-first wins over the inline comment.** The hand-written `export type UserVars = { ... }` block in `runFlowState.ts` is deleted. `UserVarsSchema` (a `z.object`, carrying the per-field JSDoc) is the vocabulary, `export type UserVars = z.infer<typeof UserVarsSchema>`, and `UserVariableChannelsSchema` is `z.looseObject(UserVarsSchema.shape).partial()`. The `satisfies { [K in keyof Required<UserVars>]-?: ... }` lockstep clause is gone because there is no longer a second declaration to keep in lockstep.
3. **Make the forced bump pay twice.** The `StateOperation` cut rides this PR's bump rather than buying one of its own.

### `SESSION_EVENT_FORMAT` 5 -> 6, once, covering both cuts

`UserVariableChannelsSchema` is reached from `SessionEventSchema` (via `StateSlicesSchema.userChannels`), and `StateOperationSchema` through `ToolResultPayloadSchema.stateMutation`. Either deletion moves the stored-shape fingerprint that `src/test-kernel/schemas/sessionEventFormat.vitest.ts` pins, and the gate then requires the version bump. **`Database` clears every store stamped with another version at open**, so this clears existing session history on first open of a build carrying this change.

The constant moves **5 -> 6 exactly once**. Both cuts are in the same unreleased PR, so only the pinned fingerprint moved a second time when the `StateOperation` commit landed; users see one format change, not two. It lands inside the 1.0 window that already carries CHANGELOG entries saying 1.0 starts with new session history and that preview builds start from a clean slate, and format 5 itself was set on 2026-09-18 in #12776 — so no second changelog entry was added. The bump is not avoidable by narrowing: it follows from the owner-ruled token deletion, not from the schema-first rewrite (a rename alone leaves the fingerprint unchanged).

## Net elements (R6)

- **Files:** 15 changed (7 production, 4 test, 1 test snapshot, 1 ratchet baseline, 1 CHANGELOG, 2 agent-creation schema docs).
- **Net LoC:** `git diff --stat origin/main` = **126 insertions, 269 deletions, net -143**.
- **Exports removed (5):** `BuildUserVarsOptions`, `resolveOutputFiles`, `getToolFlags` (all `@agent/prompt/userVars`), `formatAgentList` (`@tools/delegation/delegationAvailability`), and the hand-written `UserVars` object type (the name survives as an inferred type, the 72-line literal does not).
- **Exports added: 0.** `UserVarsSchema` is deliberately module-private, so no new export needs a consumer.
- **Type/interface declarations deleted (4):** `ModelProviderFlags` (`userVars.ts`, collapsed to an `isAnthropicModel: boolean` parameter), `FormatAgentListOptions` and `AgentListToolsStyle` (`delegationAvailability.ts`), and the `UserVars` object literal type. `UserVariableValueSchemas` was not deleted but renamed and repurposed as `UserVarsSchema`.
- **Union arms deleted (2):** the `delete` and `append` members of `StateOperationSchema` (`runLedgerEvent.ts`) and their two `mutate` branches (`runStateFold.ts`), plus the clauses of the usage docstring that existed only to rule them out.
- **Other deletions:** the `delegationAgentScope` slot on `BuildUserVarsOptions` and its pass-through in `AgentLaunchContext`; the `tools: 'inline' | 'none'` and `collapseDescriptionNewlines: false` branches of `formatAgentList`, which only the retired prompt vars ever asked for; two `config/ratchets/knip-baseline.json` rows; three unused imports in `userVars.vitest.ts` (`expectTypeOf`, `BuiltUserVars`, `TemplateVars`) and one in `DelegationAvailability.vitest.ts` (`formatAgentList`, imported but never called).

## Consumer counts (R8)

Repo-wide `rg` run in this worktree at head `d76d9c9d03`, over `src/`, `packages/`, `prompts/`, `docs/`, `config/`, excluding `node_modules` and `.git`.

| Symbol / token | Production consumers | Test consumers | Docs / YAML |
| --- | --- | --- | --- |
| `WORKFLOW_AGENTS` (as a template token) | 0 | 0 | 0 |
| `TOOL_USE_AGENTS` (as a template token) | 0 | 0 | 0 |
| `IS_OPENAI_MODEL` | 0 | 0 | 0 |
| `IS_GOOGLE_MODEL` | 0 | 0 | 0 |
| `IS_ANTHROPIC_MODEL` (kept) | 1 (`PromptBuilder.ts:33`) | 0 | 5 agent YAMLs + 2 schema docs |
| `formatAgentList` | 1, in its own file, after the userVars call sites go | 1 import, **0 calls** | 0 |
| `getToolFlags` | 0 outside `userVars.ts` | 1 direct unit test (folded) | baselined as production-dead |
| `resolveOutputFiles` | 0 outside `userVars.ts` | 1 direct unit test (folded) | baselined as production-dead |
| `BuildUserVarsOptions` | 0 outside `userVars.ts` | 0 | 0 |
| `BuildUserVarsOptions.delegationAgentScope` | 1 writer (`AgentLaunchContext`), 1 reader (`getBasicVars`, deleted here) | 1 assertion (updated) | 0 |
| `StateOperation` arm `delete` | **0** | **0** | **0** |
| `StateOperation` arm `append` | **0** | **0** | **0** |

The greps behind the last two rows, re-run in this worktree rather than taken from #12858:

- `rg -n "op: '(delete\|append)'" . -g '!node_modules' -g '!.git'` → **exit 1, zero hits**.
- `rg -n '"op"\s*:\s*"(delete\|append)"' . -g '!node_modules' -g '!.git' -g '*.json' -g '*.jsonl' -g '*.ndjson'` → **exit 1, zero hits**. This covers `src/test-kernel/**` fixtures and snapshots and `packages/*/tests`, so no stored row can carry either arm.
- `rg -n 'op:\s*.(delete\|append).' . -g '!node_modules' -g '!.git'` (quote-agnostic) → **exit 1, zero hits**.
- `rg -n "op: '" src packages -g '!node_modules'` returns every construction site in the tree: production `toolUseDispatch.ts:308` (`set`) and `:576` (`add`); tests use only `add` and `set`.
- `rg -n "literal\('(delete\|append)'\)"` after the change hits only `MemoryTool.ts:124` (`command: z.literal('delete')`, an unrelated tool input) and `runLedgerEvent.ts:205` (`kind: z.literal('append')`, a different union).

The remaining grep hits for `WORKFLOW_AGENTS` / `TOOL_USE_AGENTS` are unrelated identifiers that merely contain the substring: `BUILTIN_WORKFLOW_AGENTS_DIR`, `BUILTIN_TOOL_USE_AGENTS_DIR`, `PREFERRED_TOOL_USE_AGENTS`, `ENABLED_TOOL_USE_AGENTS` and the CLI harness's `HARNESS_VISIBLE_*` env vars. None is a template variable and none is touched.

Per AGENTS.md testing discipline this adds no test file. The two direct unit tests for the now-private userVars helpers were folded into `buildUserVars` assertions, which is where their behavior is observable: `ROUNDS` is asserted off the built vars, and the four `resolveOutputFiles` cases assert both `agentConfig.outputFiles` (the write-back) and `vars.OUTPUT_FILES`. One case had to gain `inputFiles` because it now goes through `AgentConfigSchema.parse`, which rejects more outputs than inputs; the old test bypassed the schema with a cast. The `StateOperation` cut needed no test change: no suite ever constructed either arm.

## Verified

Re-run in the worktree after the rebase onto `origin/main` `ddf75616d0` and the `StateOperation` commit. Head `d76d9c9d03`.

- `npm run typecheck` — **pass** (exit 0, all 8 project configs).
- `npm test` — **pass**, the full gate: 529 files / 5492 tests / 2 skipped, exit 0.
- `npm run test:pure` — **pass**, 173 files / 2074 tests / 1 skipped, includes the architecture ratchets.
- `npm run check:dead-code-ratchet` — **pass**, 156 findings vs 156 baselined. The baseline shrank by 2 rows here and was rebased over #12864's shrink to 158 without conflict. The run also caught one finding this lane created: unexporting the userVars call sites left `formatAgentList` production-dead. It was unexported rather than baselined, since a baseline must never widen.
- `npm run check:effect-migration-ratchet` — **pass**, no count grew, no headroom.
- `npm run check:unexecuted-effects` — **pass**, 2004 files.
- `npm run check:guidance-refs` — **pass**.
- `npx eslint` on every touched TypeScript file — **pass**, no output.
- `npx prettier --write` on all touched files — no reformatting needed; the pre-commit prettier hook passed on each commit.
- `sessionEventFormat.vitest.ts` was regenerated twice, once per shape-moving commit, and both times with `SESSION_EVENT_FORMAT` at 6. Only the pinned fingerprint moved the second time.
- Docs build: not run, and no `docs:*` script exists at the workspace root. No file under `docs/` is touched, and the commit-time `docs-root-boundary` gate reported "no files to check".

## Left out

- **N2, folding the four `failIfAborted(input.signal); failIfLaunchStopped(input.stopped);` pairs into one `failIfLaunchCancelled`.** Refuted on its own terms. The pairs in `AgentLaunchContext.ts` would collapse 8 lines to 4, but the helper plus its doc comment costs about 6, so the item is net LoC-positive. It is also element-positive in the way that matters: eight further sites (six in `AgentLaunchContext.ts`, two in `runAgent.ts`) call `failIfLaunchStopped` alone and cannot use the pair helper, so the file would carry three cancellation-check spellings where it carries two today.
- **The two prose listings.** Named in the ruling, but neither file mentions any of the four names, so there was nothing to delete. Following review, both agent-creation schema docs instead gained one clause saying `IS_ANTHROPIC_MODEL` is the only model gate and that an invented one renders as false, which closes the silent-failure gap where agent authors actually read.
- **`buildUserVarPassthrough` narrowing.** Not an item, but worth naming since it is a derived consumer: the passthrough map shrinks with the vocabulary, so an LLM-generated agent YAML emitting `{{ WORKFLOW_AGENTS }}` during agent creation no longer passes through literally. No repo template does, `AgentTemplateRenderer.vitest.ts` asserts only `ALL_CONTEXTS` / `LIST_OF_ALL_CONTEXTS` and stays green, and the creator-facing schema docs advertise only `IS_ANTHROPIC_MODEL`.


LionSR:
Addressed the TeXRA review's one finding: the orphan `How an agent's tool list is rendered inside a roster entry.` doc line left behind by the deleted `AgentListToolsStyle` type is removed in 808594eab0. No other change; typecheck still green.

LionSR:
Took the second review's non-blocking suggestion in f0a09d6925: both agent-creation schema docs now say `IS_ANTHROPIC_MODEL` is the only model gate and that an invented one renders as false, which closes the silent-failure gap at the point agent authors read.

PR #12876, opened 2026-09-20
refactor(llm): one uploaded-file provider, not two

LionSR:
## Summary

The Responses and Anthropic bindings each wired their own Files API `{ send, remove }` pair by hand. Both stated the same three rulings:

- an SDK rejection becomes this binding's enriched `ModelError`,
- a create response without a usable id is `malformed-output`,
- a 404 on delete is the deletion the caller asked for.

Only four things actually differed: the SDK create call, the SDK delete call, how the response states its expiry, and the provider's name in the message.

`filesApiUploads` in `uploadCache.ts` states the shared rulings once and takes those four as parameters. Each binding's expiry encoding stays its own concern, but moves to the boundary: the two `UploadedFileSchema`s now `.transform()` into the canonical `{ fileId, expiresAtMs }` (RFC 3339 for Anthropic, whole Unix seconds for OpenAI) instead of each call site re-deriving milliseconds after parsing.

`uploadCache` has no caller outside its own module now, so it stops being exported.

## Issue acceptance gates

No issue — opportunistic consolidation found while auditing `packages/llm` for duplicated orchestration. No gates claimed.

## Single source of truth

`filesApiUploads` (`packages/llm/src/uploadCache.ts`) now owns the Files API provider contract: the malformed-response ruling, the 404-tolerant delete, and the SDK-rejection→`ModelError` wiring. Each binding owns only its own SDK calls and its own expiry encoding.

Removed: two independently-authored Files API `{ send, remove }` pairs collapse to one. The two call sites drop from 50 and 60 lines to 21 and 29.

Added: one function, `filesApiUploads`, which owns the invariant above. It is not a pass-through — it makes the two rulings unstateable per-binding and returns the `UploadCache` the callers already wanted, so each site is one call rather than a hand-assembled pair.

## Net elements (R6)

- Files: 3 changed, 0 added, 0 deleted.
- Exported symbols: net **0** (`+export function filesApiUploads`, `-export function uploadCache` — the latter is now module-private).
- Classes/interfaces/enums: net 0.
- Net LoC: **+3** (116 insertions, 113 deletions).

Net LoC is flat, and I want to be straight about that rather than claim a line win: the ~60 lines the call sites shed are offset by the helper's body plus its rationale comments, and by the two schemas gaining `.transform()` blocks. The win claimed here is one contract instead of two copies of it, not a smaller diff.

## Consumer counts (R8)

No emit path or public export re-routed. `uploadCache` was grepped across `packages/` and `src/` before un-exporting it: **0** importers outside `packages/llm/src/uploadCache.ts` (both former importers now take `filesApiUploads`). `UploadCache` (the type) keeps its 2 importers.

## Host impact

Extension: none — provider-internal, no surface change.

Desktop: none.

CLI: none.

## Scheduled refactoring

None.

## Validation

- `corepack pnpm --filter @texra-ai/llm typecheck` — clean.
- `npx vitest run src/test-kernel/llm/AnthropicMessages.vitest.ts src/test-kernel/llm/OpenaiResponses.vitest.ts` — 126 passed.
- `npm run test:pure` — 174 files, 2066 passed / 1 skipped.
- `eslint packages/llm/src` — clean.
- `node scripts/check-dead-code-ratchet.mjs` — no new findings (154 vs 154 baselined).
- `prettier --check` on the three touched files — clean.

Not run: full `npm test` (minutes-long; the pure tier plus both adapter suites cover the changed code).

https://claude.ai/code/session_017LU6ykGraFy6ZyQfUB6vP4

---
_Generated by [Claude Code](https://claude.ai/code/session_017LU6ykGraFy6ZyQfUB6vP4)_


LionSR:
Thanks — addressed the review note in `d0203ea`. One of the two suggestions is in, one is out, with reasons:

**Taken: `Uploaded` is now exported.** The point is right and it goes to the purpose of the change — the whole case for `filesApiUploads` is that a third Files API binding could share the contract, and it couldn't annotate its own `parseUploaded` against a type it can't name. I checked the obvious objection first: exporting it does *not* trip `check:dead-code-ratchet` (still 154 vs 154 baselined), because knip already counted `Uploaded` as used via the exported signature. So this doesn't buy a new export at the cost of the "a new export needs a consumer" rule.

**Not taken: `import type { z }`.** Accurate in the abstract, but it loses to this repo's `import/order` config. Making it a type import moves it into the type-import group and eslint then fails with:

```
9:1  error  `./turn.js` import should occur before type import of `zod`  import/order
```

The only way to satisfy that is to sort the third-party `zod` type import *after* the local `./turn.js` import, which is worse to read than the cosmetic gain is worth — and every other module in `packages/llm/src` imports `z` as a value anyway. Leaving it as-is.

Re-validated on `d0203ea`: `eslint packages/llm/src` exit 0, `prettier --check` clean, `pnpm --filter @texra-ai/llm typecheck` clean, `check:dead-code-ratchet` no new findings, `test:pure` 2066 passed / 1 skipped, and the two adapter suites 126 passed.

---

Issue #12885, opened 2026-09-20
Tracking: tools and schema-surface collapse (barrel split, one probe catalog, run.fact row, ExecutionsTool as projection)


LionSR:
## Current status — 2026-09-20

Checked against `main` at `3378a967`. Parent: #12880. Proposal: `2026-09-20-tools-and-schema-surface-collapse.md` in [proposed/simplification](https://github.com/LionSR/TeXRA/tree/claude/architecture-refactoring-progress-dyyk7f/.agents/docs/proposed/simplification). Revised after the #12889 review round. Closes the open rows of the 2026-09-10 collapse-duplicate-concepts census in this territory.

## Why

`sharedSchemasDeepImportRatchet` forbids leaf imports, so 811 files import a 60-module, 9 510-line barrel of which ~3.5k lines are view-message schemas, not wire contracts; `src/shared` also carries ~10k lines of UI toolkit under a charter that says "wire contracts". The tool-call path is seven layers, two pure pass-through (`core/define.ts`, the `execute` forwarder); guards are called per tool, not by the loop. The LaTeX/image probe list is spelled five times, the five run-fact row names four times, the replacement-category universe three times. Five run-fact rows are one row with a discriminator; three singleton records are one `state.value.set`; `runRecords.ts` has five near-identical latest-row readers. `ExecutionsTool` (1 141 lines plus 1 300 around it) re-derives what the fold already holds.

## Work

- [x] Split the non-contract half out of `@shared/schemas`; move the UI toolkit out of `src/shared` or amend the charter. The settings webview keeps `stateSettings.ts` (it consumes the catalog on purpose); the progress, memory and profile views stop including it. (#12946, merged)
- [x] One dependency-probe catalog in `externalToolDefs` + `toolAvailability`; delete the other four spellings, sync comments and the completeness test. (#12896, merged)
- [x] One `run.fact` row and one `state.value.set` row with a discriminator that keeps each key family tied to its existing Zod value schema and aggregate kind; delete eight arms, four name lists, `runFactEvents.ts`, the `Database.ts` branch chain, the five readers; bump `SESSION_EVENT_FORMAT`. (#12912, merged)
- [x] `ExecutionsTool` renders from the fold; keep `wait` and `kill`. (#12921, merged)
- [x] Derive the replacement-category universe from the registry; rename the colliding type pairs; keep the ordered apply lists. (#12896, merged)
[...]

PR #12946, opened 2026-09-21
refactor(shared): split non-contract schemas and move the UI toolkit

LionSR:
## Summary

The `@shared/schemas` barrel was 9,510 lines and every importer took all of it, so roughly a third of it that is not wire contracts rode into every closure, webviews included. This lane splits the non-contract half out and moves the UI toolkit that mis-lived under `src/shared` into its own directory. 356 files changed, 1,204 insertions, 1,024 deletions; the work is module moves plus specifier rewrites, in 10 commits, plus a 16-commit run-exit and tool-union series the branch carried (owned below).

- The barrel drops from 9,510 to 6,873 lines. `stateSettings.ts` (now 1,577 lines) moves to `src/shared/state/` beside the `stateKeys.ts` it is built from; `settingsViewMessages.ts` (943), `memoryViewMessages.ts`, `profileViewMessages.ts` and their private helper `messageFactories.ts` move to the existing `src/shared/settingsView/`, beside the host-neutral handler wiring that speaks that protocol. Filenames are preserved, so git records renames.
- Importers were re-pointed by a parser-shaped rewriter, not sed: 936 `from '@shared/schemas'` statements across 82 files, then 456 `@shared/{wa,styles,transcript,markdown,copy}` specifiers across 205 files. The rewriter asserts per file that matched statements equal raw occurrences (936 of 936, zero mismatches), is idempotent on re-run, and the duplicate-import shape it leaves is the repo's prevailing value-plus-`import type` pattern.
- The toolkit lands in `src/ui/` (54 files, 8,892 lines: wa, styles, transcript, markdown, copy) behind a new `@ui/*` alias declared once in `tsconfig.json`. `scripts/aliases.mjs` derives every Vite/Vitest/ESLint alias map from that file, so no build config needed a second edit.
- Zone lists stay in sync: `src/ui` joins `VSCODE_FREE_ZONE_DIRS` in eslint.config.mjs and `VSCODE_FREE_ZONES` in dependencyDirection.vitest.ts, and the "does not grow shared-to-agent imports" test now scans `src/ui` as well as `src/shared`. `sharedSchemasDeepImportRatchet.vitest.ts` keeps its deep-import prohibition unchanged and gains a second test pinning each moved leaf absent from `src/shared/schemas/` and present at its named home, so the barrel cannot quietly re-absorb the catalog.
- Acceptance line met: no `stateSettings` / `settingsViewMessages` references under `packages/extension/src/progressView` or `packages/trace-viewer/src`. Only the settings webview, plus the desktop main process and the CLI (which are not webviews), reaches the catalog, so the progress webview bundle no longer ships the ~2,700-line settings surface it never used.
- Moves forced housekeeping: `file-size-baseline.json` and `knip-baseline.json` keys move with their files (12- and 4-line renames, counts shrink), `architecture-edges-baseline.json` relabels the renamed subsystem edges, and every prose path the guidance-refs gate resolves is re-pointed (src/README.md placement guide, docs/dev/verification.md, the code-review and find-simplification skills, five NOTICE files, comments citing moved modules). One dead re-export (`export { type GoalListItem }` from settingsViewMessages.ts) is deleted; every importer takes it from the wire-contract barrel where goal.ts defines it.
- Two findings the move exposes rather than creates, recorded here for follow-up: `architecture-edges-baseline.json` now carries `shared -> ui` and `ui -> shared` as live value edges. Both directions were internal to `src/shared` before and invisible to the ratchet; nothing regressed, but the coupling is now on the books and a follow-up can break it. And `packages/desktop/src/main/desktopAgentSettingsController.ts` would have grown past its file-size budget because a single barrel import split in two; its import was split into value and type halves, bringing it to 734 lines, rather than raising the budget.

## Out-of-lane work this branch carries

The branch tip is 26 commits past its base. The other 16 are halt/run-exit and tool-union work the proposal doc does not mention; they were already pushed and pass the full gate, so this PR owns them rather than discarding pushed work.

- `classifyHaltWriteCause` in `src/agent/runtime/loop/runExit.ts` inspects the Effect `Cause` of the halt write directly instead of routing every failure into one catch-all warn. `RunLedgerRefused` stays best-effort warn; `DatabaseWriteFailed` now fails the run through a typed `Effect<void, DatabaseWriteFailed>` channel, wired into both loops (toolUse.ts and reflection.ts, 10 lines each); unknown defects die instead of being masked. `runExit.ts` had no suite on main, so the new path-mirrored `src/test-kernel/agent/runtime/runExit.vitest.ts` covers the three resolutions. Three doc commits in the series refine that file's contract comments only.
- `refactor: inline tool command unions` deletes `src/tools/core/inputSchema.ts` (48 lines of duplicated command unions) and inlines each union in its consumer tools, net minus 51 lines.
- One commit of that series (`fix: align halt finalization state`) also deleted the `OPENROUTER_COPY` / `KIMI_CODE_COPY` / `GLM_CODING_PLAN_COPY` shared-copy tables from `stateSettings.ts` and inlined each row's copy at both use sites (settings row and Models-tab control). That removes the written-once drift guard and is where the file's 1,581-to-1,577 line change comes from; the move itself is line-count neutral. Flagged for reviewer attention.
- A validation fix de-nested a ternary in the halt cause handler (behavior identical) and prettier was applied across moved and repointed files.

## Spec

`.agents/docs/proposed/simplification/2026-09-20-tools-and-schema-surface-collapse.md`, step 1 of Changes: split the non-contract half out of `@shared/schemas` into its own surfaces, and move the UI toolkit out of `src/shared`. Steps 2 through 7 belong to other lanes.

## Assumptions

- On barrel shape, the note's "one barrel per view surface, or one non-contract surface; take the smaller" took neither. A merged non-contract barrel fails the acceptance line outright (the progress view would import the settings catalog back), and four new barrels collide with CLAUDE.md's no-convenience-barrels rule. Each module is instead published at a path whose charter fits it and imported directly: zero new barrels, two existing directories reused, and the deep-import ratchet needs no new exception because the new paths sit outside the surface it governs.
- `stateSettings.ts` went to `src/shared/state/`, not a view directory: the CLI config system, controllers and tools read the catalog, so it is not a view artifact. The settings-view message family went into the existing `src/shared/settingsView/`, already guarded against `@controllers`/`@agent`/`@model`/`@tools`/`@auth` imports by `SharedSettingsViewBoundary.vitest.ts`.
- On the UI toolkit the recommended reading was taken: move it to `src/ui` behind `@ui/*` rather than re-chartering `src/shared` in CLAUDE.md.
- `src/shared/session/sessionFold.ts` and `sessionView.ts` are touched despite the brief's exclusion of that directory, but only as a forced, disclosed, logic-free alias rename (`@shared/transcript` specifiers become `@ui/transcript`, 14 lines). The alternatives were a broken build or a re-export shim, which the repo bans.
- `BROWSER_SAFE_UTILS` needed no change: it allowlists `@utils/*` modules and says nothing about `@shared` or `@ui`.

## Behavior changes

Lane work: none. Every commit is a module move plus specifier rewrite; no schema, wire command or runtime branch changed. The one deletion is an unread type re-export. Indirectly, the progress webview bundle shrinks because it no longer contains the settings catalog and settings-view protocol it never used.

The out-of-lane series does change behavior: a failed ledger database write during halt now fails (or dies, for defects) the run instead of being downgraded to a warn. Only `RunLedgerRefused` keeps the warn.

## Skipped or reverted

- Moving `src/shared/schemas/mainView/state.ts` and `progressView/data.ts` out of the barrel: they are load-bearing for modules that stay (sessionEvent.ts imports `PermissionPayloadSchema`, traceEvent.ts imports `ToolCallStatusSchema`, streamLogEntry.ts imports progressView/data, which imports mainView/state). Moving them would create a schemas-to-new-surface-to-schemas module cycle whose members are top-level Zod constructions, exactly where a cycle yields `undefined` at evaluation time. They also carry little acceptance weight: the barrel now costs the progress view about 500 lines of view state it actually renders, not 2,700 lines of settings catalog. Splitting the contract arms out of progressView/data is its own change.
- `src/shared/highlighting/`, `litControllers/` and `monaco/` stay put. They are UI code under the same mis-charted directory, but the step names five directories (wa, styles, transcript, markdown, copy) and moving them would widen the lane. src/README.md now states where the line falls.
- Two comment-only stale paths survive, triaged and left for a follow-up: `src/controllers/modelAccess/subscriptionProviders.ts:56` and `src/shared/constants/latexConfig.ts:70` still cite `@shared/schemas/settingsViewMessages` in prose. No import or runtime path uses the old location; the fix is a two-line comment repoint.

## Validation

Central validation worktree, all six gate steps exit 0 at c659639d931e, rebased on origin/main:

- `npm run typecheck` (all 7 projects, including the desktop renderer against the new `@ui/*` alias; the lane's pre-push "Desktop renderer TS errors" failure is confirmed resolved)
- `npm run lint`
- `npx vitest run --config vitest.config.mjs --changed origin/main` (443 files, 4,474 tests)
- `npm run test:pure` (184 files, 105 architecture ratchet tests)
- `node scripts/check-guidance-refs.mjs`

The acceptance line was also independently re-verified against the pushed tip with rg.

Part of #12885


LionSR:
rebase

LionSR:
@copilot resolve the merge conflicts in this pull request

PR #12967, opened 2026-09-21
fix(deps): move the genai bump into the catalog so --frozen-lockfile works again

LionSR:
## Summary

`main` cannot install its dependencies right now. #12964 (`08ee302ea6`) rewrote the root importer's `@google/genai` specifier in `pnpm-lock.yaml` to the bare version `2.23.0` while root `package.json` still references the catalog, and left the catalog recording the old `2.22.0` resolution. Every CI job installs through `.github/actions/setup-workspace`, which runs `pnpm install --frozen-lockfile`, so a clean checkout fails before it runs a single test:

```
[ERR_PNPM_OUTDATED_LOCKFILE] Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/package.json
  specifiers in the lockfile don't match specifiers in package.json:
* 1 dependencies are mismatched:
  - @google/genai (lockfile: 2.23.0, manifest: catalog:)
```

Three concurrent rebase lanes reproduced this independently in fresh worktrees, so it is blocking the whole PR backlog, not just main.

## Fix

In this workspace the catalog is where a dependency bump belongs, so the bump moves there: `pnpm-workspace.yaml`'s catalog entry goes `^2.22.0` -> `^2.23.0`, and the lockfile is regenerated. The root importer goes back to `specifier: 'catalog:'`, the catalog's own entry resolves `2.23.0` (the version #12964 intended, and the one `packages/llm` already pins explicitly), and the now-unused `@google/genai@2.22.0` package entry is dropped.

Same shape as the fix for the identical `#12952` slip on `@openrouter/sdk` (`1fcb68941c`), except that one only needed the specifier line because its version did not move.

## Issue acceptance gates

n/a - repair of a dependency-manifest inconsistency; no behavior change.

## Single source of truth

Unchanged: `pnpm-workspace.yaml` remains the catalog for workspace-wide ranges, `packages/llm/package.json` keeps its explicit `2.23.0` pin, and no application code or schema is touched.

No new abstractions. Net effect: 2 files, +5/-27 (the deleted lines are the stale `2.22.0` resolution and its package entry).

## Net elements (R6)

No exported symbols, classes, interfaces or enums change. No test changes.

## Consumer counts (R8)

`@google/genai` has three consumers, none of their call sites change: `src/agent/export/schemas.ts`, `src/test-kernel/common/SdkErrorUtils.vitest.ts`, and `packages/llm/src/googleInteractions.ts` (already on 2.23.0 through its own pin).

## Host impact

None: this repairs installation, so every host's CI and every developer's fresh worktree can install again. Nothing in this change reaches runtime code.

## Scheduled refactoring

None.

## Validation

On a fresh worktree cut from `origin/main`:

- `corepack pnpm install --frozen-lockfile --offline` - failed with `ERR_PNPM_OUTDATED_LOCKFILE` before this change; `Done in 11.5s` after it.
- `corepack pnpm exec tsc --noEmit --checkers 8` - exit 0.
- `npx vitest run --config vitest.config.mjs src/test-kernel/common/SdkErrorUtils.vitest.ts src/test-kernel/llm/GoogleInteractions.vitest.ts` - 2 files, 126 tests passed.

This PR's own `static checks` job proves the install step end to end.


LionSR:
Superseded by #12966, which landed the same repair a few minutes earlier (`origin/main` is now `6eb322f83e`: the root importer is back to `specifier: 'catalog:'`). Closing rather than keeping two copies of the same two-line fix.

One residual from the way it was fixed: the root workspace now resolves `@google/genai` at **2.22.0** while `packages/llm/package.json` pins **2.23.0**, so the install carries two copies of that SDK. This branch's alternative - bumping the catalog range to `^2.23.0` and regenerating so both spots resolve one version - is the cleaner end state if the 2.23.0 bump was wanted at all; it is available on `fix/lockfile-genai-catalog` if someone wants to take it as a follow-up rather than let dependabot reopen the same bump.

PR #12975, opened 2026-09-21
refactor(platform): retire the last ambient platform() readers and delete the ratchet row (#12421)

LionSR:
## Summary

Retires the last three ambient `platform()` readers (#12421) and deletes the migration ratchet's `platform()` row — it reaches **0 files / 0 sites**. Each reader now takes its capability from a Context service the composition root already provides, per ruling #12073 R-1 ("Platform shrinks onto those services") and the existing `AgentResume`/`AppState`/`Secrets` precedent:

| reader | was | now |
| --- | --- | --- |
| `src/agent/index/agentRegistry.ts` | `platform().agentDirectories` | `AgentDirectories` process service |
| `src/tools/github/PollingSourceBase.ts` (+ `RunSubscriptionRegistry`, the three polling sources) | `platform().lifecycle` | `Lifecycle` process service |
| `src/utils/system/toolUtils.ts` | `platform().toolMissingHandler` | `ToolMissingReporter` via `Effect.serviceOption` — a host without the UI stays silent, exactly as the absent port did |

Both `catch:effect-importer` sites remain, now documented: the desktop process entry and the polling-listener fan-out are load-bearing boundary swallows, which is the accepted floor.

`ProcessServices` gains the three services; `installProcessRuntime` provides them; desktop, extension, CLI (built at runtime-install time rather than in `initCliPlatform`, so joining an installed runtime cannot produce a second copy), the agent-SDK embedder and the webview runtime each pass their own values. Test support provides fakes and the affected suites now provide services instead of mocking modules — every existing assertion preserved.

## Verification

- `tsc --noEmit --checkers 8`, `-p tsconfig.test-kernel.json`, `--filter @texra-ai/cli typecheck`, `--filter @texra/desktop typecheck`, `--filter @texra-ai/agent build` — clean
- The 22 affected suites pass (148 tests)
- Full suite green on this branch: **528 files / 5448 passed / 1 skipped**
- Two failures the first round of CI caught, both fixed on the branch: `fileSizeRatchet` (a touched file exceeded its budget - see the budget note below) and two `CliInitPlatform` cases that assert against the runtime *this root* installs, which must now install their own rather than join the test kernel's
- `check:effect-migration-ratchet` — `platform()` 0/0 (row deleted), no count grew; `check:unexecuted-effects`, `check:effect-test-runs` (3 exemptions), `check:dead-code-ratchet` — pass
- `prettier --check` and `eslint` clean

## Deliberately not done

- `Platform` still *declares* `lifecycle`/`agentDirectories`/`toolMissingHandler` and hosts still pass them to `initPlatform(...)`, but nothing reads them through `platform()` any more; deleting `Platform`/`initPlatform` outright is a follow-up with its own callers to convert.
- `src/platform/workspaceRoots.ts` is **not** deletable: its ambient readers are already gone and it now declares only the `WorkspaceRoots` type, which many production files import.
- No user-visible behaviour change: the capability moves, and absence of the tool-missing handler still means silence.

## Refused candidates this diff touches

`npm run check:refuted-candidates` requires a PR that touches a declaration a ruling already refused to cite that ruling's id and say whether the refusal still stands. Two are touched; neither refusal is re-proposed:

- **`RT-install-cli-process-runtime`** — "installCliProcessRuntime as an Effect program", refused because *"it is the bootstrap edge itself, so there is no runtime to run it on; an Effect face forces a bare `Effect.run*` at every citty action and grows `BARE_EFFECT_RUN_SITES` from 7 to about 20"* (`.agents/docs/proposed/simplification/2026-09-17-effect-round-trips-and-dual-systems.md#refuted-or-declined-at-implementation`). This PR keeps `installCliProcessRuntime` exactly what the refusal describes: a plain function called at the bootstrap edge. It gains one parameter (`resourcesPath`) so the *runtime install* can build the host's `lifecycle` and `agentDirectories` before a runtime exists — the same values `initCliPlatform` used to build after installing. No `Effect.run*` site is added at any citty action; the entry stays refused and unchanged.
- **`RT-run-resume-command`** — "runResumeCommand as a round trip", refused because *"not a round trip: resume.ts's citty `run(ctx)` awaits the promise at an R1 entry"* (same ruling). This PR changes only the `installCliProcessRuntime` call site in `resumeRun.ts`, to pass `resourcesPath`. No round trip is introduced; the entry stays refused and unchanged.

## File-size budget bumps (please review)

Retiring the ambient reads costs plumbing in seven files that are already on the `file-size-baseline.json` ceiling. Each number is a ceiling, so raising one is visible here rather than silent, and the growth is imports of the new `Lifecycle`/`AgentDirectories` services plus passing them through — not new logic:

| file | before | after | why |
| --- | ---: | ---: | --- |
| `runChatTui.tsx` | 651 | 653 | `resourcesPath` pass-through |
| `SupabaseAuthProvider.ts` | 540 | 548 | `agentDirectories` provision |
| `extensionHostRequests.ts` | 937 | 941 | `lifecycle` provision |
| `agentCreatorFlow.ts` | 532 | 534 | `agentDirectories` read |
| `sessionLayer.ts` | 1204 | 1235 | the three services in `ProcessServices` + `installProcessRuntime` signature |
| `PollingSourceBase.ts` | 777 | 781 | `Lifecycle` read replacing `platform()` |
| `toolUtils.ts` | 503 | 513 | `ToolMissingReporter` via `serviceOption` |

The ratchet's failure text prefers moving new code out of the file. I chose the bump because the growth is the plumbing itself, and it stays honest (a ceiling, reviewed here). If you would rather extract — `sessionLayer`'s install signature is the largest — that is a mechanical follow-up on this branch.


LionSR:
Heads-up from the CI log: the `refused refactor candidates` job is red for a documentation reason, not a code one.

Your diff touches two declarations that already have refusal rulings on file, and the gate (`scripts/check-refuted-candidates.mjs`) requires the PR body to cite the ruling id for each:

- **`RT-install-cli-process-runtime`** — `installCliProcessRuntime` as an Effect program. *Why refused:* it is the bootstrap edge itself, so there is no runtime to run it on; an Effect face forces a bare `Effect.run*` at every citty action and grows `BARE_EFFECT_RUN_SITES` from 7 to about 20.
- **`RT-run-resume-command`** — `runResumeCommand` as a round trip. *Why refused:* not a round trip — `resume.ts`'s citty `run(ctx)` awaits the promise at an R1 entry.

Both rulings: `.agents/docs/proposed/simplification/2026-09-17-effect-round-trips-and-dual-systems.md#refuted-or-declined-at-implementation`, entries also recorded in `config/ratchets/refuted-candidates.json`.

The gate never forbids the change. It asks for one of: cite the id(s) in this body (and, if the change now stands on new evidence, update or delete the entry in `config/ratchets/refuted-candidates.json` in this PR), or drop the refused hunks. The PR is also conflicting and has both kernel shards red — though those were main's own hard hang, fixed in #12985, so a rebase onto the new `main` (`34777ac5ef`) should clear them.

PR #12984, opened 2026-09-22
fix(deps): bump openai to 7.20.0 and keep the cleanup-defect contract patched

LionSR:
## Summary

Bumps `openai` from the pinned `7.10.0` to the **latest published release, `7.20.0`**, and keeps the repository's cleanup-defect contract by extending the SDK patch that already exists for this dependency. No provider code changed; no test was weakened, skipped or deleted.

The bump itself is mechanical: `packages/llm/package.json` (exact pin), the `pnpm-workspace.yaml` catalog (`^7.10.0 → ^7.20.0`, plus the `patchedDependencies` key), the ported `patches/openai@7.20.0.patch` for the existing `response.metadata` intent, and the regenerated `pnpm-lock.yaml`.

## The blocker this had to solve first

`OpenaiResponses.vitest.ts` → "retains learned acceptance in an unexpected cleanup defect" failed on 7.20.0: a stream cleanup failure was no longer surfaced. Forensics (diffs of the published tarballs):

- **`openai` 7.13.0 introduced the behaviour** - there is no 7.11/7.12 release; the line jumps `7.10.0 → 7.13.0`.
- It wraps `_iterSSEMessages` in `Stream.fromSSEResponse` (`src/core/streaming.ts`) with a `messages.return` that calls `controller.abort()` **before** the nested iterator's cleanup runs. `createAbortableSSESource.cleanup` then sees `signal.aborted === true` and skips rethrowing the underlying `ReadableStream.cancel()` rejection.
- **The masked rejection is load-bearing, not cosmetic.** Reproduced: the test's `ReadableStream.cancel()` rejects with `Error('cancel rejected')`, yet `model.background.submit(turn)` returned `{ kind: 'accepted' }` - a **failed stream cleanup was recorded as a successful run**. The repository deliberately preserves "distinct cleanup defects and abort order" (`.agents/docs/proposed/architecture/2026-09-07-effect-ai-comparison.md`), so accepting the new ordering would be a correctness regression, not a semantics update.

**Resolution:** three hunks appended to `patches/openai@7.20.0.patch` (`core/streaming.js`, `core/streaming.mjs`, `src/core/streaming.ts`) remove the abort-first wrapper so `fromSSEResponse` iterates `_iterSSEMessages(response, controller)` directly again. The provider already orders `iterator.return()` before `source.controller.abort()`, so the distinct cleanup defect is observable exactly as before. Each hunk carries an intent comment in the repo's patch style. Confirmed with `git apply --check` against a pristine 7.20.0 tarball (all hunks apply cleanly) and by `pnpm install --frozen-lockfile --offline` afterwards (only the patch hash changed in the lockfile).

## Verification

- `npx vitest run --config vitest.config.mjs src/test-kernel/llm` — 5 files / **381 tests pass** (including the cleanup-defect case)
- `npx vitest run --config vitest.config.mjs` — **528 files / 5448 passed / 1 skipped**
- `corepack pnpm --filter @texra-ai/llm typecheck` — clean; `tsc --noEmit --checkers 8` and `-p tsconfig.test-kernel.json` — clean
- `check:effect-migration-ratchet`, `check:unexecuted-effects`, `check:effect-test-runs`, `check:dead-code-ratchet` — pass
- `pnpm install --frozen-lockfile --offline` — "Already up to date"

## Residual maintenance (please note when reviewing)

The patch now depends on `Stream.fromSSEResponse` iterating `_iterSSEMessages(response, controller)` directly with no abort-before-close wrapper. On the next `openai` bump, those three hunks must be re-ported if upstream reworks that function (or `_iterSSEMessages`'s `finally` / `createAbortableSSESource.cleanup`). The `response.metadata` hunks in the same patch are independent and unaffected.

Alternative considered and rejected: staying at 7.10.0 forever (no explanation, no bump), and accepting 7.20.0's ordering (contradicts the recorded stance and silently reports failed cleanups as successes).


LionSR:
`kernel tests (linux, shard 1/2)` was failing on this branch for a main-side reason, not yours: `AgentLaunchActivation.vitest.ts` > "joins the session description fiber on the run-failure path" hung to its timeout. That test landed red with #12669 (main's CI does not run on push), and its cause was a test-wiring bug - `mocks.buildVars.mockResolvedValueOnce({})` handed the run loop a Promise where production `buildUserVars` returns an Effect, so the launch died during assembly and the description fiber was never forked. Fixed in #12985 (one line, merged).

I have rebased this branch onto the new `main` (`281b9ace65`) and force-pushed with lease so its CI can be meaningful; the repaired test passes here (5/5). Re-run/review at your convenience - no content of yours was changed.

PR #12986, opened 2026-09-22
test(kernel): the remaining Effect-native test conversions, and a gate that counts every run

LionSR:
## Summary

The remaining test-side Effect boundaries, converted to the repository's idiom. Every change is a migration: no test, assertion or scenario was added, removed or weakened.

**1. `Effect.runSync` in a test body (17 of 18 sites).**

- `tools/PRPollingSourceAnnotationPages.vitest.ts` — 3 sites: `expect(Effect.runSync(budget.tryClaim(n))).toBe(v)` became `it.effect` + `expect(yield* budget.tryClaim(n)).toBe(v)` (the test passes an explicit `nowMs`, so the test clock is irrelevant to it).
- `tools/DiagnosticsTool.vitest.ts` — 4 sites: the `Effect.runSync(session.interactions.use({...}))` context installs are now `yield*` inside the bodies' existing `Effect.gen`.
- `tools/NativeSubagentStrategy.vitest.ts` — 6 sites that sat inside `vi.fn` mock-implementation callbacks. A double's `async` callback cannot `yield*`, so the run moved into a named non-test helper (`publishRunHandle`) — the sanctioned place for a run — and each double calls it. Timing and behaviour unchanged.
- `logger/LogUtils.vitest.ts` (2) and `frontend/AgentReviewRunController.vitest.ts` (2) — bodies became `it.effect` + `yield*`.
- Left alone: `agent/runtime/RunAgentOwnership.vitest.ts:553`, whose innermost enclosing callback is a mock factory rather than the test's own body (an allowed surface, like a hook).

**2. Fake timers → the Effect test clock (1 of 6 suites in this set).** `frontend/OpenBuildDisplay.vitest.ts`'s subject is an Effect program whose `Effect.sleep` the fake timers drove, so its 8 bodies became `it.effect` + `yield* TestClock.adjust(...)`, `vi.useFakeTimers`/`advanceTimersByTimeAsync` are gone, and the suite is deterministic across 5 consecutive runs (9/9). The other five are deliberate keeps, each because the subject is not an Effect program: `SubscriptionUsageRendering` drives a Lit component's own `setInterval`/`Date.now`; `MemoryTool` already runs `it.effect` and fakes only `Date`; `DesktopFileRequests` and `UtilsCore` exercise plain `setTimeout` facades; `RunTraceStream`'s fake timer is a negative probe (`getTimerCount() === 0`).

**3. The gate now counts `runSync`** — with no test body executing a program through any run, the deliberate `runSync` exclusion lost its reason. `RUN_NAMES` and the baseline's `semantics` are updated, so the rule is complete: **no runtime boundary inside a test body**, with runs left to hooks and non-test helpers and three recorded exemptions.

## Verification

- `tsc --noEmit --checkers 8` and `-p tsconfig.test-kernel.json` — clean
- The six converted suites — 6 files / 50 tests pass; the architecture ratchets — 16 files / 109 tests pass
- `check:effect-test-runs` — no new site, **3 reasoned exemptions** (unchanged); `check:effect-migration-ratchet`, `check:unexecuted-effects`, `check:dead-code-ratchet` — pass
- `prettier` and `eslint` — clean

## Follow-up

Still in flight from the same round, to land as a second PR: the remaining fake-timer suites (`testclock-b`), the harness-owned `vi.waitFor` polls that can await a `Deferred` the fake owns (`polling-a`/`polling-b`), and the hand-rolled `deferred<T>()` gates (`deferred-helpers`).


LionSR:
`kernel tests (linux, shard 1/2)` was failing on this branch for a main-side reason, not yours: `AgentLaunchActivation.vitest.ts` > "joins the session description fiber on the run-failure path" hung to its timeout. That test landed red with #12669 (main's CI does not run on push), and its cause was a test-wiring bug - `mocks.buildVars.mockResolvedValueOnce({})` handed the run loop a Promise where production `buildUserVars` returns an Effect, so the launch died during assembly and the description fiber was never forked. Fixed in #12985 (one line, merged).

I have rebased this branch onto the new `main` (`281b9ace65`) and force-pushed with lease so its CI can be meaningful; the repaired test passes here (5/5). Re-run/review at your convenience - no content of yours was changed.

Issue #12880, opened 2026-09-20
Tracking: post-refactor architecture debts (2026-09-20 survey)


LionSR:
## Current status — 2026-09-20

Checked against `main` at `3378a967` on 2026-09-20. Umbrella for the seven proposals filed from the post-refactor architecture survey. Each child issue owns one proposal; this issue only tracks their state.

Survey record: `2026-09-20-post-refactor-architecture-survey.md` in [proposed/architecture](https://github.com/LionSR/TeXRA/tree/claude/architecture-refactoring-progress-dyyk7f/.agents/docs/proposed/architecture) on branch `claude/architecture-refactoring-progress-dyyk7f`. Status page: https://claude.ai/artifact/9aeiYZmqg1AKX5UTeoK5az

## Verdict

The engine replacement is done: `PersistedFlow`, `KVStore`, `executionLease`, `fileLocks`, the model-handler hierarchy, the run-context `AsyncLocalStorage` and `Platform.fs` have zero references; every architecture baseline moved down in the window; the Effect-migration baseline is at seven ruled-permanent sites. What remains is second copies of run state, duplicated scaffolding, an unhardened LLM package, triplicated host logic, an oversized schema barrel and tool-call path, four unused Effect families, and a test and docs corpus an autonomous agent cannot navigate.

## Child issues

- [ ] #12881 single-owner liveness and one fold (design; first)
- [ ] #12884 host-layer collapse (design; the CLI `uiHosts` ports unlock the rest)
- [ ] #12887 agent-refactorability gates (governance; the 12 ambient reads first)
- [ ] #12883 LLM package hardening (risk; the only shipped regression)
- [ ] #12882 run-program and dispatch dedup (mechanical once #12881 lands)
- [ ] #12885 tools and schema-surface collapse (mechanical)
- [ ] #12886 Effect facility adoption (mechanical)

## Sequencing

1. #12881 first: it closes the bug class that produced 13 of the 18 session and tool fixes in the window.
2. The CLI `uiHosts` ports in #12884 next: they turn `src/controllers` into a three-host layer and unlock the settings, credential and Supabase collapses.
3. The 12 ambient reads in #12887: each removal shrinks the fake-platform layer and moves suites to the fast tier.
4. Then the pure deletions and the apparatus items.

## Not re-proposed

[...]

PR #13009, opened 2026-09-22
refactor: collapse Effect run, settings, and storage layers

LionSR:
## Summary

Native child runs now keep one scope and live handle across follow-ups, settings actions compose directly as Effect programs, and project settings share their persistent database with the session. The extension now uses the same SQLite application-state implementation as the other hosts. This removes the parked-run lifecycle, backend Promise dispatch wrappers, and duplicate project database ownership.

Stacked on #13008, with `refactor/effect-native-app-state` as the base. The diff contains only these follow-up collapses.

## Issue acceptance gates

- Fresh and recovered native children use the same turn admission, accounting, and delivery driver. Idle children release their concurrency permit. Stopping an accepted turn preserves the previous completed report, and detachment changes policy through the live handle.
- Recovered children acknowledge a completed idle turn while their driver stays live. The driver alone owns the transferred recovery lease, and acknowledgment checks the admitted input against authoritative run state. Deferred delivery respects recovered consumers; startup cancellation or failure records terminal state before releasing ownership. Shutdown drains through lane cleanup, and duplicate launches cannot replace a live stop handle.
- Settings messages execute once at their native host entry. Provider-key failures distinguish a failed write from a failed refresh after the key was stored; cancellation and cleanup failures remain observable. Desktop notification failures reach the final error sink with their underlying cause.
- Project state and persistent sessions share one captured SQLite connection. Closing a session preserves a remaining state owner; closing the last owner releases the connection. Ephemeral transcripts retain persistent project settings. Extension global state uses the process database, project state uses the shared project connection, and shutdown releases both after failed activation or session cleanup. Old Memento state is left untouched and is not imported.

## Single source of truth

`RunHandle` owns live lineage; `childRunLoop` owns child-turn accounting and durable delivery. `settingsViewProgram` selects the validated command's Effect, which each host executes at its message boundary. `ProjectDatabases` is the process's native `RcMap` of persistent project databases, borrowed once by each owning scope.

Deletes parked fibers and registry entries, the nonterminal `WAITING` result type and completed-result conversion, per-turn native relaunch and interrupt rebinding, the copied run-parent field, and unused input polling. Removes internal refresh-command round trips, provider-key error callbacks, settings handler runner ports, and the unused `catchNotice` wrapper. Deletes the extension Memento state adapter and converts its remaining direct consumers; no second persistent state implementation remains for extension settings.

Two small modules are added: the shared database layer enforces connection lifetime across its two borrowers, and the backend command router serves the two graphical hosts while shared browser contracts remain dependency-light. SQLite operations continue through the existing database implementation.

## Net elements (R6)

104 files: 2,864 lines added, 4,579 removed (net -1,715). Production TypeScript: 57 files, 1,620 added and 2,160 removed (net -540). Two modules added; the obsolete strategy-test file and Memento adapter deleted. Export declaration diff: 10 added / 15 removed; class, interface, and enum declaration diff: 4 added / 2 removed. Error-channel and dead-code baselines shrink; file-size budgets are unchanged.

## Consumer counts (R8)

- Old waiting-result and parked-run APIs: zero remaining production references.
- Old backend inbound dispatcher: zero remaining consumers; `settingsViewProgram` has two host consumers.
- Provider-key controller: all four bindings converted (CLI, desktop credentials, extension settings, extension commands).
- Shared project database acquisition: two consumers, project-state acquisition and session-graph acquisition; no acquisition per state operation.
- `catchNotice`: zero consumers, declaration removed. `isUnsupported` now has a production consumer, so its dead-code baseline entry is removed.

## Host impact

Extension: settings commands and refreshes run as composed Effects; external command registrations remain available. Global and project application state use SQLite, including directory selection, inline criticism, and onboarding. SecretStorage and worktree-sharing policy remain unchanged.

Desktop: direct settings dispatch and one project scope for state and session disposal.

CLI: typed provider-key failures and one project scope retaining persistent state through lazy session startup.

All hosts: native child runs retain their live scope between turns and use the same recovery driver.

## Dispatcher ruling revision

Implements the owner's [explicit shared-contract revision](https://github.com/LionSR/TeXRA/pull/13009#issuecomment-5779248081): one backend selector and one native execution boundary per host replace the old dispatcher and per-arm runners together. The architecture rulings ledger now records the remaining synchronous frontend consumers and failure ownership. The local #12880/#12884 notes are reconciled without claiming those broader trackers complete.

## Scheduled refactoring

None required for these changes.

## Validation

- Full suite plus isolated worker retry: all 530 suites covered, 5,457 tests passed and one skipped. The full run passed 529 suites but timed out starting the worker for unchanged `cli/RetryRequest.vitest.ts`; rerunning that suite alone passed all four tests.
- Full `npm run typecheck` passed on the final sources.
- Full `npm run lint` passed; scoped lint also passed after the final error-channel annotation and startup-finalization edits.
- `npm run compile:fast`, `npm run format`, and `npm run format:check` passed.
- Effect migration, Effect test execution, unexecuted Effect, dead-code, and guidance-reference gates passed; architecture ratchets passed in the full suite.
- Refuted-candidate gate: no refused declaration touched.
- `git diff --check` and staged diff check passed.


LionSR:
Clarification to the architecture review: treating #13005's historical Promise-boundary ruling as an automatic blocker was too strong. The owner's agreement with #13005's overall direction did not separately ratify every historical ruling it references.

The original rationale was to avoid a parallel Effect dispatcher or pass-through adapters when host callbacks already require Promise/void. This PR changes the evidence: it moves execution to one native host message entry and removes the per-arm runners. That can be the cleaner design if it replaces the old backend routing completely, preserves failure/interruption behavior, and leaves the frontend dispatcher contract coherent.

The principle to preserve is **one execution boundary, no duplicate dispatch machinery or pass-through adapters**. Requiring every individual registry arm to be that boundary is an implementation decision we should reconsider against this diff, not an intrinsic compatibility requirement of #13005.

Please document an explicit revision of the shared-contract ruling, including the remaining frontend consumers and rejection handling, rather than treating this as incidental cleanup. The earlier instruction in #12880/#12884 to remove the settings-dispatch slice is therefore not a final technical rejection; those tracker notes should be reconciled with the explicit decision. This comment does not claim the replacement has completed that validation.

Separately, the recovered-child resume issue identified at `1f4b06f` remains a correctness concern: `resumeRun.ts` joins the continuous child's lifetime fiber, so returning to WAITING no longer settles the resume attempt. Resume acknowledgment must be distinguished from lifetime completion. Revising the dispatcher ruling does not resolve that issue.

